PR #5, opened 2021-06-12
Add network module

chimp1984:
Based on https://github.com/bisq-network/misq/pull/4


chimp1984:
Merged dependent PRs and rebased.

jmacxx (approved):
utACK

network/src/main/java/network/misq/network/http/common/ClearNetHttpClient.java
```diff
@@ -0,0 +1,126 @@
...
+                connection.setRequestProperty(header.first(), header.second());
+            });
+
+            if (httpMethod == HttpMethod.POST) {
+                connection.setDoOutput(true);
+                connection.getOutputStream().write(param.getBytes(StandardCharsets.UTF_8));
+            }
+
+            int responseCode = connection.getResponseCode();
+            if (responseCode == 200) {
```
kaiwolfram: What about other success codes like 204? Same in Tor Client

chimp1984: Good point. Yes I guess we should consider all 2xx codes as success case.

ghubstan: Resolved by commit 521de33087c3fbd3d52b651620930e169a1ada8b.

ghubstan:
utACK

PR #9, opened 2021-06-21
Upgrade Gradle to v7.1

ghubstan:
The most relevant change in v7.1 is the [caching of incremental compilation analysis](https://docs.gradle.org/current/release-notes.html#incremental-java) to speed up incremental builds. (I am seeing some <200ms inc. builds.)

A _"Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0"_ warning is sometimes logged during a build, but `./gradlew build --warning-mode all` does not tell us what in our build is deprecated.  I think the intermittent warning could be ignored until Gradle tells us something specific, and this warning might be due to some 3rd party plugin(s) we are using.


jmacxx (approved):
ACK

PR #10, opened 2021-06-28
Create gradle network-platform

ghubstan:
The `network-platform` defines version constraints for dependencies common to (or may become common to) network modules.  

The first commit defines platform constraints for commons-compress, httpclient, and jsocks libraries.  Dependencies such as the tor libs are unique to the tor module, and are excluded from network-platform.

Following commits are related to usage (consumption) of the network-platform, and declaring needed network-platform dependencies w/out version specs.

Information about Gradle platforms is [here](https://docs.gradle.org/current/userguide/java_platform_plugin.html).

There is also a switch from double-quotes to single quotes.  Working with a mix of single and double quotes in build files can get annoying if one does not know why one type is chosen over the other.  

Double quotes are needed for string interpolation -- a groovy thing.  For example:

		"group:artifactId:$someVersion"

In gradle, single quotes are the convention when not using string interpolation:

		'group:artifactId:2.0.0'

We might save someone a little future guess-work over this if we can keep double quote usage to a minimum (for string interpolation only).


ghubstan:
I tried to request @chimp1984 to review this, but he doesn't pop up in the search results.

platforms/network-platform/settings.gradle
```diff
@@ -0,0 +1 @@
+rootProject.name = 'network-platform'
```
chimp1984: Seems there is missing a new line.

chimp1984 (commented):
utACK

chimp1984 (approved):

platforms/network-platform/settings.gradle
ghubstan: Newline added by https://github.com/bisq-network/misq/pull/10/commits/4f96b561c1a84fcb0e266b80763a1f6b60bba8a4.

chimp1984 (approved):
utACK

jmacxx (approved):
ACK
(I only checked that it clean builds ok)

PR #16, opened 2021-12-21
Integrate Bitcoind as a wallet backend into Misq

alvasw:
## Tasks
- [x] Bitcoind Process Manager (startup & shutdown)
- [x] RPC Connection to Bitcoind
- [x] RPC calls:
    - [x] `addmultisigaddress`
    - [x] `createwallet`
    - [x] `finalizepsbt`
    - [x] `generatetoaddress`
    - [x] `getaddressinfo`
    - [x] `getbalance`
    - [x] `getnewaddress`
    - [x] `importaddress`
    - [x] `listtransactions`
    - [x] `listunspent`
    - [x] `loadwallet`
    - [x] `sendrawtransaction`
    - [x] `sendtoaddress`
    - [x] `signmessage`
    - [x] `stop`
    - [x] `unloadwallet`
    - [x] `verifymessage`
    - [x] `walletcreatefundedpsbt`
    - [x] `walletpassphrase`
    - [x] `walletprocesspsbt`
- [x] Integration test for all RPC calls

Fixes bisq-network/projects#62


wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindBackend.java
```diff
@@ -0,0 +1,212 @@
...
+import network.misq.wallets.bitcoind.psbt.PsbtOutput;
+import network.misq.wallets.bitcoind.psbt.WalletCreateFundedPsbtOptions;
+import network.misq.wallets.bitcoind.responses.*;
+import network.misq.wallets.rpc.RpcClient;
+
+import java.nio.file.Path;
+
+public class BitcoindBackend {
+
+    public static final int MAX_WALLET_TIMEOUT = 100000000;
```
chimp1984: I assume 100000000 is ms. Thats 27.77 hours. Is that intended to be that long?
I think its better to use `public static final int MAX_WALLET_TIMEOUT = TimeUnit.HOURS.toMillis(24)`; for better readability

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindBackend.java
```diff
@@ -0,0 +1,212 @@
...
+
+    public WalletCreateFundedPsbtResponse walletCreateFundedPsbt(PsbtInput[] inputs,
+                                                                 PsbtOutput psbtOutput,
+                                                                 WalletCreateFundedPsbtOptions psbtOptions) {
+        return walletRpcClient.invoke(
+                BitcoindRpcEndpoint.WALLET_CREATE_FUNDED_PSBT,
+                new Object[]{
+                        inputs,
+                        psbtOutput.toPsbtOutputObject(),
+                        0, // locktime
```
chimp1984: I am not familiar with the API and PSBT spec. Would it make sense to add lockTime to the arguments or must it be 0 in that use case?

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindBackend.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    private void createWallet(String passphrase) {
+        String absoluteWalletPath = walletPath.toAbsolutePath().toString();
+        CreateOrLoadWalletResponse response = rpcClient.invoke(
+                BitcoindRpcEndpoint.CREATE_WALLET,
+                new Object[]{
+                        absoluteWalletPath, // filename
+                        false, // Is watch only wallet?
+                        false, // Create blank wallet (no seed / no HD seed)
```
chimp1984: Would it make sense to add params for those values?

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindProcess.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private Process bitcoindProcess;
+
+    public BitcoindProcess(RpcConfig rpcConfig, Path dataDir) {
+        this.rpcConfig = rpcConfig;
+        this.dataDir = dataDir;
+    }
+
+    public void startAndWaitUntilReady() {
+        dataDir.toFile().mkdirs();
```
chimp1984: Not sure what happens if `mkdirs()` fail. Probably an exception in some follow up code. Maybe better make it explicit by use e.g. `FileUtils.makeDirs( dataDir.toFile());` which throws an `IOException`.

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindProcess.java
```diff
@@ -0,0 +1,98 @@
...
+                    "-rpcallowip=" + rpcConfig.hostname(),
+                    "-rpcuser=" + rpcConfig.user(),
+                    "-rpcpassword=" + rpcConfig.password(),
+                    "-fallbackfee=0.00000001",
+                    "-txindex=1"
+            ).start();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        return null;
```
chimp1984: I think it's better to add the exception to the method signature. A null return value will likely screwup follow up code anyway.

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindProcess.java
```diff
@@ -0,0 +1,98 @@
...
+        } catch (IOException | InterruptedException e) {
+            e.printStackTrace();
+        }
+    }
+
+    private Process createAndStartProcess() {
+        try {
+            return new ProcessBuilder(
+                    "bitcoind",
+                    rpcConfig.networkType() == NetworkType.REGTEST ? "-regtest" : "",
```
chimp1984: I assume mainnet is default but other networks might have as well a custom param.

wallets/src/main/java/network/misq/wallets/bitcoind/psbt/PsbtOutput.java
```diff
@@ -0,0 +1,27 @@
+package network.misq.wallets.bitcoind.psbt;
+
+import lombok.Setter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class PsbtOutput {
```
chimp1984, line 8: I am not sure about the usage of it, but would it be possible to use a record instead, so its always immutable data?

wallets/src/main/java/network/misq/wallets/bitcoind/psbt/WalletCreateFundedPsbtOptions.java
```diff
@@ -0,0 +1,13 @@
+package network.misq.wallets.bitcoind.psbt;
+
+import lombok.Getter;
+import lombok.Setter;
+
+public class WalletCreateFundedPsbtOptions {
```
chimp1984: I am not sure about the usage of it, but would it be possible to use a record instead, so its always immutable data?

wallets/src/main/java/network/misq/wallets/bitcoind/responses/BitcoindResponse.java
```diff
@@ -0,0 +1,18 @@
+package network.misq.wallets.bitcoind.responses;
+
+import lombok.Getter;
+import lombok.Setter;
+
+public class BitcoindResponse {
```
chimp1984: If it is not intended to be used directly but only for extending, we should make it abstract

wallets/src/main/java/network/misq/wallets/rpc/RpcCallFailureException.java
```diff
@@ -0,0 +1,11 @@
+package network.misq.wallets.rpc;
```
chimp1984: Is the rpc package generic enough to leave it outside of bitcoind or is it specific for bitcoind? If latter better move it inside bitcoind package

wallets/src/test/java/network/misq/wallets/bitcoind/BitcoindRegtestSetup.java
```diff
@@ -0,0 +1,118 @@
...
+        minerBackend.sendToAddress(receiverAddress, amount);
+
+        mineOneBlock(minerBackend);
+        return receiverAddress;
+    }
+
+    /* Source:
+       https://stackoverflow.com/questions/15022219/does-files-createtempdirectory-remove-the-directory-after-jvm-exits-normally/20280989
+     */
+    private static void recursiveDeleteOnShutdownHook(final Path path) {
```
chimp1984: We dont use final in params. Maybe we can move that to a common util? Maybe to FileUtils?

wallets/src/test/java/network/misq/wallets/bitcoind/BitcoindRegtestSetup.java
```diff
@@ -0,0 +1,118 @@
...
+
+public class BitcoindRegtestSetup {
+    public static final RpcConfig RPC_CONFIG = new RpcConfigBuilder()
+            .setNetworkType(NetworkType.REGTEST)
+            .setHostname("127.0.0.1")
+            .setUser("bisq")
+            .setPassword("bisq")
+            .build();
+
+    public static final String WALLET_PASSPHRASE = "My super secret passphrase that nobody can guess.";
```
chimp1984, line 19: Better move fields to top

wallets/src/test/java/network/misq/wallets/bitcoind/BitcoindRegtestSetup.java
```diff
@@ -0,0 +1,118 @@
...
+
+    public static Path createTempDir() throws IOException {
+        Path tempDirPath = Files.createTempDirectory(null);
+        recursiveDeleteOnShutdownHook(tempDirPath);
+        return tempDirPath;
+    }
+
+    public static BitcoindBackend createTestWalletBackend(Path tmpDirPath, String walletName, RpcClient rpcClient) throws MalformedURLException {
+        Path receiverWalletPath = tmpDirPath.resolve(walletName);
+        RpcClient receiverWalletRpc = BitcoindRegtestSetup.createWalletRpcClient(receiverWalletPath);
```
chimp1984: `BitcoindRegtestSetup.` can be omitted

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindBackend.java
```diff
@@ -0,0 +1,212 @@
...
+        return walletRpcClient.invoke(
+                BitcoindRpcEndpoint.SIGN_MESSAGE,
+                new Object[]{address, message},
+                String.class
+        );
+    }
+
+    public void unloadWallet() {
+        String absoluteWalletPath = walletPath.toAbsolutePath().toString();
+        UnloadWalletResponse response = rpcClient.invoke(
```
chimp1984: Shouldn't that be using `walletRpcClient`?

wallets/src/test/java/network/misq/wallets/bitcoind/BitcoindSigningTests.java
```diff
@@ -0,0 +1,18 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class BitcoindSigningTests extends SharedBitcoindInstanceTests {
+    private static final String MESSAGE = "my message";
+
+    @Test
+    public void signAndVerifyMessage() {
+        String address = minerBackend.getNewAddress(AddressType.LEGACY, "");
+        String signature = minerBackend.signMessage(address, MESSAGE);
+        boolean isValid = minerBackend.verifyMessage(address, signature, MESSAGE);
```
chimp1984: Does verifyMessage also work from another wallet? I assume not as the pubKey is not know to another wallet as long the address was not used in a tx and the pubKey revealed.

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindBackend.java
```diff
@@ -0,0 +1,212 @@
...
+    private final RpcClient walletRpcClient;
+    private final Path walletPath;
+
+    public BitcoindBackend(RpcClient rpcClient, RpcClient walletRpcClient, Path walletPath) {
+        this.rpcClient = rpcClient;
+        this.walletRpcClient = walletRpcClient;
+        this.walletPath = walletPath;
+    }
+
+    public AddMultisigAddressResponse addMultisigAddress(int nRequired, String[] keys) {
```
chimp1984: Is there a reason why to use arrays rather than Lists? I would consider Lists more comfortable to work with.

wallets/src/main/java/network/misq/wallets/bitcoind/psbt/WalletCreateFundedPsbtOptions.java
chimp1984: The name is a bit confusing for me. Would just `PsbtOptions` be sufficient?

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindProcess.java
```diff
@@ -0,0 +1,98 @@
...
+        try {
+            while (exitCode != 0) {
+                Process p = new ProcessBuilder(
+                        "bitcoin-cli",
+                        rpcConfig.networkType() == NetworkType.REGTEST ? "-regtest" : "",
+                        "-rpcuser=" + rpcConfig.user(),
+                        "-rpcpassword=" + rpcConfig.password(),
+                        "stop"
+                ).start();
+                exitCode = p.waitFor();
```
chimp1984: Is the while loop with the exitCode needed? The `waitFor` is blocking until the process is completed, now?

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindProcess.java
```diff
@@ -0,0 +1,98 @@
...
+
+                bitcoindProcess.waitFor();
```
chimp1984: When running the tests it takes quite a while until the process is terminated.  Not sure what happens if bitcoind gets shut down too quickly, probably for mainnet not good, but for tests with regtest might be ok to speed up test runs. 

We could pass a timeout and use the matching method signature.  Maybe have one `stopAndWaitUntilStopped()` signature with timeout and one without?

`if (!bitcoindProcess.waitFor(100, TimeUnit.MILLISECONDS)) {
           log.warn("Terminated bitcoindProcess by a timeout");
 }`

wallets/src/test/java/network/misq/wallets/bitcoind/BitcoindSendTests.java
```diff
@@ -0,0 +1,169 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.nio.file.Path;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class BitcoindSendTests {
```
chimp1984, line 21: I get the second test failed as bitcoind shuts down in between.

wallets/src/test/java/network/misq/wallets/bitcoind/BitcoindPsbtMultiSigTests.java
```diff
@@ -0,0 +1,81 @@
...
+
+        aliceBackend.importAddress(aliceMultiSigAddrResponse.getAddress(), "");
+        bobBackend.importAddress(bobMultiSigAddrResponse.getAddress(), "");
+        charlieBackend.importAddress(charlieMultiSigAddrResponse.getAddress(), "");
+
+        minerBackend.sendToAddress(aliceMultiSigAddrResponse.getAddress(), 5);
+        BitcoindRegtestSetup.mineOneBlock(minerBackend);
+
+        // Create PSBT (send to Alice without Alice)
+        String aliceReceiveAddr = aliceBackend.getNewAddress(AddressType.BECH32, "");
```
chimp1984, line 55: This test runs until this line and then hangs for me (OSX, bitcoind v22.0.0)

chimp1984 (changes requested):

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindBackend.java
alvasw: That's in seconds and the max timeout value for `walletpassphrase` but I like your solution more! (https://developer.bitcoin.org/reference/rpc/walletpassphrase.html) I didn't think much about the timeout (copy & paste from documentation).

wallets/src/main/java/network/misq/wallets/bitcoind/psbt/PsbtOutput.java
alvasw: Making it a record would add getter and expose all variables. This class converts the data to the rpc call format while hiding the details. You create an instance of a `PsbtOutput` register all outputs and pass it to the backend which calls `toPsbtOutputObject()` to get the rpc format.

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindBackend.java
alvasw: No, that's correct. "Specifying the wallet name on a wallet endpoint is invalid." (https://developer.bitcoin.org/reference/rpc/unloadwallet.html)

wallets/src/test/java/network/misq/wallets/bitcoind/BitcoindSigningTests.java
alvasw: No, it won't. This is a Util RPC call to help developers (https://developer.bitcoin.org/reference/rpc/unloadwallet.html).

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindProcess.java
alvasw: I added this but if I use it in the tests they never complete. Starting a new regtest instance before the previous has terminated seems to be a bad idea. We could use different ports but this would start a lot of parallel instances. Do we want this?

wallets/src/test/java/network/misq/wallets/bitcoind/BitcoindSendTests.java
alvasw: Did you had a running regtest instance before starting the tests? I do encounter the same situation if an instance was running before starting the tests. I do not want to kill a running instance. Who knows why the user has bitcoind running.

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindProcess.java
chimp1984: No, I think 1 single instance of bitcoind is better. But maybe it was a bad suggestion anyway. I was just wondering why the shutdown takes so long. I just tested with normal Bitcoin core regtest and at quit it shuts down very fast. So might be some other issue. Also the test I could run seemed to be quite slow. So maybe there is some blocking issues somewhere which also causes my issues running the tests. How long does it take for you to run for instance BitcoindWalletCreationTests? For me after the test is green it takes about 10 sec until the test runner exits.

wallets/src/test/java/network/misq/wallets/bitcoind/BitcoindSendTests.java
chimp1984: No i did not had one running.

wallets/src/main/java/network/misq/wallets/FileUtils.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+
+public class FileUtils {
```
chimp1984: In the common module there is a FileUtils class which can be used (and you can add that new method there).

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindProcess.java
```diff
@@ -0,0 +1,139 @@
...
+    private String getBitcoindParamForNetworkType(NetworkType networkType) {
+        String arg;
+        switch (networkType) {
+            case MAINNET -> arg = "";
+            case REGTEST -> arg = "-regtest";
+            case SIGNET -> arg = "-signet";
+            case TESTNET -> arg = "-testnet";
+            default -> throw new IllegalArgumentException("Invalid Network Type: " + networkType);
+        }
+        return arg;
```
chimp1984: Could also return directly in the cases....

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindProcess.java
alvasw: BitcoindWalletCreationTests runs for 8 seconds but the shutdown takes the most time. Yes, I'm not sure why it takes so long. Hmm I'll try to find out why it takes so long.

wallets/src/main/java/network/misq/wallets/bitcoind/rpc/config/RpcConfigBuilder.java
```diff
@@ -0,0 +1,71 @@
+package network.misq.wallets.bitcoind.rpc.config;
+
+import network.misq.wallets.NetworkType;
+
+import java.util.Optional;
+
+public class RpcConfigBuilder {
```
chimp1984: We could move the RpcConfigBuilder inside RpcConfig and rename it to Builder. So its more contained as a part of RpcConfig.

alvasw:
Can you test it again? I fixed a bug that caused the long shutdown.

wallets/src/main/java/network/misq/wallets/bitcoind/BitcoindProcess.java
```diff
@@ -0,0 +1,114 @@
...
+                "-txindex=1"
+        ).start();
+    }
+
+    private void waitUntilReady() {
+        waitUntilLogFileContainsLine("init message: Done loading");
+    }
+
+    private void waitUntilLogFileContainsLine(String lineToMatch) {
+        try {
```
chimp1984: I think that try can be removed and the catch used for the try below

chimp1984: Actually I think the Scanner API would make the code much more readable

```
 try (Scanner scanner = new Scanner(bitcoindProcess.getInputStream())) {
            while (scanner.hasNextLine()) {
                // The bitcoind log starts with a timestamp, so we check only for expected string with `endsWith`
                if (scanner.nextLine().endsWith(lineToMatch)) {
                    return;
                }
            }
        } catch (Exception e) {
            log.error("Bitcoind didn't start correctly.", e);
            throw e;
        }
```

chimp1984 (approved):
ACK
See small suggestion in the review... but all in all ACK and tests are running for me now.

chimp1984 (approved):
ACK

PR #18, opened 2021-12-25
Demo lightbend / typesafe configuration library

ghubstan:
Based on https://github.com/bisq-network/misq/pull/17

This is a simple example of how Misq might use the [typesafe / lightbend library](https://github.com/lightbend/config) (~295Kb) to load all global and sub-module configurations from a single misq.conf file.

- Dependency constraint added to common-platform's build.gradle

- Dependency (implemention) declared in common and network build files.

- Global misq.conf file added to common module's resource folder (might need to move it).  This sample configuration contains a network service sub config, but only the  "networkIOPool {}" block has real values.

- Added new MisqConfig class to load misq.conf, and provide a getConfig(String path) method to give any object a typesafe Config instance of any scope: from global, to library, down to smallest grouping of scalar configuration values.  

- Added test to check "networkIOPool" config is loaded with correct values.

- Replaced hard-coded NETWORK_IO_POOL parameters with values loaded from misq.conf

TODO 
  
The typesafe library can probably support everying Misq will need in terms of configuration, including the ability to merge distinct dev and production configs (sub-module level), as well as property files and json files into a global config.


chimp1984 (approved):
utACK

PR #27, opened 2022-01-02
Use typesafe Misq.conf in NetworkServiceConfigFactory 

ghubstan:
-  Some `TODO` remarks are added to `NetworkServiceConfigFactory`.
- Config variable names are verbose to disambiguate from non-typesafe Config types.

Based on `main`.


chimp1984 (approved):
utACK

PR #31, opened 2022-01-03
Remove String[] args from network service class constructors

ghubstan:
Lightbend config library can use jvm-opts `-Dmisq.x.y` to override conf file values instead of `String[] programArguments`.


chimp1984 (approved):
utACK

PR #50, opened 2022-01-30
ApplicationOptions: Get Default Locale From LocaleRepository

alvasw:
On some systems the country is not defined.
LocaleRepository.getDefaultLocale() calls Locale.getDefault() first and
handles the special cases afterwards. We should rely on it instead of
calling Locale.getDefault() directly.

### DIFF
--- a/application/src/main/java/bisq/application/ApplicationOptions.java
+++ b/application/src/main/java/bisq/application/ApplicationOptions.java
@@ -17,6 +17,7 @@
 
 package bisq.application;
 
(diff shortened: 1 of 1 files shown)


chimp1984 (approved):
utACK

alkum:
Works :+1:

alvasw:
@alkum What does `Locale.getDefault()` return?

alkum:
Ever since 330e6512d2d6a2e67e491693afabedf69d2cc734 this is not an issue anymore.

PR #51, opened 2022-01-30
BisqTextField: Set Label Float by default

alvasw:
After setting a prompt text, most of our text fields call setLabelFloat(true). Setting label float to true by default could reduce boilerplate code. Ideally, I would like to call setLabelFloat(true) when setPromptText() is called, but sadly the method is final.


alkum (approved):
ACK

PR #58, opened 2022-02-02
Change I2P layer from SAM to streaming API

alkum:
Interact with the I2P network using the I2P streaming API. This exposes methods for managing sockets, abstracting away the session setup, handshake, etc.

Moving away from SAM also means the SAM bridge is not required anymore.


chimp1984 (approved):
utACK

PR #59, opened 2022-02-02
Document run configs

alkum:
Document sample run configs for running the seeds and several desktop nodes locally.

Builds on top of #58


docs/intellij-idea-sample-run-configs.md
```diff
@@ -0,0 +1,94 @@
...
+-Dbisq.networkServiceConfig.supportedTransportTypes.0=CLEAR 
+-Dbisq.networkServiceConfig.seedAddressByTransportType.clear.0=127.0.0.1:8000 
+-Dbisq.networkServiceConfig.seedAddressByTransportType.clear.1=127.0.0.1:8001
+```
+
+
+#### Run Config: `Alice_tor`
+
+Copy the `Alice_clear` run configuration, rename it to `Alice_tor` and change:
+- Program Arguments to `--appName=bisq_Alice_tor`
```
chimp1984, line 75: I added support for appName in the type safe config (using JVM args). I guess its better to not mix progArgs and JVM args but only use one type. Still not 100% happy with that typesafe framework but I guess for now we stick with it. You still can leave it, the code supports both, just wanted to let you know

chimp1984 (approved):
utACK

PR #62, opened 2022-02-03
Complete RoboHash implementation

### LINKED ISSUE #61: Complete RoboHash implementation
Any dev interested to complete the open TODOs in  the RoboHash implementation?

- Use low level image composition not StackPane and ImageView
- Add set2 and set3
- Maybe add backgrounds?

See https://github.com/bisq-network/bisq2/blob/main/desktop/src/main/java/bisq/desktop/robohash/RoboHash.java

xyzmaker123:
Fixes #61

### DIFF
--- a/desktop/src/main/java/bisq/desktop/common/utils/ImageUtil.java
+++ b/desktop/src/main/java/bisq/desktop/common/utils/ImageUtil.java
@@ -17,8 +17,8 @@
 
 package bisq.desktop.common.utils;
 
-import javafx.scene.image.Image;
-import javafx.scene.image.ImageView;
+import javafx.scene.image.*;
+import javafx.scene.paint.Color;
 import javafx.stage.Screen;
 import lombok.extern.slf4j.Slf4j;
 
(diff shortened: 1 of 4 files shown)


chimp1984 (approved):
The image creation does result in more pixelated images. We need to look further how to improve the images. Maybe to render them in original size and then scale it down with smoothing enabled.

We should also get all 3 sets enabled, so that derived from the hash data the set is selected.

I will merge it but might change back to the stackview version as long the render quality is not better.

xyzmaker123:
Thanks - if you'd like me to make some further work regarding pixelated images or other issues just let me know.

chimp1984:
For now thats done. See my latest PR...

xyzmaker123:
Got it. 
We could possibly think about extracting common logic from SetXConfiguration classes into higher level of abstraction.

chimp1984:
I tried out the other sets but I am not convinced that they are good enough quality. At least one set is rather poor image quality and also style. I left set 1 for now. Once we have the designer worked on the whole concept we can see how to move further. Maybe he comes up with his own robo hash set ;-)

xyzmaker123:
Sounds good - we'll see what the future brings :)

PR #63, opened 2022-02-03
Improve I2P log handling

alkum:
Customize and document the i2p log manager.

Also handle an edge case whereby a closed socket was being closed again.

### DIFF
--- a/i2p/src/main/java/bisq/i2p/I2pClient.java
+++ b/i2p/src/main/java/bisq/i2p/I2pClient.java
@@ -19,6 +19,7 @@ package bisq.i2p;
 
 import bisq.common.util.FileUtils;
 import lombok.extern.slf4j.Slf4j;
+import net.i2p.I2PAppContext;
 import net.i2p.I2PException;
 import net.i2p.client.streaming.I2PSocketManager;
 import net.i2p.client.streaming.I2PSocketManagerFactory;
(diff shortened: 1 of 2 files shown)


i2p/src/main/java/bisq/i2p/I2pClient.java
```diff
@@ -59,6 +60,28 @@ public static I2pClient getI2pClient(String dirPath) {
...
+          See https://geti2p.net/spec/configuration -> "Logger (logger.config)"
+
+          Note: A custom config file name and location can be loaded using
+          I2PAppContext.getGlobalContext().logManager().setConfig("desktop/src/main/resources/bisq.properties");
+          using a path relative to the current working directory.
+
+          2) Using the exposed setters on the log manager:
+          I2PAppContext.getGlobalContext().logManager().set*
+         */
+        I2PAppContext.getGlobalContext().logManager().setBaseLogfilename("logs/i2p-@.log");
```
chimp1984, line 82: Whats the root here? The project root? I think we should use the data directory which is only available at startup.
The I2PTransport could do that as it gets passed the data dir.

chimp1984 (approved):
We should continue to look how to get the logs into our data directory and not in the project root.

i2p/src/main/java/bisq/i2p/I2pClient.java
alkum: Good point, using data dir with #66

PR #66, opened 2022-02-04
Store i2p logs in client data-dir

alkum:
Configure i2p to store its logs in the i2p client's data directory.


chimp1984 (approved):
utACK

chimp1984 (approved):
utACK

PR #67, opened 2022-02-04
Autoclose file inputstream when creating i2p socket manager

alkum:
Ensure inputstream is closed in case of an IO error.


chimp1984 (approved):
utACK

PR #73, opened 2022-02-06
Make RpcCallFailureException a RuntimeException

### LINKED ISSUE #72: Make RpcCallFailureException a RuntimeException
The `RpcClient` will be called later in `CompletableFuture`s. A checked exception forces us to explicitly handle the exception in the `WalletService`, making the code more complex. A `RuntimeException` can be handled by the caller using `CompletableFuture.handle(...)` or a related method. This allows the `WalletService` consumer to decide how to respond to a failed call. Furthermore, the `WalletService` is not smart enough to handle the exception because it doesn't know the context in which it was triggered.

alvasw:
The RpcClient will be called later in CompletableFutures. A checked
exception forces us to explicitly handle the exception in the
WalletService, making the code more complex. A RuntimeException can be
handled by the caller using CompletableFuture.handle(...) or a related
method. This allows the WalletService consumer to decide how to respond
to a failed call. Furthermore, the WalletService is not smart enough to
handle the exception because it doesn't know the context in which it
was triggered.

Fixes #72


chimp1984 (approved):
utACK

PR #74, opened 2022-02-06
Fix: Prompt text cutoff in BisqTextField

alvasw:
When the user enters text into a text field, the prompt text floats
above the text field. There is no padding by default which takes this
into account. This results in the prompt text clashing with the control
above.


chimp1984 (approved):
utACK

PR #81, opened 2022-02-07
Create WalletService (1/2)

alvasw:
The `WalletService` is the public API that other modules call. It exposes all wallet functionalities. All methods of the `WalletService` are asynchronous and return a `CompletableFuture`.


chimp1984 (approved):
utACK

PR #85, opened 2022-02-08
Enforce socket timeout in I2P client

alkum:
Pass on the optional timeout argument when initializing the i2p sockets.


chimp1984 (approved):
utACK

PR #92, opened 2022-02-09
Check roles using mempool tx lookup.

### LINKED ISSUE #68: Add Entitlement verifications
For the social trading features we use entitlements.
An entitlement is the right to execute certain roles.

Here the enum entries and descriptions (provided numbers are TBD):
```
LIQUIDITY_PROVIDER=Liquidity provider as BTC Seller 
  Requirement: Burn at least 50 BSQ/DAO cycle for earning reputation
CHANNEL_ADMIN=Channel administrator 
  Requirement: Bonded DAO role as channel administrator
CHANNEL_MODERATOR=Channel moderator 
  Requirement: Invitation from the channel owner and burn 100 BSQ/DAO cycle. 
  Entitled to make a compensation request for 500 BSQ/DAO cycle
MEDIATOR=Mediator 
  Requirement: Bonded DAO role as mediator
``` 

There are 2 types for providing proofs: Burned BSQ and bonded roles.

For adding an entitled role to a user profile one need to fulfill the requirements of the role.

Process for burned BSQ is:

User starts to create new user profile. They get displayed the pubkey hash when they have choosen a robo hash icon.
Next they copy that pubkey hash, open Bisq1 DAO section at proof of burn and paste it as pre-image for the proof of burn tx.
They need to burn at least the required amount. 
They publish the tx wait until the tx is confirmed.
After that they copy the tx ID and move back to Bisq 2 to paste it into the `proof` input text field and click the `Verify` button.
This triggers a request to BSQ explorer(s) and returns the tx data. If its a valid BSQ tx and the opReturn data matches the pubkey hash everything is correct and the txId gets attachted to the entitlement data and the entitlement role added to the user profile.
The user could add multiple different entitled roles.

After having created the user profile the entitled role is part of the profile and can be verified by other users.

Verification of proof of burn based entitlement:

Request the tx ID taken from the peers public user profile data from a BSQ explorer.
Check the opReturn data if it matches the pubkey hash. If so the linkage of the pub key to the burned BSQ is correct.
Any interaction with the peer is based on the key pair from which the pub key is derived, so any relevant interaction will only work if the pubkey was not fake. Such interactions are private signed and encrypted messages and data added to the P2P network like offers or chat messages.  It must not be possible to use a foreign pubkey hash and publish data to other users.

For entitlements based  on bonded roles the process is as follows:

Before creating a new user profile the user need to do a DAO request for a bonded role. We will likely re-use existing roles like Forum admin for the channel admin role or the current mediator role for mediators in the social trading area (there can be overlap of mediators who cover both areas).
We need to add a signing and verification tool for the bonded roles similar as we provide it for proof of burn.
With that tool the bonded role owner can sign their Bisq 2 pubkey hash and use the signature as provided proof next to the tx ID of the bond request tx. 

For verifying the user can request the tx data from a BSQ explorer, take the EC pub key from the first BSQ input and verify the provided signature with that pub key hash of the user profile.

Once the DAO is integrated into Bisq 2 those explorer requests are not required anymore.

Here a screenshot of the WIP screen:

Screenshot 2022-02-05 at 20 52 43

jmacxx:
Closes #68

### DIFF
--- a/social/src/main/java/bisq/social/userprofile/BsqTxValidator.java
+++ b/social/src/main/java/bisq/social/userprofile/BsqTxValidator.java
@@ -27,6 +27,8 @@ import com.google.gson.JsonSyntaxException;
 
 import lombok.extern.slf4j.Slf4j;
 
+import java.util.Optional;
+
 @Slf4j
 public class BsqTxValidator {
 
@@ -68,21 +70,29 @@ public class BsqTxValidator {
         return txType.getAsString().equalsIgnoreCase("LOCKUP");
     }
 
-    public static byte[] getOpReturnData(String jsonTxt) {
+    public static long getBurntAmount(String jsonTxt) {
+        JsonObject json = new Gson().fromJson(jsonTxt, JsonObject.class);
+        JsonElement burntFee = json.get("burntFee");
+        if (burntFee == null) {
+            return 0;   // no json element, assume zero burnt amount
+        }
+        return burntFee.getAsLong();
+    }
+
+    public static Optional<String> getOpReturnData(String jsonTxt) {
         try {
             Pair<JsonArray, JsonArray> vinAndVout = getVinAndVout(jsonTxt);
             JsonArray voutArray = vinAndVout.second();
             for (JsonElement x : voutArray) {
                 JsonObject y = x.getAsJsonObject();
                 if (y.get("txOutputType").getAsString().matches(".*OP_RETURN.*")) {
-                    String opReturnHexString = y.get("opReturn").getAsString();
-                    return Hex.decode(opReturnHexString);
+                    return Optional.of(y.get("opReturn").getAsString());
                 }
             }
         } catch (JsonSyntaxException e) {
             log.error("json error:", e);
         }
-        return null;
+        return Optional.empty();
     }
 
     private static Pair<JsonArray, JsonArray> getVinAndVout(String jsonTxt) throws JsonSyntaxException {
--- a/social/src/main/java/bisq/social/userprofile/BsqTxValidator.java
+++ b/social/src/main/java/bisq/social/userprofile/BsqTxValidator.java
@@ -1,0 +1,100 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.social.userprofile;
+
+import bisq.common.data.Pair;
+import bisq.common.encoding.Hex;
+import com.google.gson.Gson;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonSyntaxException;
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class BsqTxValidator {
+
+    public static boolean initialSanityChecks(String txId, String jsonTxt) {
+        if (jsonTxt == null || jsonTxt.length() == 0) {
+            return false;
+        }
+        JsonObject json = new Gson().fromJson(jsonTxt, JsonObject.class);
+        // there should always be "id" string element at the top level
+        if (json.get("id") == null) {
+            return false;
+        }
+        // txid should match what we requested
+        if (!txId.equals(json.get("id").getAsString())) {
+            return false;
+        }
+        return true;
+    }
+
+    public static boolean isBsqTx(String url, String txId, String jsonTxt) {
+        return url.matches(".*bisq.*") && initialSanityChecks(txId, jsonTxt);
+    }
+
+    public static boolean isProofOfBurn(String jsonTxt) {
+        JsonObject json = new Gson().fromJson(jsonTxt, JsonObject.class);
+        JsonElement txType = json.get("txType");
+        if (txType == null) {
+            return false;
+        }
+        return txType.getAsString().equalsIgnoreCase("PROOF_OF_BURN");
+    }
+
+    public static boolean isLockup(String jsonTxt) {
+        JsonObject json = new Gson().fromJson(jsonTxt, JsonObject.class);
+        JsonElement txType = json.get("txType");
+        if (txType == null) {
+            return false;
+        }
+        return txType.getAsString().equalsIgnoreCase("LOCKUP");
+    }
+
+    public static byte[] getOpReturnData(String jsonTxt) {
+        try {
+            Pair<JsonArray, JsonArray> vinAndVout = getVinAndVout(jsonTxt);
+            JsonArray voutArray = vinAndVout.second();
+            for (JsonElement x : voutArray) {
+                JsonObject y = x.getAsJsonObject();
+                if (y.get("txOutputType").getAsString().matches(".*OP_RETURN.*")) {
+                    String opReturnHexString = y.get("opReturn").getAsString();
+                    return Hex.decode(opReturnHexString);
+                }
+            }
+        } catch (JsonSyntaxException e) {
+            log.error("json error:", e);
+        }
+        return null;
+    }
+
+    private static Pair<JsonArray, JsonArray> getVinAndVout(String jsonTxt) throws JsonSyntaxException {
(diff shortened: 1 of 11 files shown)


social/src/main/java/bisq/social/userprofile/BsqTxValidator.java
```diff
@@ -0,0 +1,100 @@
...
+                JsonObject y = x.getAsJsonObject();
+                if (y.get("txOutputType").getAsString().matches(".*OP_RETURN.*")) {
+                    String opReturnHexString = y.get("opReturn").getAsString();
+                    return Hex.decode(opReturnHexString);
+                }
+            }
+        } catch (JsonSyntaxException e) {
+            log.error("json error:", e);
+        }
+        return null;
```
chimp1984: We try to avoid null as far as possible in Bisq 2. Better propagate the exception if the exception is not an expected case or return Optional.

social/src/main/java/bisq/social/userprofile/BtcTxValidator.java
```diff
@@ -0,0 +1,82 @@
...
+        return true;
+    }
+
+    public static String getFirstInputPubkey(String jsonTxt) {
+        try {
+            Pair<JsonArray, JsonArray> vinAndVout = getVinAndVout(jsonTxt);
+            JsonArray vinArray = vinAndVout.first();
+            for (JsonElement x : vinArray) {
+                JsonObject vin = x.getAsJsonObject();
+                JsonArray witnesses = vin.getAsJsonArray("witness");
```
chimp1984, line 60: Can it be a non-segwit tx as well?

social/src/main/java/bisq/social/userprofile/BtcTxValidator.java
```diff
@@ -0,0 +1,82 @@
...
+                    return pubKey;
+                }
+            }
+        } catch (JsonSyntaxException e) {
+            log.error("json error:", e);
+        }
+        throw new JsonSyntaxException("could not find pubkey");
+    }
+
+    private static Pair<JsonArray, JsonArray> getVinAndVout(String jsonTxt) throws JsonSyntaxException {
```
chimp1984, line 81: Maybe extract duplicated methods (also initialSanityChecks)

social/src/main/java/bisq/social/userprofile/UserProfileService.java
```diff
@@ -126,53 +135,63 @@ public void selectUserProfile(UserProfile value) {
...
-                } else {
+                BaseHttpClient httpClient = getBsqApiHttpClient();
+                String jsonTxLookup = httpClient.get("tx/" + proofOfBurnTxId, Optional.of(new Pair<>("User-Agent", httpClient.userAgent)));
+                if (!BsqTxValidator.initialSanityChecks(proofOfBurnTxId, jsonTxLookup)) {
                     return Optional.empty();
                 }
+                boolean isBsqTx = httpClient.getBaseUrl().matches(".*//bisq.*");
+                boolean isProofOfBurn = BsqTxValidator.isProofOfBurn(jsonTxLookup);
+                byte[] publicKeyHash = Hex.decode(pubKeyHash);
+                byte[] opReturn = BsqTxValidator.getOpReturnData(jsonTxLookup);
```
chimp1984: As opReturn is hex encoded we could use string comparison instead converting both to bytes

social/src/main/java/bisq/social/userprofile/UserProfileService.java
```diff
@@ -126,53 +135,63 @@ public void selectUserProfile(UserProfile value) {
...
+                boolean opReturnMatches = Arrays.equals(publicKeyHash, opReturn);
+                if (isBsqTx && isProofOfBurn && opReturnMatches) {
```
chimp1984: We need to check also if burned amount > required. I think I added that to a more recent commit.

social/src/main/java/bisq/social/userprofile/UserProfileService.java
```diff
@@ -126,53 +135,63 @@ public void selectUserProfile(UserProfile value) {
...
-                long minBurnAmount = getMinBurnAmount(type); //todo verify if tx burned amount is >= minBurnAmount
-                if (isBsqTx && isProofOfBurn && pubKeyHash.equals(opReturn)) {
-                    return Optional.of(new Entitlement.ProofOfBurnProof(proofOfBurnTxId));
-                } else {
+                BaseHttpClient httpClient = getBsqApiHttpClient();
+                String jsonTxLookup = httpClient.get("tx/" + proofOfBurnTxId, Optional.of(new Pair<>("User-Agent", httpClient.userAgent)));
+                if (!BsqTxValidator.initialSanityChecks(proofOfBurnTxId, jsonTxLookup)) {
                     return Optional.empty();
                 }
+                boolean isBsqTx = httpClient.getBaseUrl().matches(".*//bisq.*");
```
chimp1984: We control the url used for the request anyway, so i think that check is not adding much. Though maybe its good to keep it to have it more explicit and later when using DAO code it will gets its proper checks.

social/src/main/java/bisq/social/userprofile/UserProfileService.java
```diff
@@ -126,53 +135,63 @@ public void selectUserProfile(UserProfile value) {
...
+                BaseHttpClient httpClientBsq = getBsqApiHttpClient();
+                BaseHttpClient httpClientBtc = getBtcApiHttpClient();
+                String jsonBsqTx = httpClientBsq.get("tx/" + txId, Optional.of(new Pair<>("User-Agent", httpClientBsq.userAgent)));
+                String jsonBtcTx = httpClientBtc.get("tx/" + txId, Optional.of(new Pair<>("User-Agent", httpClientBtc.userAgent)));
+                checkArgument(BsqTxValidator.initialSanityChecks(txId, jsonBsqTx), "bsq tx sanity checks");
+                checkArgument(BtcTxValidator.initialSanityChecks(txId, jsonBtcTx), "btc tx sanity checks");
+                checkArgument(BsqTxValidator.isBsqTx(httpClientBsq.getBaseUrl(), txId, jsonBsqTx), "isBsqTx");
+                checkArgument(BsqTxValidator.isLockup(jsonBsqTx), "is lockup transaction");
+                String identityKeyHash = Hex.encode(DigestUtil.hash(identityPublicKey.getEncoded()));
+                String signingPubkeyTxt = BtcTxValidator.getFirstInputPubkey(jsonBtcTx);
```
chimp1984: Is the TxT at signingPubkeyTxt meaning text? Maybe better signingPubkeyAsHex?

social/src/main/java/bisq/social/userprofile/UserProfileService.java
```diff
@@ -126,53 +135,63 @@ public void selectUserProfile(UserProfile value) {
...
+                PublicKey signingPubKey = KeyGeneration.generatePublicFromCompressed(Hex.decode(signingPubkeyTxt));
+                boolean signatureMatches = SignatureUtil.verify(Hex.decode(identityKeyHash), Base64.decode(signature), signingPubKey);
```
chimp1984: We can extract the DigestUtil.hash(identityPublicKey.getEncoded()) to a var to be reused

chimp1984: Do we use Base64 for the signing tools in Bisq1?

social/src/main/java/bisq/social/userprofile/UserProfileService.java
```diff
@@ -197,4 +216,49 @@ public long getMinBurnAmount(Entitlement.Type type) {
...
+        } else if (supportedTransportTypes.contains(Transport.Type.TOR)) {
+            transportType = Transport.Type.TOR;
+        } else {
+            throw new RuntimeException("I2P is not supported yet");
+        }
+        BaseHttpClient httpClient = networkService.getHttpClient(url, userAgent, transportType);
+        return httpClient;
+    }
+
+    private BaseHttpClient getBtcApiHttpClient() {
```
chimp1984: As only providers are different we can use a common method for both.

social/src/main/java/bisq/social/userprofile/UserProfileService.java
```diff
@@ -197,4 +216,49 @@ public long getMinBurnAmount(Entitlement.Type type) {
...
+        String userAgent = "Bisq 2";
+        //todo add providers to Bisq.config
+        List<String> providers = List.of(   // note the BSQ urls are different than the BTC urls
```
chimp1984: We should move the providers to Bisq.config in application modules resource dir and add a inner static class Config to UserProfileService where we take those 2 provider lists. See DefaultApplicationService for other examples how its used. Typesafe is used for the config handling, will require prob. the dependency to get added to that module.

chimp1984 (changes requested):
Contentwise it looks all good! Some smaller nits...

chimp1984 (approved):
utACK

jmacxx:
After 2 days fighting `KeyGeneration.java` and `SignatureUtil.java` I cannot get Bisq2 to successfully verify Bisq1 (bitcoin) signatures.  Leaving it as is for now.

chimp1984 (approved):
utACK

chimp1984:
No worry I will look into it later...

PR #94, opened 2022-02-11
Fix Gradle Run Task

alvasw:
We cannot start the desktop app using Gradle. This change fixes this.

Changes:
- Add JVM arguments for jfoenix
- Load resources using `ClassLoader`


chimp1984 (approved):
utACK

PR #95, opened 2022-02-13
Add support for gradle args when starting desktop client

alkum:
Make sure to propagate system properties to the started desktop application. Add documentation indicating how to do this.


chimp1984 (approved):
utACK

PR #97, opened 2022-02-14
Distinguish between wallet not running and wrong rpc credentials

alvasw:
At the moment, it's not easy for the UI to check whether the wallet RPC call failed because the external wallet was not running or the credentials were wrong.


chimp1984 (approved):
utACK

PR #98, opened 2022-02-14
I2P b32 address support

alkum:
Add support for b32.i2p address formats.


chimp1984 (commented):
When using seed node addresses in b32.i2p format the client need to do the network lookup, right? So that will slow down the initial connection. 
I think its great to add the support, but for seed nodes I think the trade off is not worth it as the user do not deal anyway with the addresses (comes from config). But for displaying addresses we should definitely use the shorter versions.

PR #99, opened 2022-02-15
Modularize Common Module (Java 9 Module)

alvasw:
Add proper Java 9 Module support.

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -16,6 +16,7 @@ dependencies {
 
     implementation 'com.google.guava:guava'
     implementation 'com.typesafe:config'
+    implementation 'org.jetbrains:annotations:23.0.0'
 }
 
 test {
--- a/common/src/main/java/bisq/common/annotations/LateInit.java
+++ b/common/src/main/java/bisq/common/annotations/LateInit.java
@@ -15,9 +15,8 @@
  */
 package bisq.common.annotations;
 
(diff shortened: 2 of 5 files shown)


chimp1984 (commented):
Can you provide information whats the difference between the 2 Nullable packages?

alvasw:
There is not really a difference between both packages. At first, the `@Nullable` and `@NonNull`/`@NotNull` annotations were part of multiple individual libraries, e.g., FindBugs. Sun, Google, JetBrains, and a couple others wanted to standardize it in JSR 305 (Annotations for Software Defect Detection) [1]. `com.google.code.findbugs:jsr305` was the reference implementation [2].
However, the JSR's status has been dormant for many years. The library is usually loaded on the classpath, and Java 9 modules cannot require anything from the unnamed module. In the meantime, Findbugs was abandoned, and a project called SpotBugs [3] is recommended now.
I didn't want to add a dependency to SpotBugs because we never applied static analysis to look for bugs. We only use the annotations to give the IDE hints. For this purpose, JetBrain's annotation library is sufficient.

[1] https://jcp.org/en/jsr/detail?id=305
[2] https://mvnrepository.com/artifact/com.google.code.findbugs/jsr305
[3] https://spotbugs.github.io/

chimp1984:
Thanks for giving the background info.

PR #102, opened 2022-02-17
Add :seed:run gradle task

alkum:
Add support for and document usage of :seed:run task.

### DIFF
--- a/docs/sample-run-configs.md
+++ b/docs/sample-run-configs.md
@@ -103,15 +103,52 @@ Copy the `Alice_clear` run configuration, rename it to `Alice_i2p` and change:
 
 ### Command line: Gradle run configs
 
-Start a default desktop client with:
+Start a seed with:
 
 ```
+# Using default settings
+./gradlew seed:run
+```
+
+For example, to start two local seeds, `bisq2_seed1` and `bisq2_seed2`, reachable on all 3 networks:
+
+```
+# Seed 1
+./gradlew seed:run \
+    -Dbisq.application.appName=bisq2_seed1 \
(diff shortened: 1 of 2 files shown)


docs/sample-run-configs.md
```diff
@@ -103,15 +103,52 @@ Copy the `Alice_clear` run configuration, rename it to `Alice_i2p` and change:
...
 ```
+# Using default settings
+./gradlew seed:run
+```
+
+For example, to start two local seeds, `bisq2_seed1` and `bisq2_seed2`, reachable on all 3 networks:
+
+```
+# Seed 1
+./gradlew seed:run \
```
chimp1984, line 117: Maybe its better to either limit supportedTransportTypes to clearnet as tor and i2p require local manual setup for the local seed addresses.
Once we use protobuf for serialisation we can start those seeds on a server so its easier to run but currently with java serialisation it carries a slight security risk.

chimp1984: So I would recommend to remove 
```
-Dbisq.networkServiceConfig.supportedTransportTypes.0=TOR \
-Dbisq.networkServiceConfig.supportedTransportTypes.1=I2P \
```

chimp1984: Actually not sure if typesafe works with `supportedTransportTypes.2=CLEAR` alone or if index need to be 0 in that case. Propably more clear with 0.

chimp1984 (changes requested):

docs/sample-run-configs.md
alkum: Yes good point, I removed tor and i2p, kept only the args for clearnet.

PR #103, opened 2022-02-18
Add simple framework to start local seeds / clients

alkum:
Combine make and screen to allow the launch of different combinations of seeds and clients. Include a scenario with two local seeds and n clients.

Depends on the `:seed:run` gradle task in #102.


chimp1984 (approved):
utACK

PR #105, opened 2022-02-19
Add local I2P test scenario

alkum:
Extend makefile with a target that launches a local I2P test environment (2 seeds and N clients).

Builds on top of #104


chimp1984 (approved):
utACK

PR #107, opened 2022-02-20
fix rename bug on WinOS

salokiam:
fix for WinOS, renaming a file must be after closing that file.

### DIFF
--- a/persistence/src/main/java/bisq/persistence/Persistence.java
+++ b/persistence/src/main/java/bisq/persistence/Persistence.java
@@ -101,30 +101,38 @@ public class Persistence<T extends Serializable> {
                 File tempFile = File.createTempFile("temp_" + fileName, null, new File(directory));
                 FileUtils.deleteOnExit(tempFile);
                 File storageFile = new File(storagePath);
+                boolean tempSuccess = true;
                 try (FileOutputStream fileOutputStream = new FileOutputStream(tempFile);
                      ObjectOutputStream objectOutputStream = new ObjectOutputStream(fileOutputStream)) {
                     objectOutputStream.writeObject(serializable);
                     objectOutputStream.flush();
                     fileOutputStream.flush();
                     fileOutputStream.getFD().sync();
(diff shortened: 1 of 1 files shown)


persistence/src/main/java/bisq/persistence/Persistence.java
```diff
@@ -101,30 +101,38 @@ public boolean persist(T serializable) {
...
                      ObjectOutputStream objectOutputStream = new ObjectOutputStream(fileOutputStream)) {
                     objectOutputStream.writeObject(serializable);
                     objectOutputStream.flush();
                     fileOutputStream.flush();
                     fileOutputStream.getFD().sync();
+                    // can't set tempSuccess to true here, because the close() is coming implicitly after closing brackets.
+                } catch(IOException ex) {
+                    log.error("Error at read for " + storagePath+ " msg: "+ex.getMessage(), ex);
+                    tempSuccess = false;
+                }// need to end the catch here to have the temp file autoclose
```
chimp1984: I assume the tempFile gets closed because its used in `new FileOutputStream(tempFile)`. Maybe we should make it more explicit, for instance to close it in a finally clause or add a bit more info to the comment.

chimp1984: Just checked the code in Bisq (1) at PersistenceManager.
```
// Close resources before replacing file with temp file because otherwise it causes problems on windows
// when rename temp file
fileOutputStream.close();
```

chimp1984 (changes requested):

persistence/src/main/java/bisq/persistence/Persistence.java
salokiam: hmm, yes. Usually a good idea to use try-resource, but in this case it does make the code less readable.
I think its better handled in busq1 I will change it.

chimp1984: Yes, agree.

chimp1984 (approved):
utACK

PR #108, opened 2022-02-22
corrected name for root bundle

salokiam:
According to bisq.i18n.Res#initialize, there should be no 'en' properties file, instead a 'ROOT' Locale which results in default.prroperties without '_en'. That's the way its done with displayStrings.properties in bisq1.


chimp1984 (approved):
utACK

Issue #115, opened 2022-03-15
Error handling failed when Bitcoin core has an unencrypted wallet, but wallet password is passed


ripcurlx:
When trying to connect to Bitcoin core passing a passphrase and the wallet is unencrypted it fails with
 
```
Caused by: java.lang.NullPointerException: Cannot invoke "java.lang.Throwable.toString()" because the return value of "java.lang.Throwable.getCause()" is null
Caused by: java.lang.NullPointerException: Cannot invoke "java.lang.Throwable.toString()" because the return value of "java.lang.Throwable.getCause()" is null

	at bisq.wallets.bitcoind.rpc.RpcClient.rpcAuthenticationFailed(RpcClient.java:77)
	at bisq.wallets.bitcoind.rpc.RpcClient.invoke(RpcClient.java:54)
	at bisq.wallets.bitcoind.BitcoindWalletBackend.walletPassphrase(BitcoindWalletBackend.java:141)
	at bisq.wallets.bitcoind.BitcoindWallet.initialize(BitcoindWallet.java:60)
	at bisq.wallets.WalletService.lambda$initialize$0(WalletService.java:47)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1800)
```

Because `Error: running with an unencrypted wallet, but walletpassphrase was called.` is returned and 
```
private boolean rpcAuthenticationFailed(Throwable t) {
        return t.getCause().toString().contains("401");
    }
```
doesn't handle the case properly when no cause exists.

ripcurlx:
The weird thing is that I actually have protected my wallet in Bitcoin core with a passphrase.

ripcurlx:
> Bitcoin Core supports multiple wallets (you can load and use many simultaneously). In Bisq 2, we create a new wallet. We don't touch any existing/loaded wallets.

Ah - just recognized the newly created wallet. 👍

alvasw:
Fixed in #121

PR #121, opened 2022-03-18
Wallet: Support not encrypted wallets (#115)

alvasw:
At the moment, the wallet module raises a NullPointerException if the
wallet is not encrypted. This shouldn't happen.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/primary/main/content/wallet/config/WalletConfigPopup.java
+++ b/desktop/src/main/java/bisq/desktop/primary/main/content/wallet/config/WalletConfigPopup.java
@@ -85,7 +85,9 @@ public class WalletConfigPopup extends Popup {
         }
 
         private void onConnectToWallet() {
-            String passphrase = model.walletPassphraseProperty.get();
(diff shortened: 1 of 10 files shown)


desktop/src/main/java/bisq/desktop/primary/main/content/wallet/config/WalletConfigPopup.java
```diff
@@ -85,7 +85,9 @@ public void onViewDetached() {
         }
 
         private void onConnectToWallet() {
-            String passphrase = model.walletPassphraseProperty.get();
+            String passphraseValue = model.walletPassphraseProperty.get();
+            Optional<String> passphrase = passphraseValue.isEmpty() ? Optional.empty()
+                    : Optional.of(passphraseValue);
             model.walletPassphraseProperty.setValue(""); // Wipe passphrase from memory
```
chimp1984, line 91: Actually to avoid that the passphrase sticks in memory (until GC cleans up) we should use a char array instead of a String and clear the array afterwards. But no need to do it now, just for hardening once we get more production ready.

chimp1984 (approved):
utACK

PR #142, opened 2022-03-22
Update params to support embedded I2P router

alkum:
Prepare codebase for the use of embedded routers.


i2p/src/main/java/bisq/i2p/I2pClient.java
```diff
@@ -177,6 +190,75 @@ private I2PSocketManager maybeCreateClientSession(String sessionId) {
         return sessionMap.get(sessionId);
     }
 
+    private Properties getPropertiesForEmbeddedRouter() throws IOException {
+        Properties p = new Properties();
```
chimp1984, line 194: I would prefer to use "properties" instead of one char variable names. Improves code readability....

chimp1984 (approved):
utACK

PR #143, opened 2022-03-22
use UserNameGenerator for use name in chat CreateUserProfile

NodinYakari:


desktop/src/main/java/bisq/desktop/primary/main/content/social/profile/components/CreateUserProfile.java
```diff
@@ -174,10 +168,12 @@ private View(Model model, Controller controller, Pane entitlementSelection) {
             userNameInputField = new BisqTextField();
             double minWidth = 300;
             userNameInputField.setMinWidth(minWidth);
+            userNameInputField.setEditable(false);
```
chimp1984, line 171: We could use a simple Label now... but no worry, can be changed later.

chimp1984 (approved):
utACK

PR #144, opened 2022-03-22
Added first pass of open / create Private Channel

salokiam:
unfinished work on opening a private channel.
Synced with main today 19:00 CET


salokiam:
instead of merging, I will send a new pr

PR #158, opened 2022-03-24
Update list of Active Connections when peer lost

alkum:
Correctly update the list of active nodes when a peer gets disconnected.


chimp1984 (approved):
utACK

PR #160, opened 2022-03-25
Allow starting of clearnet seeds in makefile tests

alkum:
To simplify testing, allow starting only the clearnet seeds.


chimp1984 (approved):
utACK

PR #161, opened 2022-03-25
Add scrollbars to createofferview

alkum:
Embed content in a VBox, and that in a ScrollPane, in order to show scrollbars when the entire view doesn't fit in the window.


chimp1984 (approved):
utACK

PR #163, opened 2022-03-26
Persist to disk persistedPeers in PeerGroup

alkum:
Add persistence store and handling for persistedPeers.

### DIFF
--- a/network/src/main/java/bisq/network/p2p/services/peergroup/PeerGroup.java
+++ b/network/src/main/java/bisq/network/p2p/services/peergroup/PeerGroup.java
@@ -19,13 +19,13 @@ package bisq.network.p2p.services.peergroup;
 
 import bisq.common.util.MathUtils;
 import bisq.network.p2p.node.*;
+import bisq.persistence.Persistence;
+import bisq.persistence.PersistenceClient;
+import bisq.persistence.PersistenceService;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
-import java.util.Collection;
-import java.util.Comparator;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.stream.Stream;
 
@@ -33,7 +33,7 @@ import java.util.stream.Stream;
  * Maintains different collections of peers and connections
  */
 @Slf4j
-public class PeerGroup {
+public class PeerGroup implements PersistenceClient<PersistedPeersStore>  {
 
     @Getter
     public static class Config {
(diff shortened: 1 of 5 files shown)


alkum:
Marked as draft until proto refactoring is in.

network/src/main/java/bisq/network/p2p/services/peergroup/PeerGroup.java
```diff
@@ -19,21 +19,21 @@
...
+import java.util.*;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.stream.Stream;
 
 /**
  * Maintains different collections of peers and connections
  */
 @Slf4j
-public class PeerGroup {
+public class PeerGroup implements PersistenceClient<PersistedPeersStore>  {
```
chimp1984, line 36: I think the PeerGroupService should be responsible for persistence. PeerGroup is more of a model class.

network/src/main/java/bisq/network/p2p/services/peergroup/PeerGroup.java
```diff
@@ -85,14 +87,29 @@ public PeerGroup(Node node, Config config, List<Address> seedNodeAddresses, BanL
 
     public void addReportedPeers(Set<Peer> peers) {
         reportedPeers.addAll(peers);
+        addPersistedPeers(peers); // New reported peers are also persisted
```
chimp1984, line 90: I guess we need a bit of refactoring here so that the caller is using the service for updating the model and the service is calling persist() after the PeerGroup model has been changed.

chimp1984: Alternatively we could redefine PeerGroup as sub-servive and PeerGroupService is more the coordination/managing parent service and a new model will be created for the PeerGroup data. I will have a look to the code to get a better idea, whats the best option here.

network/src/main/java/bisq/network/p2p/services/peergroup/PersistedPeersStore.java
```diff
@@ -0,0 +1,39 @@
+package bisq.network.p2p.services.peergroup;
+
+import bisq.persistence.PersistableStore;
+import lombok.Getter;
+
+import java.util.Collection;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+
+public class PersistedPeersStore implements PersistableStore<PersistedPeersStore> {
```
chimp1984, line 10: This will require the protobuf implementation. I will post soon a readme about all, then it should be easier to implement it, otherwise just look at some other store like KeyPairStore to see the patterns used.

chimp1984 (changes requested):
PeerGroup will likely require some refactoring. Protobuf support can be added now.

PR #164, opened 2022-03-26
Minor optimization to RoboHash template calculation

alkum:
Get rid of 1 string replacement per template lookup.


alkum:
Marked as draft until proto change is in.

chimp1984 (approved):
utACK

PR #167, opened 2022-03-27
Add init status on splash screen

alkum:
Add status line which indicates which service or component is currently being initialized.

### DIFF
--- a/application/src/main/java/bisq/application/DefaultApplicationService.java
+++ b/application/src/main/java/bisq/application/DefaultApplicationService.java
@@ -93,6 +93,11 @@ public class DefaultApplicationService extends ServiceProvider {
     private final DaoBridgeService daoBridgeService;
     private final AccountAgeWitnessService accountAgeWitnessService;
 
+    /**
+     * Initialization status, updated as each required service bootstraps
+     */
+    private final Observable<String> initStatus = new Observable<>();
+
     public DefaultApplicationService(String[] args) {
         super("Bisq");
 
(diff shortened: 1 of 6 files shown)


alkum:
Builds on https://github.com/bisq-network/bisq2/pull/173

chimp1984 (approved):
utACK

application/src/main/java/bisq/application/DefaultApplicationService.java
```diff
@@ -96,6 +97,11 @@
     private final SocialService socialService;
     private final SecurityService securityService;
 
+    /**
+     * Initialization status, updated as each required service bootstraps
+     */
+    private final Observable<String> initStatus = new Observable<>();
```
chimp1984: I guess we will need some more finegrained state info. But I think that should be preovided by the domain services directly, so to have only one state here seems ok.

application/src/main/java/bisq/application/DefaultApplicationService.java
```diff
@@ -152,19 +158,32 @@ public DefaultApplicationService(String[] args) {
...
      * Initializes all domain objects, services and repositories.
      * We do in parallel as far as possible. If there are dependencies we chain those as sequence.
      */
     @Override
     public CompletableFuture<Boolean> initialize() {
-        return securityService.initialize()
+        return
+                securityService.initialize()
+
+                .thenCompose(result -> setInitStatus("Bootstrapping network..."))
```
chimp1984: Ah you set the different states here...
The I think we should use an enum inside a Observable instead to keep the core domain out from presentation concerns. The View can map then the enum entries to display strings via `Res.get(state.name())` and we use by convention the enum name as key in the resources.

alkum: Done, now tracking the state via an enum.

application/src/main/java/bisq/application/DefaultApplicationService.java
alkum: Depending on how granular we want to go, we can add a 2nd line with a more detailed substatus message and icons.

For now the main status is added as an enum.

application/src/main/java/bisq/application/DefaultApplicationService.java
```diff
@@ -271,4 +274,30 @@ private boolean isRegtestRun() {
                 .build();
         return Optional.of(walletConfig);
     }
+
+    private CompletableFuture<Boolean> setState(State newState) {
+        checkArgument(state.get().ordinal() < newState.ordinal(),
+                "New state %s must have a higher ordinal as the current state %s", newState, state.get());
+        state.set(newState);
+        log.info("New state {}", newState);
+        runAsync(() -> listeners.forEach(e -> e.onStateChanged(newState)), DISPATCHER);
```
chimp1984, line 283: I think wrapping the state in an Observeable would be enough here.

application/src/main/java/bisq/application/DefaultApplicationService.java
```diff
@@ -271,4 +274,30 @@ private boolean isRegtestRun() {
...
+
+        return CompletableFuture.completedFuture(true);
+    }
+
+    private CompletableFuture<Boolean> setStateAfter(CompletableFuture<Boolean> initTask, State stateAfter) {
+        return initTask.thenCompose(result -> setState(stateAfter));
```
chimp1984, line 289: I think the setState could be handled in a whenComplete instead of staring a new CompletableFuture with thenCompose

chimp1984 (approved):
utACK

Suggested changes can be done in a follow up PR as well...

PR #168, opened 2022-03-27
Refactor channel selection

salokiam:
1. refactore PrublicChannelSelection and PrivateChannelSelection to hae a base class and reduce code duplication
2. removed ChannelSelection.ListItem as that class didnt do anything for 'keep it simple'-principle.
3. fixed issue where 2 channels were selectable at the same time.


desktop/src/main/java/bisq/desktop/primary/main/content/social/chat/components/PublicChannelSelection.java
```diff
@@ -41,130 +42,23 @@
...
-                            if (item != null && !empty) {
-                                label.setText(item.channelName);
-                                setGraphic(label);
-                            } else {
-                                setGraphic(null);
-                            }
-                        }
-                    };
+        super();
+        controller = new ChannelSelection.Controller(chatService, Res.get("social.publicChannels")) {
```
chimp1984, line 50: We could pass only the display string in the super call and initiate the controller in the base class and make it final.

desktop/src/main/java/bisq/desktop/primary/main/content/social/chat/components/PublicChannelSelection.java
```diff
@@ -51,12 +52,11 @@ public PublicChannelSelection(ChatService chatService) {
             @Override
             public void onViewAttached() {
                 super.onViewAttached();
-                channelsPin = FxBindings.<PublicChannel, ChannelSelection.ListItem> bind(model.channels)
-                        .map(ChannelSelection.ListItem::new)
+                channelsPin = FxBindings.<PublicChannel, Channel> bind(model.channels)
```
chimp1984, line 55: Channel is missing the generic type

chimp1984 (changes requested):
I get `AuthenticatedDataStorageService: Signature is invalid at add.` when testing sending messages. 
I will merge soon the protobug changes, that might require small adoptions anyway and maybe even resolve that issue.

chimp1984:
> I get `AuthenticatedDataStorageService: Signature is invalid at add.` when testing sending messages. I will merge soon the protobug changes, that might require small adoptions anyway and maybe even resolve that issue.

Might have been a local issue on my side, got the same with my branch but it was working before, so I guess I had an idle seed node with another branch running causing that... so if you don't get that error, ignore my comment. Still prefer to merge it after protobuf as the generic types might have changed a bit and its missing in your PR at some points.

chimp1984 (approved):
utACK

PR #169, opened 2022-03-28
Store seeds in Sets to avoid duplicates

alkum:
In some tests, there were multiple connections to the same seed. Eliminate duplicates by storing and passing seeds in hashsets.


chimp1984 (approved):
utACK

PR #170, opened 2022-03-28
Preserve I2P manager for default session when one peer times out

alkum:
Keep the session manager alive if it has other connected sockets, when one of its sockets encounters an IOException.


chimp1984 (approved):
utACK

PR #171, opened 2022-03-28
Extend PeerExchangeService with optimistic peer exchange

alkum:
Extend the peer exchange logic with ability to complete future when first peer exchange succeeds (optimistic scenario). Alternatively, wait until all peer connections have been tried.


chimp1984 (commented):
`anyOfBooleanMatchingFilterOrAll` is not very intuitive, but I think I also dont have a better suggestion. 
Not sure if its a too specific case to use as Util. But fine to me to keep it as it is for now.

chimp1984 (approved):
utACK

Issue #178, opened 2022-03-30
Gradle fixes and improvements


alkum:
This is an "umbrella task" for various gradle-related fixes and improvements:

- [ ] Consolidate dependency versions in `gradle/libs.versions.toml`
- [x] Find mechanism to enforce single version for dependencies, incl. transitive
  - solves issues such as https://github.com/bisq-network/bisq/issues/4086
  - see constraints in `platforms/common-platform/build.gradle` (constraints not being enforced?)
  - see (recommended?) approach in https://docs.gradle.org/current/userguide/dependency_constraints.html
  - see `ResolutionStrategy` in https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.ResolutionStrategy.html
- [ ] Add dependency verification + doc
  - similar to https://github.com/bisq-network/bisq/pull/5824
- [ ] Add ability to check for unused dependencies + doc usage
  - option 1: https://www.baeldung.com/gradle-finding-unused-dependencies
  - option 2: https://github.com/gradle-dependency-analyze/gradle-dependency-analyze
  - further discussion: https://stackoverflow.com/questions/19379517/how-to-find-remove-unused-dependencies-in-gradle
- [x] Build platform-specific binaries
  - https://github.com/bisq-network/bisq2/pull/218

I plan to address the above with small incremental PRs, to simplify reviews and to detect potential issues early.

alkum:
> Can you also have a look if binary build for JavaFx is working? It had issues when we started with the setup.

You mean integrating and using https://gluonhq.com/products/javafx/ ?

chimp1984:
is JavaFx now a product of Gluon? I just mean to create a binary as we do in Bisq 1. Java packager had issues with our java version as far I remember.

alkum:
AFAIK JavaFX packaging was tricky, it only worked with the Bisq 1 packager because the openjdk version used also included the JavaFX modules and OS-native libraries.

I can look into it again and see if things changed.

chimp1984:
I guess that is superseeded by the improvements @alvasw did over the past weekss.

PR #179, opened 2022-03-30
Consolidate guava lib version

alkum:
Define guava lib version in libs.versions.toml

Related to https://github.com/bisq-network/bisq2/issues/178


chimp1984 (approved):
utACK

PR #180, opened 2022-03-30
Add DaoBridgeData as authorized data

chimp1984:
Add DaoBridgeData as authorized data (requires a signature and pubkey against a hard coded list of pub keys like used for Filer, Alert, Arbitrator in Bisq 1).
DaoBridgeData could be used for providing DAO data to Bisq 2 while not having integrated the DAO.


chimp1984:
Add DaoBridgeData as authorized data (requires a signature and pubkey against a hard coded list of pub keys like used for Filer, Alert, Arbitrator in Bisq 1).
DaoBridgeData could be used for providing DAO data to Bisq 2 while not having integrated the DAO.

PR #187, opened 2022-03-31
Persist to PeerGroupStore when new peers are known

alkum:
Add any newly reported peers to the peer group store.

### DIFF
--- a/network/src/main/java/bisq/network/p2p/services/peergroup/PeerGroup.java
+++ b/network/src/main/java/bisq/network/p2p/services/peergroup/PeerGroup.java
@@ -65,16 +65,16 @@ public class PeerGroup {
     @Getter
     private final Set<Address> seedNodeAddresses;
     private final BanList banList;
-    private final PeerGroupStore peerGroupStore;
+    private final PeerGroupService peerGroupService;
     @Getter
     private final Set<Peer> reportedPeers = new CopyOnWriteArraySet<>();
 
-    public PeerGroup(Node node, Config config, Set<Address> seedNodeAddresses, BanList banList, PeerGroupStore peerGroupStore) {
+    public PeerGroup(Node node, Config config, Set<Address> seedNodeAddresses, BanList banList, PeerGroupService peerGroupService) {
         this.node = node;
         this.config = config;
         this.seedNodeAddresses = seedNodeAddresses;
         this.banList = banList;
-        this.peerGroupStore = peerGroupStore;
+        this.peerGroupService = peerGroupService;
     }
 
     ///////////////////////////////////////////////////////////////////////////////////////////////////
@@ -83,6 +83,7 @@ public class PeerGroup {
 
     public void addReportedPeers(Set<Peer> peers) {
         reportedPeers.addAll(peers);
+        peerGroupService.addPersistedPeers(peers);
     }
 
     public void removeReportedPeers(Collection<Peer> peers) {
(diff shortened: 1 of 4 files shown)


network/src/main/java/bisq/network/p2p/services/peergroup/PeerGroup.java
```diff
@@ -65,16 +65,16 @@ public static Config from(com.typesafe.config.Config typesafeConfig) {
     @Getter
     private final Set<Address> seedNodeAddresses;
     private final BanList banList;
-    private final PeerGroupStore peerGroupStore;
+    private final PeerGroupService peerGroupService;
     @Getter
     private final Set<Peer> reportedPeers = new CopyOnWriteArraySet<>();
 
-    public PeerGroup(Node node, Config config, Set<Address> seedNodeAddresses, BanList banList, PeerGroupStore peerGroupStore) {
+    public PeerGroup(Node node, Config config, Set<Address> seedNodeAddresses, BanList banList, PeerGroupService peerGroupService) {
```
chimp1984: PeerGroupService creates PeerGroup so it should not have a bi-directional dependency. If needed we can use loose coupling via Interface or handler.
See PeerExchangeService how it uses PersistedPeersHandler

network/src/test/java/bisq/network/p2p/services/peergroup/exchange/BasePeerExchangeServiceTest.java
```diff
@@ -41,24 +44,32 @@
     void test_peerExchange(Node.Config nodeConfig) throws InterruptedException, ExecutionException {
         int numSeeds = 2;
         int numNodes = 2;
-        PeerGroupStore peerGroupStore = new PeerGroupStore();
         BanList banList = new BanList();
+        Node tempNode = new Node(banList, nodeConfig, "node-id");
+        PeerGroupStore peerGroupStore = new PeerGroupStore();
+        PersistenceService persistenceService = new PersistenceService(getBaseDirName());
```
chimp1984, line 50: For really testing persistence I guess it requires more to setup the persistence framework.

network/src/main/java/bisq/network/p2p/services/peergroup/PeerGroup.java
```diff
@@ -83,6 +83,7 @@ public PeerGroup(Node node, Config config, Set<Address> seedNodeAddresses, BanLi
 
     public void addReportedPeers(Set<Peer> peers) {
         reportedPeers.addAll(peers);
+        peerGroupService.addPersistedPeers(peers);
```
chimp1984: when using the persistedPeersHandler interface it would be: persistedPeersHandler.addPersistedPeers(peers);

chimp1984 (changes requested):

network/src/main/java/bisq/network/p2p/services/peergroup/PeerGroup.java
alkum: Good point, I didn't realize that.

network/src/test/java/bisq/network/p2p/services/peergroup/exchange/BasePeerExchangeServiceTest.java
alkum: Yes, that was only to make the test compile after the constructor change.

Haven't found a way yet to fully test persistence, although some tests worked (like `ClearNetPeerExchangeServiceTest` when enabled).

network/src/main/java/bisq/network/p2p/services/peergroup/PeerGroup.java
alkum: Done

PR #196, opened 2022-04-02
Consolidate lombok versioning

alkum:
Add lombok to the version catalog.

Relates to https://github.com/bisq-network/bisq2/issues/178


chimp1984 (approved):
utACK

PR #203, opened 2022-04-04
Gradle Regtest: Add networkServiceConfig JVM args

alvasw:
The runWithBitcoindRegtestWallet is broken. This change adds the
following JVM arguments:
-Dbisq.networkServiceConfig.seedAddressByTransportType.clear.0=127.0.0.1:8000
-Dbisq.networkServiceConfig.seedAddressByTransportType.clear.1=127.0.0.1:8001


chimp1984 (approved):
utACK

PR #206, opened 2022-04-05
Bitcoind: ZeroMQ Support + Auto-Update Balance in DesktopApp

alvasw:
Consumers can register listeners when: 
- a new block gets mined
- a UTXO is spent
- a receive address received money

### DIFF
--- a/application/src/main/java/bisq/application/DefaultApplicationService.java
+++ b/application/src/main/java/bisq/application/DefaultApplicationService.java
@@ -166,7 +166,6 @@ public class DefaultApplicationService extends ServiceProvider {
 
         Optional<WalletConfig> walletConfig = !isRegtestRun() ? Optional.empty() : createRegtestWalletConfig();
         walletService = new WalletService(persistenceService, walletConfig);
-        walletService.tryAutoInitialization();
 
         daoBridgeService = new DaoBridgeService(networkService, identityService, getConfig("bisq.oracle.daoBridge"));
     }
@@ -207,13 +206,12 @@ public class DefaultApplicationService extends ServiceProvider {
                 .thenCompose(result -> setStateAfter(accountAgeWitnessService.initialize(), State.ACCOUNT_AGE_WITNESS_SERVICE_INITIALIZED))
                 .thenCompose(result -> setStateAfterList(protocolService.initialize(), State.PROTOCOL_SERVICE_INITIALIZED))
                 .thenCompose(result -> CompletableFutureUtils.allOf(
-                        walletService.tryAutoInitialization(),
                         userProfileService.initialize()
                                 .thenCompose(res -> chatService.initialize()),
                         openOfferService.initialize(),
                         offerBookService.initialize(),
                         tradeIntentService.initialize(),
-                        walletService.tryAutoInitialization()))
+                        walletService.initialize()))
                 // TODO Needs to increase if using embedded I2P router (i2p internal bootstrap timeouts after 5 mins)
                 .orTimeout(5, TimeUnit.MINUTES)
                 .thenApply(list -> list.stream().allMatch(e -> e))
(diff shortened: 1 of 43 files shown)


application/src/main/java/bisq/application/DefaultApplicationService.java
```diff
@@ -137,7 +137,8 @@ public DefaultApplicationService(String[] args) {
         protocolService = new ProtocolService(networkService, identityService, persistenceService, openOfferService);
 
         Optional<WalletConfig> walletConfig = !isRegtestRun() ? Optional.empty() : createRegtestWalletConfig();
-        walletService = new WalletService(walletConfig);
+        walletService = new WalletService(persistenceService, walletConfig);
+        walletService.tryAutoInitialization();
```
chimp1984: We only create the services in the constructor but any initialisation should happen in the initialize method inside the completable future calls

chimp1984: Ah just saw you call it already there. So I assume that is a left over and need to be removed.

desktop/src/main/java/bisq/desktop/primary/main/content/wallet/receive/WalletReceiveModel.java
```diff
@@ -22,11 +22,7 @@
...
 
 public class WalletReceiveModel implements Model {
-    private final ObservableList<String> listItems = FXCollections.observableArrayList();
-
-    public void addNewAddress(String address) {
-        listItems.add(address);
-    }
+    final ObservableList<String> listItems = FXCollections.observableArrayList();
 
     public ObservableList<String> getListItems() {
```
chimp1984: Better use the lombok @Getter for reducing boilerplate

wallets/src/main/java/bisq/wallets/bitcoind/zeromq/BitcoindZeroMqTopicProcessors.java
```diff
@@ -0,0 +1,67 @@
...
+
+import bisq.wallets.bitcoind.rpc.BitcoindDaemon;
+import bisq.wallets.bitcoind.rpc.responses.BitcoindDecodeRawTransactionResponse;
+import com.google.common.io.BaseEncoding;
+
+import java.util.Set;
+
+public class BitcoindZeroMqTopicProcessors {
+
+    public static final String TOPIC_HASHBLOCK = "hashblock";
```
chimp1984: Wouldn't be a enum more appropriate here?

wallets/src/main/java/bisq/wallets/WalletService.java
```diff
@@ -121,9 +160,21 @@ public boolean isWalletReady() {
...
+
+                listUnspent().thenAccept(utxos -> {
+                    utxoTxIds.clear();
+                    utxos.stream()
+                            .map(Utxo::getTxId)
+                            .forEach(utxoTxIds::add);
+                });
+            }
+
+            return (long) walletBalance;
```
chimp1984: Is walletBalance in satoshis? Weird why they use double then... support for sub-satoshis?

chimp1984 (changes requested):
Looks good overall, just a few small issues and questions.
Have not tested it yet.

wallets/src/main/java/bisq/wallets/WalletService.java
alvasw: It's in BTC (see https://developer.bitcoin.org/reference/rpc/getbalance.html).

application/src/main/java/bisq/application/DefaultApplicationService.java
alvasw: Yes, that's a left over. Will fix that.

wallets/src/main/java/bisq/wallets/WalletService.java
chimp1984: But then `return (long) walletBalance;` is wrong, no? As well as `observableBalanceAsCoin.get().getValue() != walletBalance`

alvasw: You're correct. I was distracted.

chimp1984 (approved):
utACK

PR #211, opened 2022-04-05
Gradle consolidate various libs

alkum:
Consolidate gson, logback, slf4j, kotlin-dsl, jetbrains-annotations libs in the version catalog.

Relates to https://github.com/bisq-network/bisq2/issues/178


chimp1984 (approved):
utACK

PR #220, opened 2022-04-09
Remove usage of discouraged gson-internal class

alkum:
Update `MarketPriceService` to avoid using gson-internal `LinkedTreeMap`.

The removed class was in the package `com.google.gson.internal` which had as description

> Do NOT use any class in this package as they are meant for internal use in Gson. These classes will very likely change incompatibly in future versions. You have been warned.


chimp1984 (approved):
utACK

PR #221, opened 2022-04-09
Document notes on java modularization

alkum:
Document findings that could be helpful in future efforts of modularizing Bisq 2.


chimp1984 (approved):
utACK

PR #225, opened 2022-04-09
remove PrivatChatMessages, when they expire

salokiam:
first pass on removing PrivateChatMessages. Empty PrivateChannels not included yet. (Message expiration is be default 10Days.
When selecting a private channel, old messages are silently pruned.


chimp1984 (approved):
utACK

PR #241, opened 2022-04-13
Gradle: Regtest Task Configuration Avoidance (lazy init)

alvasw:
At the moment, every Gradle run hits the disk to check for the regtest files and initializes these if they don't exist. This is bad for performance, and we shouldn't do it. This change lazily initializes all regtest tasks.
For more information, see: https://docs.gradle.org/current/userguide/task_configuration_avoidance.html


chimp1984 (approved):
utACK

PR #244, opened 2022-04-14
Advance bootstrap when any network ready

alkum:
Advance through the initialization process as soon as any network is bootstrapped.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/primary/onboarding/initUserProfile/InitUserProfileController.java
+++ b/desktop/src/main/java/bisq/desktop/primary/onboarding/initUserProfile/InitUserProfileController.java
@@ -75,6 +75,13 @@ public class InitUserProfileController implements Controller {
         model.createProfileButtonDisable.set(true);
         model.showProcessingPopup.set(true);
         String profileId = model.profileId.get();
+        // Notes on the first run, when a new user profile is created at startup:
+        // Unfortunately the new profile can only be initialized after the network is fully initialized, which is after
+        // all supported network types have a local address. This translates to longer display times for the processing
+        // popup ("Preparing network node for new identity...").
+        // The good news is, the user is likely to spend some time on the Create Profile popup before clicking Next,
+        // which triggers the new profile initialization. The network layer is being initialized in parallel. This means
+        // in practice, the processing popup won't be shown for long.
         userProfileService.createNewInitializedUserProfile(profileId,
                         model.getNickName().get(),
                         model.tempKeyId,
--- a/network/src/main/java/bisq/network/NetworkService.java
+++ b/network/src/main/java/bisq/network/NetworkService.java
@@ -208,12 +208,13 @@ public class NetworkService implements PersistenceClient<NetworkIdStore> {
         log.info("bootstrapToNetwork");
         String nodeId = Node.DEFAULT;
         return serviceNodesByTransport.bootstrapToNetwork(getDefaultPortByTransport(), nodeId)
-                .whenComplete((result, throwable) -> {
+                .thenApply(result -> { // We use thenApply to return a future that completes after this function
                     // If networkNode has not been created we create now one with the default pubKey (default keyId)
                     // and persist it.
-                    if (throwable == null) {
+                    if (result == false) {
                         persistNetworkId(nodeId, keyPairService.getDefaultPubKey());
                     }
+                    return result;
                 });
     }
 
--- a/network/src/main/java/bisq/network/p2p/ServiceNodesByTransport.java
+++ b/network/src/main/java/bisq/network/p2p/ServiceNodesByTransport.java
@@ -111,7 +111,8 @@ public class ServiceNodesByTransport {
     }
 
     public CompletableFuture<Boolean> bootstrapToNetwork(Map<Transport.Type, Integer> portByTransport, String nodeId) {
-        return CompletableFutureUtils.allOf(map.entrySet().stream()
+        return CompletableFutureUtils.anyOf( // We use anyOf to complete as soon as any network bootstraps
+                map.entrySet().stream()
                 .map(entry -> {
                     int port = portByTransport.get(entry.getKey());
                     ServiceNode serviceNode = entry.getValue();
--- a/network/src/main/java/bisq/network/p2p/node/transport/TorTransport.java
+++ b/network/src/main/java/bisq/network/p2p/node/transport/TorTransport.java
@@ -82,7 +82,13 @@ public class TorTransport implements Transport {
     }
 
     public Optional<Socks5Proxy> getSocksProxy() throws IOException {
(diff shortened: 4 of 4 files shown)


network/src/main/java/bisq/network/NetworkService.java
```diff
@@ -208,12 +208,13 @@ public CompletableFuture<Boolean> bootstrapToNetwork() {
         log.info("bootstrapToNetwork");
         String nodeId = Node.DEFAULT;
         return serviceNodesByTransport.bootstrapToNetwork(getDefaultPortByTransport(), nodeId)
-                .whenComplete((result, throwable) -> {
+                .thenApply(result -> { // We use thenApply to return a future that completes after this function
                     // If networkNode has not been created we create now one with the default pubKey (default keyId)
                     // and persist it.
-                    if (throwable == null) {
+                    if (result == false) {
```
chimp1984, line 214: Shouldn't it be `result==true`?

desktop/src/main/java/bisq/desktop/primary/onboarding/initUserProfile/InitUserProfileController.java
```diff
@@ -75,6 +75,13 @@ void onCreateUserProfile() {
         model.createProfileButtonDisable.set(true);
         model.showProcessingPopup.set(true);
         String profileId = model.profileId.get();
+        // Notes on the first run, when a new user profile is created at startup:
+        // Unfortunately the new profile can only be initialized after the network is fully initialized, which is after
+        // all supported network types have a local address. This translates to longer display times for the processing
+        // popup ("Preparing network node for new identity...").
+        // The good news is, the user is likely to spend some time on the Create Profile popup before clicking Next,
+        // which triggers the new profile initialization. The network layer is being initialized in parallel. This means
+        // in practice, the processing popup won't be shown for long.
```
chimp1984, line 84: Currently the addresses are created when the user clicks next so the time being on that screen does not help us as the user can change the keypair any time, thus we cannot create the addresses upfront. But we should consider to bake something into the UI which allows us to do that in the background. Maybe we dont need the popup but delay it to the time when the address is actually needed (at create offer later, until then the address shoudld be created, and only if not we need to show the user a "in-progress" screen

chimp1984 (changes requested):

network/src/main/java/bisq/network/NetworkService.java
alkum: Comment says _"If networkNode has not been created..."_ so my understanding is in that case `result` is `false`?

desktop/src/main/java/bisq/desktop/primary/onboarding/initUserProfile/InitUserProfileController.java
alkum: > Currently the addresses are created when the user clicks next

Correct, but the address creation is relatively fast compared to network init. Also, address creation waits on complete network initialization (e.g. clearnet + tor + i2p).

> ...so the time being on that screen does not help us as the user can change the keypair any time.

In my tests (clearnet + tor) it helped. It didn't eliminate the need to show the popup because as you said address creation takes a bit of time. But it took way less time if some time was spent on that screen vs if user tried to click next right away. 

I spent some time trying to optimize it (make address creation depend on network bootstrap, which advances after first network is ready, vs network init, which completes when all networks are ready). But I couldn't, because the addresses indeed need an initialized network:

https://github.com/bisq-network/bisq2/blob/4f1cf9068e40d41efbb9b0a64a2e6a8fdb952d58/identity/src/main/java/bisq/identity/IdentityService.java#L156-L160

So I added that comment to spare future efforts spent in this direction.

> Maybe we dont need the popup but delay it to the time when the address is actually needed (at create offer later, until then the address shoudld be created, and only if not we need to show the user a "in-progress" screen

Good idea. I'll prepare a PR to do that.

chimp1984: Yes, I think best is to remvoe that popup and check if network addresses have been created once they are needed. The identity creating needs its so we need to keep a temporary state until we know all addresses.

network/src/main/java/bisq/network/NetworkService.java
chimp1984: Hm... dont remember exaclty. Its a bit unclear and should be improved. I think it was meaning that if the network was not initialized already (I guess we call it just in case and after once initialized it returns fast) we need to persist the keypair. Best to put a breakpoint in to see what the value of result is and if its called repeatedly...

chimp1984: The `bootstrapToNetwork` method below seems to return always true if successful.

chimp1984:
Could you rebase that PR.

alkum:
Rebased, but there are still 2 open points in the review.

network/src/main/java/bisq/network/NetworkService.java
alkum: This specific line (`if (result == false)`) is called only once, even when using all 3 networks, with `result` being `true`. So the if condition is never true.

If I change it to `if (result == true)`, persisting the network in the very next line fails with
```
NetworkId for default must be present
```

This makes sense, because bootstrapping != initializing the network. The network ID is only available after the networkService is initialized, but here we only do bootstrapping.

NetworkService.initialize is relatively long, because it only finishes when _all_ enabled networks are up and running.

Would it make sense then to remove the whole `thenApply` block?

```
return serviceNodesByTransport.bootstrapToNetwork(getDefaultPortByTransport(), nodeId)
        .thenApply(result -> { // We use thenApply to return a future that completes after this function
            // If networkNode has not been created we create now one with the default pubKey (default keyId)
            // and persist it.
            if (result == false) {
                persistNetworkId(nodeId, keyPairService.getDefaultPubKey());
            }
            return result;
        });
```

Because its not executed (boolean check is called only once with result = true). And persisting the network would fail because right after bootstrapping the network is too early (it would work after network initialize).

desktop/src/main/java/bisq/desktop/primary/onboarding/initUserProfile/InitUserProfileController.java
alkum: Should I do that in this PR or a separate one?

network/src/main/java/bisq/network/NetworkService.java
chimp1984: I checked out your PR and looked into that. The `if (result == false) ` check is wrong as we always return true from `serviceNodesByTransport.bootstrapToNetwork`. That API is questionable in itself and would require more refactoring work...

chimp1984:
I think that PR is too old now to be easy to get merged. Will close it for now.

PR #249, opened 2022-04-18
fixed issue where 2 channels at the same time where marked as selected

salokiam:
fixed issue where 2 channels at the same time where marked as selected


chimp1984 (changes requested):
NACK

It does not keep selection when switching between "exchange" and "discussion" tabs. 
It should pin the channel on both screens to the last selected. We need 2 different selectedChannel fields in the ChatStore for that.

chimp1984:
I will work myself on the channel selection task. We can drop that PR.

PR #255, opened 2022-04-21
Gradle: Global JUnit 5 Setup

alvasw:
For more information, see: https://junit.org/junit5/docs/current/user-guide/#running-tests-build-gradle


chimp1984 (approved):
utACK

PR #258, opened 2022-04-21
Add market channel chooser

chimp1984:


alkum:
ACK. Tested on clearnet, looks good.

A few minor things, but can be handled later (and not sure if caused by this change) :
- when Alice starts a private chat with Bob, if Bob is already in the Chat view before the private chat begins, he will not see a new entry under Private Chats (he'd have to navigate away and back to see it)
- slight misalignments of panels in Chat view if window width is around 1000px (when clicking Search, Notification Bell or Info button)
- Chat view: when switching from Exchange to Discuss tabs, the previously selected chat room contents (from prev. tab) are still shown, instead of showing the contents of a new room in the new category.
- Community > pressing the plus sign on Market Channels opens a very big popup with only 1 dropdown (popup oversized)
- Community view and Trade > Satoshi Square view have almost entirely the same content (Market Channels and Private Channels). Perhaps one of them can go? To simplify menu on the left?
- Found no way to create an offer via the Create Offer button in the chat views, the button didn't respond. Going via Trade > Bisq Multisig > top right Create Offer worked though.

Other than that looks very thorough and comprehensive.

chimp1984:
Thanks for the review and feedback. Yes there is lot still WIP.... just wanted to get continuous merges to avoid that code base of devs gets too far apart...

PR #267, opened 2022-04-23
Refactor public discussion channel id

xyzmaker123:
- use UUID for `PublicDiscussionChannel -> id` during creation of default channels
- display `PublicDiscussionChannel -> channelName` instead of `PublicDiscussionChannel -> id`


chimp1984:
I fear we have been working in the same area... I did quite a lot of changes in the channel management in the meantime

PR #272, opened 2022-04-25
Toggle button auto-slide component

### LINKED ISSUE #270: Make autoSlide Togglebox component
As we got rid of the JFoenix dependency we need a replacement for the autoSlide Togglebox component.
At first its ok to only have it graphically equvalent, no animations needed. Minimal animations can be added in a second step.
Does not need to be a full JavaFx compliant component but can be a simple higher level composition component.

@xyzmaker123 Interested in that task?

xyzmaker123:
Fixes #270

toggle-button

### DIFF
--- a/desktop/src/main/java/bisq/desktop/components/controls/BisqToggleButton.java
+++ b/desktop/src/main/java/bisq/desktop/components/controls/BisqToggleButton.java
@@ -1,0 +1,189 @@
+/*
+ * This file is part of Bisq.
+ *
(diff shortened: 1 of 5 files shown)


chimp1984:
@xyzmaker123 You need to set up commit signing

chimp1984 (approved):
utACK

xyzmaker123:
@chimp1984 OK - I just pushed signed commit

chimp1984:
The failed build is likely from a failed test in that branch (fixed in more recent main branch).

PR #275, opened 2022-04-25
Fix unit test

alkum:
Adapt check to new `Market.toString()` implementation.


chimp1984:
I fixed those already

PR #280, opened 2022-04-27
Wallet: Save WalletFile by default in data directory

alvasw:


application/src/main/java/bisq/application/DefaultApplicationService.java
```diff
@@ -160,7 +160,8 @@ public DefaultApplicationService(String[] args) {
         protocolService = new ProtocolService(networkService, identityService, persistenceService, openOfferService);
 
         Optional<WalletConfig> walletConfig = !isRegtestRun() ? Optional.empty() : createRegtestWalletConfig();
-        walletService = new WalletService(persistenceService, walletConfig);
+        Path walletsDataDir = Path.of(applicationConfig.baseDir() + File.separator + "wallets");
```
chimp1984, line 163: If its makes sense to keep that path configurable I think its better to add it to the typesafe config. Otherwise to move it into the wallet service as its more an internal property. But I guess users should be able to point to any external wallet directory as well.

chimp1984 (approved):
utACK

application/src/main/java/bisq/application/DefaultApplicationService.java
alvasw: When I implemented Bitcoin Core and Elements Core, I didn't think about the case where the daemon runs on another host. I'll have to get rid of the path altogether.

PR #281, opened 2022-04-27
Wallet: Remove NetworkType Enum (Regtest, Signet, Testnet, ...)

alvasw:
It doesn't matter on which network Bisq is running. The code works the same everywhere.


wallets/src/main/java/bisq/wallets/bitcoind/BitcoindProcess.java
```diff
@@ -53,12 +52,11 @@ public BitcoindProcess(RpcConfig rpcConfig, Path dataDir) {
 
     @Override
     public ProcessConfig createProcessConfig() {
-        String networkArg = getParamForNetworkType(rpcConfig.networkType());
         int zmqPort = NetworkUtils.findFreeSystemPort();
         return new ProcessConfig(
                 "bitcoind",
                 List.of(
-                        networkArg,
+                        "-regtest",
```
chimp1984, line 59: Is that correct?

wallets/src/main/java/bisq/wallets/elementsd/process/ElementsdProcess.java
```diff
@@ -49,7 +48,7 @@ public ProcessConfig createProcessConfig() {
         return new ProcessConfig(
                 "elementsd",
                 List.of(
-                        "-chain=" + getNetworkName(),
+                        "-chain=elementsregtest",
```
chimp1984, line 50: Is that correct?

chimp1984 (changes requested):

wallets/src/main/java/bisq/wallets/bitcoind/BitcoindProcess.java
alvasw: Yes, that's correct but you're right. It's confusing. I renamed it to `BitcoindRegtestProcess` and moved it into the tests module.

wallets/src/main/java/bisq/wallets/elementsd/process/ElementsdProcess.java
alvasw: Yes, that's correct but you're right. It's confusing. I renamed it to `ElementsdRegtestProcess` and moved it into the tests module.

wallets/src/main/java/bisq/wallets/bitcoind/BitcoindProcess.java
chimp1984: Ah thanks.

chimp1984 (approved):
utACK

PR #285, opened 2022-04-27
Amend seed program args

alkum:
Update seed args to avoid endless loop in a network thread, whereby the seed tries to connect to itself on all supported network types.


chimp1984 (approved):
utACK

Issue #238, opened 2022-04-12
Seed setup


alkum:
This is an umbrella task covering the seed setup.

- [x] Initial setup: Package runnable jar + document how to create it
  - https://github.com/bisq-network/bisq2/pull/239
- [x] Clearnet seeding: Fix issue with IP binding
  - https://github.com/bisq-network/bisq2/pull/240
- [x] Tor seeding: Fix issue with port selection
  - Fixed by giving 2nd seed args to change default tor port: `-Dbisq.networkServiceConfig.supportedTransportTypes.1=TOR -Dbisq.networkServiceConfig.defaultNodePortByTransportType.tor=1001`
  - Same applies for clearnet
- [x] Maintenance
    - [x] Document full setup on a fresh VM (including dependencies, like `jdk17`, `i2p`)
      - #286
    - [x] Document how to install seed as service (using runnable jar)
      - #274
    - [x] Document how to update seed, once installed as service
      - #274

PR #286, opened 2022-04-27
Document I2P service setup for seed installation

alkum:
Document the installation of the I2P service on Debian / Ubuntu.

This is typically required in a seed environment.

Part of #238


chimp1984 (approved):
utACK

PR #287, opened 2022-04-28
Chat message -  more options menu refactoring

xyzmaker123:


desktop/src/main/java/bisq/desktop/primary/main/content/social/components/ChatMessagesComponent.java
```diff
@@ -403,6 +402,7 @@ private static class Model implements bisq.desktop.common.view.Model {
         private final boolean isDiscussionsChat;
         private final ObservableList<String> customTags = FXCollections.observableArrayList();
         private final BooleanProperty allowEditing = new SimpleBooleanProperty();
+        private final ObjectProperty<ChatMessage> moreOptionsVisibleMessage = new SimpleObjectProperty<>(null);
```
chimp1984, line 405: If we do not use any binding we dont need a bindable property. But I think its better to keep it that way as the whole view part might need a refactoring at some point and then it might be useful. But for now as long the sceens are not more final and confirmed by Pedro we are ok.

chimp1984 (approved):
utACK

PR #289, opened 2022-04-29
fixed API to actually deliver JSON instead of String of json

salokiam:
the json delivered before was actualy a strng with quoted json, so parsed that to really deliver json. toto is to optimize the json generation by changing the serialization process.


api/src/main/java/bisq/api/rest/controller/ApiChatController.java
```diff
@@ -49,17 +50,17 @@ public ApiChatController(ApiApplicationService apiApplicationService) {
     }
 
     @GetMapping(path = "/api/chat/get-public-trade-channels")
-    public List<String> getPublicTradeChannels() {
+    public List<JsonNode> getPublicTradeChannels() {
```
ghost: What do you think about returning [DTO domain objects](https://www.baeldung.com/entity-to-and-from-dto-for-a-java-spring-application) in controllers? It should give as more flexibility regarding which fields we want to return by API and which to hide. Also swagger documentation would be more robust then.

salokiam: There are DTO Objects already which we are sending over the wire. That are the Protobuf Objects. Right now we are trying to send the proto buf objects as json, but it may happen in the future that we need to tweek that. There are serveral option to do that 1.) manipulate the generated JsonNode objects. 2.) change the serialisation pocess in Spring /Jackson. 3.) use DTOs. DTOs have the advantage that you are not accidently exposing fields that you didnt want to expose. That risk is here minimal, because we anyway only expose the proto buf data, which is send throuth the network anyway. Other that that DTO has the most boilerplate effort and usually result is a lot of code that does nothing. So i prefer 1+2 if needed. Also, I am awware that there are cases where you want to change a lot for the json, then it might be better to use DTOs to have the ultimative flexibility. However, according to the KISS principle I would deplay these decisions until needed and do them on a case by case basis. Actually I already reworked the stuff in question to use method 2, altering the spring serialization process, which I will push the the PR asap.

ghost: Got it, I'm okay with postponement of the decision to introduce DTOs.

salokiam: ah, I did answer the argument with the swagger documentation: Thats very true. We probably dont have any other place to put the documentation other than DTOs, when going the swagger way. So the DTO will probably happen soon than later, also when I am looking at the amount of renamings and changes coming in today.. :)

ghost: Yes, I'm pretty sure we'll finally use DTOs. I don't want to put any pressure on the pace - replacing String with JsonNode is a step in the right direction anyway :)

chimp1984 (approved):
utACK

chimp1984:
Sorry have not checked before my merge for open PRs... I hope its not much for conflict resolving...

api/src/main/java/bisq/api/rest/controller/ApiChatController.java
chimp1984: I would like to avoid to get a 3rd data representation. We have already te java objects and the protobuf specs. Once we go live all protobuf data will be hard to change anyway as its used for network and/or persistence, so I think thats good for enforcing some stability for the API as well. Where needed we can tweak with the suggested options but by default creating 100s of DTOs is a lot of overhead and adds lot of maintainance costs.

salokiam:
pushed the conflict resolution and some more changes. Please ACK again. Sorry for the confusion, will make 2 branches next time.

chimp1984 (approved):
utACK

PR #296, opened 2022-05-04
Welcome screen - redesign

xyzmaker123:
Preview:
welcome-screen-after-redesign


ripcurlx:
I don't know the current design, but shouldn't the green line under the selected tab be closer to the text?

desktop/src/main/java/bisq/desktop/primary/main/content/social/SocialView.java
```diff
@@ -17,30 +17,50 @@
...
     protected void onViewAttached() {
+        model.getView().addListener(viewChangeListener);
     }
 
     @Override
     protected void onViewDetached() {
+        model.getView().removeListener(viewChangeListener);
+    }
+    
+    private void setupHeadline(Object view) {
```
ghost: > I don't know the current design, but shouldn't the green line under the selected tab be closer to the text?

"Welcome" headline on [newest design](https://xd.adobe.com/view/88ec6f20-15ca-4326-81f4-30a30cc1aa0a-ddbf/screen/2003f6b1-f8dd-4b46-87ac-c80b6c92bb2b) for this specific screen is bigger. It currently causes small UI issues like this mentioned by @ripcurlx. Fixing them requires reorganization of [TabView](https://github.com/bisq-network/bisq2/blob/1f3507ec170121e736a2c0f53db8997553f7895e/desktop/src/main/java/bisq/desktop/common/view/TabView.java#L57) - I'm aware of it however I didn't wanted to make it in this pull request - I predict that designs may still evolve. If you thinks I should do it here just let me know.

CC: @chimp1984

desktop/src/main/java/bisq/desktop/primary/main/content/social/SocialView.java
```diff
@@ -17,30 +17,50 @@
...
         super(model, controller);
 
-        headlineLabel.setText(Res.get("welcome"));
-
         addTab(Res.get("social.gettingStarted"), NavigationTarget.GETTING_STARTED);
         addTab(Res.get("social.discuss"), NavigationTarget.DISCUSS);
         addTab(Res.get("social.learn"), NavigationTarget.LEARN);
         addTab(Res.get("social.connect"), NavigationTarget.CONNECT);
+
+        setupHeadline(model.getView()); // TODO: check why view is null at the beginning
```
chimp1984: At onViewAttached it would not be null. But better use EasyBind.subscribe() instead of listeners as that will call the callback at subscribe time, thus not requiring the extra setupHeadline(model.getView()) call.

desktop/src/main/java/bisq/desktop/primary/main/content/social/SocialView.java
chimp1984: Can you ask Pedro if the large distance of the line is intended? It looks to me to much distance, but he is our designer, so if he wants it that way we do it. Yes we can do that in an upcoming PR.

desktop/src/main/resources/bisq.css
```diff
@@ -109,7 +109,23 @@
 }
 
 .super-large-text {
-    -fx-font-size: 4.1em;
+    -fx-font-size: 4.1em !important;
```
chimp1984: Why is `!important` needed here?

desktop/src/main/resources/bisq.css
```diff
@@ -233,6 +249,24 @@
     -fx-font-size: 1.35em;
 }
 
+.bisq-text-2 {
+    -fx-text-fill: -bisq-grey-13;
+    -fx-font-family: "IBM Plex Sans Regular";
```
chimp1984: I guess we dont need to define the -fx-font-family as its the default

desktop/src/main/resources/bisq.css
```diff
@@ -432,6 +466,20 @@
     -fx-background-color: -bisq-content-bg;
 }
 
+/* Primary green button */
+.bisq-primary-button {
```
chimp1984: Can we apply that to the `default` pseudo class?
So we dont need to add a style class as in Bisq 1 but apply it to all buttons which are set as default buttons.

chimp1984: Ah just saw its not the defult button... For the welcome screen buttons we need a css class as they are larger as normal buttons. So maybe just rename the class to make it more clear. I misinterpreted bisq-primary-button as general css for primary buttons.

desktop/src/main/resources/images.css
```diff
@@ -100,3 +100,11 @@
 #governance {
     -fx-image: url("images/navigation/governance.png");
 }
+
+#welcome-community {
+    -fx-image: url("images/welcome-community@2x.png");
```
chimp1984: The @2x postfix should not be added to let the javafx framework apply the right image based on the resolution.
We just need to have both images provided in the resources. One with `@2x` postfix for high resolution and one without.

chimp1984 (changes requested):

chimp1984:
The layout does not match the design. I think one issue is the usage of @2x but also the fonts are a bit too large. Best to make screenshots and compare in Photoshop if design is matching the template.

Screenshot 2022-05-05 at 11 31 30

desktop/src/main/java/bisq/desktop/primary/main/content/social/SocialView.java
ghost: [Fixed](https://github.com/bisq-network/bisq2/pull/296/commits/ef575be7cd4503c8abf8cfc7adac29d1a3e3577f)

desktop/src/main/resources/bisq.css
ghost: [`headlineLabel`](https://github.com/bisq-network/bisq2/blob/24e8e0a8b75060b168abc660017303668e4b8dbd/desktop/src/main/java/bisq/desktop/common/view/TabView.java#L51) has [`.bisq-content-headline-label`](https://github.com/bisq-network/bisq2/blob/24e8e0a8b75060b168abc660017303668e4b8dbd/desktop/src/main/resources/bisq.css#L307-L311) class with `-fx-font-size: 1.78em;`. On getting started screen `headlineLabel` should be bigger. I decided to manage it by toggling `.super-large-text`  utility class. Unfortunately without `!important` javafx uses font size from `.bisq-content-headline-label`. 

Using `!important` for utility classes is rather [good practice](https://css-tricks.com/when-using-important-is-the-right-choice/).

desktop/src/main/resources/bisq.css
ghost: Good point. I see two "Regular" font faces defined on the top of `bisq.css` - [Sans](https://github.com/bisq-network/bisq2/blob/24e8e0a8b75060b168abc660017303668e4b8dbd/desktop/src/main/resources/bisq.css#L1-L3) and [Mono](https://github.com/bisq-network/bisq2/blob/24e8e0a8b75060b168abc660017303668e4b8dbd/desktop/src/main/resources/bisq.css#L17-L19) - I'll check how javafx chooses one of them. I wonder if we don't need to define one of theme somewhere in the root level.

desktop/src/main/resources/bisq.css
ghost: [Fixed](https://github.com/bisq-network/bisq2/pull/296/commits/b7fb5589cd442125fe59724d87fa30a41b97ac3e)

desktop/src/main/java/bisq/desktop/primary/main/content/social/SocialView.java
```diff
@@ -37,30 +36,27 @@ public SocialView(SocialModel model, SocialController controller) {
...
     protected void onViewAttached() {
-        model.getView().addListener(viewChangeListener);
+        selectedViewSubscription = EasyBind.subscribe(model.getView(), view -> {
+            if (view instanceof GettingStartedView) {
+                headlineLabel.setText(Res.get("welcome"));
+                if (!headlineLabel.getStyleClass().contains("super-large-text")) {
+                    headlineLabel.getStyleClass().add("super-large-text");
+                }
+                line.setOpacity(0);
+            } else {
```
chimp1984, line 50: If view is null it would fall into that branch but likely we dont want to ignore null values or clear the view (not needed here). Its likely not happening, just wanted to mention it.

desktop/src/main/resources/bisq.css
chimp1984: Ok.

desktop/src/main/resources/bisq.css
ghost: Seems like you're right - [fixed](https://github.com/bisq-network/bisq2/pull/296/commits/402cf2f1436736ce55cddb6c448be217f65b5963). 

Unfortunately I have no idea how JavaFX load default font - hope I'll figure it out one day :)

desktop/src/main/java/bisq/desktop/primary/main/content/social/SocialView.java
ghost: I'll fix distance between line and "Welcome" label in an upcoming PR - it's a bit challenging issue.

desktop/src/main/resources/bisq.css
chimp1984: Its in the css file on top.

desktop/src/main/java/bisq/desktop/primary/main/content/social/SocialView.java
chimp1984: Ok. We can delay it also until large distance is confirmed by Pedro.

desktop/src/main/resources/images.css
ghost: Fixed

chimp1984 (approved):
utACK

PR #300, opened 2022-05-06
Update font configuration

xyzmaker123:
Refers to: https://github.com/bisq-network/bisq2/pull/296#discussion_r865683868

I went a bit deeper into font configuration and I realized that [System Regular](https://github.com/openjdk/jfx/blob/14b23bcf69a321c44b09e584b30f39e1fa3e6948/modules/javafx.graphics/src/main/java/javafx/scene/text/Font.java#L85) font was used by default in application instead of `IBM Plex Sans`. Adding:
```
.root {
    -fx-font-family: "IBM Plex Sans";
    ...
 }   
```
fixed it.

I also extended `@font-face` declaration about `font-family`. It doesn't have any impact on UI but could be helpful for developers. I retrieved family names this way:
```
fc-scan IBMPlexSans-Medium.ttf
fc-scan IBMPlexSans-Bold.ttf
fc-scan IBMPlexSans-Regular.ttf
fc-scan IBMPlexSans-Light.ttf
fc-scan IBMPlexMono-Regular.ttf
```

UI:
1) Before
before-font-setup

2) After
after-font-setup

### DIFF
--- a/desktop/src/main/resources/bisq.css
+++ b/desktop/src/main/resources/bisq.css
@@ -1,24 +1,32 @@
 @font-face {
     src: url("/fonts/IBMPlexSans-Regular.ttf");
+    font-family: "IBM Plex Sans";
+    font-weight: normal;
 }
 
 @font-face {
     src: url("/fonts/IBMPlexSans-Bold.ttf");
+    font-family: "IBM Plex Sans";
+    font-weight: bold;
 }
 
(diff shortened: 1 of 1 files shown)


chimp1984 (approved):
utACK

Ah thanks for fixing that. Did not look closer into it myself

PR #306, opened 2022-05-09
Extend logging with per-module log files

alkum:
Add ability to direct a specific module (logger) to a specific appender, which can result in a per-module log file.

The custom appender is based on the appender settings[^1] used in Bisq 1.

The `module-info.java` file was removed from `bisq/common` because I couldn't find another way to make it allow the creation of a custom appender. It also doesn't actually enforce any java module behavior, since its the only subproject with a `module-info.java`.

[^1]: https://github.com/bisq-network/bisq/blob/master/common/src/main/java/bisq/common/app/Log.java


chimp1984 (approved):
utACK

chimp1984:
As @alvasw has added the modules file I want to wait for feedback from him regarding the removal of the module file. 
@alvasw: Do you have any comments on that?

alvasw:
The version we're using does not support Java 9 Modules..

> The 1.3.x series is Jigsaw/Java 9 modularized and requires slf4j-api version 2.0.x. Moreover, the 1.3.x series requires Java 8 or later at runtime whereas building logback from source requires Java 9. 
https://logback.qos.ch/news.html

See: http://mailman.qos.ch/pipermail/announce/2022/000172.html

The 1.3.x series are alpha versions. I started adding Java 9 Module to the wallets module, but I don't mind if we wait until logback is ready. It doesn't have a high priority at the moment.

@chimp1984 What do you think? We don't depend on jfoenix anymore, so this is the only blocker for Java 9 Jigsaw support.

chimp1984:
> The version we're using does not support Java 9 Modules..
> 
> > The 1.3.x series is Jigsaw/Java 9 modularized and requires slf4j-api version 2.0.x. Moreover, the 1.3.x series requires Java 8 or later at runtime whereas building logback from source requires Java 9.
> > https://logback.qos.ch/news.html
> 
> See: http://mailman.qos.ch/pipermail/announce/2022/000172.html
> 
> The 1.3.x series are alpha versions. I started adding Java 9 Module to the wallets module, but I don't mind if we wait until logback is ready. It doesn't have a high priority at the moment.
> 
> @chimp1984 What do you think? We don't depend on jfoenix anymore, so this is the only blocker for Java 9 Jigsaw support.

If it requires to use an alpha verison I would prefer to wait until its out of alpha. So far I have not seen a real benefit from Java modularisation but plenty of problems. To me it seems like its a not well thought out concept (in ragards to reality that many libs don't update and then requiring lot of ugly hacks,...)

alvasw:
> If it requires to use an alpha verison I would prefer to wait until its out of alpha. So far I have not seen a real benefit from Java modularisation but plenty of problems. To me it seems like its a not well thought out concept (in ragards to reality that many libs don't update and then requiring lot of ugly hacks,...)

I agree. Then we can delete `module-info.java`.

PR #307, opened 2022-05-09
Fix startup CSS warnings related to fonts

alkum:
One warning was fixed by having the `font-family` prop defined first in the `@font-face` blocks. A second warning was used by using the full font family name, and thus removing the font-weight.

The full font family name was found by opening the `tff` in IntelliJ. The font viewer showed the font name as the first line.


chimp1984 (approved):
utACK

PR #308, opened 2022-05-10
Split WalletService into BitcoinWalletService and LiquidWalletService

alvasw:
- Add RpcConfig verification
- Split RpcConfig into DaemonRpcConfig and WalletRpcConfig
- Add Wallet submenu in UI for both cryptocurrencies
- Split WalletStore into BitcoinWalletStore and LiquidWalletStore
- Persist RPC config after verifying whether they are correct


wallets/src/main/java/bisq/wallets/RpcConfig.java
```diff
@@ -15,62 +15,68 @@
...
-    public WalletStore() {
-    }
-
-    public WalletStore(BitcoinWalletStore bitcoinWalletStore, LiquidWalletStore liquidWalletStore) {
-        this.bitcoinWalletStore = bitcoinWalletStore;
-        this.liquidWalletStore = liquidWalletStore;
-    }
+@Builder
+@Getter
+public class RpcConfig implements PersistableStore<RpcConfig> {
```
chimp1984: I would prefer to stick with the convention to call classes used for persistence store -> RpcConfigStore.

wallets/src/main/java/bisq/wallets/stores/BitcoinWalletStore.java
```diff
@@ -21,34 +21,46 @@
...
 
 import java.util.HashSet;
+import java.util.Optional;
 import java.util.Set;
 
 @Slf4j
 public class BitcoinWalletStore implements PersistableStore<BitcoinWalletStore> {
+    @Getter
+    @Setter
+    private Optional<RpcConfig> rpcConfig = Optional.empty();
```
chimp1984, line 38: Ah the RpcConfig is using inside the walletstores. In that case ignore my previous comment, but I think RpcConfig does not require to implement PersistableStore if it is not stored itself. PersistableStore is used for the classes which used for the storage file.

chimp1984 (changes requested):

chimp1984:
Does the initialize call on both wallets cause some delay/resources (beside creating directories)?
As those wallets are optional I guess we should delay initialisation to the moment when the user decides to select a wallet for usage. We can leave it for the moment and once the UI is more final see how to deal with it best.

wallets/src/main/java/bisq/wallets/RpcConfig.java
```diff
@@ -20,15 +20,29 @@
...
-
 @Builder
 @Getter
-public class WalletConfig {
-    private final WalletBackend walletBackend;
-    private final String hostname;
-    private final int port;
-    private final String user;
-    private final String password;
+public class RpcConfig {
```
chimp1984: Should `implements Proto` to mark it as class used for protobuf

wallets/src/main/java/bisq/wallets/RpcConfig.java
alvasw: You're right the `RpcConfig` does not need to implement `PersistableStore`. I updated the PR.

wallets/src/main/java/bisq/wallets/stores/BitcoinWalletStore.java
alvasw: You're right the `RpcConfig` does not need to implement `PersistableStore`. I updated the PR.

alvasw:
> Does the initialize call on both wallets cause some delay/resources (beside creating directories)? As those wallets are optional I guess we should delay initialisation to the moment when the user decides to select a wallet for usage. We can leave it for the moment and once the UI is more final see how to deal with it best.

No, the initialize call does nothing except creating the directories. It only instantiates a wallet implementation if a persisted RpcConfig exists for a wallet. Otherwise, the UI later calls `initializeWallet` to instantiate a wallet implementation.

alvasw:
I rebased this PR with main and resolved the merge conflict.

chimp1984:
> I rebased this PR with main and resolved the merge conflict.

I think there is still an unresolved remark: `Should implements Proto to mark it as class used for protobuf` for RpcConfig

alvasw:
> I think there is still an unresolved remark: `Should implements Proto to mark it as class used for protobuf` for RpcConfig

`RpcConfig` implements `Proto` now.

chimp1984 (approved):
utACK

desktop/src/main/java/bisq/desktop/primary/main/left/LeftNavView.java
```diff
@@ -298,20 +311,30 @@ private void maybeAnimateMark() {
             double targetY = menuTop + selectedLeftNavButton.getBoundsInParent().getMinY();
             if (selectedLeftNavButton instanceof LeftNavSubButton) {
                 targetY += tradeSubMenuItems.getLayoutY();
+                targetY += walletSubMenuItems.getLayoutY();
```
ghost, line 314: @alvasw I see a few UI issues with wallet submenu:
- when Trade menu is selected Wallet menu is incorrectly highlighted
- when any protocol from Trade submenu is selected then green marker moves too low

Preview:
navigation-issue

alvasw: Oops, I just saw that. You're right. Can you fix it? I have a lot of other things to do at the moment.

ghost: Yes - I'll be updating this menu soon.

alvasw: Thank you!

PR #309, opened 2022-05-10
Left navigation redesign

xyzmaker123:
Redesign of left navigation to match [designs](https://xd.adobe.com/view/9cc76e75-baf3-4f1e-be2c-469d46530590-c994/screen/2003f6b1-f8dd-4b46-87ac-c80b6c92bb2b):
- font size and icons update
- submenu highlight fixes
- small refactoring
- a couple of smaller things

NOTE: Not including network info redesign - I'll prepare it in an upcoming pr


alkum:
I took it for a test drive, looks good. ACK from my side.

Here are a few small things I noticed though:
- the protocols in the Trade view have a slightly different order than the entries in the left menu
- Trade > Grid view: the view needs more vertical space than a full-screen window provides (in my case 1920x1080). As I compare with the List view, I'd say if the Grid view cards would use the same font size as the rows in List view, height wouldn't be a problem. Plus it would be more consistent.
- The top-right inactive tab titles are very hard to notice. The dark grey is too dark. Maybe they could use the same shade of gray as left menu nav inactive elements?

Not sure if these areas were part of your redesign, but figured I'd mention them just in case.

chimp1984:
@alkum 
>     The top-right inactive tab titles are very hard to notice. The dark grey is too dark. Maybe they could use the same shade of gray as left menu nav inactive elements?

Could you tell that pedro directly? on OSX it looks ok to me but i guess on linux its a bit different.

chimp1984 (approved):
utACK

PR #321, opened 2022-05-22
Gradle: Add all libraries to central version catalog

alvasw:
See for more information: https://docs.gradle.org/current/userguide/platforms.html#sub:central-declaration-of-dependencies


chimp1984 (approved):
utACK

PR #331, opened 2022-05-24
Chat view - improve nav UI

xyzmaker123:
- adapt font size and weight to design
- render market icons (icons gathered from screenshots with gimp so not looks perfect atm)

Preview:
nav-preview


chimp1984 (approved):
utACK

PR #334, opened 2022-05-27
REST API without spring PoC

salokiam:
This PR is WIP and serve as Proof of Concept. It show how we can expose a REST api and documentation using swagger. This approach uses standalone JDK HTTP Server and Jersey. It uses DTOs as suggested by @xyzmaker123  for json parsing and swagger resolution.
It creates a new gradle module 'rest' and disables module 'api'. I am leaving 'api' in the code base for reference and will delete it in the next PR.
MainClass to start is 'RestApplication', then point you browser to [http://localhost:8082/doc/v1/index.html](url)


chimp1984 (approved):
utACK

PR #378, opened 2022-06-12
Add Electrum RPC and Notification Support

alvasw:
- [x] broadcast
- [x] create
- [x] deserialize
- [x] getbalance
- [x] getseed
- [x] gettransaction
- [x] getunusedaddress
- [x] listunspent
- [x] loadwallet
- [x] notify
- [x] password
- [x] payto
- [x] signmessage
- [x] signtransaction
- [x] verifymessage

### DIFF
--- a/wallets/electrum/build.gradle
+++ b/wallets/electrum/build.gradle
@@ -1,0 +1,25 @@
+plugins {
+    id 'bisq.java-library'
+    id 'bisq.protobuf'
+}
+
+ext {
+    swaggerVersion = '2.2.0'
+    jerseyVersion = '3.0.4'
+}
+
+dependencies {
+    implementation("bisq:persistence")
+    implementation project(':core')
+    implementation libs.jackson.databind
+
+    implementation("org.glassfish.jersey.containers:jersey-container-jdk-http:$jerseyVersion")
+    implementation("org.glassfish.jersey.media:jersey-media-json-jackson:$jerseyVersion")
+    implementation("org.glassfish.jersey.inject:jersey-hk2:$jerseyVersion")
+    implementation("org.glassfish.jaxb:jaxb-runtime:3.0.2")
+    implementation("io.swagger.core.v3:swagger-jaxrs2-jakarta:$swaggerVersion")
+
+    testImplementation project(':bitcoind')
+    testImplementation project(':regtest')
+    testImplementation libs.assertj.core
+}
--- a/wallets/electrum/src/main/java/bisq/wallets/electrum/notifications/ElectrumNotifyApi.java
+++ b/wallets/electrum/src/main/java/bisq/wallets/electrum/notifications/ElectrumNotifyApi.java
@@ -1,0 +1,57 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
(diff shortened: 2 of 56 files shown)


wallets/electrum/src/main/java/bisq/wallets/electrum/notifications/ElectrumNotifyApi.java
```diff
@@ -0,0 +1,57 @@
...
+@Path("/electrum-notify")
+public class ElectrumNotifyApi {
+
+    public static final String ENDPOINT_NAME = "electrum-notify";
+
+    public interface Listener {
+        void onAddressStatusChanged(String address, String status);
+    }
+
+    private static final CopyOnWriteArrayList<Listener> sListeners = new CopyOnWriteArrayList<>();
```
chimp1984, line 39: Is there a reason to call it `sListeners` instead of `listeners`?
We also use as type usually the base type (List instead of CopyOnWriteArrayList)

wallets/electrum/build.gradle
```diff
@@ -0,0 +1,25 @@
...
+    swaggerVersion = '2.2.0'
+    jerseyVersion = '3.0.4'
+}
+
+dependencies {
+    implementation("bisq:persistence")
+    implementation project(':core')
+    implementation libs.jackson.databind
+
+    implementation("org.glassfish.jersey.containers:jersey-container-jdk-http:$jerseyVersion")
```
chimp1984, line 16: Is a servlet container needed or is only the http server needed? We use for the Rest API the jdk http server to limit dependencies. Maybe we can synchronize with REST API dependencies. @salokiam will have better overview in that area...

chimp1984 (approved):
utACK

The NITs can be addressed in an upcoming PR... Maybe good to discuss the http server dependencies with @salokiam

wallets/electrum/build.gradle
salokiam: `implementation("org.glassfish.jersey.containers:jersey-container-jdk-http:$jerseyVersion")`
from the name it looks like a servlet container, but it is not. Its just an small adapter to use JDK http inside the Jersey framework. Actually, I am using the dependency for the Rest-Api

wallets/electrum/src/main/java/bisq/wallets/electrum/notifications/ElectrumNotifyRequest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @XmlElement(name = "address")
+    public String getAddress() {
+        return address;
+    }
+
+    @XmlElement(name = "status")
+    public String getStatus() {
+        return status;
+    }
```
salokiam, line 39: Using the @Xml... annotations here doesnt make sense to me. In this case they are useless. Besides that, if you want to rename the property, why dont you use @JsonProperty as you have been using in some other places. The @Xml annotation would only make sense if you want to use the JAXB standard, but we are locked to json and jackson anyway.

wallets/electrum/build.gradle
chimp1984: Ah ok.

wallets/electrum/src/main/java/bisq/wallets/electrum/notifications/ElectrumNotifyRequest.java
alvasw: @salokiam Thank you, I'll update it! I'm not familiar with the swagger library. This is from their "Getting Started" sample.

PR #385, opened 2022-06-14
Electrum RPC support follow up

alvasw:
- [ElectrumNotifyApi: Change to List<> base type and rename field](https://github.com/bisq-network/bisq2/commit/ac70ec36912c1739d0081b694d1023b935879376)
- [ElectrumNotifyRequest: Switch from XML to JsonProperty](https://github.com/bisq-network/bisq2/commit/fc4f9006e8d87276af1ee166183761d0b45a95dc)
- [Gradle: Move glassfish and swagger dependencies to version catalog](https://github.com/bisq-network/bisq2/commit/1aecacb3aa70ac395253281ec572531a361859d5)

### DIFF
--- a/wallets/electrum/src/main/java/bisq/wallets/electrum/notifications/ElectrumNotifyApi.java
+++ b/wallets/electrum/src/main/java/bisq/wallets/electrum/notifications/ElectrumNotifyApi.java
@@ -17,34 +17,36 @@
 
 package bisq.wallets.electrum.notifications;
 
-import io.swagger.v3.oas.annotations.Parameter;
 import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
-import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.core.MediaType;
 import lombok.extern.slf4j.Slf4j;
 
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 
+import static bisq.wallets.electrum.notifications.ElectrumNotifyApi.ENDPOINT_NAME;
+
 @Slf4j
-@Path("/electrum-notify")
+@Path("/" + ENDPOINT_NAME)
 public class ElectrumNotifyApi {
 
     public static final String ENDPOINT_NAME = "electrum-notify";
-
-    public interface Listener {
-        void onAddressStatusChanged(String address, String status);
-    }
-
     private static final List<Listener> listeners = new CopyOnWriteArrayList<>();
 
+    public interface Listener {
+
+        void onAddressStatusChanged(String address, String status);
+
+    }
+
     @POST
-    @Consumes("application/json")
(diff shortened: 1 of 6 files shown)


chimp1984 (approved):
utACK
@salokiam Can you review as well?

wallets/electrum/src/main/java/bisq/wallets/electrum/notifications/ElectrumNotifyRequest.java
```diff
@@ -17,27 +17,19 @@
...
-    }
-
-    @XmlElement(name = "status")
-    public String getStatus() {
-        return status;
-    }
-
     @Override
     public String toString() {
         return "ElectrumNotifyRequest{" +
```
salokiam: why not use lombok @ToString ?

salokiam (commented):
regarding  ElectrumNotifyApi.ENDPOINT_NAME;
you could use this constant in the annotation as well, like @Path("/" + ENDPOINT_NAME)
that way you can dont need the string twice.

wallets/electrum/src/main/java/bisq/wallets/electrum/notifications/ElectrumNotifyApi.java
```diff
@@ -36,22 +37,22 @@ public interface Listener {
...
-    private static final CopyOnWriteArrayList<Listener> sListeners = new CopyOnWriteArrayList<>();
+    private static final List<Listener> listeners = new CopyOnWriteArrayList<>();
 
     @POST
     @Consumes("application/json")
     public Response notifyEndpoint(@Parameter(required = true) ElectrumNotifyRequest request) {
         log.info(request.toString());
-        sListeners.forEach(listener -> listener.onAddressStatusChanged(request.getAddress(), request.getStatus()));
+        listeners.forEach(listener -> listener.onAddressStatusChanged(request.getAddress(), request.getStatus()));
         return Response.ok().entity("SUCCESS").build();
```
salokiam: return "SUCCESS"  would do the same. You notation is usefull if you may want to send back anything other than Response.ok() in special case.

wallets/electrum/src/main/java/bisq/wallets/electrum/notifications/ElectrumNotifyApi.java
```diff
@@ -36,22 +37,22 @@ public interface Listener {
         void onAddressStatusChanged(String address, String status);
     }
 
-    private static final CopyOnWriteArrayList<Listener> sListeners = new CopyOnWriteArrayList<>();
+    private static final List<Listener> listeners = new CopyOnWriteArrayList<>();
 
     @POST
     @Consumes("application/json")
     public Response notifyEndpoint(@Parameter(required = true) ElectrumNotifyRequest request) {
```
salokiam: why are you using swagger annotation here? Not needed if swagger is not used.
Also MediaType.APPLICATION_JSON could be used instead of the String "application/json"

salokiam (approved):

chimp1984 (approved):
utACK

Issue #52, opened 2022-01-30
Reduce Wallet Unlocked Timeout


alvasw:
At the moment the default wallet timeout (wallet unlocked) is set to 24 hours. We have to reduce it later to something like 5 minutes.
https://github.com/bisq-network/bisq2/blob/2b6a99ed7b33b51f4cd13c09842800c4b46c7bf9/wallets/src/main/java/bisq/wallets/bitcoind/BitcoindWalletBackend.java#L17
#### Tasks
- [x] Identify RPC Calls that need wallet to be unlocked
- [x] Unlock wallet before making RPC call (if locked)
- [x] Lock wallet after RPC call

PR #386, opened 2022-06-14
Bitcoind: Unlock and lock wallet between RPC calls

alvasw:
Fixes https://github.com/bisq-network/bisq2/issues/52

### DIFF
--- a/desktop/src/main/java/bisq/desktop/primary/main/content/wallet/send/WalletSendController.java
+++ b/desktop/src/main/java/bisq/desktop/primary/main/content/wallet/send/WalletSendController.java
@@ -21,6 +21,8 @@ import bisq.desktop.common.view.Controller;
 import bisq.wallets.core.WalletService;
 import lombok.Getter;
 
+import java.util.Optional;
+
 public class WalletSendController implements Controller {
     private final WalletService walletService;
     private final WalletSendModel model;
@@ -42,6 +44,8 @@ public class WalletSendController implements Controller {
     }
 
     public void onSendButtonClicked() {
-        walletService.sendToAddress(model.getAddress(), Double.parseDouble(model.getAmount()));
+        String passphraseString = model.getPassphrase();
+        Optional<String> passphrase = Optional.ofNullable(passphraseString);
+        walletService.sendToAddress(passphrase, model.getAddress(), Double.parseDouble(model.getAmount()));
     }
 }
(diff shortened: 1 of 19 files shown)


desktop/src/main/java/bisq/desktop/primary/main/content/wallet/send/WalletSendController.java
```diff
@@ -42,6 +44,8 @@ public void onDeactivate() {
     }
 
     public void onSendButtonClicked() {
-        walletService.sendToAddress(model.getAddress(), Double.parseDouble(model.getAmount()));
+        String passphraseString = model.getPassphrase();
+        Optional<String> passphrase = passphraseString.isEmpty() ? Optional.empty() : Optional.of(passphraseString);
```
chimp1984: Could use: `Optional.ofNullable(passphraseString);`

desktop/src/main/java/bisq/desktop/primary/main/content/wallet/send/WalletSendView.java
```diff
@@ -37,6 +38,10 @@ public WalletSendView(WalletSendModel model, WalletSendController controller) {
         TextField amountTextField = createTextField(Res.get("amount") + ":");
         amountTextField.textProperty().bindBidirectional(model.amountProperty());
 
+        PasswordField passphraseTextField = new PasswordField();
+        passphraseTextField.setPromptText(Res.get("passphrase") + ":");
+        passphraseTextField.textProperty().bindBidirectional(model.passphraseProperty());
```
chimp1984: Bindings should be set in the `onViewAttached` method and unbind called at `onViewDetached` to avoid potential memory leaks. Most of the time it would not cause a leak as the reference resource gets removed as well but there might be edge cases where its not the case and then its very hard to find if we dont follow that more strict pattern.

desktop/src/main/java/bisq/desktop/primary/main/content/wallet/send/WalletSendModel.java
```diff
@@ -24,6 +24,7 @@
 public class WalletSendModel implements Model {
     private final StringProperty addressProperty = new SimpleStringProperty(this, "address");
     private final StringProperty amountProperty = new SimpleStringProperty(this, "amount");
+    private final StringProperty passphraseProperty = new SimpleStringProperty(this, "passphrase");
 
     public StringProperty addressProperty() {
```
chimp1984, line 29: I don't follow the JavaFx convention of offering properties and the protpery value in getters as it causes more boilerplate code and Lombok does not support those types of getters unfortunately. But this version is for sure more the JavaFx standard way...

desktop/src/main/java/bisq/desktop/primary/main/content/wallet/send/WalletSendController.java
alvasw: I cannot use it here. Passing an empty string to `Optional.ofNullable("")` does not return an empty `Optional`. It returns an `Optional<String>` with `""` as its value.

alvasw: You're right. This is unnecessarily complicated. I simplified this by switching to `Optional.ofNullable(passphraseString)` and handling the empty string case in the Wallet implementation.

chimp1984 (approved):
utACK

PR #389, opened 2022-06-14
Chat API and Offer API, templates for using Dto-classes in constructing further web APIs

salokiam:
Showing 2 more examples for web api and swagger documentation: Chat API and Offer API. Added automated coping of bean properties. Also this codesshows the handlng of Dto-classes using the base class read-method.


restApi/src/main/java/bisq/restApi/dto/OfferDto.java
```diff
@@ -0,0 +1,40 @@
...
+@Slf4j
+@Data
+@ToString
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false)
+@Schema(name = "Offer")
+public class OfferDto extends BaseDto<OfferDto> {
+
+    @EqualsAndHashCode.Include
+    protected String id;
+    protected long date; //long ???
```
chimp1984: We use the value in case of wrapper types -> date as long value in dto/protobuf context, only in java domain we use Date object. Similar with Coin where we use the satoshi value as long and Coin only in higher level APIs.

restApi/src/main/java/bisq/restApi/dto/OfferDto.java
```diff
@@ -0,0 +1,40 @@
...
+    protected String makeNodeId; // instead of network ID
```
chimp1984: Should be `makerNodeId` - might be wrong in the offer domain object as well... its old code and might change...

restApi/src/main/java/bisq/restApi/dto/PublicDiscussionChannelDto.java
```diff
@@ -0,0 +1,21 @@
...
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false)
+@Data
+@Schema(name = "PublicDiscussionChannel")
+public class PublicDiscussionChannelDto extends BaseDto<PublicDiscussionChannelDto> {
+
+    @EqualsAndHashCode.Include
+    protected String id;
+    private String channelName;
+    private String description;
+    private String channelAdminId; // TODO do we need the admin and moderater in the web api?
```
chimp1984: Not sure yet if that will stay... I guess easiest is to mirror 1:1 the domain in the API dto classes.

chimp1984: But I guess admin and moderators will be handled differently as the channel should be immutable and moderators could change

restApi/src/main/java/bisq/restApi/dto/BaseDto.java
```diff
@@ -0,0 +1,13 @@
+package bisq.restApi.dto;
+
+import lombok.Data;
+
+@Data
+public class BaseDto<T extends BaseDto> {
```
chimp1984: Should be `BaseDto<T extends BaseDto<?>>` to avoid IDE complaining...

restApi/src/main/java/bisq/restApi/dto/BaseDto.java
```diff
@@ -0,0 +1,13 @@
...
+
+    public T read(Object domainObject) {
```
chimp1984: Maybe we can use a different name here. It is more like filling one object with the content from another. Usually we use static methods with `from` as method name but I think that works better in a static context. Could be refactored all as well to use it statically... But no need to change, just a more clear name would be good as read does not imply that we also write what we read to ourself.

restApi/src/main/java/bisq/restApi/dto/DtoUtil.java
```diff
@@ -0,0 +1,45 @@
+package bisq.restApi.dto;
+
+import java.beans.BeanInfo;
+import java.beans.Introspector;
+import java.beans.PropertyDescriptor;
+import java.lang.reflect.Method;
+
+public class DtoUtil {
+
+    public static <T> T copyBeanProperties(Object fromObject, T toObject) {
```
chimp1984: I am not a fan of interfaced where arguments get altered, is against functional programming rules.
Also we return the same value we pass... 
As the toObject is used as class only, we could pass the class and also construct the object (i think it is always used with no args) here. Then I think a static context from the caller would also work better.

restApi/src/main/java/bisq/restApi/dto/BaseDto.java
salokiam: In my first version I had a static creation method. I changed it because it has better way of reusing base class implementation (and therfore less boilerplate code)

restApi/src/main/java/bisq/restApi/dto/DtoUtil.java
salokiam: yep, was my first thought as well. Basically its a tradeoff, use static create methods, then we need to have it always,even if copyBeanProperties for simple Dto would be enough. Because the static creation method cant be inherited. Which way shall we go?

restApi/src/main/java/bisq/restApi/dto/BaseDto.java
salokiam: change the name to 'loadFieldsFrom(...)'

restApi/src/main/java/bisq/restApi/dto/KeyPairDto.java
```diff
@@ -22,14 +22,17 @@
 @Getter
 @Schema(name = "KeyPair")
 public class KeyPairDto {
-    byte[] publicKey;
-    byte[] privateKey;
+    protected byte[] publicKey;
+    protected byte[] privateKey;
 
-    public KeyPairDto(java.security.KeyPair k) {
+    public static KeyPairDto from(java.security.KeyPair k) {
```
chimp1984: Prefer to not use single char names beside in loops.

restApi/src/main/java/bisq/restApi/dto/KeyPairDto.java
```diff
@@ -22,14 +22,17 @@
...
         if (k != null) {
```
chimp1984: Not sure if we should do parameter null checks. By default they have to be non null. If its a valid case then the caller should do the check. 
In the UI code we have that more frequently as the JavaFX components use null commonly (unfortunately).

restApi/src/main/java/bisq/restApi/dto/KeyPairDto.java
```diff
@@ -22,14 +22,17 @@
 @Getter
 @Schema(name = "KeyPair")
 public class KeyPairDto {
```
chimp1984: I think we should make all dto classes final.

restApi/src/main/java/bisq/restApi/dto/KeyPairDto.java
```diff
@@ -22,14 +22,17 @@
...
-    byte[] publicKey;
-    byte[] privateKey;
+    protected byte[] publicKey;
```
chimp1984: I guess that can be private, no?

restApi/src/main/java/bisq/restApi/dto/KeyPairDto.java
```diff
@@ -22,14 +22,17 @@
...
+    protected byte[] privateKey;
 
-    public KeyPairDto(java.security.KeyPair k) {
+    public static KeyPairDto from(java.security.KeyPair k) {
         if (k != null) {
-            publicKey = k.getPublic().getEncoded();
-            privateKey = k.getPrivate().getEncoded();
+            KeyPairDto dto = new KeyPairDto();
+            dto.privateKey = k.getPublic().getEncoded();
```
chimp1984: Should be `dto.publicKey`

chimp1984 (changes requested):

chimp1984 (approved):
utACK

PR #408, opened 2022-06-28
Set Java 11 as Compilation Target

alvasw:
We do still use the Java 17 toolchain because we need jpackage to create the binaries. Since version 9, the Java compiler can be configured to produce bytecode for an older Java version while making sure the code does not use any APIs from a more recent version.

See https://docs.gradle.org/current/userguide/building_java_projects.html#sec:java_cross_compilation

Other changes:
- [Java 11: Replace Stream.toList() by Stream.collect(Collectors.toList())](https://github.com/bisq-network/bisq2/commit/5fa89b54daec1aa91ded5699789129bfcd4da21c)
- [JavaFX 11: Remove GraphicsContext2D.setImageSmoothing(...)](https://github.com/bisq-network/bisq2/commit/6617a1219e5a59969e19c2eb316794066e782bd4)

### DIFF
--- a/build-logic/commons/src/main/groovy/bisq.java-conventions.gradle
+++ b/build-logic/commons/src/main/groovy/bisq.java-conventions.gradle
@@ -9,8 +9,13 @@ repositories {
 
 java {
     toolchain {
+        // We use the Java 17 toolchain to use jpackage to create the binaries.
         languageVersion = JavaLanguageVersion.of(17)
     }
+}
+
+compileJava {
+    options.release = 11
 }
 
 dependencies {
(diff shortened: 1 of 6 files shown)


chimp1984:
Gradle build worked out of the box. For IDE build I needed re-import gradle dependencies and did a invalidate cache and restart (that alone did not fix it), after changing java version in module settings to Java 11. Now it works.

chimp1984:
I have not tried out java packager...
With `languageVersion = JavaLanguageVersion.of(17)` we stick to java 17 packager, right?

PR #414, opened 2022-06-30
Bundle Electrum Binaries with Bisq

alvasw:
Package the Electrum binaries after verifying the signatures and copy it to the Java resources.


chimp1984 (approved):
utACK

PR #418, opened 2022-07-05
Run Bitcoind Integration Tests on CI

alvasw:
Check whether bitcoind is installed and run integration tests depending on its result.


chimp1984 (approved):
utACK

PR #428, opened 2022-07-08
Refactor Bitcoin Core and Elements Core Installer

alvasw:
- Create Installer Interface
- Create Bitcoin Core Installer
- Override properties in Elements Core Installer


chimp1984 (approved):
utACK

PR #436, opened 2022-07-15
desktop/desktopapp: Target Java 17

alvasw:
JPackage cannot sign the DMG image on Mac because the JavaFX 11 Mac dylib libraries are not signed. It's less work to target Java 17 (and JavaFx 17) instead of adapting the build process to sign the dylib libraries before creating and signing the DMG image.


chimp1984 (approved):
utACK

PR #443, opened 2022-07-23
375-I2P-Improvements

c4talys7:
Initial commit for changes related to improvements on I2P layer
Improvements to the I2P Layer + Implementation of log mechanism for I2P local router (enabled by config variable)
Extracting I2pRouter to it's own class and making a singleton out of it.
From now on, whenever using the embedded router, only one router will be started.
Making the client always connect to the configured port for the external router
Adding code to check if port is in use.

Dev: @c4talys7

### DIFF
--- a/i2p/src/main/java/bisq/i2p/I2pEmbeddedRouter.java
+++ b/i2p/src/main/java/bisq/i2p/I2pEmbeddedRouter.java
@@ -1,0 +1,282 @@
+package bisq.i2p;
+
+import bisq.common.timer.Scheduler;
+import bisq.i2p.util.I2PLogManager;
+import lombok.extern.slf4j.Slf4j;
+import net.i2p.client.streaming.I2PSocketManager;
+import net.i2p.client.streaming.I2PSocketManagerFactory;
+import net.i2p.router.CommSystemFacade;
+import net.i2p.router.Router;
+import net.i2p.router.RouterContext;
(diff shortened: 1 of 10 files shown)


c4talys7:
@chimp1984 , find the PR for the I2P improvements discussed so far.

chimp1984 (approved):
utACK

i2p/src/main/java/bisq/i2p/I2pEmbeddedRouter.java
```diff
@@ -0,0 +1,282 @@
...
+        p.put("i2np.bandwidth.inboundKBytesPerSecond", inboundKBytesPerSecond);
+        p.put("i2np.bandwidth.inboundBurstKBytesPerSecond", inboundKBytesPerSecond);
+        p.put("i2np.bandwidth.outboundKBytesPerSecond", outboundKBytesPerSecond);
+        p.put("i2np.bandwidth.outboundBurstKBytesPerSecond", outboundKBytesPerSecond);
+        p.put("router.sharePercentage", bandwidthSharePercentage);
+
+        p.put("i2cp.disableInterface", "true");
+        p.put("i2np.ntcp.nodelay", "true");
+        p.put("router.encType","4");
+        p.put("router.useShortTBM","true");
```
zlatinb, line 254: These 3 settings became defaults a while ago, it is safe to remove them.

PR #450, opened 2022-07-29
I2P - Fix session issue

c4talys7:
Fixed issue where whenever the sessions would start on the desktop application, the port assigned would be different, thus recreating the session on every run of the app.

Related to feature/375-I2P-Improvements

Dev: @c4talys7

### DIFF
--- a/application/src/main/resources/default.conf
+++ b/application/src/main/resources/default.conf
@@ -56,7 +56,7 @@ application {
     }
     
     network = {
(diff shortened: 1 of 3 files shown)


c4talys7:
@chimp1984,

This is the fix to the issue related to the sessions getting recreated over each port change.. It was a simple one in the end.

Regards,

c4talys7

chimp1984 (approved):
utACK

PR #453, opened 2022-08-01
ElectrumDownloader: Skip download if the file exists already

alvasw:
Verify the hash and perform the signature check if the file exists.


chimp1984 (approved):
utACK

PR #461, opened 2022-08-04
GitHub Actions: Install ElectrumX

alvasw:
- Download and unpack ElectrumX
- Install Python 3.10
- Install leveldb on MacOS
- Install ElectrumX dependencies
- Add binaries to path


chimp1984 (approved):
utACK

PR #466, opened 2022-08-07
I2P - Autodetect External I2P Router

c4talys7:
This change checks if there's an I2P router running locally on port 7654, if there's one and the user is trying to run an embedded router, it will ignore the parameter for the embedded router and use the external router instead.

Developer: @c4talys7


c4talys7:
@chimp1984 , follows the PR to check if the local I2P router is running automagically

chimp1984 (approved):
utACK

PR #471, opened 2022-08-27
Electrum: Implement WindowsElectrumRegtestSetup

alvasw:
- Run tests against external Bitcoin Core and ElectrumX
- Wait for exact blockHashes instead of number of mined blocks


chimp1984 (approved):
utACK

PR #473, opened 2022-08-27
Electrum: Create configFile before start

alvasw:
- Enable RPC endpoint
- Enable logging (needed to parse RPC "ready" signal)


chimp1984 (approved):
utACK

PR #480, opened 2022-08-28
Task 2 (CW): Reputation text overhaul

uranuspucksaxophone:


i18n/src/main/resources/default.properties
```diff
@@ -1015,25 +1017,28 @@ reputation.accountAge.tab3=How-to
 reputation.accountAge.infoHeadline=Provide trust
 reputation.accountAge.info=By linking your Bisq 1 'account age' you can provide some level of trust.\n\n\
   There are some reasonable expectations that a user who has used Bisq for some time is an honest user. But of course \
-  that expectation is rather weak compared to other forms of reputation where the user provides more "skin in the game" with \
+  that expectation is rather weak compared to other forms of reputation where the user provides stronger proof with \
   allocation of some financial resources.
 reputation.accountAge.infoHeadline2=Privacy implications
-reputation.accountAge.info2=Linking your Bisq 1 account with Bisq 2 has some implications on your privacy. \
+reputation.accountAge.info2=
+  Linking your Bisq 1 account with Bisq 2 has some implications on your privacy. \
```
chimp1984: Those line breaks require a `\` otherwise the string is not parsed correctly. If linebreak is not needed better dont use any at all...

chimp1984 (changes requested):

i18n/src/main/resources/default.properties
ghost: ok, this is a bit confusing, let me try to fix it...

chimp1984 (approved):
utACK

PR #482, opened 2022-08-29
Electrum: Watch log files for "ready" signal

alvasw:
We wait until the Electrum daemon is ready to accept RPC calls by waiting until the Electrum process outputs a ready signal in its process output. However, this only happens on Mac and Windows. This change parses the log file for the "ready" signal.


chimp1984 (commented):
utACK

How does it deal with Linux?

PR #486, opened 2022-08-31
Gradle: Fix compiler mismatch bug (bug in Gradle)

alvasw:
Currently, it's impossible to target any Java version other than 1.8
because it is hardcoded into Gradle. This bug has been fixed. The
compiler warning "'compileJava' task (current target is 17) and
'compileKotlin' task (current target is 1.8) jvm target compatibility
should be set to the same Java version." will disappear now.

Changes:
- Update to Gradle 7.5.1
- Update the kotlin-dsl plugin


chimp1984 (approved):
utACK

PR #488, opened 2022-09-01
[WIP] CW: Academy text

ghost:


chimp1984 (changes requested):
NACK
Does not match language style used on the webpage and elsewhere.

PR #490, opened 2022-09-09
Produce CI/CD artifact

ghost:


chimp1984:
Which artefacts do you want to upload? Does not feel right to me. @alvasw Can you have a look?

chimp1984 (changes requested):
NACK

PR #493, opened 2022-09-12
Add pow to network layer

chimp1984:


chimp1984:
Missing pieces:
- Send Load to peers when change of load exceeds a certain threshold
- Maintain an inxed of the load changes and add that data to the token to look up for a past load object to avoid race conditions when we receive a message with a previous load
- Impl. formula for difficulty from cost and load
- Setup cost values for all messages
- Get meaningful load figures

PR #614, opened 2022-12-28
Downgraded Protobuf Java to Protobuf Lite

niyid:
This is to ensure protocol compatibility across all devices (including PC and Android).

The rationale is to keep the code for PC and Android (and other devices) the same across all the devices to prevent code fragmentation.

Android has several limitations one of which it only supports Protobuf Lite. And Protobuf Lite limits the message type to `com.google.protobuf.MessageLite`. The standard `com.google.protobuf.Message` is not available.

Also `com.google.protobuf.Any` in Protobuf Lite is missing helper methods `pack()` and `unpack()`.

Without harmonizing the protocol, PC and Android devices will not be able to communicate as the PC will use the standard Protobuf Java while the Android will use Protobuf Lite.

### DIFF
--- a/account/src/main/java/bisq/account/AccountStore.java
+++ b/account/src/main/java/bisq/account/AccountStore.java
@@ -61,7 +61,7 @@ public final class AccountStore implements PersistableStore<AccountStore> {
     public ProtoResolver<PersistableStore<?>> getResolver() {
         return any -> {
             try {
-                return fromProto(any.unpack(bisq.account.protobuf.AccountStore.class));
+                return fromProto(bisq.common.util.ProtobufUtils.unpack(any, bisq.account.protobuf.AccountStore.class));
             } catch (InvalidProtocolBufferException e) {
                 throw new UnresolvableProtobufMessageException(e);
             }
--- a/account/src/main/java/bisq/account/accounts/CountryBasedAccountPayload.java
+++ b/account/src/main/java/bisq/account/accounts/CountryBasedAccountPayload.java
@@ -17,7 +17,7 @@
 
 package bisq.account.accounts;
 
-import com.google.protobuf.Message;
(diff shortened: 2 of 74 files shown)


HenrikJannsen:
This comes with lot of changes. I prefer to not add that to the current code base but @niyid should work on his branch and keep it up to date with master.

niyid:
My question is, how will an Android app whose protocol is based on Protobuf
Lite be able to communicate with other applications with protocol based on
Protobuf Java? Remember the limitation in the first place was Protobuf Java
does not work with Android; hence the adoption of Protobuf Lite.

I have noted the risk you pointed out. But that risk can be avoided by
including a Unit Test that runs ABI compatibility test in the build; so
that when a newer version breaks compatibility, it is rejected.


Secondly Protobuf Lite is optimized for size and speed. Protobuf Java
carries a lot of unnecessary bells and whistles.

The point is an Android version is totally essential to grow adoption.

I am curious to learn of any other pathways or alternative solutions to
support Android (and other devices with limited resources such as mobile
and SBUs).

niyid:
> I'm not sure if I understand your question. There's no problem in using Protobuf Lite only on Android. It's the same wireformat (both libraries serialize to the same bytes). You can use the Lite runtime on Android and parse all messages from other nodes.
> 
> For more details see: https://developers.google.com/protocol-buffers/docs/encoding

Thanks a lot for pointing this out. I had discussed with @salokiam and we both worried that using Protobuf Lite on Android would result in not being able to join the network and communicate with the PC apps running on Protobuf Java.

This is good news as now the Java (PC) app does not need to be forcefully "downgraded". If the work I have done is considered useful and will serve any future purpose, I will maintain it as an Android fork. But once again, if it is considered to be of any value.

Regards.

PR #617, opened 2022-12-30
Fix wallet init for osx

HenrikJannsen:


HenrikJannsen:
Based on https://github.com/bisq-network/bisq2/pull/616

PR #621, opened 2023-01-02
Fix Electrum empty password failures

alvasw:
When the Electrum wallet is not encrypted we have to ommit the password field in all JSON requests. Sending an empty string causes the RPC calls to fail.


HenrikJannsen (approved):
ACK

HenrikJannsen:
> @HenrikJannsen Does this work on your OS?

Yes

PR #678, opened 2023-03-07
Linux notification

DrColver:
Version was needed for jpackage to work for linux.

Icon and description show after installing the deb.

Using the the app icon, notification show the icon, but app name is not visible. I don't know where it should show though, it doesn't show when notify-send is run manually either.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/common/notifications/linux/LinuxNotifications.java
+++ b/desktop/src/main/java/bisq/desktop/common/notifications/linux/LinuxNotifications.java
@@ -33,12 +33,11 @@ public class LinuxNotifications implements NotificationsDelegate {
 
     @Override
     public void notify(String title, String message) {
-        List<String> command = new ArrayList<String>();
+        List<String> command = new ArrayList<>();
         command.add("notify-send");
 
-        // todo add icon support
-        //command.add("-i");
-        //command.add("...");
+        command.add("-i");
+        command.add(System.getProperty("java.home") + "/../Bisq_2.png");
 
         command.add("--app-name");
         command.add("Bisq");
@@ -46,7 +45,7 @@ public class LinuxNotifications implements NotificationsDelegate {
         command.add(title);
         command.add(message);
         try {
-            Runtime.getRuntime().exec(command.toArray(new String[command.size()]));
+            Runtime.getRuntime().exec(command.toArray(new String[0]));
         } catch (IOException e) {
             throw new RuntimeException("Unable to notify with Notify OSD", e);
         }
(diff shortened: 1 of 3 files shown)


desktop/src/main/java/bisq/desktop/common/notifications/linux/LinuxNotifications.java
```diff
@@ -33,20 +33,19 @@ public static boolean isSupported() {
...
     public void notify(String title, String message) {
-        List<String> command = new ArrayList<String>();
+        List<String> command = new ArrayList<>();
         command.add("notify-send");
 
-        // todo add icon support
-        //command.add("-i");
-        //command.add("...");
+        command.add("-i");
+        command.add(System.getProperty("java.home") + "/../Bisq_2.png");
```
alvasw, line 40: We cannot assume that the icon is outside the JRE directory. It would be better to load it from the installation or data directory.

desktop/src/main/java/bisq/desktop/common/notifications/linux/LinuxNotifications.java
```diff
@@ -33,20 +33,19 @@ public static boolean isSupported() {
...
 
         command.add("--app-name");
         command.add("Bisq");
 
         command.add(title);
         command.add(message);
         try {
-            Runtime.getRuntime().exec(command.toArray(new String[command.size()]));
+            Runtime.getRuntime().exec(command.toArray(new String[0]));
```
alvasw, line 48: This change drops all other arguments. What was the reason that you did this change?

desktop/src/main/java/bisq/desktop/common/notifications/linux/LinuxNotifications.java
DrColver: This takes it from the installation directory, which only works when installed. Maybe it would be better to include the icon in the data dir?

desktop/src/main/java/bisq/desktop/common/notifications/linux/LinuxNotifications.java
DrColver: `List.toArray(T[] a)` will make a new array if `a` is no big enough, so it won't drop any argument.

I changed it to avoid warnings from the IDE. Since the warnings are useful sometimes it's worth keeping them to a minimum if possible.

alvasw: You're correct. Sorry, my bad!

desktop/src/main/java/bisq/desktop/common/notifications/linux/LinuxNotifications.java
alvasw: It's a misunderstanding, you're code works! Usually, `System.getProperty("java.home")` points to the installation directory of the JRE (not Bisq), see [1]. You're so works because of `jpackage`. We'll move away from it later, but for now this is good.

[1] https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html

desktop/src/main/java/bisq/desktop/common/notifications/linux/LinuxNotifications.java
HenrikJannsen: Which warning in the IDE was displayed? Do you use IntelliJ? I do not see any...

PR #687, opened 2023-03-10
Fix notifications constructor

DrColver:
Switch doesn't work when support fails and it falls through to the next OS. Remove tray icon when using AwtNotifications


HenrikJannsen (approved):
utACK

HenrikJannsen:
Good find! ;-)

PR #690, opened 2023-03-15
Disable privateMsgButton in trade channel

DrColver:
An exception is thrown when trying to open a private chat in the trade channel. Private chat is only allowed after taking a trade offer.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/primary/main/content/chat/ChatController.java
+++ b/desktop/src/main/java/bisq/desktop/primary/main/content/chat/ChatController.java
@@ -116,6 +116,7 @@ public abstract class ChatController<V extends ChatView, M extends ChatModel> ex
             userProfileSidebar.setOnMentionUserHandler(chatMessagesComponent::mentionUser);
             model.setChatUserDetails(Optional.of(userProfileSidebar));
             model.getChatUserDetailsRoot().set(userProfileSidebar.getRoot());
+            userProfileSidebar.setChannelDomain(chatMessagesComponent.getChannelDomain());
         });
 
         searchTextPin = EasyBind.subscribe(model.getSearchText(), searchText -> {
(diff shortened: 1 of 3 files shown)


HenrikJannsen:
Hm... I think we should permit in trade channel private messages also in case no offer is taken. A taker might have questions to the maker and might prefer to do that as private message. I would need to look into the code what causes the exception and what could be required to support non-trade PMs. Maybe it complicate things too much if we allow that...

HenrikJannsen:
@DrColver 
I had quite some larger pending work which I just put into a PR (https://github.com/bisq-network/bisq2/pull/694) to not get too far away from main branch. Did not had much time recently to work on it... If you have any changes in the notifications area maybe better to base the PR on my branch if it takes longer until merged.

Are you in the matrix channels as well?

DrColver:
The exception is thrown on purpose when someone tries to private message in the trade channel, so it looks like it is not intended to be allowed.

If it should be allowed, maybe it would be better to have a different button on the offer message to just ask a question and not show intention to trade. Then the offer owner knows what the message is in reference to.

Not on matrix. How to join?

DrColver:
With the changes in https://github.com/bisq-network/bisq2/pull/700 it would be possible to include the related offer when opening a new private channel. The offer could be used as a reference when sending a private message without taking the offer.

This PR should not be merged and usage of the correct private message handlers should be used.

HenrikJannsen:
> With the changes in #700 it would be possible to include the related offer when opening a new private channel. The offer could be used as a reference when sending a private message without taking the offer.
> 

Yes, that would be good. The UX is not fully worked out atm in that area. Feel free to suggest what makes most sense to you.

> This PR should not be merged and usage of the correct private message handlers should be used.

You can convert it to a draft (somewhere at the top right in GH).

DrColver:
Fixed in #805

PR #693, opened 2023-03-20
Change channel on different call stack

DrColver:
Channel change is listened to in multiple places. Causing another channel change before the previous ones are processed will result in the changes being processed in the wrong order by the listeners.

In particular, this fixes the default loading of the PublicTradeChannel after the last PrivateTradeChannel was closed.

Cleanup

### DIFF
--- a/desktop/src/main/java/bisq/desktop/primary/main/content/chat/channels/PublicTradeChannelSelection.java
+++ b/desktop/src/main/java/bisq/desktop/primary/main/content/chat/channels/PublicTradeChannelSelection.java
@@ -27,6 +27,7 @@ import bisq.common.currency.MarketRepository;
 import bisq.common.data.Pair;
 import bisq.common.observable.Pin;
 import bisq.desktop.common.observable.FxBindings;
+import bisq.desktop.common.threading.UIThread;
 import bisq.desktop.common.utils.Icons;
 import bisq.desktop.common.utils.Transitions;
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/primary/main/content/chat/channels/PublicTradeChannelSelection.java
```diff
@@ -309,8 +311,7 @@ protected ListCell<ChannelItem> getListCell() {
                 @Override
                 protected void updateItem(ChannelItem item, boolean empty) {
                     super.updateItem(item, empty);
-                    if (item != null && !empty && item.getChannel() instanceof PublicTradeChannel) {
-                        PublicTradeChannel publicTradeChannel = (PublicTradeChannel) item.getChannel();
+                    if (item != null && !empty && item.getChannel() instanceof PublicTradeChannel publicTradeChannel) {
```
HenrikJannsen: We want to support Java 11 for Android, so we avoided to use features not supported by Java 11.

HenrikJannsen (changes requested):

desktop/src/main/java/bisq/desktop/primary/main/content/chat/channels/PublicTradeChannelSelection.java
DrColver: Understood, will revert all such changes.

PR #694, opened 2023-03-21
Add notifications to chat

HenrikJannsen:
This is still work in progress, but I prefer to get that current state merge to not get too far off from main branch.


i18n/src/main/resources/default.properties
```diff
@@ -46,8 +46,8 @@ markets=Markets
 dashboard=Dashboard
 grid=Grid
 list=List
-chat=Discussions
-learn=Learn
+discussion=Discussions
+learn=LearnPu
```
DrColver, line 50: Mistyped?

DrColver (commented):
Notifications work on Linux. Settings are not respected, but seeing as this is work in progress that should not block merging.

i18n/src/main/resources/default.properties
HenrikJannsen: Ups, yes. Thanks.

DrColver:
ACK

This doesn't break the client, but would not be a good merge once bisq2 is released.

What is the process for merging larger changes? What is the plan after release?

PR #698, opened 2023-03-28
Update top pane userprofile combobox width properly

DrColver:
When userprofile is changed from other places than the top pane the width does not update if the widthproperty listener is removed.


DrColver:
I have been playing around with the bisq easy interface and fixing issues as I go along. I have been trying to add some additional features that will come as a larger PR. I don't expect those features to be accepted as is, it's just a good way to get familiar with the code and fixing minor bugs has been a nice bonus.

Look for drcolver on matrix.

HenrikJannsen (approved):
utACK

PR #699, opened 2023-03-30
Keep previous userprofile selection on nothing selected

DrColver:
When clicking the combobox for user profile selection and then clicking outside without selecting a new userprofile, the combobox is left with no userprofile selected. This fix resets the combobox to the globally selected userprofile.


HenrikJannsen (approved):
utACK

PR #700, opened 2023-04-02
Send unused btc address

DrColver:
## Description
Add feature to send an unused bitcoin address from the internal wallet to the trade peer when completing a trade in the private trade chat.

## Details
This was an exercise in learning a bit more about the code base. If the feature is not appropriate, please disregard. The other PRs are a result of working on this feature.

- Refactor TradeChatOffer
- Add TradeChatOfferInterface for public and private trade messages to use
- Add TradeChatOfferInterface to private messages to have the details of the offer within the private chat
- Add button to automatically get an unused bitcoin address and send message to the bitcoin seller

Further features related to this work would be adding the same functionality for bitcoin sellers. Display fiat payment details with a button push to increase ease of use and avoid errors.

## Questions
- Is bitcoin supposed to be one side of each trade or can other pairs be traded? I looks like bitcoin is intended to be one side of every trade but that seems quite restrictive as a design choice.
- Is default.properties meant to be translated? It looks like it and will affect design when sending generated text messages as text instead of as operations.
- There looks to be a work in process framework for a more extensive trade protocol. Is this discussed somewhere?

### DIFF
--- a/chat/src/main/java/bisq/chat/trade/TradeChatOffer.java
+++ b/chat/src/main/java/bisq/chat/trade/TradeChatOffer.java
@@ -1,4 +1,4 @@
-package bisq.chat.trade.pub;
+package bisq.chat.trade;
 
 
 import bisq.common.currency.Market;
--- a/chat/src/main/java/bisq/chat/trade/TradeChatOfferMessage.java
+++ b/chat/src/main/java/bisq/chat/trade/TradeChatOfferMessage.java
@@ -1,0 +1,26 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.chat.trade;
+
+import java.util.Optional;
+
+public interface TradeChatOfferMessage {
+    Optional<TradeChatOffer> getTradeChatOffer();
+
+    boolean hasTradeChatOffer();
+}
--- a/chat/src/main/java/bisq/chat/trade/priv/PrivateTradeChannelService.java
+++ b/chat/src/main/java/bisq/chat/trade/priv/PrivateTradeChannelService.java
@@ -21,10 +21,12 @@ import bisq.chat.channel.BasePrivateChannelService;
 import bisq.chat.channel.ChannelDomain;
 import bisq.chat.message.MessageType;
 import bisq.chat.message.Quotation;
+import bisq.chat.trade.TradeChatOffer;
 import bisq.common.data.Pair;
 import bisq.common.observable.ObservableArray;
 import bisq.common.util.CompletableFutureUtils;
 import bisq.common.util.StringUtils;
+import bisq.network.NetworkId;
 import bisq.network.NetworkIdWithKeyPair;
 import bisq.network.NetworkService;
 import bisq.network.p2p.message.NetworkMessage;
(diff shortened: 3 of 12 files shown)


HenrikJannsen:
> ## Questions
> 
>     * Is bitcoin supposed to be one side of each trade or can other pairs be traded? I looks like bitcoin is intended to be one side of every trade but that seems quite restrictive as a design choice.

Yes, Bitcoin is considered to be one part of the trade pair. But the user could if they wish trade whatever they like. But to design it multi-currency based would distract from our intended user group and would add more complexity, specially with regards to wallet integration.
I think we should do nothing with restricts hardly to Bitcoin, so if users want to trade any altcoin they can do it but they do not get the full feature support (e.g. wallet integration). As well the language would not be adjusted to non-Bitcoin usage.
Target group are newbies and confusing them with 1000s of altcoins where >90 have questionable reputation would contradict the goal of being "Easy".

> 
>     * Is default.properties meant to be translated? It looks like it and will affect design when sending generated text messages as text instead of as operations.

Yes, it will be translated probably similar to Bisq 1 (using transifex). Error messages from exceptions are usually not translated. Only messages which are expected in normal use cases.

> 
>     * There looks to be a work in process framework for a more extensive trade protocol. Is this discussed somewhere?

Yes there are other trade protocols planned. It has been started to work on designs for a generic protocol framework but has been stalled. Latest conclusion was to use completeable futures statemachine a basic model but avoid to create too much framework overhead. There is some demo code at `bisq.protocol.liquidswap`. But not sure if that approach would be used finally. The Bisq Multisig protocol will be likely the next we will work on and that will come with quite some complexity and will battle test if the currently considered approach holds.

HenrikJannsen (approved):
ACK

HenrikJannsen:
Btw. regarding fiat details info:
We have to decide if we consider it a must have for release. If so if we want to make it sufficiently open to be used for other trade protocols (risking that we do not know for sure the requirements before impl. those), or if we keep it Bisq Easy local, risking that later users might complain to have to maintain 2 diff. places for payment infos.
Payment accounts are quite a large boilerplate code base in Bisq 1, so better to be sure to do it right...

I am personally not sure if we really need it for initial launch, or maybe make a simple one, where the seller can manage their free text payment accounts in the settings. Once we have it for other protocols we can replace it. 
The account module has some starting code, but I guess its better to wait until the relevant protocols are in place.

HenrikJannsen:
Ah there is one issue:
You use the BTC amount in the offer, but that is just for informational purpose at the created offer. The Fiat amount is considered the "fixed" amount and the BTC amount is subject to negotiation of the seller (e.g. the seller will tell the buyer how much btc he gets for xxx fiat). Maybe better to remove the amount from the message and send only the BTC address. If its plain BTC address its also easier with copy/paste. Maybe sending 2 messages would be a cheap way to achieve that (e.g. message 1: "Please send me the Bitcoin to following address:" -> Message 2: " bcrt1qw90wvrydkdgv5em0jfdr6p28tt0k0fzd8c5n3j"

DrColver:
Thank you for your comments.

I will update the message to not include the amount.

Moving the button outside of the offer message could be a problem if you have a second trade with the same peer. If no amount is included, then just having a button to send an unused bitcoin address would not lead to any confusion. That would however not work nicely if there is a fiat payment method attached to the offer which should be displayed.

I have implemented a simple fiat account settings page that allows free text. I could be extended to more descriptive account features later on if there is a need. I will clean it up and make another pull request after this one has been merged. I didn't want to include too much in one PR since it's harder to review and also not clear that the features are needed.

HenrikJannsen:
> Moving the button outside of the offer message could be a problem if you have a second trade with the same peer. If no amount is included, then just having a button to send an unused bitcoin address would not lead to any confusion. That would however not work nicely if there is a fiat payment method attached to the offer which should be displayed.

Yes true. The same problem is actually for the open dispute button. Maybe we need to place those buttons inside the chat view. Above or below the guide might be an option. 

I think multiple trades with same peer do not need to be supported from our side as I guess that is rather unlikely and users need to take care then by themself to not mix up the multiple trades. But multiple trades with multiple peers we need to make sure to support well from UX, thus moving the open dispute buttons would be good.

> 
> I have implemented a simple fiat account settings page that allows free text. I could be extended to more descriptive account features later on if there is a need. I will clean it up and make another pull request after this one has been merged. I didn't want to include too much in one PR since it's harder to review and also not clear that the features are needed.

Ah great. Yes better a new PR for the payment account. I guess having a simple free text format is best for now.

PR #703, opened 2023-04-03
Avoid running multiple instances on one directory

DrColver:
Acquire an exclusive lock on a file in the data directory to avoid multiple instances using the same directory. This works on Linux, other systems might handle file locks differently.


HenrikJannsen (approved):
utACK

DrColver:
Yes, it should be released on crash. As an aside, it looks like tor is using the same mechanism to on the `tor/lock` file in the data directory.

PR #723, opened 2023-04-06
Implement leave private channel for non trade channels

HenrikJannsen:


DrColver:
NACK
Channel is closed but view is still open if it's the last private chat. Should switch to default public channel the same way private trade chat does.

HenrikJannsen:
> NACK Channel is closed but view is still open if it's the last private chat. Should switch to default public channel the same way private trade chat does.

Thanks for testing and sorry my sloppy testing ;-).
The new commits should fix the issue.

DrColver:
ACK

I noticed that the last one to leave a channel will send a leave message causing the channel to get reopened on the other side. Fixed in #742

PR #724, opened 2023-04-06
Create base dir if not exists. The checkInstanceLock method throws an…

HenrikJannsen:
… exception otherwise with a fresh data dir.


DrColver:
ACK

PR #725, opened 2023-04-06
Fix left navigation bug

HenrikJannsen:
Sometimes at startup the left navigation does not select the menu item.
After quite some time of debugging I could not find the exact reason why that happens. It is likely a timing issue with runLater used in some places. When starting multiple apps at the same time the issue happens more frequently, otherwise its hard to reproduce.


DrColver:
ACK
I was unable to reproduce the original issue, but the fix doesn't break anything and should handle the described issue.

PR #742, opened 2023-04-11
Don't send leave channel message if peer left channel

DrColver:
Sending a leave message after the peer left the channel reopens the channel for the peer. Refrain from sending a leave message if the last message in the channel was a leave message.


HenrikJannsen (approved):
ACK

HenrikJannsen:
Can you add that to the non-trade channels as well?

PR #743, opened 2023-04-11
Move trade buttons

DrColver:
A separate Pane is used for trade buttons and the action button is now a pure create offer button to minimize confusion with multiuse buttons.

The last trade in the chat is used to decide the action of the complete trade button.


desktop/src/main/java/bisq/desktop/primary/main/content/trade/bisqEasy/chat/BisqEasyChatController.java
```diff
@@ -143,6 +163,45 @@ protected void handleChannelChange(Channel<? extends ChatMessage> channel) {
...
+            boolean isBtcSeller = isMyMessage(lastOfferMessage) == offer.getDirection().isBuy();
+            boolean noWallet = walletService.isEmpty();
+            boolean isDisabled = isBtcSeller || noWallet;
+            model.getCompleteTradeDisabled().set(isDisabled);
+
+            String tooltip = isBtcSeller ? "Coming soon" :
+                    noWallet ?
+                            Res.get("wallet.unavailable") :
+                            Res.get("bisqEasy.completeOffer.tooltip");
+            view.updateCompleteTradeTooltip(isDisabled, tooltip);
```
HenrikJannsen: In the MCV pattern we use we do not call the view from the controller, but change state in the model and the view is listening to those property changes.

HenrikJannsen: In this case you have already the disable property, just missing the tooltip property and a binding in the view.

desktop/src/main/java/bisq/desktop/primary/main/content/trade/bisqEasy/chat/BisqEasyChatController.java
```diff
@@ -156,26 +215,70 @@ protected Optional<? extends Controller> createController(NavigationTarget navig
...
+        if (chatMessage == null) return;
+
+        TradeChatOffer offer = chatMessage.getTradeChatOffer().orElse(null);
+        if (offer == null) return;
+
+        if (isMyMessage(chatMessage) == offer.getDirection().isSell()) {
+            if (walletService.isEmpty()) return;
+
+            // I'm buying BTC, send unused address
+            // Format: bitcoin:<address>[?amount=<amount>][?label=<label>][?message=<message>]
```
HenrikJannsen: The amount is not used anymore

HenrikJannsen (changes requested):

desktop/src/main/java/bisq/desktop/primary/main/content/trade/bisqEasy/chat/BisqEasyChatController.java
DrColver: Thank you for the pointers. This is what I should've done and it made me realize the same Tooltip can be installed for different Nodes. I have pushed a fix.

There are now conflicts with main. How is that handled in the pull request?

desktop/src/main/java/bisq/desktop/primary/main/content/trade/bisqEasy/chat/BisqEasyChatController.java
```diff
@@ -281,4 +269,15 @@ void onCompleteTrade() {
     boolean isMyMessage(ChatMessage chatMessage) {
         return userIdentityService.isUserIdentityPresent(chatMessage.getAuthorId());
     }
+
+    boolean isBtcSeller() {
+        checkArgument(lastOfferMessage.getTradeChatOffer().isPresent());
+        var offer = lastOfferMessage.getTradeChatOffer().get();
```
HenrikJannsen, line 276: We could use here also `.orThrow` if the value need to be present.

HenrikJannsen (approved):
utACK

desktop/src/main/java/bisq/desktop/primary/main/content/trade/bisqEasy/chat/BisqEasyChatController.java
DrColver: I prefer check or assert to indicate there is a bug if the check triggers. Is there a convention in the project to use throw here?

DrColver:
@alvasw how are conflicts handled? Do I merge main into this branch or some other way?

PR #778, opened 2023-04-29
Move leaveChannel to Base class

DrColver:
@alvasw there is an issue with the network PR https://github.com/bisq-network/bisq2/pull/764. I based this branch on the last main commit that works to be able to test.

This PR generalizes https://github.com/bisq-network/bisq2/pull/742 to all private channels.


HenrikJannsen (approved):
utACK

PR #782, opened 2023-04-30
Revert "Switch to NIO implementation"

alvasw:
This reverts commit 2d0254dad6f89a69c81687af2796fd5345aca03b.

### DIFF
--- a/network/network/src/main/java/bisq/network/p2p/node/PeerConnectionsManager.java
+++ b/network/network/src/main/java/bisq/network/p2p/node/PeerConnectionsManager.java
@@ -18,22 +18,23 @@
 package bisq.network.p2p.node;
 
 import bisq.network.p2p.node.authorization.AuthorizationService;
-import bisq.network.p2p.node.transport.TorTransport;
 import bisq.network.p2p.node.transport.Transport;
-import bisq.network.p2p.node.transport.socketchannel.ClearNetSocketChannelFactory;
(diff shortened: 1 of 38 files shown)


network/network/src/main/java/bisq/network/p2p/node/PeerConnectionsManager.java
```diff
@@ -52,19 +52,21 @@ public PeerConnectionsManager(Node.Config config,
...
-    public CompletableFuture<Void> start(Node node, int port) {
-        return createServerAndListenAsync(node, port)
-                .thenAccept(myCapability -> createAndStartOutboundConnectionMultiplexer(myCapability, node));
+    public void start(Node node, int port) {
+        try {
+            Capability myCapability = createServerAndListen(node, port);
+            createAndStartOutboundConnectionMultiplexer(myCapability, node);
+        } catch (IOException e) {
+            log.error("Couldn't start PeerConnectionsManager", e);
+        }
```
DrColver, line 69: @alvasw I think this was part of the issue with the NIO PR, as far as I noticed.

The `outboundConnectionMultiplexer` was not initialized before being accessed after the change to using an async initialization. I haven't looked through the startup sequence and procedure in detail so I can't suggest a fix.

alvasw: Thank you!

PR #805, opened 2023-05-02
Various bug fixes

HenrikJannsen:


desktop/src/main/resources/bisq.css
```diff
@@ -585,6 +585,15 @@
     -fx-font-family: "IBM Plex Sans Thin";
 }
 
+#base-amount-text-field-small {
+    -fx-background-color: transparent;
+    -fx-border-style: none;
+    -fx-text-fill: -bisq-white;
+    -fx-highlight-fill: -bisq-green;
+    -fx-font-size: 4em;
```
DrColver: I suggest something smaller. Just tried creating an offer with VND (Vietnam) and it doesn't fit in the box due to the two decimal places. I tried with 3em, it would likely fit most cases.

Another consideration is that such low value currencies likely don't use decimal places.

desktop/src/main/java/bisq/desktop/primary/overlay/createOffer/amount/AmountController.java
```diff
@@ -118,8 +117,11 @@ public void onActivate() {
 
         baseAmount.setAmount(null);
         if (model.getQuoteSideAmount().get() == null) {
-            //todo we need to adjust to diff fiat exchange rates
-            quoteAmount.setAmount(Fiat.parse("100", model.getMarket().getQuoteCurrencyCode()));
+            // We use 0.004 BTC as default value converted with the price to the fiat amount
+            Quote fixPrice = price.fixPriceProperty().get();
+            Monetary baseCurrencyAmount = Coin.asBtc(400000);
+            Monetary exactAmount = fixPrice.toQuoteMonetary(baseCurrencyAmount);
```
DrColver: This causes NPE when `fixPrice` is null. Happens when amount is not set/deleted by the user.

DrColver (commented):
Not introduced in this PR. but I noticed the vertical collapse button is visible in the original position when horizontally collapsed.

desktop/src/main/java/bisq/desktop/primary/overlay/createOffer/amount/AmountController.java
HenrikJannsen: How can you reproduce that?
Here is a fix - assuming its a rare edge case: https://github.com/bisq-network/bisq2/pull/805/commits/b2d5de3dbe4fb823d9d6f34bdc2b6c3f0b7709b1

DrColver: It happens when I change the currency (testing with IRR) and delete the amount string. Pressing the back button triggers the else statement in your fix. With the fix it doesn't thrown any more exceptions.

desktop/src/main/resources/bisq.css
HenrikJannsen: Ok, reduced to 3em via https://github.com/bisq-network/bisq2/pull/805/commits/7db0c455f64b64e2e35f1344dcffa5812cf24ec4

Yes, true regarding decimals. But I guess thats not that high prio. Can be improved later.

desktop/src/main/java/bisq/desktop/primary/overlay/createOffer/amount/AmountController.java
HenrikJannsen: Thanks for testing!

HenrikJannsen:
> Not introduced in this PR. but I noticed the vertical collapse button is visible in the original position when horizontally collapsed.

I cannot reproduce that. Can you add an issue with a screenshot?

HenrikJannsen:
Leave that PR now. Have WIP for the mediation use cases but will do that in a new PR. That one is already huge...

DrColver:
@HenrikJannsen I added a fix for the expand/collapse button issue, PR https://github.com/bisq-network/bisq2/pull/823

PR #824, opened 2023-05-03
Implement group chat

HenrikJannsen:
This PR changes the private trade channels.
- If the channel was created by a taker offer action we use the offerID as channelID and store the offer in the channel. Multiple offers with same users will get multiple private channels. This help to avoid mix of different offers.
- A private trade channel is a group channel with the mediator and the 2 traders as members.
- Channels created by private messages (PM / text messages) will be treated as 2-party channels. The combination of the peers userProfileIDs is the channel ID, thus only one such private channel can exist.


HenrikJannsen:
There is still open work but I will do some refactoring work before and I prefer to split that into a new PR. So that PR is ready for review/merge now. There will be issues open which will get fixed after refactoring.

PR #826, opened 2023-05-04
Update visible chat messages

DrColver:
Improve performance when switching between channels with a lot of messages.


HenrikJannsen (approved):
utACK

namloan:
@DrColver apologies for chiming in this 5-month issue, just stumbled upon this code. If you still remember - mind sharing some details about how is this improving performance?

PR #830, opened 2023-05-06
Improve chat domain

HenrikJannsen:


HenrikJannsen:
There will be still more refactoring in follow up PRs, but I think the changes are stable enough to get merged to avoid too large change sets.

DrColver:
@HenrikJannsen this refactoring has broken some channel handling. It's not possible to add more trade channels when clicking the + button for new currency.

HenrikJannsen:
@DrColver Will be fixed with upcoming PR

PR #834, opened 2023-05-08
Userprofile selection

DrColver:
> Setting selection via selectionModel.select displays the label from the base class

My suspicion is that the issue was `selectedUserIdentity` not being one of the instances in the `userIdentities` set and `equals()` not working as expected. That would cause `select()` to fail. `UserIdentity.equals()` now work but I still think it's better to use the same instances rather than a copy for `selectedUserIdentity`.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/primary/main/content/settings/userProfile/UserProfileView.java
+++ b/desktop/src/main/java/bisq/desktop/primary/main/content/settings/userProfile/UserProfileView.java
@@ -149,14 +149,7 @@ public class UserProfileView extends View<HBox, UserProfileModel, UserProfileCon
         });
 
         selectedChatUserIdentityPin = EasyBind.subscribe(model.getSelectedUserIdentity(),
-                userIdentity -> {
-                    if (userIdentity != null) {
-                        // Setting selection via selectionModel.select displays the label from the base class. 
-                        // Did not find out how to avoid that... ;-(
-                        // With setting our selection to the editorTextField it behaves as expected.
-                        UIThread.runOnNextRenderFrame(() -> comboBox.getEditorTextField().setText(comboBox.getConverter().toString(userIdentity)));
-                    }
-                });
+                userIdentity -> comboBox.getSelectionModel().select(model.getSelectedUserIdentity().get()));
     }
 
     @Override
--- a/user/src/main/java/bisq/user/identity/UserIdentityStore.java
+++ b/user/src/main/java/bisq/user/identity/UserIdentityStore.java
@@ -44,7 +44,9 @@ public final class UserIdentityStore implements PersistableStore<UserIdentitySto
 
     private UserIdentityStore(UserIdentity selectedUserIdentity,
                               Set<UserIdentity> userIdentities) {
-        this.selectedUserIdentity.set(selectedUserIdentity);
+        this.selectedUserIdentity.set(userIdentities.stream()
+                .filter(userIdentity -> userIdentity.equals(selectedUserIdentity))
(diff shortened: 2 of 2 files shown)


desktop/src/main/java/bisq/desktop/primary/main/content/settings/userProfile/UserProfileView.java
```diff
@@ -149,14 +149,7 @@ protected void onViewAttached() {
...
         selectedChatUserIdentityPin = EasyBind.subscribe(model.getSelectedUserIdentity(),
-                userIdentity -> {
-                    if (userIdentity != null) {
-                        // Setting selection via selectionModel.select displays the label from the base class. 
-                        // Did not find out how to avoid that... ;-(
-                        // With setting our selection to the editorTextField it behaves as expected.
-                        UIThread.runOnNextRenderFrame(() -> comboBox.getEditorTextField().setText(comboBox.getConverter().toString(userIdentity)));
-                    }
-                });
+                userIdentity -> comboBox.getSelectionModel().select(model.getSelectedUserIdentity().get()));
```
HenrikJannsen: `model.getSelectedUserIdentity().get()` can be replaced by the param `userIdentity`

user/src/main/java/bisq/user/identity/UserIdentityStore.java
```diff
@@ -44,7 +44,9 @@ public UserIdentityStore() {
 
     private UserIdentityStore(UserIdentity selectedUserIdentity,
                               Set<UserIdentity> userIdentities) {
-        this.selectedUserIdentity.set(selectedUserIdentity);
+        this.selectedUserIdentity.set(userIdentities.stream()
```
HenrikJannsen, line 53: We should find the reason why a different instance of (the same) userIdentity is used. I guess we might get it from another persisted object (channelSelection?).

HenrikJannsen (changes requested):

desktop/src/main/java/bisq/desktop/primary/main/content/settings/userProfile/UserProfileView.java
DrColver: Yes, fixed.

user/src/main/java/bisq/user/identity/UserIdentityStore.java
DrColver: When userIdentityStore is loaded from file it creates selectedUserIdentity from protobuf and the userIdentities Set from protobuf separately. I think this is ok, as long as the selectedUserIdentity instance is only used to select the instance from the Set.

I haven't seen any other usages where selectedUserIdentity would be set to an instance not in the Set, but it would be possible. I added a setter to make sure an instance from the Set is used.

HenrikJannsen: Ah ok. We should replace the selectedUserIdentity with an selectedUserIdentityId (pubkeyHash) and use that to lookup the one from the set. We have a similar pattern used for channels.

HenrikJannsen: But this can be done a new PR...

HenrikJannsen (approved):
utACK

user/src/main/java/bisq/user/identity/UserIdentityStore.java
DrColver: I don't think it's necessary to use the hash. Saving the object makes it easier to see what is saved if you look at the save files and it needs to be handled when loaded in any case.

PR #837, opened 2023-05-09
Useridentitystore getter

DrColver:
This builds on #834 

I don't want to prevent #834 from being merged and the risk of conflicts between #830 and this PR. This can be seen as a study of the benefits and drawbacks of the suggestions in #835


user/src/main/java/bisq/user/identity/UserIdentityService.java
```diff
@@ -214,7 +214,7 @@ public boolean hasUserProfiles() {
     }
 
     public Observable<UserIdentity> getSelectedUserIdentity() {
-        return persistableStore.getSelectedUserIdentity();
+        return persistableStore.getSelectedUserIdentityObservable();
```
HenrikJannsen, line 222: If we follow the JavaFx Property pattern it would be `selectedUserIdentityObservable()` . I was never a big fan of that convention but I guess it might be still the better approach to keep it similar to Properties and apply also the JavaFx convention where we use properties (a lot). I don't think its an urgent refactoring task, but can be done over time. Maybe ChatGPT can do it ;-) its really a code monkey job ;-).

desktop/src/main/java/bisq/desktop/primary/main/content/components/ChatMessagesComponent.java
```diff
@@ -235,7 +235,7 @@ private void onSendMessage(String text) {
 
         private void doSendMessage(String text) {
             ChatChannel<? extends ChatMessage> chatChannel = model.selectedChannel.get();
-            UserIdentity userIdentity = userIdentityService.getSelectedUserIdentity().get();
+            UserIdentity userIdentity = userIdentityService.getSelectedUserIdentity();
```
HenrikJannsen: Maybe good to add checkNotNull here as well to make it explicit that the value is expected to be not null.

HenrikJannsen (approved):
utACK

HenrikJannsen (approved):
utACK

PR #839, opened 2023-05-11
Refactor channel selection UI clases

HenrikJannsen:


DrColver:
https://github.com/bisq-network/bisq2/pull/830#issuecomment-1542121474 is still not fixed with this PR. Create offer also does not work, throws an exception when clicking next on the PAYMENT METHOD screen.

HenrikJannsen:
> [#830 (comment)](https://github.com/bisq-network/bisq2/pull/830#issuecomment-1542121474) is still not fixed with this PR. Create offer also does not work, throws an exception when clicking next on the PAYMENT METHOD screen.

Can you try it with latest PR?

PR #848, opened 2023-05-15
Add equals and hashCode to observables

DrColver:


HenrikJannsen (changes requested):
We could use Lombok for it to avoid boilerplate code. But I am not sure if we need equals/hashcode in those classes as they are not intended to be used as value objects.

DrColver:
After discussion, it's better to not add equals and hashCode to observables.

PR #894, opened 2023-05-29
Add leave channel button when peer left channel

DrColver:
I have added three examples of code.

Example 1 works without warning since all types are known.

Example 2 works but with warnings.

Example 3 doesn't work since the implementations of leaveChannel take different types.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/primary/main/content/components/ChatMessagesComponent.java
+++ b/desktop/src/main/java/bisq/desktop/primary/main/content/components/ChatMessagesComponent.java
@@ -27,11 +27,15 @@ import bisq.chat.bisqeasy.message.BisqEasyPrivateTradeChatMessage;
 import bisq.chat.channel.ChatChannel;
 import bisq.chat.channel.ChatChannelDomain;
 import bisq.chat.channel.ChatChannelSelectionService;
+import bisq.chat.channel.priv.PrivateChatChannel;
+import bisq.chat.channel.priv.PrivateChatChannelService;
 import bisq.chat.channel.priv.TwoPartyPrivateChatChannel;
 import bisq.chat.channel.pub.CommonPublicChatChannel;
 import bisq.chat.channel.pub.PublicChatChannel;
 import bisq.chat.message.ChatMessage;
+import bisq.chat.message.ChatMessageType;
 import bisq.chat.message.Citation;
+import bisq.chat.message.PrivateChatMessage;
 import bisq.common.observable.Pin;
 import bisq.common.util.StringUtils;
 import bisq.desktop.common.observable.FxBindings;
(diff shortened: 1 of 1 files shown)


HenrikJannsen:
That's the same problem I run into a while back and it took me a while to understand the problem. I will try to explain:

The `ChannelService` class gets the `Channel` as a generic type `C`. So all implementations of `ChannelService` can use the concrete type for the channel as the type is passed in the implementation. That type `C` is used for the collection holding the channels.

But if a method is using the channel as parameter `C` the compiler cannot know if the type of the parameter is matching the type defined in the `ChannelService` class. You can call the method passing the concrete type for the methods generic type but then your calling code need to know which implementation is used and that is mostly not the case, it deals with a `<? extends Channel>` type not with the `C` type.

Think if it how to provoke a type mismatch: 
You could pass another impl. of `Channel `fullfulling the `<? extends Channel>` condition but not being the same types which was passed in the concrete impl. of `ChannelService`. The compiler cannot know if the types matches. In other words the `C` type of the method is not the `C` type of the class. So when calling a method in `ChannelService` using the `C` type for a parameter and calling that method with an unknown implementation type (`<? extends Channel>`) the compiler complains about the incompatible types.

One way to deal with it is do `instanceof` checks and cast but that's boilerplate and contradicts the reason why we use generics. Another is to just do an unsafe cast but that's risky to get exceptions and you get a compiler warning.
 
The correct way IMO is to lookup the channel in the `ChannelService` which returns the correct generic type (`C`). That way if the method would be called with an incorrect channel (e.g. passing a `TwoPartyPrivateChatChannel` to `CommonPublicChatChannelService`)  the lookup would not find the channel and the `ifPresent` does not get executed. 

Here an example dealing with the same problem in `PrivateChannelSelectionMenu`:
```
 public void onLeaveChannel(ChatChannel<?> chatChannel) {
            if (chatChannel instanceof PrivateChatChannel) {
                // We map the ChatChannel<?> chatChannel to our generic type
                chatChannelService.findChannel(chatChannel.getId())
                        .ifPresent(privateChatChannel -> {
                            String dontShowAgainId = "leavePrivateChannel";
                            if (DontShowAgainService.showAgain(dontShowAgainId)) {
                                new Popup().warning(Res.get("social.privateChannel.leave.warning",
                                                privateChatChannel.getMyUserIdentity().getUserName()))
                                        .dontShowAgainId(dontShowAgainId)
                                        .closeButtonText(Res.get("cancel"))
                                        .actionButtonText(Res.get("social.privateChannel.leave"))
                                        .onAction(() -> doLeaveChannel(privateChatChannel))
                                        .show();
                            } else {
                                doLeaveChannel(privateChatChannel);
                            }
                        });
            }
        }
```

PR #896, opened 2023-05-30
Add leave channel button when peer left channel

DrColver:
Leave channel by id, let services handle generics.

Cleanup

@HenrikJannsen I see what you mean with the generics in #894. I let the generic classes themself handle the generics by adding a convenience function that takes a String. Similar to what you did in your usage of leaveChannel().

This works and is cleaner than the other suggestions as I see it.

Again, I have added an unsolicited feature. If it's not welcome, please close the PR.

### DIFF
--- a/desktop/src/main/java/bisq/desktop/primary/main/content/components/ChatMessagesComponent.java
+++ b/desktop/src/main/java/bisq/desktop/primary/main/content/components/ChatMessagesComponent.java
@@ -27,10 +27,12 @@ import bisq.chat.bisqeasy.message.BisqEasyPrivateTradeChatMessage;
 import bisq.chat.channel.ChatChannel;
 import bisq.chat.channel.ChatChannelDomain;
 import bisq.chat.channel.ChatChannelSelectionService;
+import bisq.chat.channel.priv.PrivateChatChannel;
 import bisq.chat.channel.priv.TwoPartyPrivateChatChannel;
 import bisq.chat.channel.pub.CommonPublicChatChannel;
 import bisq.chat.channel.pub.PublicChatChannel;
 import bisq.chat.message.ChatMessage;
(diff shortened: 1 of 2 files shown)


desktop/src/main/java/bisq/desktop/primary/main/content/components/ChatMessagesComponent.java
```diff
@@ -308,8 +313,18 @@ private void privateTradeMessagesChanged(BisqEasyPrivateTradeChatChannel chatCha
                     model.getSendBtcAddressButtonVisible().set(walletService.isPresent());
                 }
             });
+            twoPartyPeerLeft(chatChannel);
+        }
+
+        private void twoPartyPeerLeft(PrivateChatChannel<?> chatChannel) {
```
HenrikJannsen: As its also called from `privateTradeMessagesChanged(BisqEasyPrivateTradeChatChannel chatChannel)` the method name `twoPartyPeerLeft` is not optimal, the BisqEasyPrivateTradeChatChannel is a groupChannel not a 2-party channel.

HenrikJannsen: Also its not the case that the peer left, just that the leaveChannelButton visible state gets updated. Maybe call it `updateLeaveChannelButtonState`?

HenrikJannsen:
I will work on a new UI element in the trade chat and will move the buttons over there once its more clear how it looks like.
To have 3 buttons might become quite overloaded. I am not 100% sure if we need that leave button, but I am not against the merge. We can drop if it we figure out its too much later...

DrColver:
The buttons in the left column was much better than next to the send message button. It's better to separate them from the text input and chat message area for clarity.

HenrikJannsen (approved):
utACK

PR #955, opened 2023-07-02
Use protobuf aarch64 binaries for ARM processors

namloan:
aarch64 binaries have been available since version 3.5.0, therefore it shouldn't be needed to default to x86_64.

I could successfully run the app in an Apple M2 Pro. Would be great if someone with M1 could test and verify.


namloan:
@alvasw sure, will sign the commit.

Based on your comment I was under the impression that it was absolutely needed to also make the swtich in the Tor and Electrum binaries or otherwise this change would break those integrations. Did I misunderstand? I only tested this on clearnet.

Issue #909, opened 2023-06-03
Add payment methods


HenrikJannsen:
Add missing Bisq1 payment methods to `bisq.account.settlement.FiatSettlement`

namloan:
I'll take this one.

The code was refactored after the creation of this issue and ```FiatSettlement``` doesn't exist anymore. Looks like the payment methods will need to be added to [FiatPaymentRail](https://github.com/bisq-network/bisq2/blob/main/account/src/main/java/bisq/account/payment_method/FiatPaymentRail.java). The payment methods for Bisq 1 can be found in [bisq / core / payment](https://github.com/bisq-network/bisq/tree/8b233866f2930982df129f5be90334dd03100de0/core/src/main/java/bisq/core/payment).

HenrikJannsen:
Yes. I hope the current structure is ok. I have not spent too much time on it... 

Goal for Bisq Easy is to show the matching payment methods on the create offer screen according to the selected currency. As we do not provide information about the country, we cannot filter for that. So there might be cases where it does not work. E.g. user selects CHF and no SEPA is shown, though their bank support SEPA. as there is the custom option they can add it anyway... There might be more such situations. But I think we should favor simplicity for Bisq Easy and not try to provide additional data which would make the selection more accurate.

For Bisq Multisig though we should aim for a similar model as in Bisq 1 but with more options so that users can have multiple methods in an offer. There might be some complexities with that (diff. max trade amounts,...), but I guess that can be delegated for later.

HenrikJannsen:
@namloan Any progress on that?

namloan:
Couldn't spend much time on this, expect some progress between today and tomorrow.

HenrikJannsen:
I guess we can close that issue as most important methods are implemented.

PR #1044, opened 2023-07-29
Add F2F payment method

namloan:
Sending this PR as part of #909 to gather early feedback and validate that it's aligned with the expectations. Attaching a screenshot where F2F can be seen as the first payment method.

f2f_screenshot


HenrikJannsen (commented):
ACK

HenrikJannsen:
Looks good!

PR #1052, opened 2023-07-30
Add ACH payment method

namloan:
ach_screenshot


HenrikJannsen (approved):
utACK

HenrikJannsen:
Thanks for the PR!
Is the icon related to an official ACH icon? I could not find any in the web. If there is no kind of official icon, we can leave it as you suggested. Looks good!

Could you add the next accounts according to the usage in bisq 1 from the trade statistics table? so if we do not get all implemented that we cover the most used ones.

namloan:
@HenrikJannsen I created the icon from [the logo set](https://www.nacha.org/content/ach-network-logo-set) provided by [Nacha](https://en.wikipedia.org/wiki/NACHA) which is the organization that governs the ACH Network.

> Could you add the next accounts according to the usage in bisq 1 from the trade statistics table?

Sure.

namloan:
@HenrikJannsen can you point me to the stats table? the closest I could find is the table that shows the number of open offers by payment method in the Bisq1 app itself.

PR #1092, opened 2023-08-02
Unit test payment method accounts

namloan:
Discovered and fixed some issues along the way. The most important is the removal of HRK as supported currency in Revolut and Wise. HRK was replaced by the Euro in 2023. The presence of HRK was preventing the Revolut and Wise FiatPaymentRails enums from being properly initialized (they were throwing NullPointerException).

#909


HenrikJannsen (approved):
utACK

account/build.gradle
```diff
@@ -11,4 +11,6 @@ dependencies {
 
     implementation 'network:network'
     implementation libs.google.guava
+
+    testImplementation libs.assertj.core
```
alvasw: This is not needed because `AssertJ` is already defined in: https://github.com/bisq-network/bisq2/blob/10c085a02e98e455629c828ae265dfe2845377e7/build-logic/commons/src/main/groovy/bisq.java-conventions.gradle#L33

PR #1106, opened 2023-08-07
Add Zelle and National Bank payment accounts

namloan:
These ones were already selectable in Bisq Easy, so I'm just completing their implementation to match the other payment methods.

#909

### DIFF
--- a/account/src/main/java/bisq/account/accounts/Account.java
+++ b/account/src/main/java/bisq/account/accounts/Account.java
@@ -71,6 +71,9 @@ public abstract class Account<P extends AccountPayload, M extends PaymentMethod<
 
     public static Account<?, ?> fromProto(bisq.account.protobuf.Account proto) {
         switch (proto.getMessageCase()) {
+            case ZELLEACCOUNT: {
+                return ZelleAccount.fromProto(proto);
+            }
             case USERDEFINEDFIATACCOUNT: {
                 return UserDefinedFiatAccount.fromProto(proto);
             }
--- a/account/src/main/java/bisq/account/accounts/AccountPayload.java
+++ b/account/src/main/java/bisq/account/accounts/AccountPayload.java
@@ -50,6 +50,9 @@ public abstract class AccountPayload implements Proto {
 
     public static AccountPayload fromProto(bisq.account.protobuf.AccountPayload proto) {
         switch (proto.getMessageCase()) {
+            case ZELLEACCOUNTPAYLOAD: {
+                return ZelleAccountPayload.fromProto(proto);
+            }
             case COUNTRYBASEDACCOUNTPAYLOAD: {
                 return CountryBasedAccountPayload.fromProto(proto);
             }
--- a/account/src/main/java/bisq/account/accounts/BankAccount.java
+++ b/account/src/main/java/bisq/account/accounts/BankAccount.java
@@ -17,6 +17,8 @@ public abstract class BankAccount<P extends BankAccountPayload> extends CountryB
         switch (proto.getCountryBasedAccount().getBankAccount().getMessageCase()) {
             case ACHTRANSFERACCOUNT:
                 return AchTransferAccount.fromProto(proto);
+            case NATIONALBANKACCOUNT:
+                return NationalBankAccount.fromProto(proto);
             case MESSAGE_NOT_SET: {
                 throw new UnresolvableProtobufMessageException(proto);
             }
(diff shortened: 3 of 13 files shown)


HenrikJannsen (changes requested):
I would prefer that we use `Optional` instead of `@Nullable`. In Bisq 2 we try to avoid nullable fields in most cases (beside trivial stuff in UI).

I don't see much reason for the tests. We do not test applying protobuf data as there is hardly any logic or complexity in it.

namloan:
@HenrikJannsen I'm used to follow the 'rule' of only using Optional for only for return types and never for method arguments / class fields. Here's a discussion about it: [Why should Java 8's Optional not be used in arguments](https://stackoverflow.com/questions/31922866/why-should-java-8s-optional-not-be-used-in-arguments). I'll adapt to whatever code convention is planned for Bisq2 though, is using Optional in method arguments something we'd like to enforce?

Regarding the tests: I see them as a way to validate that something works as expected and that it keeps working over time. I've been finding them useful as they've already uncovered some issues - in [this unit testing PR](https://github.com/bisq-network/bisq2/blob/d1e8293edf4015b5c26ebf4d529abf833062513a/account/src/main/java/bisq/account/payment_method/FiatPaymentRail.java#L91) I found a couple of mistakes in the protobuf transformation logic + the Revolut and Wise ```FiatPaymentRails``` were throwing ```ExceptionInInitializerError``` because of the HRK currency not being found which btw I suspect it was probably the root cause of [this exception when sorting currencyCodes](https://github.com/bisq-network/bisq2/blob/d1e8293edf4015b5c26ebf4d529abf833062513a/account/src/main/java/bisq/account/payment_method/FiatPaymentRail.java#L91).

I believe that unit tests, even when the underlying logic is simple, does increase development output in the long run but of course it can be subjective. If not writing unit tests for protobuf transformations is a code convention that we'd like to enforce here, happy to drop them / stop writing them.

HenrikJannsen:
> I'm used to follow the 'rule' of only using Optional for only for return types and never for method arguments / class fields. Here's a discussion about it: [Why should Java 8's Optional not be used in arguments](https://stackoverflow.com/questions/31922866/why-should-java-8s-optional-not-be-used-in-arguments). I'll adapt to whatever code convention is planned for Bisq2 though, is using Optional in method arguments something we'd like to enforce?

I read about those discussions but do not get convinced. In Java we have no way to avoid passing null. I am not a fan of using `@NotNull` as it becomes very boilerplate and I prefer to have it as default that any parameter must not be null (unfortunately JavaFX components propagate null values so this cannot be applied to the whole code base).
The cases where absence of a value is a valid case are IMO best represented by Optional (Kotlin made a better design choice by making in null a type and delegate the problem to the type system). 
Following that convention to not user Optionals in fields or parameters would lead to a mix of `@Nullable` annotated fields/parameters and Optionals which I find not good.

So as far there are no other convincing arguments I prefer to stick with convention to use Optional and it also mirrors protobuf optional fields better.

@alvasw What is your opinion regarding that topic?


> Regarding the tests: I see them as a way to validate that something works as expected and that it keeps working over time. I've been finding them useful as they've already uncovered some issues - in [this unit testing PR](https://github.com/bisq-network/bisq2/blob/d1e8293edf4015b5c26ebf4d529abf833062513a/account/src/main/java/bisq/account/payment_method/FiatPaymentRail.java#L91) I found a couple of mistakes in the protobuf transformation logic + the Revolut and Wise FiatPaymentRails were throwing ExceptionInInitializerError because of the HRK currency not being found which btw I suspect it was probably the root cause of [this exception when sorting currencyCodes](https://github.com/bisq-network/bisq2/blob/d1e8293edf4015b5c26ebf4d529abf833062513a/account/src/main/java/bisq/account/payment_method/FiatPaymentRail.java#L91).
I believe that unit tests, even when the underlying logic is simple, do increase development output in the long run but of course it can be subjective. If not writing unit tests for protobuf transformations is a code convention that we'd like to enforce here, happy to drop them / stop writing them.

I was hesitating with my comment about the tests, and I see the value if it helped to find a lurking bug. So I take my comments back.

PS: Are you on Matrix as well? Maybe sometimes easier to discuss in a chat interface...

namloan:
> I was hesitating with my comment about the tests

Please don't hesitate. I've literally never written production code without tests as I've always worked in environments where test automation is paramount (unit tests, integration tests, UI tests). Therefore, I can be biased and I find it interesting to hear about other perspectives.

> Are you on Matrix as well?

Yes, happy to chat there.

Issue #883, opened 2023-05-26
Finish implementation for input validators


HenrikJannsen:
Input validators are implemented but missing the handling in the UI components to show the error messages and adopt the css.
It should follow the pattern used in JavaFX components/JFoenix.

namloan:
taking a look at this one

namloan:
been offline for a week, I'm retaking this today.

HenrikJannsen:
@namloan Are there still some open? Maybe can can add a bullet point list of the UI screens where validation is open and set priorities to each?

namloan:
@HenrikJannsen we're still missing some validations in 'User options':
* **Reputation**. I started working on this but halted it until https://github.com/bisq-network/bisq2/pull/1320/commits is merged as there will be conflicts. Validating the numbers are >= 0 and that the score isn't calculated when amounts are invalid.
* **Bonded roles**. Not sure what validations we want here but maybe at least text length size.
* **Nodes**. Same as above.

HenrikJannsen:
We can consider the Pro user/contributor sections as lower prio.
Feel free to pause that task and pick anything else with higher prio if you like.

HenrikJannsen:
@namloan Are you still working on that?

namloan:
Nope, it's free. Show we merge with https://github.com/bisq-network/bisq2/issues/1153?

PR #1126, opened 2023-08-12
Add password validations to UnlockView

namloan:
Adapted validation controls from JFoenix

Gathering early feedback, let me know if you like it. I think we may need to make the red color a bit brighter.

unlockview

#883


HenrikJannsen (approved):
utACK

HenrikJannsen:
Great thanks for working on that. One important part which was missing for the release milestone!

Issue #895, opened 2023-05-30
Do not use String in MaterialPasswordField


HenrikJannsen:
For security reasons we need to modify `MaterialPasswordField` so that no `String` object gets created for the password.
We should use `CharSequence` instead or the 'bisq.security.SecureString' class and clean up the byte array once the password is not needed anymore. 
The client code must handle it as well accordingly.

String objects could stick around in memory until the GC cleans it up, which is not determined when and if that happen.

namloan:
will tackle this

namloan:
@HenrikJannsen did you already have something in mind for this? It seems like by design JavaFX is only supporting String for text input through TextInputControl. I guess it should be possible to write our own custom input control that uses CharSequence, however it looks like a lot of work so I wanted to double check.

Should this be a feature request to JavaFX instead?

HenrikJannsen:
It it gets too difficult we can postpone it as well. Currently the password protection is only for protecting access to the UI screen and the user db file is encrypted. As long we don't have the wallet the security implications are not that critical.
Also getting access to the application RAM is already a severe breach of a users system. For Bisq 1 we considered it as well but as BitcoinJ holds the private keys as objects in RAM secure password protection would not make the situation better. Bisq 2 has still not yet decided on the lightweight wallet to be used (currently Electrum but we want to check out other alternatives like BDK or Sparrow). So it will also depend on that if that threat scenario is covered by the wallet or not.

namloan:
Thanks for the Sparrow suggestion. Looks like Sparrow hasn't solved the problem either - it's loading the password in a String and converting it to SecureString. A couple of code pointers:

* [LoadWallet.java](https://github.com/sparrowwallet/sparrow/blob/0a469a380bd11424a96214499fd0ced13e213212/src/main/java/com/sparrowwallet/sparrow/terminal/wallet/LoadWallet.java#L71)
* [WalletDialog.java](https://github.com/sparrowwallet/sparrow/blob/0a469a380bd11424a96214499fd0ced13e213212/src/main/java/com/sparrowwallet/sparrow/terminal/wallet/WalletDialog.java#L82)

I can try a similar approach, but the security threat will still remain.

HenrikJannsen:
If we cannot avoid that the entered password gets filled in a java object (String) at the input component level I guess it has no value to try to treat it as secure string further down the road. Just makes code more complex and cumbersome and creates fake sense of security. 
[...]

PR #1130, opened 2023-08-15
Validate inputs in confirm password screens

namloan:
#883 

Added a new validator to check that two text fields are equal. The unlock button is enabled / disabled based on whether both input fields passed all validations.

confirm_password


namloan:
@alvasw Interesting, looks like flakiness (this is a UI test). Can the build be retried?

namloan:
@alvasw I see that it has passed now. First time I see this test being flaky, may be worth adding a retry strategy there. Feel free to remove it or disable it if it starts timing out frequently, it's just a very basic test I added while I was exploring #895

PR #1131, opened 2023-08-15
Add Cash Deposit, UPI and Bizum payment methods

namloan:
#909 

screenshot2
screenshot1


account/src/main/java/bisq/account/payment_method/FiatPaymentRail.java
```diff
@@ -25,6 +25,7 @@
 import lombok.Getter;
 
 import javax.annotation.Nullable;
+import java.sql.Array;
```
alvasw: I guess you accidentally imported `java.sql.Array`.

namloan: oops

PR #1148, opened 2023-08-26
Refactor amount input components

namloan:
SmallAmountInput and BigAmountInput had a considerable amount of duplicate code which I've abstracted into a new AmountInput class.

No change in functionality.


HenrikJannsen (approved):
utACK

namloan:
Reverted the refactor of the `onAmount` method as it didn't match the original behavior.

HenrikJannsen (commented):
utACK

PR #1150, opened 2023-09-02
Add amount validation to CreateOfferAmount screen

namloan:
#883 

amount_validation


HenrikJannsen:
I tested it and got an exception at `AmountParser.parse` at AmountInput line 159. I could not reproduce the error but it happened while quickly entering at both fields invalid values.
This is not a bug from your changes but was not handled before but would be good to catch the exception.

Also saw that the classed in `bisq.desktop.components.controls.validator` lack the Bisq license header. You can add the header to IDE so its included at each new class by default. Would be good to add that to the .idea files and commit it as well.

HenrikJannsen (approved):
ACK

namloan:
Sure, I'll tackle those issues in next PRs. Thanks for reviewing!

PR #1159, opened 2023-09-05
Aspire2dev/adjust enums

aspire2dev:
@alvasw and @HenrikJannsen , here's the code for issue 1111


account/src/main/java/bisq/account/protocol_type/LoanProtocolType.java
```diff
@@ -20,15 +20,18 @@
 import bisq.common.util.ProtobufUtils;
 
 public enum LoanProtocolType implements ProtocolType {
+    UNSPECIFIED,
```
HenrikJannsen: We do not want to have the `UNSPECIFIED` entry in the java code.

common/src/main/java/bisq/common/util/ProtobufUtils.java
```diff
@@ -41,6 +41,25 @@ public static <E extends Enum<E>> E enumFromProto(Class<E> enumType, String name
         return result;
     }
 
+    public static <E extends Enum<E>> E enumFromProto(Class<E> enumType, String name, E fallBack) {
+        String info = "Enum type= " + enumType.getSimpleName() + "; name=" + name + "; fallBack=" + fallBack;
```
HenrikJannsen, line 48: The `name == null` check must be before the `if (name.endsWith("_UNSPECIFIED"))`.

Convert the enum class name to the protobuf prefix (instead at each caller):
```
String protobufEnumPrefix = StringUtils.capitalizeAll(enumType.getSimpleName()) + "_";
        String enumName = name.replace(protobufEnumPrefix,"");
        E result = Enums.getIfPresent(enumType, enumName).orNull();
```

Add that method to StringUtils: 
```
 public static String capitalizeAll(String value) {
        if (value == null || value.isEmpty()) {
            return value;
        } else {
            return value.toUpperCase();
        }
    }
```

The method with no fallback should throw an exception if no valid enum can be resolved.

Methods should be:
```
public static <E extends Enum<E>> E enumFromProto(Class<E> enumType, String name, E fallBack) {
        String info = "Enum type= " + enumType.getSimpleName() + "; name=" + name + "; fallBack=" + fallBack;
        if (name == null) {
            log.error("Enum name is null. We use the fallback value instead. {}", info);
            return fallBack;
        }
        
        if (name.endsWith("_UNSPECIFIED")) {
            log.warn("Unspecified enum. We use the fallback value instead. {}", info);
            return fallBack;
        }

        String protobufEnumPrefix = StringUtils.capitalizeAll(enumType.getSimpleName()) + "_";
        String enumName = name.replace(protobufEnumPrefix,"");
        E result = Enums.getIfPresent(enumType, enumName).orNull();
        if (result == null) {
            log.error("Enum could not be resolved. We use the fallback value instead. {}", info);
            return fallBack;
        }
        return result;
    }

// If we do not provide a fallback value we throw an exception
public static <E extends Enum<E>> E enumFromProto(Class<E> enumType, String name) {
        String info = "Enum type= " + enumType.getSimpleName() + "; name=" + name;
        checkNotNull(name, "Enum name must not be null. "+info);
        checkArgument(!name.endsWith("_UNSPECIFIED"), "Unspecified enum. "+info);
       
        String protobufEnumPrefix = StringUtils.capitalizeAll(enumType.getSimpleName()) + "_";
        String enumName = name.replace(protobufEnumPrefix,"");
        E result = Enums.getIfPresent(enumType, enumName).orNull();
        checkNotNull(result, "Enum could not be resolved. "+info);
        return result;
    }
```

account/src/main/java/bisq/account/protocol_type/LoanProtocolType.java
```diff
@@ -20,15 +20,18 @@
...
+
     @Override
     public bisq.account.protobuf.LoanProtocolType toProto() {
-        return bisq.account.protobuf.LoanProtocolType.valueOf(name());
+        return bisq.account.protobuf.LoanProtocolType.valueOf(protobufPrefix + name());
     }
 
     public static LoanProtocolType fromProto(bisq.account.protobuf.LoanProtocolType proto) {
-        return ProtobufUtils.enumFromProto(LoanProtocolType.class, proto.name());
+        return ProtobufUtils.enumFromProto(LoanProtocolType.class, proto.name().replace(protobufPrefix, ""));
```
HenrikJannsen: Do name conversion generically in ProtobufUtils (see comment there)

HenrikJannsen (changes requested):

account/src/main/java/bisq/account/protocol_type/LoanProtocolType.java
aspire2dev: I changed this further to remove the hardcoded prefix name and use the class simple_name to get it

common/src/main/java/bisq/common/util/ProtobufUtils.java
aspire2dev: Changes applied

account/src/main/java/bisq/account/protocol_type/LoanProtocolType.java
aspire2dev: Done

chat/src/main/java/bisq/chat/channel/ChatChannelNotificationType.java
```diff
@@ -19,22 +19,22 @@
...
 
 public enum ChatChannelNotificationType implements ProtoEnum {
-    UNSPECIFIED,
     GLOBAL_DEFAULT, // As defined in settings
     ALL,
     MENTION,
     OFF;
 
-    private static final String protobufPrefix = "CHATCHANNELNOTIFICATIONTYPE_";
+    private static final String protobufPrefix = StringUtils.capitalizeAll(ChatChannelNotificationType.class.getSimpleName()) + "_";
```
HenrikJannsen: Can be removed as protobufPrefix is not used anymore.

HenrikJannsen (changes requested):

chat/src/main/java/bisq/chat/channel/ChatChannelNotificationType.java
aspire2dev: Hello, it is still used on the toProto method for all of the modified enums. For this specific class, line 34 references the attribute protobufPrefix.

account/src/main/java/bisq/account/protocol_type/LoanProtocolType.java
```diff
@@ -18,17 +18,21 @@
 package bisq.account.protocol_type;
 
 import bisq.common.util.ProtobufUtils;
+import bisq.common.util.StringUtils;
 
 public enum LoanProtocolType implements ProtocolType {
     COLLATERALIZED,
     REPUTATION;
 
+    private static final String protobufPrefix = StringUtils.capitalizeAll(LoanProtocolType.class.getSimpleName()) + "_";
```
HenrikJannsen: This can be moved to the ProtoEnum and the static method reused by ProtobufUtils

```
public interface ProtoEnum {
    ProtocolMessageEnum toProto();

    default String getProtobufPrefix() {
        return getProtobufPrefix(this.getClass());
    }

    static String getProtobufPrefix(Class<?> clazz) {
        return StringUtils.capitalizeAll(clazz.getSimpleName()) + "_";
    }
}

Usage:
public bisq.offer.protobuf.Direction toProto() {
        return bisq.offer.protobuf.Direction.valueOf(getProtobufPrefix() + name());
}

in ProtobufUtils.enumFromProto:
 String enumName = name.replace(ProtoEnum.getProtobufPrefix(enumType),"");
       
```

HenrikJannsen (changes requested):

account/src/main/java/bisq/account/protocol_type/LoanProtocolType.java
aspire2dev: Great suggestion. Will add it to the code.

HenrikJannsen (changes requested):
Some enums are not updated with the latest change request. FeeOption.FeeType
    Direction,..

HenrikJannsen (approved):
ACK

PR #1175, opened 2023-09-14
Disable userprofile combobox in private channels

DrColver:
It should not be possible to change the user profile for private channels. Disable the userprofile combobox when in private channels to avoid this issue.

The current solution was to reset the selection after it occurred, causing a null pointer exception since the combobox is not meant to handle resets like that.


HenrikJannsen (approved):
ACK

HenrikJannsen:
I think it would be even better to switch from the combobox to the text field version (same as if only one profile exist). Currently the disable style for the profile component is a bit weird. But can be done in a followup PR.

PR #1178, opened 2023-09-15
Initial commit to BisqText removal

aspire2dev:
@alvasw and @HenrikJannsen , here's the code for issue 1112.
As the text is enclosed on a VBox, I couldn' t find an easy way to make it wrap independent of the width of the view. So I moved to a Text enclosed in a TextFlow object.


HenrikJannsen (approved):
utACK

PR #1187, opened 2023-09-17
Subscribe profile appearance to private channel status

DrColver:
As suggested in https://github.com/bisq-network/bisq2/pull/1175#issuecomment-1719745345 the combobox is switched to the text field version in private channel mode.

With the new bisq easy style, is there a way to close private channels?

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/content/components/UserProfileSelection.java
+++ b/desktop/src/main/java/bisq/desktop/main/content/components/UserProfileSelection.java
@@ -113,7 +113,6 @@ public class UserProfileSelection {
         private Pin userProfilesPin;
         private Pin chatChannelSelectionPin;
         private Pin navigationPin;
-        private Pin disabledPin;
 
         private Controller(ServiceProvider serviceProvider, int iconSize, boolean useMaterialStyle) {
             this.userIdentityService = serviceProvider.getUserService().getUserIdentityService();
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/main/content/components/UserProfileSelection.java
```diff
@@ -291,13 +284,22 @@ protected void onViewAttached() {
                 }
             });
             comboBoxWidthPin = EasyBind.subscribe(model.comboBoxWidth, w -> comboBox.setComboBoxWidth(w.doubleValue()));
+
+            isComboBoxPin = FxBindings.subscribe(model.isPrivateChannel, isPrivate -> {
```
HenrikJannsen, line 288: I think it would be more clear if we call IsPrivateChannel field in model `useComboBox` and adopt that to the local variable and pin names.

HenrikJannsen (approved):
utACK

desktop/src/main/java/bisq/desktop/main/content/components/UserProfileSelection.java
DrColver: I think that would be more confusing, for example in this case
```
                if (model.isPrivateChannel.get() && selectedUserIdentity != null) {
                    new Popup().warning(Res.get("chat.privateChannel.changeUserProfile.warn", selectedUserIdentity.getUserProfile().getUserName()))
                            .onClose(() -> {
                                model.selectedUserProfile.set(null);
                                model.selectedUserProfile.set(new ListItem(selectedUserIdentity));
                            })
                            .show();
```

The field is describing if it's a private channel rather than if the combobox should be used.

With the name `isComboBoxPin` for the pin it's quite clear that the status as private or public channel is what dictates whether the profile should be a combobox or a text field.

Issue #1177, opened 2023-09-14
Layout bug with TextList component


HenrikJannsen:
Sometimes the layout is messed up. I guess its caused by some delayed rendering cycles.

<img width="947" alt="Screenshot 2023-09-14 at 23 17 01" src="https://github.com/bisq-network/bisq2/assets/116298498/f16a1bdf-14a3-4358-aa70-f9bc65f013ee">

HenrikJannsen:
An alternative implementation using TextFlow and Text might be an option.

aspire2dev:
I'll take this one

aspire2dev:
@HenrikJannsen and @alvasw, follows a PR for review

PR #1200, opened 2023-09-19
Switching TextList implementation from Labels to Text and TextFlow.

aspire2dev:
Fixes bisq-network/bisq2#1177


HenrikJannsen (approved):
ACK

Issue #794, opened 2023-05-01
[deferred] Issues when deleting user profile


HenrikJannsen:
When a user profile is deleted and there are still open private chats, the previous user profile will be selected once the channel is selected.
We should not allow to delete a user profile as long the profile is used in any private chat or is used by offers. The user need to delete the offers and close the private channels before being able to delete the user profile.
The user profile is not deleted from the identity layer but only retired, so we could still receive private messages.
Maybe easiest would be to not allow to delete a user profile as it comes with quite some complexities.

HenrikJannsen:
For now we hide the delete button.
Impl. with https://github.com/bisq-network/bisq2/pull/805

PR #1198, opened 2023-09-19
Add bisq easy module

HenrikJannsen:

PR #1192, opened 2023-09-18
Reenable delete profile button

DrColver:
Fix https://github.com/bisq-network/bisq2/issues/794

Check if delete profile allowed and disable button if not allowed

Add delete profile button tooltip when button is disabled

Set callback from ChatService to UsedIdentityService to test if there are any open private channels or messages in public channels from the profile that is about to be deleted.

Setting the callback from ChatService for the check of user profile activity doesn't really follow the style of the project. Is there another recommended way to handle this?

Messages to a deleted profile are not handled. I tried handling that but it doesn't make much sense. If a profile is deleted it's meant to not be active anymore so all messages to the deleted profile are ignored.

PR #1201, opened 2023-09-19
Delete profile

DrColver:
Replaces PR https://github.com/bisq-network/bisq2/pull/1192

Based on PR https://github.com/bisq-network/bisq2/pull/1198

Fix https://github.com/bisq-network/bisq2/issues/794

### DIFF
--- a/bisq_easy/src/main/java/bisq/bisq_easy/BisqEasyService.java
+++ b/bisq_easy/src/main/java/bisq/bisq_easy/BisqEasyService.java
@@ -1,0 +1,47 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.bisq_easy;
+
+import bisq.common.application.Service;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.concurrent.CompletableFuture;
+
+@Slf4j
+@Getter
+public class BisqEasyService implements Service {
+
+    public BisqEasyService() {
+    }
+
+
+    ///////////////////////////////////////////////////////////////////////////////////////////////////
+    // Service
+    ///////////////////////////////////////////////////////////////////////////////////////////////////
+
+    public CompletableFuture<Boolean> initialize() {
+        log.info("initialize");
+        return CompletableFuture.completedFuture(true);
+    }
+
+    public CompletableFuture<Boolean> shutdown() {
+        log.info("shutdown");
+        return CompletableFuture.completedFuture(true);
+    }
+}
--- a/bisq_easy/src/main/java/bisq/bisq_easy/BisqEasyService.java
+++ b/bisq_easy/src/main/java/bisq/bisq_easy/BisqEasyService.java
@@ -22,8 +22,10 @@ import bisq.bonded_roles.BondedRolesService;
 import bisq.chat.ChatService;
 import bisq.common.application.Service;
 import bisq.contract.ContractService;
+import bisq.i18n.Res;
 import bisq.identity.IdentityService;
 import bisq.network.NetworkService;
+import bisq.network.p2p.services.data.DataService;
 import bisq.offer.OfferService;
 import bisq.persistence.PersistenceService;
 import bisq.presentation.notifications.NotificationsService;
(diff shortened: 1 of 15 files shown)


chat/src/main/java/bisq/chat/ChatService.java
```diff
@@ -205,7 +207,22 @@ public Optional<TwoPartyPrivateChatChannel> createAndSelectTwoPartyPrivateChatCh
             optionalChannel.ifPresent(channel -> getChatChannelSelectionService(chatChannelDomain).selectChannel(channel));
             return optionalChannel;
         }
+    }
+
+    public boolean isIdentityUsed(UserIdentity userIdentity) {
```
HenrikJannsen, line 212: Maybe better `isUserIdentityUsed`?

desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
```diff
@@ -110,11 +114,12 @@ public UserIdentity fromString(String string) {
         saveButton = new Button(Res.get("action.save"));
         saveButton.setDefaultButton(true);
 
-        deletedButton = new Button(Res.get("user.userProfile.deleteProfile"));
+        deleteButton = new Button(Res.get("user.userProfile.deleteProfile"));
+        deleteWrapper = new SplitPane(deleteButton);
```
HenrikJannsen, line 118: Why is a SplitPane used here?

user/src/main/java/bisq/user/identity/UserIdentityService.java
```diff
@@ -215,14 +215,11 @@ public CompletableFuture<DataService.BroadCastDataResult> editUserProfile(UserId
                 .thenCompose(result -> networkService.publishAuthenticatedData(newUserProfile, oldIdentity.getNodeIdAndKeyPair().getKeyPair()));
     }
 
-    public CompletableFuture<DataService.BroadCastDataResult> deleteUserProfile(UserIdentity userIdentity) {
-        //todo add more checks if deleting profile is permitted (e.g. not used in trades, PM,...)
-        if (getUserIdentities().size() <= 1) {
```
HenrikJannsen, line 220: Maybe we should keep that check? The app would be in an unexpected/untested state without any UserIdentity. I know the callers check, but that module does not know about that. The rule that at least one identity must exist is inside its domain so I think we should have the check here.

bisq_easy/src/main/java/bisq/bisq_easy/BisqEasyService.java
```diff
@@ -105,4 +111,16 @@ public CompletableFuture<Boolean> shutdown() {
...
     }
+
+    public boolean isDeleteUserIdentityProhibited(UserIdentity userIdentity) {
+        return chatService.isIdentityUsed(userIdentity) ||
+                !userIdentityService.hasMultipleUserIdentities();
+    }
+
+    public CompletableFuture<DataService.BroadCastDataResult> deleteUserIdentity(UserIdentity userIdentity) {
+        if (isDeleteUserIdentityProhibited(userIdentity)) {
+            return CompletableFuture.failedFuture(new RuntimeException(Res.get("user.userProfile.deleteProfile.cannotDelete")));
```
HenrikJannsen: I think we should not reuse a UI text here as error message. Error messages are so far mostly in plain English and not from Resources and should only be escalated to the user by the error popup which can be filled with more technical data. As we check anyway at the caller that error should never be thrown anyway (but good to leave it at the service level).

HenrikJannsen (changes requested):
Some nits...

chat/src/main/java/bisq/chat/ChatService.java
DrColver: I think that would be less readable considering the argument is UserIdentity.

desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
DrColver: A Control is needed to add the Tooltip. When the button is disabled it doesn't receive mouse events so it wouldn't show the Tooltip. Other Controls would work as well as wrappers.

user/src/main/java/bisq/user/identity/UserIdentityService.java
DrColver: Good point. I added it again.

bisq_easy/src/main/java/bisq/bisq_easy/BisqEasyService.java
DrColver: Fixed.

chat/src/main/java/bisq/chat/ChatService.java
HenrikJannsen: Ok

desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
HenrikJannsen: Yes, I was just wondering why Splitpane was selected. I guess Pane would be the cheapest one as it does not support auto-layout (if that is not needed as I assume so).

HenrikJannsen (approved):
utACK

desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
DrColver: A `Control` is needed for the tooltip support. Maybe there is a better choice than SplitPane but from what I've seen it's reasonable. It's meant to take Nodes as argument and doesn't decorate it when it's just a single node. Maybe ScrollPane would be better, not sure, don't think it makes a big difference.

Issue #1194, opened 2023-09-18
Do not add messages on other markets to notidication badge updates


HenrikJannsen:
With https://github.com/bisq-network/bisq2/pull/1193 there are the notification badges on the Bisq Easy tabs added.
The user gets all the message notification of any market. They should be only notified about the selected market.

DrColver:
I will take this.

PR #1208, opened 2023-09-20
Show notification for selected channel only

DrColver:
Fix https://github.com/bisq-network/bisq2/issues/1194

There are some other oddities concerning the badges when switching between markets. They show the total amount of messages in a market in the dropdown, but I suspect the idea is to show number of unread messages for each market channel.

I don't think it's a good idea to track this for all markets. Probably best to track only for some selected. The UI would become overwhelming if all markets show a lot of unread messages, and likely it's also irrelevant since most users would be interested in just one market, or maybe two.

### DIFF
--- a/chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
+++ b/chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
@@ -128,16 +128,12 @@ public class ChatNotificationService implements Service {
                 onChatChannelsChanged(bisqEasyOfferbookChannelService.getChannels()));
 
         chatService.getCommonPublicChatChannelServices().values()
-                .forEach(commonPublicChatChannelService -> {
-                    commonPublicChatChannelService.getChannels().addListener(() ->
-                            onChatChannelsChanged(commonPublicChatChannelService.getChannels()));
-                });
+                .forEach(commonPublicChatChannelService -> commonPublicChatChannelService.getChannels().addListener(() ->
+                        onChatChannelsChanged(commonPublicChatChannelService.getChannels())));
 
         chatService.getTwoPartyPrivateChatChannelServices().values()
-                .forEach(twoPartyPrivateChatChannelService -> {
-                    twoPartyPrivateChatChannelService.getChannels().addListener(() ->
-                            onChatChannelsChanged(twoPartyPrivateChatChannelService.getChannels()));
-                });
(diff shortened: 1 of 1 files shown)


HenrikJannsen:
> They show the total amount of messages in a market in the dropdown, but I suspect the idea is to show number of unread messages for each market channel.

No the idea is to show the offers per market. But we should use a different icon style (maybe orange background or an icon instead of the circle).

chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
```diff
@@ -245,7 +246,8 @@ private <M extends ChatMessage> void chatNotificationAdded(ChatChannel<M> chatCh
         }
 
         // Only notify for the currently selected channel
-        if (!chatService.getChatChannelSelectionService(chatChannel.getChatChannelDomain()).getSelectedChannel().get().equals(chatChannel)) {
+        if (chatChannel instanceof PublicChatChannel &&
```
HenrikJannsen: With PublicChatChannel we apply it the the common channels as well. I guess there its ok to show all notifications. Users can disable them anyway per channel. If applied only for BisqEasyOfferbookChannel the instanceof check should be using BisqEasyOfferbookChannel. I think only applying the filtering to BisqEasyOfferbookChannel is better.

HenrikJannsen (changes requested):

chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
DrColver: Implemented as suggested.

I think it would be better to just show number of unread messages in non selected channels, but as you say, it's possible to disable per channel. We could set the default for public channels to not show notifications, unless the intention is to get users engaged with the app and feel there is something happening. Maybe something to consider for a later release when notifications start being a nuisance rather than engaging.

HenrikJannsen (approved):
ACK

chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
HenrikJannsen: I just made a PR which changed the red color to grey for the default notification, so they do not get that annoying. 
I think in the beginning its better to boost the other common channels so ppl use it. As you said if it gets too much over time we can change default setting to turn them off. Also there will be further refinement for the whole notification design, but atm I think its better to focus on the most important...

Issue #1195, opened 2023-09-19
Persist expand/collapse state of left navigation


HenrikJannsen:
Use the Cookie class to persist if user collapsed or expanded left navigation

aspire2dev:
I'll take this one

PR #1212, opened 2023-09-20
Persist expand/collapse state of left navigation

aspire2dev:
Resolves bisq-network/bisq2#1195


desktop/src/main/java/bisq/desktop/main/left/LeftNavController.java
```diff
@@ -106,7 +107,9 @@ void onNavigationTargetSelected(NavigationTarget navigationTarget) {
     }
 
     void onToggleHorizontalExpandState() {
-        model.getMenuHorizontalExpanded().set(!model.getMenuHorizontalExpanded().get());
+        boolean newState = !model.getSettingsService().getCookie().asBoolean(CookieKey.MENU_HORIZONTAL_EXPANDED).orElse(false);
```
HenrikJannsen: We are setting the state from the cookie in the model. we should not read it again but use our model state and only apply it to the cookie here. The false value as fallback would be confusing otherwise.

aspire2dev: Replaced false by the value contained in the model. That was a good catch. Thanks!

HenrikJannsen (changes requested):
Suggestion: 

```
Index: desktop/src/main/java/bisq/desktop/main/left/LeftNavController.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/desktop/src/main/java/bisq/desktop/main/left/LeftNavController.java b/desktop/src/main/java/bisq/desktop/main/left/LeftNavController.java
--- a/desktop/src/main/java/bisq/desktop/main/left/LeftNavController.java	(revision 45e55294b8d8a9f4803332cb8422391e88f8ac92)
+++ b/desktop/src/main/java/bisq/desktop/main/left/LeftNavController.java	(date 1695396337185)
@@ -69,6 +69,8 @@
 
         releaseNotificationPin = updaterService.getReleaseNotification().addObserver(releaseNotification ->
                 UIThread.run(() -> model.getNewVersionAvailable().set(releaseNotification != null)));
+
+        model.getMenuHorizontalExpanded().set(model.getSettingsService().getCookie().asBoolean(CookieKey.MENU_HORIZONTAL_EXPANDED).orElse(false));
     }
 
     @Override
@@ -107,8 +109,7 @@
     }
 
     void onToggleHorizontalExpandState() {
-        boolean newState = !model.getSettingsService().getCookie().asBoolean(CookieKey.MENU_HORIZONTAL_EXPANDED)
-                .orElse(model.getMenuHorizontalExpanded().get());
+        boolean newState = ! model.getMenuHorizontalExpanded().get();
         model.getSettingsService().setCookie(CookieKey.MENU_HORIZONTAL_EXPANDED, newState);
         model.getMenuHorizontalExpanded().set(newState);
     }
     ```

HenrikJannsen (approved):
utACK

PR #1221, opened 2023-09-23
Improve network status in Splash screen

### LINKED ISSUE #1165: Improve Splash screen
We show below the progress indicator the network type and a % value (e.g. TOR 50%).
We should improve the display of connection to multiple networks and use more readable strings.
E.g. multiple lines with:
[TOR ICON] Connecting to Tor network: 50%
[I2P ICON] Connecting to I2P network: 75%

mrblue313:
Enhance display of network connection progress
in Splash screen:

* Add a new line for each network.
* Add icon and improved string.

Resolve #1165

Animation_01

Animation_02


desktop/src/main/java/bisq/desktop/splash/SplashView.java
```diff
@@ -55,17 +68,29 @@ public SplashView(SplashModel model, SplashController controller) {
         progressBar.setMaxHeight(3);
         progressBar.setMinWidth(535);
 
-        transportStatusLabel = new Label("");
-        transportStatusLabel.getStyleClass().add("bisq-small-light-label-dimmed");
+        clearStatusLabel = new Label();
+        torStatusLabel = new Label();
+        i2pStatusLabel = new Label();
 
+        clearStatusHBox = new HBox(new Label("Connecting to ClearNet  "), clearStatusLabel);
```
HenrikJannsen: Those Strings should be resolved from translation strings (`Res.get(key...)`)

desktop/src/main/java/bisq/desktop/splash/SplashView.java
```diff
@@ -77,9 +102,68 @@ protected void onViewAttached() {
...
+        this.unbindNetworkState(clearStatusHBox, clearStatusLabel, clearIcon);
+        this.unbindNetworkState(torStatusHBox, torStatusLabel, torIcon);
+        this.unbindNetworkState(i2pStatusHBox, i2pStatusLabel, i2pIcon);
         progressBar.progressProperty().unbind();
         progressBar.setProgress(0);
         appStateSubscription.unsubscribe();
     }
+
+    private HBox createNetworkStatusBox() {
+        final double height = 24;
```
HenrikJannsen: According to our code conventions we do not use `final` in local variables.

HenrikJannsen (approved):
ACK (beside small NITs)

HenrikJannsen:
Looks great, thanks!

I am wondering if it would be better to keep the % string in white instead of green and to use uppercase for the strings.

Screenshot 2023-09-23 at 23 45 49

Another small thing is that the 2 spaces at the end of the string could be removed and a spacing in the hbox used instead. When translation strings are used we usually try to avoid to add layout-related spaces, specially at the end of the string as the IDE would mark that as a potential bug.

desktop/src/main/java/bisq/desktop/splash/SplashView.java
mrblue313: Do you refer to any docs for code conventions?

mrblue313:
I tested different ways on how to set this.

I think lower case with status in green is the most readable format (especially when having multiple lines).
The green color on the status is mainly for two reasons:
- It shows that it represents a variable.
- It makes sense to me to have % in green just like the progress bar above.

As for the lower case, I think this way it represents a subtitle from the main status above.

HenrikJannsen (approved):
utACK

Issue #1184, opened 2023-09-16
Handle case of different identity used with same peer in private chat


HenrikJannsen:
If a user has multiple profiles set up and start a PM with the same peer with the different profiles selected, there will be multiple different private chats opened. Currently we do not display the information which identity was use. Only in the profile selection we would see it as the selection changes with the selected private channel. 

See those 2 screenshot with that case:
<img width="1125" alt="Screenshot 2023-09-17 at 01 16 36" src="https://github.com/bisq-network/bisq2/assets/116298498/de8f6232-4749-4fba-93a2-0988871f17a3">
<img width="1123" alt="Screenshot 2023-09-17 at 01 16 43" src="https://github.com/bisq-network/bisq2/assets/116298498/ec0f2aa2-8fe5-4170-ac13-b6f40de207cb">

DrColver:
This is an issue for the open trades list as well where space is more limited. Will take a look at that as well.

PR #1231, opened 2023-09-25
Show own profile info in private chat list

DrColver:
Add column to show own user name in private chat list, fix https://github.com/bisq-network/bisq2/issues/1184


HenrikJannsen (approved):
ACK

PR #1233, opened 2023-09-27
Fix bug in ComboBoxWithSearch menu

mrblue313:
Fix the following bugs:

* When search result is empty, the placeholder text should stay within the menu. (At the moment it was overflowing).

Before:
overflow-bug

After:
overflow-bug-fixed

* When clicking outside the menu, the menu should close. (At the moment it was getting stuck - only closing when clicking in very specific places - and if clicking on the parent, multiple instances of it were being created).

Before:
closing-bug

After:
closing-bugfix


HenrikJannsen (approved):
utACK

PR #1235, opened 2023-09-29
Introduce new icons for Bisq Easy waiting states

### LINKED ISSUE #1209: Change waiting animation to new design
@pedromvpg delivered those new designs for the waiting animation in the trade state screens.
The ring should be animated.  JavaFx has some performance issues with animations (progress indicator takes way too much cpu), so the animation should be implemented with that problem in mind (e.g. a simple rotation of the image might be ok).

image
image2

Here are pngs from the screenshots (not great quality, waiting for Pedro to deliver pngs).

animation-circle@2x
animation-block@2x
animation-file@2x

axpoems:
- Design new icons for all of the waiting states in the Bisq Easy protocol.
- Change `WaitingAnimation` component to use a rotation of a png image of a spinning circle instead to improve CPU performance.
- Introduce the different icons into each buyer and seller view states accordingly.

At the moment, `FIAT_PAYMENT` and `FIAT_PAYMENT_CONFIRMATION` use the same icon. I think it works well as is, but will explore that some more.

Fix #1209

buy-newicons+

sell-newicons+


desktop/src/main/java/bisq/desktop/components/controls/WaitingState.java
```diff
@@ -0,0 +1,10 @@
+package bisq.desktop.components.controls;
```
HenrikJannsen: I think that should be moved inside into the bisq_easy components package as its not generic but special for bisq easy.

HenrikJannsen (approved):
utACK

desktop/src/main/java/bisq/desktop/components/controls/WaitingState.java
axpoems: Good point! Should we make the `WaitingAnimation` specific to Bisq Easy then as well? since it depends on this state to decide the `innerIcon`.
Another option could be that we keep just a record of all the possible `WaitingStates` for any protocol in common even if we don´t necessarily use them all.

HenrikJannsen: > Good point! Should we make the WaitingAnimation specific to Bisq Easy then as well? since it depends on this state to decide the innerIcon.

Yes, I guess thats better.

> Another option could be that we keep just a record of all the possible WaitingStates for any protocol in common even if we don´t necessarily use them all.

I fear thats too premature as long we dont have the other protocols. Over time we will surely generalize things...

PR #1236, opened 2023-09-29
Fix remove password

namloan:
This isn't working because confirmedPassword is always false (as there's no confirmedPassword in the remove screen), so an exception is thrown.

This check doesn't add any value as password validation already happens on text input, so it's safe to remove.


HenrikJannsen (approved):
utACK

namloan:
Closing this PR without merging as I'll send a new one with many more changes that override this one

Issue #1211, opened 2023-09-20
Refactor and clean up css/styles and standardize styles in app


HenrikJannsen:
- cleanup and generalize css (atm lot of duplicates, changes not adopted to style names,...)
- get a set of styles we want to use and apply them to all screens. atm there might be small differences in font size/weight,...
- adjust all images/icons to the set of colors we use. the current green gets an update. the greys are not consistently applies
- standardize the hover/pressed/disabled states of buttons and related controls
- improve styles (some fonts are to small, less readible,..)
- improve or renew icons which do not have a good quality (some icons got derived from screenshots and do not have proper transparancy)
- organize the image directory better (got messed up over time)

axpoems:
I will start working on this.

HenrikJannsen:
I guess we can close that. It is an ongoing effort but I guess large parts are already improved.

PR #1237, opened 2023-09-30
Fix encoding issues in UnorderedList component

axpoems:
Represent the bullet point "•" as a Unicode escape sequence to avoid encoding mismatch.

Towards #1211

Some examples below. Testing on Windows.

before1
after1

before2
after2

before3
after3

before4
after4

before5
after5


HenrikJannsen (approved):
utACK

Issue #1153, opened 2023-09-03
Improve input validation


HenrikJannsen:
We should show a error message below the text input component. 
I guess the help field of the MaterialTextField would be the right place and style (just need to be in red). The help field is rarely used, but if its used we might prefer to not show both in case of invalid input but toggle the visibility of help and error field instead.

I think a popup is not needed as it creates to much attention IMO.

We call the validation at focus out but that concept needs to be reconsidered as if the user has invalid input and not moving the focus out of the input field they will not get the feedback if the invalid input.
One option would be to add a delay of about 2 sec. after last input activity to call the invalidate method additional to the focus out.

To call validate at each input might be annoying as while entering a valid value there might be intermediary invalid inputs (e.g. if min value is 100 and user enter 2 -> 20 -> 200 they would get at first 2 inputs a error message.

HenrikJannsen:
@namloan Would you be available for that issue?

namloan:
Agreed on Matrix to do the following:

* Let "next" buttons be enabled and call validate on click. if invalid input show error msg at invalid input component. if there is none (like payment method chip buttons) show a popup (e.g. you need to select at least 1 payment method).
* On focus out call validate.
* No delayed validation if user leaves focus in field. when he clicks next he gets the error.

HenrikJannsen:
Related to https://github.com/bisq-network/bisq2/issues/1152

HenrikJannsen:
@namloan Are you still working on that?

namloan:
nope, free for anyone to take. However does this still make sense? we may need to review the places where we're still missing validations or whether we'd like the current ones to behave differently.

rodvar:
@namloan @HenrikJannsen What would you guys say is the status on this one? can I help? The PR referred as related (https://github.com/bisq-network/bisq2/pull/1152) has been dropped

HenrikJannsen:
There are some deprecated Validator classes still around. Those should be replaced. I think the rest can be done on demand when we see a validation is missing.

HenrikJannsen:
@rodvar I took the issue myself...

PR #1238, opened 2023-09-30
Improve password validation

namloan:
#1153

* All password screens - onboarding, set up, remove, unlock
* Now validates on focus out and button click. No more validation on text input.
* Added error messages below textboxes.
* Removed error message tooltips.

improve_password


HenrikJannsen (approved):
ACK

HenrikJannsen:
I think we should not use the generic error popup but either the warning or make the error popup more customizable to not show the bug report text. Maybe we should rename the existing error to exception and add an error which is same like warning just red error icon. 
 
Screenshot 2023-09-30 at 21 37 27

PR #1244, opened 2023-10-02
Adapting the Trade Apps view to use the GridPaneUtil class. 

aspire2dev:
Changing the structure of the view, whereas before we had two gridpanes per line of protocols, now only one pane with 6 columns is in place to avoid misalignment of artifacts.

This is a follow-up for the aspire2dev/fix_learn_issues branch.


HenrikJannsen (approved):
ACK

PR #1245, opened 2023-10-03
Fix missing dashboard content at first start

djing-chan:
At a fresh startup the Dashboard did not get initialized as NONE was defined as default navigation target in ContentModel.


HenrikJannsen (approved):
utACK

Issue #1218, opened 2023-09-23
Make sure that important text fields are not truncated


HenrikJannsen:
When scaling the app down to min. size some labels/textfields are truncated (e.g. header of tabviews).
We need to ensure that the important fields get not truncated. Using min width would be easiest but has the drawback that it does not support diff. sizes with translates strings. Better to use hBox growth ALWAYS if possible or other means. But if too complicated minWidth is better than nothing, if small part is truncated its acceptable as well. Also there will be tabs truncated and we need to make the trade off what to truncate. We could also hide the tab headline when its too small width.

aspire2dev:
I volunteer for this task

PR #1247, opened 2023-10-03
Adjusting the TabView text.

aspire2dev:
Fixes non-wrapping text at User Options view.

bisq-network/bisq2#1218


HenrikJannsen (approved):
NACK

We should minWidth instead of wrapText

HenrikJannsen (changes requested):
NACK

HenrikJannsen:
Actually I think its better to use left aligned tabs like at Bisq Easy instead and by that remove the headline at all.

aspire2dev:
> Actually I think its better to use left aligned tabs like at Bisq Easy instead and by that remove the headline at all.

Change was applied. The headline was removed and the tabs were moved to the left.

HenrikJannsen (approved):
utACK

PR #1249, opened 2023-10-03
Improve amount validation

namloan:
* Stopped validating at text input
* Next button always enabled

Requesting feedback about further improvements to do in this screen. Should we add more friction when any of the amounts are invalid?

improve_amount_val_2


djing-chan:
I think we should set the value from one of the other fields (either the btc amount or the slider) at focus out if the input was invalid. If the user clicks next without prior focus out we should show a popup that the input was invalid and apply a valid input after popup is closed. Alternatively we could leave the invalid input but I guess that causes more problems in the component.
Leaving it empty does not feel right to me.

namloan:
Thanks for the feedback @djing-chan. It makes sense. Will update.

namloan:
@djing-chan I've updated to always autocorrect on focus out. When clicking Next, the amount is autocorrected before moving to the next screen (because the amount loses the focus when clicking the button). Do you still think a popup is needed in this case?

For context - we're being careful with popups as we're only using them when the validation is perform on the backend or in very specific cases like no input provided in the payment method chip buttons screen.

new_amount_validation

djing-chan:
If the input is a valid number but out of range (e.g. 500 EUR) then it might be a bit surprising that at next and the review screen another value is used. But I think both is ok, but a popup might be a bit more on the safe side.

djing-chan (approved):
ACK

HenrikJannsen (approved):
utACk

Issue #1210, opened 2023-09-20
UI bugs when animations are turned off in Settings


HenrikJannsen:
When animations are turned of in Settings there are some issues. One is that the green tab marker lines are not positioned. There might be more...

PR #1251, opened 2023-10-04
Adjusting tab bug when animations are off

aspire2dev:
There were two issues affecting the rendering of the green tab markers lines. One of them was related to the maybeAnimateMark function only being called when transitionStarted was true. After removing the condition, rendering was breaking when the user view was the first to be opened, and that was adjusted increasing the timing of the transition painting method by 150ms

bisq-network/bisq2#1210


HenrikJannsen (approved):
ACK

PR #1250, opened 2023-10-04
[TradeWizard] Show warning popup when no payment methods are selected

namloan:
#1153

validation_pm

PR #1253, opened 2023-10-06
Add validation popup inside an overlay

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1250


HenrikJannsen (approved):
utACK

Issue #1041, opened 2023-07-27
Improve network info screen


HenrikJannsen:
Adopt screen to new design and UI elements

aspire2dev:
I volunteer for this task

aspire2dev:
Submitting the code now... This is my suggestion for the Network Info screen. (I left clearnet and tor running to be able to show how it maps with more than one protocol, as we expect tor and i2p in production)

![image](https://github.com/bisq-network/bisq2/assets/143408636/64a1241c-ea9a-4018-bf8d-b1cd3b342cfb)

aspire2dev:
@HenrikJannsen , I think we can close this issue, no?

PR #1254, opened 2023-10-07
Improve network info screen

aspire2dev:
Removed the BisqGridPane code.
Moved the functions to add the text boxes to GridPaneUtil. Changed slightly the styling of the network info view to make it look like the remainder of the newer views.

bisq-network/bisq2#1041


HenrikJannsen (approved):
ACK

HenrikJannsen:
It could e further improved by increasing column width of number of connections so that the headline does not truncate.
Further to reduce column width of address in first table and use toopltips for all data which can be truncated when using smaller width.

aspire2dev:
> It could e further improved by increasing column width of number of connections so that the headline does not truncate. Further to reduce column width of address in first table and use toopltips for all data which can be truncated when using smaller width.

I'll work on this next, once back.

PR #1258, opened 2023-10-08
Improve notification panel and tab view menu

axpoems:
Change the notification panel to a single line/smaller band.
* This is a simpler implementation that optimizes for visibility and used space.
* To achieve a better user experience, the menu in tab view is modified to an always visible position. Now, only content is scrollable. In this way, it is possible to jump to other tabs as well as know where you are at any position. This change is also necessary to implement the new design for the notification panel.

After
notification-after

Before
notification-before


HenrikJannsen:
I think a white button looks better then green on yellow.

HenrikJannsen:
Thanks for working on that!
I played a bit around with it. Here my feedback:

- top padding of dashboard is 20 px too high
- there is a 20 px gap in the chat views (discussion,...)
- if content scrolls up there is a padding in the tab views but not at dashboard. I think it should behave the same for all views. I am not sure if its better to let the content disappear at the tab line or better with a padding as you did in the tab views. I guess more natural would be if there would be a different background but that would break design in non scrolled state. 
- tabline should have left and right padding like content
- at scrolling at dashboard the right padding of the notification and the content are not aligned. At tab views its not an issue as the visual distance is larger. So if we use some gap at dashboard that might be enough as fix.
- if we reduce the tab paddings I think the padding of the tab buttons to the line should be the same then as in the tabs used in the popups (e.g. trade guide) (see screenshot below). I think we should also use the light grey line (as in popups) instead of the dark grey line.
- IMO the green button on yellow background does not work that well. We could use a hyperlink with same text size as the text (see screenshot below)
- I would prefer to not have the scrollbar handling in all content child views but get it generalized in content view. I will have a look into that and might send you a suggestion.


Screenshot 2023-10-09 at 22 13 34

Screenshot 2023-10-09 at 22 26 22

Screenshot 2023-10-09 at 22 52 12

axpoems:
> * IMO the green button on yellow background does not work that well. We could use a hyperlink with same text size as the text (see screenshot below)

I agree on both the green button on yellow and the use of hyperlink. I have updated it now:

notification-after2

axpoems:
Thank you for the review.

> * top padding of dashboard is 20 px too high

Fixed.

> * there is a 20 px gap in the chat views (discussion,...)

Fixed.

> * if content scrolls up there is a padding in the tab views but not at dashboard. I think it should behave the same for all views. I am not sure if its better to let the content disappear at the tab line or better with a padding as you did in the tab views. I guess more natural would be if there would be a different background but that would break design in non scrolled state.

Good point. Perhaps we could add the same padding that we have under the tab line?

> * tabline should have left and right padding like content

I agree with that, it looks better that way. I could work on that on a follow-up PR.

> * at scrolling at dashboard the right padding of the notification and the content are not aligned. At tab views its not an issue as the visual distance is larger. So if we use some gap at dashboard that might be enough as fix.

Can you clarify what you mean with "if we use some gap at dashboard that might be enough as fix"?

> * if we reduce the tab paddings I think the padding of the tab buttons to the line should be the same then as in the tabs used in the popups (e.g. trade guide) (see screenshot below). I think we should also use the light grey line (as in popups) instead of the dark grey line.

Could be. I tried keeping everything else as it was. We could improve pop-ups in another PR. I also think that buttons in pop-up which follow a flow (like in this case the "next" button) should always be found in the same position. Then the flow feels more natural.

> * IMO the green button on yellow background does not work that well. We could use a hyperlink with same text size as the text (see screenshot below)

Fixed.

> * I would prefer to not have the scrollbar handling in all content child views but get it generalized in content view. I will have a look into that and might send you a suggestion.

I considered that option (and implemented it in code as well) while deciding which solution to choose. IMO it makes sense that every content child view chooses how to use scrolling. For some views it might not even be necessary to have it in the first place. I also didn't like the fact that I had to have the "optional" header in the `MainView`, when having a header doesn't apply to all of the views at the moment.

HenrikJannsen:
> I agree with that, it looks better that way. I could work on that on a follow-up PR.

Wouldn't it be just a small fix? I guess it would be better to have all the changes in one PR.

> Good point. Perhaps we could add the same padding that we have under the tab line?

Yes, I guess that might be the best option as long we don't find a better one. I looked at other webpages/apps (Github scrolles the tabs with the content, IntelliJ scrolls content to the line without gap) to find inspiration but I guess with a gap it fits best in our case.

 > Can you clarify what you mean with "if we use some gap at dashboard that might be enough as fix"?

I meant if we add a gap between the line and the scrollable area in dashboard the mis-alignment of the right paddings due the scrollbar width will not be that problematic as it is now where the scroll area touches the notification pane and the mis-alignment is very visible.

>Could be. I tried keeping everything else as it was. 

The vertical padding between tab buttons and line was much larger in the original design as well as left and right paddings. I liked it as it gave lot of space but i fear on small screens we are wasting too much of content space, so I am fine with your suggestion to reduce those spaces. But the popups have a slightly larger vertical padding and I think we should stick wither with Pedros design to not change too much of his original design.

> We could improve pop-ups in another PR. I also think that buttons in pop-up which follow a flow (like in this case the "next" button) should always be found in the same position. Then the flow feels more natural.

Yes agree. The headlines in the popups are not all the same I guess and the next/back buttons should be always same position.

> I considered that option (and implemented it in code as well) while deciding which solution to choose. IMO it makes sense that every content child view chooses how to use scrolling. For some views it might not even be necessary to have it in the first place.

Yes I agree that the view itself should decide on the scroll policy. Maybe we can use a ScrollableTabView subclass which handles all the scollpane related things and layout and which all concrete TabViews which use scolling extend from instead of TabView. That way we avoid the code duplication and keep the responsibilities intact.

> I also didn't like the fact that I had to have the "optional" header in the MainView, when having a header doesn't apply to all of the views at the moment.

Would need to look into that in the code. But yes sounds like a candidate for refactoring. Feel free to touch that as well.

HenrikJannsen:
- The hyperlink normal state is regular font and hover bold. That causes a re-layout and the background size changes a bit. I think it would be better to use bold in both cases and make hover with color change. The white has little contrast. Maybe we can consider to use a dark grey tone for the text and black for the hyperlink?

What about: 
Screenshot 2023-10-10 at 13 03 48

```

.notification-headline {
    -fx-fill: -bisq-black;
    -fx-text-fill: -bisq-black;
    -fx-font-size: 1.15em;
    -fx-font-family: "IBM Plex Sans";
}

.notification-hyperlink, .hyperlink:visited {
    -fx-fill: -bisq-black;
    -fx-text-fill: -bisq-black;
    -fx-font-size: 1.15em;
    -fx-font-family: "IBM Plex Sans Medium";
}

.notification-hyperlink:hover, .hyperlink:visited:hover {
    -fx-fill: derive(-bisq-black, 20%);
    -fx-text-fill: derive(-bisq-black, 20%);
    -fx-font-size: 1.15em;
    -fx-font-family: "IBM Plex Sans Medium";
}
```

HenrikJannsen (approved):
ACK

Some issues are open but I leave it to you to include it in that PR or do it in a new one

axpoems:
> > I agree with that, it looks better that way. I could work on that on a follow-up PR.
> 
> Wouldn't it be just a small fix? I guess it would be better to have all the changes in one PR.
> 
> > Good point. Perhaps we could add the same padding that we have under the tab line?
> 
> Yes, I guess that might be the best option as long we don't find a better one. I looked at other webpages/apps (Github scrolles the tabs with the content, IntelliJ scrolls content to the line without gap) to find inspiration but I guess with a gap it fits best in our case.

In response to these two. To be honest, I am not sure either of the *better* solution for this particular use case. So, let's start with this proposal (no paddings underneath and the tab line with no paddings on the sides, separating the tabs with the actual content - as the examples you point out).
Then, while testing we can iterate again on this. See at the end the gif for ref on how it looks like now.

> I meant if we add a gap between the line and the scrollable area in dashboard the mis-alignment of the right paddings due the scrollbar width will not be that problematic as it is now where the scroll area touches the notification pane and the mis-alignment is very visible.

I see what you mean now. Definitely something to round off.  As before, I need to look more into the current solution. IMO ideally the target should be a solution that works for every case and not fix it specifically for some.

> > Could be. I tried keeping everything else as it was.
> 
> The vertical padding between tab buttons and line was much larger in the original design as well as left and right paddings. I liked it as it gave lot of space but i fear on small screens we are wasting too much of content space, so I am fine with your suggestion to reduce those spaces. But the popups have a slightly larger vertical padding and I think we should stick wither with Pedros design to not change too much of his original design.
> 
> > We could improve pop-ups in another PR. I also think that buttons in pop-up which follow a flow (like in this case the "next" button) should always be found in the same position. Then the flow feels more natural.
> 
> Yes agree. The headlines in the popups are not all the same I guess and the next/back buttons should be always same position.

Let's leave popups in general as follow-up.

> Yes I agree that the view itself should decide on the scroll policy. Maybe we can use a ScrollableTabView subclass which handles all the scollpane related things and layout and which all concrete TabViews which use scolling extend from instead of TabView. That way we avoid the code duplication and keep the responsibilities intact.

> > I also didn't like the fact that I had to have the "optional" header in the MainView, when having a header doesn't apply to all of the views at the moment.
> 
> Would need to look into that in the code. But yes sounds like a candidate for refactoring. Feel free to touch that as well.

Thanks for the suggestion! I agree, so let's put this as well as follow-up. I can make a refactor of the code once we decide on the final design.

> I will test and review later today...

Thanks for the review and the comments!

---

> * The hyperlink normal state is regular font and hover bold. That causes a re-layout and the background size changes a bit. I think it would be better to use bold in both cases and make hover with color change. The white has little contrast. Maybe we can consider to use a dark grey tone for the text and black for the hyperlink?

Again, to being honest, I had exactly the same concerns as you are pointing out here. I explored many different options but I wasn't convinced 100% by any.
I like what you are suggesting. I tried something similar, but the opposite configuration (first black, second grey). Considering this, I added a new proposal.

after+++2

---

To summarise, I would say this PR is ready to be merged, and coming next is:

- [ ] Test the current tab line and implement modifications for a better design, if needed
- [ ] Look into fine tuning the pop-ups
- [ ] Refactor the scrollable tab views at the end

PR #1259, opened 2023-10-09
Use left sided tabs by default

djing-chan:
Invert default value for isRightSide as we use now left sided tabs everywhere.


HenrikJannsen (approved):
utACK

PR #1260, opened 2023-10-09
Optimize waiting animation

djing-chan:
The rotation animation consumes about 5% of CPU on a high end laptop and about 30% on an older laptop.

This commit changes the infinite rotation to a 1 time rotation and adds a fade transition. The animation gets triggered at window focus again.


axpoems:
> The rotation animation consumes about 5% of CPU on a high end laptop and about 30% on an older laptop.

Which benchmarks were utilized to obtain these figures?

djing-chan:
> > The rotation animation consumes about 5% of CPU on a high end laptop and about 30% on an older laptop.
> 
> Which benchmarks were utilized to obtain these figures? and which method did you use to run the tests?

I realized that the UI was laggy and then looked into system monitor and saw roughly those numbers when the animation was running. When switching screen the cpu usage went back to a very low value.

PR #1261, opened 2023-10-12
Add wallet guide

HenrikJannsen:


counterweightoperator:
@alvasw We are going to make some final adjustments to the content before merging as well.

counterweightoperator:
Adjustments made, unless @HenrikJannsen wants to change something, I think we are good for merging.

PR #1264, opened 2023-10-13
Fix padding when notification is visible

axpoems:
Before, the notification padding was not being updated properly while navigating through the different pages while the notification was visible. Proper padding handling for the special case of the tabView is achieved by changing the responsibility of handling this to tabView instead of the notification panel.

After
fix-notification-after

Before
fix-notification-before


HenrikJannsen (changes requested):
Here is my suggestion to avoid singletons: https://github.com/HenrikJannsen/bisq2/tree/fix-notification-padding_suggested_changes Feel free to pick out from it what you like. I don't care about authorship, I just created a branch as it was a more complex change and I needed to try out myself to see if that approach is feasible.

axpoems:
Thanks for the suggestion. I have reviewed your commits and code and like your solution.
I had also considered both options: passing the `NotificationPaneController` or using the `NotificationService` to achieve this.
Getting inspiration from what you mention about persisting the close state and looking at your code, I have picked up the branch I was working prior to this and improved it - so this is my new proposal: https://github.com/axpoems/bisq2/tree/bugfix-notification-panel

HenrikJannsen:
https://github.com/axpoems/bisq2/tree/bugfix-notification-panel looks good to me.

I am not so 100% sure if its best to let the TabController/View know so much about the environment. To let the parent container view handle it would sound more correct to me. But I am fine with that as well. We don't need to over-think it either....

I am also not so sure if the previous version with having a gap between tab line and scroll area was better. But I leave it up to you. Not too strong onion here. Both version do not feel perfect. Maybe we come up later with an alternative solution...

HenrikJannsen:
After another thought I think doing the handling inside tabview/controller does not work as we use tabviews inside popups as well. I think it should be handled in the container (contentview)

axpoems:
> After another thought I think doing the handling inside tabview/controller does not work as we use tabviews inside popups as well. I think it should be handled in the container (contentview)

I agree. The way the code is setup it makes sense that notification handling is done from parent of tabview. I have updated the code with your proposal.

> I am also not so sure if the previous version with having a gap between tab line and scroll area was better. But I leave it up to you. Not too strong onion here. Both version do not feel perfect. Maybe we come up later with an alternative solution...

Agree as well that non of these solutions look perfect. I have reverted it to previous version. Will look more into this. I think the problem here might be that background color is the same for tabview and content.

HenrikJannsen (approved):
ACK

PR #1281, opened 2023-10-19
Improve key event handling in popups

djing-chan:
Remove KeyHandlerUtil.handleDevModeKeyEvent method.
Add useEscapeKeyHandler and enterKeyHandler fields to OverlayController. 
Apply key handler to owner scene and overlay scene. 
Remove key handlers in custom popups and/or use the new fields in OverlayController. 
Set focusTraversable to 'false' to avoid (invisible) focus change by tabbing. 

We do not support tabbed controls navigation as focus stye is disabled at the moment. If we want to support that in future this might get changed again.


HenrikJannsen (approved):
ACK

PR #1292, opened 2023-10-22
Fix reference to unexisting image path

axpoems:
On Windows, a null pointer exception is thrown due to missing image path.
Set reference to the previous used image (128x128).


HenrikJannsen (approved):
utACK

PR #1297, opened 2023-10-23
Fix bug with not clearing the chat messages in the model.

djing-chan:
We used the binding previously which internally clear the model collection at registering the binding. Now we use addObserver which does not have that implicit behavior, thus the clearing of the messages was missing.


HenrikJannsen (approved):
ACK

Issue #1266, opened 2023-10-15
Add notification badge to trade item in open trades list view


HenrikJannsen:
If there are multiple trades it is not clear which trade triggered the notification. We should add the badge to the trade row.
<img width="1431" alt="Screenshot 2023-10-15 at 17 45 56" src="https://github.com/bisq-network/bisq2/assets/116298498/a2cc065c-fe06-4a9c-8d9b-6d21713ac8bc">

djing-chan:
I am working on that.
![Screenshot](https://github.com/bisq-network/bisq2/assets/146422224/fdd11ac5-2ab4-4d7c-8686-b12f53d5551c)

PR #1301, opened 2023-10-24
Add notification badge to trade items

djing-chan:
Implements https://github.com/bisq-network/bisq2/issues/1266


HenrikJannsen (approved):
ACK

PR #1302, opened 2023-10-25
Clean up and update the Bisq2 grey palette

axpoems:
An analysis was carried out to examine the existing usage of grey colours in Bisq2.
This analysis led to the development of an updated grey palette.

This was the state of greys:
before

The updated Bisq grey palette is based on the current grey distribution and achieves a coherent and succinct set through the whole spectrum. The following was taken into account:
- Unify same or very similar existing tones
- Preserve the current design
- Establish a base to further build and structure the CSS upon
- Attain symmetry in dark and light tones

A summary of the results obtained:
mapping

The palette contains two layers of depthness, and are named accordingly to be easily understandable.
* Five colours tones:
  * Black, dark-grey, grey, light-grey and white.
* Three additional shades:
  * Dim, medium and bright. 

Final palette:
after

Towards #1211


HenrikJannsen:
Great thanks for work on that!
Maybe if we swap the grey-light to light-grey it would not sound that weird when the bright postfix gets added?

HenrikJannsen (approved):
utACK

axpoems:
> Great thanks for work on that! Maybe if we swap the grey-light to light-grey it would not sound that weird when the bright postfix gets added? Same for dark...

I agree with you; it does indeed sound better. I opted for this approach to maintain some hierarchy in the nomenclature and easiness when searching. It is quite difficult to find a straightforward solution without some form of compromise. Nevertheless building upon your suggestion, I also considered that employing an abbreviation for the suffix might be effective and give more clarity:

```
-dark-grey-dim
-dark-grey-mid
-dark-grey-brt

-medium-grey-dim
-medium-grey-mid
-medium-grey-brt

-light-grey-dim
-light-grey-mid
-light-grey-brt
```

otherwise:

```
-dark-grey-dimmed
-dark-grey-medium
-dark-grey-bright

-grey-dimmed
-grey-medium
-grey-bright

-light-grey-dimmed
-light-grey-medium
-light-grey-bright
```

@HenrikJannsen - What do you think?

HenrikJannsen:
> I opted for this approach to maintain some hierarchy in the nomenclature and easiness when searching. It is quite difficult to find a straightforward solution without some form of compromise.  

Yes I agree and also saw that as benefit. I have no strong opinion. leave it to you what you find best.
maybe low/mid/hi might be an alternative. dim/mid sounds good but btr might be less clear...

axpoems:
Updated naming map:
mapping2

HenrikJannsen (approved):
utACK

djing-chan (changes requested):
NACK some grey tones have been changed to the current version

djing-chan:
Some grey tones are not the same as before.
Left is PR version, right is main branch version.

Screenshot from 2023-10-27 14-16-49
Screenshot from 2023-10-27 14-17-25
Screenshot from 2023-10-27 14-17-57

axpoems:
> Some grey tones are not the same as before.

Of course they are not all the same. There were 29 shades of grey, now there are 15, if you can count. Approximations and simplifications had to be made, preserving as much as possible the initial design. These are building blocks that should have been defined in the past, so that's why now we're cleaning-up and adjusting.
To those greys you are pointing out, some will just simple cease to exist very soon. As for the header of the program, I only tested it on Windows, which has it in white (no change in this case). Since you are using another OS, you're welcomed to update those you can see and I can't, using the updated pallet.
I've tested it and it looks good and cleaner. Nonetheless, some changes can and will be made in future UI reviews.

djing-chan:
Sorry, I fear my NACK and short comments came over unfriendly. It was not meant at all that way. I highly appreciate your work on that, just wanted to point out changes which might have been unintended (seems not).
If the changes have been intended, please ignore my NACK.

axpoems:
> Sorry, I fear my NACK and short comments came over unfriendly. It was not meant at all that way. I highly appreciate your work on that, just wanted to point out changes which might have been unintended (seems not).
> If the changes have been intended, please ignore my NACK.

Certainly, there are no hard feelings about that. As I mentioned before, I expected some changes due to the reasons I explained earlier. This is part of an initial stage of cleanup, unification, and adjustment process for the UI. I appreciate your clarification and always welcome constructive feedback.

PR #1303, opened 2023-10-25
Adding effect to TabView

aspire2dev:
Added the transitLeftOut and transitRightOut animations to the TabView class. Also re-factored some code that got triggered on checkstyle evaluations.

bisq-network#1216


HenrikJannsen (approved):
utACK

PR #1307, opened 2023-10-27
Avoid -1 value at progress bar at startup

djing-chan:
Set progress at bootstrap start to a value of 1-24% instead of keeping -1 which would trigger the progress bar animation which is very CPU intense.


HenrikJannsen (approved):
ACK

Issue #948, opened 2023-06-27
Check performance of chat with massive amount of messages


HenrikJannsen:

namloan:
What's the success criteria? Is this about bursts of messages in a short period of time or about whether a long history of messages can degrade loading times? In a single chat window or adding up the messages from different chats?

namloan:
> But probably best to set up a target scenario (e.g. 100 offers, 500 text messages) and then use a profiler to see where the bottlenecks are.

I'll do this

HenrikJannsen:
I have seen that again a while back when I have many private messages in a trade that the UI got quite slow. So there is definitely an issue.

namloan:
@HenrikJannsen what things have you noticed that go slower? With 100 offers and 1000+ messages there's a lag when clicking on 'offerbook' but once loaded it seems to work fine. I may need to give it a try on a VM with lower specs.

Early results from the profiling show that the heap keeps increasing every time the offerbook is loaded even if there are no new messages.

namloan:
Heap going from 56 MB to 1 GB in ~5 mins:

https://github.com/bisq-network/bisq2/assets/138292162/a0b46be7-afd7-44e2-8922-d783b0d02893

HenrikJannsen:
Seems there are some memory leaks. I remember to have seen in the item renderers some missing listener removals or other cleanups.
Which tool do you use?

namloan:
I still need to analyze the profiling results, but looking at missing listener removals is a good hint.

> Which tool do you use?

[JDK Mission Control](https://www.oracle.com/java/technologies/jdk-mission-control.html)

namloan:
From the automated analysis results:

The live set on the heap seems to increase with a speed of about 4.27 MiB per second during the recording.
An analysis of the reference tree found 2 leak candidates. The main candidate is bisq.desktop.main.content.bisq_easy.open_trades.BisqEasyOpenTradesController.view Referenced by this chain:
```
bisq.desktop.main.content.bisq_easy.open_trades.BisqEasyOpenTradesController$$Lambda$1721+0x0000000801701ab8.1781279162.arg$1
bisq.common.observable.collection.CollectionChangeHandler.listener
java.lang.Object[]
java.util.concurrent.CopyOnWriteArrayList.array
bisq.common.observable.collection.ObservableSet.observers
bisq.trade.bisq_easy.BisqEasyTradeStore.trades
[...]

PR #1310, opened 2023-10-28
Fix chat performance issues

namloan:
Metrics based on a 1 minute session scrolling and switching back and forth an offerbook channel with a large amount of messages (similar test as [here](https://github.com/bisq-network/bisq2/issues/948#issuecomment-1753631855)).

Now:
performancenew

Before:
performanceold

#948

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/content/components/ChatMessagesListView.java
+++ b/desktop/src/main/java/bisq/desktop/main/content/components/ChatMessagesListView.java
@@ -459,7 +459,7 @@ public class ChatMessagesListView {
             return channel.getChatMessages().addObserver(new CollectionObserver<>() {
                 @Override
                 public void add(M chatMessage) {
-                    UIThread.run(() -> {
+                    UIThread.runOnNextRenderFrame(() -> {
                         model.chatMessages.add(new ChatMessageListItem<>(chatMessage, userProfileService, reputationService,
                                 bisqEasyTradeService, userIdentityService, networkService));
                     });
@@ -468,7 +468,7 @@ public class ChatMessagesListView {
                 @Override
                 public void remove(Object element) {
                     if (element instanceof ChatMessage) {
-                        UIThread.run(() -> {
+                        UIThread.runOnNextRenderFrame(() -> {
                             ChatMessage chatMessage = (ChatMessage) element;
                             Optional<ChatMessageListItem<? extends ChatMessage>> toRemove = model.chatMessages.stream()
                                     .filter(item -> item.getChatMessage().getId().equals(chatMessage.getId()))
@@ -483,7 +483,7 @@ public class ChatMessagesListView {
 
                 @Override
                 public void clear() {
-                    UIThread.run(() -> {
+                    UIThread.runOnNextRenderFrame(() -> {
                         model.chatMessages.forEach(ChatMessageListItem::dispose);
                         model.chatMessages.clear();
                     });


HenrikJannsen:
I want to analyze if further to figure out why the frame delay makes that difference.

HenrikJannsen:
Did you find out the reason why the `run` to `runOnNextRenderFrame` makes such a difference?

namloan:
> As the run is called from a non UI thread it will run anyway on the next render frame

This isn't what I'm observing - it's always being called from the application thread. Therefore, the application thread immediately executes each ```model.chatMessages.add```, which in turn triggers the expensive ```updateItem()```, all while the rest of the UI is waiting as everything happens within the ```onActivate()```.

I set up a breakpoint  on ```model.chatMessages.add``` and took a thread dump - see the ```UIThread.java:25```:

```
"JavaFX Application Thread@4231" prio=5 tid=0x1f nid=NA runnable
  java.lang.Thread.State: RUNNABLE
	  at bisq.desktop.main.content.components.ChatMessagesListView$Controller$1.lambda$add$0(ChatMessagesListView.java:463)
	  at bisq.desktop.main.content.components.ChatMessagesListView$Controller$1$$Lambda$1798/0x00000008016b04d8.run(Unknown Source:-1)
	  at bisq.desktop.common.threading.UIThread.run(UIThread.java:25)
	  at bisq.desktop.main.content.components.ChatMessagesListView$Controller$1.add(ChatMessagesListView.java:462)
	  at bisq.desktop.main.content.components.ChatMessagesListView$Controller$1.add(ChatMessagesListView.java:459)
	  at bisq.common.observable.collection.CollectionObserver$$Lambda$768/0x000000080146e9a8.accept(Unknown Source:-1)
	  at java.util.concurrent.CopyOnWriteArrayList.forEach(CopyOnWriteArrayList.java:807)
	  at java.util.concurrent.CopyOnWriteArraySet.forEach(CopyOnWriteArraySet.java:425)
	  at bisq.common.observable.collection.CollectionObserver.addAll(CollectionObserver.java:26)
	  at bisq.common.observable.collection.ObservableCollection.addObserver(ObservableCollection.java:47)
	  at bisq.desktop.main.content.components.ChatMessagesListView$Controller.bindChatMessages(ChatMessagesListView.java:459)
	  at bisq.desktop.main.content.components.ChatMessagesListView$Controller.lambda$onActivate$3(ChatMessagesListView.java:201)
	  at bisq.desktop.main.content.components.ChatMessagesListView$Controller$$Lambda$1239/0x00000008015ccfe8.run(Unknown Source:-1)
	  at bisq.desktop.common.threading.UIThread.run(UIThread.java:25)
	  at bisq.desktop.main.content.components.ChatMessagesListView$Controller.lambda$onActivate$4(ChatMessagesListView.java:188)
	  at bisq.desktop.main.content.components.ChatMessagesListView$Controller$$Lambda$1238/0x00000008015ccdb0.accept(Unknown Source:-1)
	  at bisq.common.observable.Observable.addObserver(Observable.java:40)
	  at bisq.desktop.main.content.components.ChatMessagesListView$Controller.onActivate(ChatMessagesListView.java:187)
```

HenrikJannsen:
That's when the messages get added due the observer registration which happens at onActive on the UI thread. 
When a message comes from the network its a network thread.
Screenshot 2023-10-31 at 17 53 57


But I see that in the use case where it gets called from the UI thread we are calling the runnable directly and by delaying it a render frame we decrease load.
Let me think a bit more about it as that situation will occur elsewhere as well and it should be handled in the observer framework.
I am not much online until next week, so allow me a bit of time to get back to that.

HenrikJannsen:
I saw that the `addAll` method in `CollectionChangeMapper` is not using `addAll` internally but iterates over the list with `add` calls.

Can you run your performance test with that patch applied?

```
Index: common/src/main/java/bisq/common/observable/collection/CollectionChangeMapper.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/common/src/main/java/bisq/common/observable/collection/CollectionChangeMapper.java b/common/src/main/java/bisq/common/observable/collection/CollectionChangeMapper.java
--- a/common/src/main/java/bisq/common/observable/collection/CollectionChangeMapper.java	(revision 293d13d07aacbc582dab615e65f86026c3705c53)
+++ b/common/src/main/java/bisq/common/observable/collection/CollectionChangeMapper.java	(date 1698751558152)
@@ -21,6 +21,7 @@
 import lombok.ToString;
 
 import java.util.Collection;
+import java.util.List;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -62,22 +63,23 @@
 
     @Override
     public void addAll(Collection<? extends S> values) {
-        executor.accept(() -> values.forEach(element -> {
-            T item = mapFunction.apply(element);
-            if (!targetCollection.contains(item)) {
-                targetCollection.add(item);
-            }
-        }));
+        executor.accept(() -> {
+            List<T> toAdd = values.stream()
+                    .map(mapFunction)
+                    .filter(item -> !targetCollection.contains(item))
+                    .collect(Collectors.toList());
+            targetCollection.addAll(toAdd);
+        });
     }
 
     @Override
     public void setAll(Collection<? extends S> values) {
         executor.accept(() -> {
             targetCollection.clear();
-            values.forEach(element -> {
-                T item = mapFunction.apply(element);
-                targetCollection.add(item);
-            });
+            List<T> toAdd = values.stream()
+                    .map(mapFunction)
+                    .collect(Collectors.toList());
+            targetCollection.addAll(toAdd);
         });
     }
 
```

The FX collection should behave much better if the items are not added individually but in a batch.

HenrikJannsen (approved):
utACK

HenrikJannsen:
I think to delay here one render frame when it comes from activate is a good idea anyway to reduce a bit of load at screen activation. We could do that at all screens where heavier listviews get filled at activate.

PR #1311, opened 2023-10-29
Fix UI bugs in Bisq Easy

mrblue313:
* **Quote price not updated when market changes.**

Before

After

* **Base price not updated when seller as a taker changes the trade price.**

Before

After


HenrikJannsen (approved):
utACK

PR #1313, opened 2023-10-31
Fixing truncating button text issue

aspire2dev:
Created a new css element to pad the buttons that are truncating differently from the other buttons without changing their aspect when in a larger size view and some other adjustments

bisq-network#1246

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/content/user/bonded_roles/nodes/tabs/registration/NodeRegistrationView.java
+++ b/desktop/src/main/java/bisq/desktop/main/content/user/bonded_roles/nodes/tabs/registration/NodeRegistrationView.java
@@ -44,7 +44,7 @@ public class NodeRegistrationView extends BondedRolesRegistrationView<NodeRegist
 
         importNodeAddressButton = new Button(Res.get("user.bondedRoles.registration.node.importAddress"));
         importNodeAddressButton.getStyleClass().addAll("outlined-button", "button-reduced-padding");
-        importNodeAddressButton.setPrefWidth(200);
+        importNodeAddressButton.setPrefWidth(180);
 
         showKeyPairButton = new Button(Res.get("user.bondedRoles.registration.node.showKeyPair"));
         showKeyPairButton.getStyleClass().add("button-reduced-padding");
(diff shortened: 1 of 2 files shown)


desktop/src/main/java/bisq/desktop/main/content/user/bonded_roles/nodes/tabs/registration/NodeRegistrationView.java
```diff
@@ -44,8 +44,12 @@ public NodeRegistrationView(NodeRegistrationModel model,
 
         importNodeAddressButton = new Button(Res.get("user.bondedRoles.registration.node.importAddress"));
         importNodeAddressButton.getStyleClass().add("outlined-button");
+        importNodeAddressButton.getStyleClass().add("button-reduced-padding");
```
HenrikJannsen: Better use `addAll`

HenrikJannsen (changes requested):
NACK

HenrikJannsen:
Left button border gets swallowed in min width layout.
Hyperlink should not have line break IMO.

Screenshot 2023-10-31 at 21 02 12

aspire2dev:
> Left button border gets swallowed in min width layout. Hyperlink should not have line break IMO.
> Screenshot 2023-10-31 at 21 02 12

@HenrikJannsen, this issue is not a regression, as it happens already on main (see image bellow):
image

Should it be worked on this issue or can I open another one to review all of the other places where this may be occurring?

HenrikJannsen:
I would reduce padding and accept ellipsis in "Learn more".
Border issue i only have seen in that screen. I think a 1 or 2 px left padding should fix it.

desktop/src/main/java/bisq/desktop/main/content/user/bonded_roles/nodes/tabs/registration/NodeRegistrationView.java
aspire2dev: Done

aspire2dev:
> I would reduce padding and accept ellipsis in "Learn more". Border issue i only have seen in that screen. I think a 1 or 2 px left padding should fix it.

Done

HenrikJannsen (approved):
utACK

PR #1320, opened 2023-11-03
Aspire2dev/fix popups user options

aspire2dev:
The popup windows from the user section (reputation, all buttons) are actually breaking on a too narrow screen.
Merged the Burn BSQ score simulation into the BurnBsqTab2View and moved the controller methods to the BurnBsqTab2Controller. Then added a GridPane to distribute equally the available space on the view independent of the size.

Tried making the popup resize when the underlying window is changed in size, but even binding the prefWidthView that's not working.

Popup size is calculated during runtime depending on size of the window.


aspire2dev:
@HenrikJannsen, this is a first shot at the changes on the popups from the Reputation view of Users. Please let me know if I can proceed with the same fixes for the score calculation of the BSQ Bonds, Account Age and Signed account age witness.

HenrikJannsen (approved):
NACK

HenrikJannsen:
Several layout bugs (did not review in depth) but those have been obvious at a 5 second check. Please test better yourself in future, it takes too much effort for review when it goes often back and forth and there are obvious and easy to spot issues! Some issues might have been before already but if your task is to fix that area it included to fix all past bugs.

Screenshot 2023-11-05 at 13 29 13
Screenshot 2023-11-05 at 13 29 24
Screenshot 2023-11-05 at 13 29 52

aspire2dev:
> 
> Wrong top padding at sub headline: Screenshot 2023-11-05 at 13 29 13
>
Sorry, the intention here was to demonstrate the POC regarding the utilization of the GridPane and the changes to the code. Next time I'll make sure to also properly format the screen for POCs as well.
>
> Too much space between text and input controls: Screenshot 2023-11-05 at 13 29 24
> 
I'll remove the space, but I didn't change anything on this view.
>
> Space at linebreak: Screenshot 2023-11-05 at 13 29 52
>
Do you mean the space before the text " Enter the amount of BSQ ..."?

HenrikJannsen:
> Do you mean the space before the text " Enter the amount of BSQ ..."?

Yes.

HenrikJannsen:
If its POC it should not be a PR but a draft.

aspire2dev:
When switching the implementation from Label to Text + Text Flow there are slight differences in the way the text is rendered. As an example, bellow is a screenshot taken from main:

And bellow a screenshot taken replacing Label by Text + TextFlow (instead of using wrap on the Label, I exchanged the implementation to TextFlow)
image

Notice how there's a difference in the position of the next button, this is due to the text from the "What is the recommended amount to burn" text section. Is that acceptable or it should be pixel perfect to what it was before?

aspire2dev:
> > Do you mean the space before the text " Enter the amount of BSQ ..."?
> 
> Yes.
Here I added a #4 item:
image

Here I just adjusted the spaces:
image

Which one should I use?

aspire2dev:
Here a video of the actual implementation of the Burning BSQ view.. Notice that when the window is narrower, both WHY and SCORE tabs get some scrolling. Should I reduce the spacing here (which will affect looks elsewhere) or keep it as is?


https://github.com/bisq-network/bisq2/assets/143408636/84cf03eb-9175-457d-b0dd-a426e14532c2

HenrikJannsen:
> Notice how there's a difference in the position of the next button, this is due to the text from the "What is the recommended amount to burn" text section. Is that acceptable or it should be pixel perfect to what it was before?

I cannot see any difference. So looks good to me.

HenrikJannsen:
> Which one should I use?

The one with 6 items.

HenrikJannsen:
> Here a video of the actual implementation of the Burning BSQ view.. Notice that when the window is narrower, both WHY and SCORE tabs get some scrolling. Should I reduce the spacing here (which will affect looks elsewhere) or keep it as is? If possible, reducing the blank space on left and right would be the best solution. Other views also scroll when in a narrower view, and in some cases, hide the continue button.
> view_gridpane_resize-2023-11-07_08.38.43.mp4

Yes the left and right space should be reduced to maybe 20 px in small width case.

aspire2dev:
> > Here a video of the actual implementation of the Burning BSQ view.. Notice that when the window is narrower, both WHY and SCORE tabs get some scrolling. Should I reduce the spacing here (which will affect looks elsewhere) or keep it as is? If possible, reducing the blank space on left and right would be the best solution. Other views also scroll when in a narrower view, and in some cases, hide the continue button.
> > view_gridpane_resize-2023-11-07_08.38.43.mp4
> 
> Yes the left and right space should be reduced to maybe 20 px in small width case.

On my submitted solution I did test all narrower sizes I could and did reduce as much as needed to get a comfortable view. If I only left 20 pixels it would not look very nice. Let me know if it should be only 20 px.

Just a reminder, for the view size to change according to the window size, the view has to be opened after the window resizing. For some reason I was not able to figure out, binding is not working to this views.

HenrikJannsen (changes requested):
NACK

HenrikJannsen:
Scrollbar not displayed at that window height. It gets displayed when smaller height.
Please squash all the commits with the same commit message.

Screenshot 2023-11-10 at 08 01 48

aspire2dev:
> Scrollbar not displayed at that window height. It gets displayed when smaller height. Please squash all the commits with the same commit message.
> Screenshot 2023-11-10 at 08 01 48

I tested the possible dimensions and space left over for the width and the height. Then I took the minimum margins that would not trigger scrolling on the view. This was then applied whenever the screen is narrower than either 1100 width or 880 height. So, this way, we don' t have to worry about scrolling in these views.
Also squashed the commits.

HenrikJannsen (approved):
ACK

desktop/src/main/java/bisq/desktop/main/content/user/reputation/accountAge/AccountAgeView.java
```diff
@@ -38,12 +40,16 @@ public class AccountAgeView extends TabView<AccountAgeModel, AccountAgeControlle
...
-        root.setMaxHeight(height);
+        Window window = Stage.getWindows().stream().filter(Window::isShowing)
+                .findFirst().orElse(null);
+        double prefWidth = DesktopModel.PREF_WIDTH - DesktopModel.MIN_WIDTH - 50;
+        double prefHeight = DesktopModel.PREF_HEIGHT - DesktopModel.MIN_HEIGHT + 40;
+        if (window.getWidth() <= 1100 || window.getHeight() <= 880) {
+            prefWidth = 105;
+            prefHeight = 70;
+        }
+        root.prefWidthProperty().bind(window.widthProperty().subtract(prefWidth));
```
HenrikJannsen, line 51: Binding have to be in onViewAttached and unbind called in onViewDetached

aspire2dev:
As my PRs are not up to the standard required and take too much time to review, I'm dropping off the project. Sorry for all the inconvenience...

PR #1321, opened 2023-11-03
[ChatMessagesListView] Improve readability

namloan:
#948

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/content/components/ChatMessagesListView.java
+++ b/desktop/src/main/java/bisq/desktop/main/content/components/ChatMessagesListView.java
@@ -741,60 +741,7 @@ public class ChatMessagesListView {
                             if (isMyMessage) {
                                 buildMyMessage(isBisqEasyPublicChatMessageWithOffer, userProfileIconVbox, chatMessage);
                             } else {
-                                // Peer
-                                HBox userNameAndDateHBox = new HBox(10, userName, dateTime);
-                                message.setAlignment(Pos.CENTER_LEFT);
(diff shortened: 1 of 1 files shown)


desktop/src/main/java/bisq/desktop/main/content/components/ChatMessagesListView.java
```diff
@@ -704,227 +704,232 @@ private void hideReactionsBox() {
...
-                                if (flow != null && !flow.isVisible())
-                                    return;
+                            subscriptions.clear();
+                            ChatMessage chatMessage = item.getChatMessage();
 
-                                boolean hasTradeChatOffer = model.hasTradeChatOffer(chatMessage);
-                                boolean isBisqEasyPublicChatMessageWithOffer = chatMessage instanceof BisqEasyOfferbookMessage && hasTradeChatOffer;
-                                boolean isMyMessage = model.isMyMessage(chatMessage);
+                            Node flow = this.getListView().lookup(".virtual-flow");
+                            if (flow != null && !flow.isVisible())
```
HenrikJannsen: Single line returns have caused bugs in the past in Bisq1, so we prefer to use always curly brackets. But not super strong opinion about it, so leave it to you.

HenrikJannsen (approved):
ACK

desktop/src/main/java/bisq/desktop/main/content/components/ChatMessagesListView.java
namloan: I agree! will update

Issue #1255, opened 2023-10-07
Decide on the green colore tone used in Bisq 2


HenrikJannsen:
The green color used in Bisq 1 is quite intense on Linux (I assume Windows use the same color profile). On OSX it looks ok.

In Bisq 1 those are the base colors:
-bs-color-primary: #25b135; (left menu, progr. indicator,...)
-bs-color-primary-dark: #2ea33c; (right menu, action button,...)

In Bisq 2 the base color was changed to #56ae48. The color is a good compromise which looks ok both on Linux and Mac.
For large buttons we need to adjust to a darker/less intense variation. A `derive(#56ae48, -15%)` worked for me on both OSX and Linux.

See also: 
https://github.com/bisq-network/bisq-website/issues/478

HenrikJannsen:
@axpoems 
From your last PR it seems that we stick with the Bisq1 green color for the logo (which makes sense).
The green tone in the app can then be different depending on the usage.
For small areas where its good to have a strong signal effect I would suggest to stick with the bisq1 green.
For default state of standard buttons i would suggest the bisq2 green from Pedro. 
If the button green is still too intense for large buttons we can adjust that further (IMO it is fine - one has to compare with all OS, they look quite different and those color suggestions look like a good compromise to me).

Suggestion:
```
-bisq-green: #25b135;
-fx-default-button: derive(#56ae48, -15%);
-bisq-green-hover: derive(-fx-default-button, 10%);
-bisq-green-pressed: derive(-fx-default-button, -5%);
```

If those suggestions dont work well for you can you offer your suggestions?

Bisq 1 green: #25b135 
Bisq 2 green: #56ae48

axpoems:
@HenrikJannsen - As you mentioned, I believe it's a good idea to keep the logo colour consistent at `#25b135`, while using another shade of green with some variations for the rest of the application. The green already implemented, `#56ae48`, seems like a good choice.

So far, I've tested it on Windows, and it appears to work well. I'm planning to test it on Linux as well to ensure uniformity.

Regarding the variations, I'd like to propose a slightly different approach. Instead of using dynamic derived variations, which can sometimes be less intuitive, I suggest choosing two or three specific shades of green from the palette. These selections could be related to your initial proposal, like this:

```
-fx-default-button: #3c942e;  /* (-10% from #56ae48) / derive(#56ae48, -15%) */
[...]

PR #1324, opened 2023-11-04
Introduce bisq2 green palette

axpoems:
As discussed in issue #1255, introducing a static set of green shades and renaming the existing greens accordingly.

&#12288;
56ae48_pallet2_map
&#12288;

Towards #1211


HenrikJannsen (approved):
utACK

PR #1332, opened 2023-11-07
Unify tab transitions with the other transitions

### LINKED ISSUE #1328: Harmonize tab transitions to other transitions
The new transitions added to tab view changes is not the same style as the transition used at normal navigation changes. Also the new left/right change is confusing (did a usability test and the candidate was confused why it comes sometimes left and sometimes right, he thought its a bug).

axpoems:
content_slide3

Resolve #1328

### DIFF
--- a/desktop/src/main/java/bisq/desktop/common/Transitions.java
+++ b/desktop/src/main/java/bisq/desktop/common/Transitions.java
@@ -402,10 +402,10 @@ public class Transitions {
         }
     }
 
-    public static void transitRightOut(Region nodeIn, Region nodeOut, int customDuration) {
+    public static void transitRightOut(Region nodeIn, Region nodeOut) {
         nodeIn.setOpacity(0);
         UIScheduler.run(() -> slideInLeft(nodeIn, () -> {
-        })).after(customDuration);
+        })).after(DEFAULT_DURATION / 4);
         slideOutRight(nodeOut, () -> {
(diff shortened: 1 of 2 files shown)


HenrikJannsen (approved):
ACK

The transition used for the content views (`Transitions.transitContentViews`) has `DEFAULT_DURATION/2` duration. I think we should adjust that as well to have both the same. The `DEFAULT_DURATION/4` seems better to me.

axpoems:
Added the change. @HenrikJannsen were you referring to `fadeIn` transition?

HenrikJannsen (approved):
utACK

HenrikJannsen:
> Added the change. @HenrikJannsen were you referring to `fadeIn` transition?

Both the fade and the side animation.

axpoems:
> Both the fade and the side animation.

Added it for the slide out transition as well. Can you confirm those are the ones you were referring?

HenrikJannsen (approved):
utACK

djing-chan (approved):
ACK

The trade wizard/ create offer popups use a different transition style (slower, left/right). The other popups with tabs (reputation) use the also the fast right out transition. I think we should harmonize all those transitions (not as part of that PR) and use the same methods in the Transitions class. I find the left/right transition when using back/next buttons more intuitive. Not sure if that is also the best option for tab views without next/back buttons.

PR #1333, opened 2023-11-07
Update the waiting state icons

axpoems:
* Change the colour to Bisq2 green
* Improve some of the icon's design
* Fix cut pixels in spinning circle

Towards #1211


desktop/src/main/java/bisq/desktop/main/content/bisq_easy/components/WaitingAnimation.java
```diff
@@ -47,7 +47,11 @@ public WaitingAnimation(WaitingState waitingState) {
     public WaitingAnimation() {
         setAlignment(Pos.CENTER);
         ImageView spinningCircle = ImageUtil.getImageViewById("spinning-circle");
-        spinningCircle.setOpacity(0);
+
+        spinningCircle.setFitHeight(78);
```
HenrikJannsen, line 51: The image has 67 px, right? Wouldn't it be better to provide the image itself in the size we want to use?

HenrikJannsen (approved):
utACK

desktop/src/main/java/bisq/desktop/main/content/bisq_easy/components/WaitingAnimation.java
axpoems: This is the legacy size. I can re-do it to the display size.

HenrikJannsen: Yes I think would be better to have the image in the correct size so no codewise scaling is needed.

axpoems: Looking closer into it, there are actually two sizes of images: the smaller 67px and the @2x 134px.  The image being picked up is the one with 134px. The new icons were design to fit inside this circle size.
The reason I added `setFitHeight` is to prevent an existing cropping which looks bad. I tried redoing the images to that size, but the issue with the cropping still persists. I have spent some time and I wasn't able to find the root cause. I can follow-up with a PR and investigate more.

HenrikJannsen: Ah ok. Yes not worth to spent too much time on it. Sometimes a 1 px extra transparent border helps from my experience. Graphic rendering is not the strenght of JavaFx.

djing-chan (approved):
ACK

PR #1335, opened 2023-11-08
[User Profile view] Improve input validation

namloan:
https://github.com/bisq-network/bisq2/issues/1153

user_profile_validation

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
+++ b/desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
@@ -132,8 +132,6 @@ public class UserProfileView extends View<HBox, UserProfileModel, UserProfileCon
         statement.textProperty().bindBidirectional(model.getStatement());
         terms.textProperty().bindBidirectional(model.getTerms());
         roboIconImageView.imageProperty().bind(model.getRoboHash());
-        saveButton.disableProperty().bind(model.getSaveButtonDisabled());
-        deleteButton.disableProperty().bind(model.getDeleteButtonDisabled());
 
         useDeleteTooltipPin = EasyBind.subscribe(model.getUseDeleteTooltip(), useDeleteTooltip ->
                 deleteWrapper.setTooltip(useDeleteTooltip ? deleteTooltip : null));
--- a/desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
+++ b/desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
@@ -27,6 +27,7 @@ import bisq.desktop.components.controls.MaterialTextArea;
 import bisq.desktop.components.controls.MaterialTextField;
 import bisq.desktop.components.controls.validator.TextMaxLengthValidator;
 import bisq.desktop.components.controls.validator.ValidatorBase;
+import bisq.desktop.components.overlay.Popup;
 import bisq.i18n.Res;
 import bisq.user.identity.UserIdentity;
 import javafx.geometry.Insets;
@@ -158,11 +159,7 @@ public class UserProfileView extends View<HBox, UserProfileModel, UserProfileCon
             }
         });
 
-        deleteButton.setOnAction(e -> {
-            if (comboBox.validate()) {
-                controller.onDeleteProfile();
-            }
-        });
+        deleteButton.setOnAction(e -> onDeleteButtonPressed());
         saveButton.setOnAction(e -> onSaveButtonPressed());
         createNewProfileButton.setOnAction(e -> controller.onAddNewChatUser());
         comboBox.setOnChangeConfirmed(e -> {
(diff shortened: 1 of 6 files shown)


desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
```diff
@@ -165,6 +174,19 @@ protected void onViewAttached() {
                 userIdentity -> comboBox.getSelectionModel().select(userIdentity));
     }
 
+    private void onSaveButtonPressed() {
+        if (runOnSaveValidations()) {
+            controller.onSave();
+        }
+    }
+
+    private boolean runOnSaveValidations() {
```
HenrikJannsen, line 215: Maybe isValid() would be a better method name?

desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
```diff
@@ -172,6 +173,18 @@ protected void onViewAttached() {
...
         selectedChatUserIdentityPin = EasyBind.subscribe(model.getSelectedUserIdentity(),
                 userIdentity -> comboBox.getSelectionModel().select(userIdentity));
+
+        isValidSelectionPin = EasyBind.subscribe(comboBox.getIsValidSelection(), isValidSelection -> UIThread.run(() -> {
+            if (!isValidSelection) {
+                statement.setEditable(false);
+                terms.setEditable(false);
+                controller.resetSelection();
+            }
+            else {
```
HenrikJannsen: the `else` should be in the same line as the closing }

HenrikJannsen (approved):
utACK

Some Nits...

djing-chan:
When using a incorrect pw for remove password the error popup should not be a generic exceptions popup but a user friendly error string. Maybe we add an exception popup style for those generic error popups and use the error popup with handled error.


Screenshot from 2023-11-09 12-04-00

After using an incorrect pw and using the correc pw I got that error. After restart the pw worked. So there is a bug (can be fixed in another PR).
Screenshot from 2023-11-09 12-04-58

I think the VHox spacing could be even smaller as before (15 instead of 20, now 25). There is a rendering bug in the Component when the error message is getting displayed the height of the component does not update immediately but with some 500 ms delay.
Here with 15 px spacing and after delayed correct rendering:
Screenshot from 2023-11-09 12-17-49

desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
```diff
@@ -158,11 +159,7 @@ protected void onViewAttached() {
             }
         });
 
-        deleteButton.setOnAction(e -> {
-            if (comboBox.validate()) {
-                controller.onDeleteProfile();
-            }
-        });
+        deleteButton.setOnAction(e -> onDeleteButtonPressed());
```
djing-chan, line 162: I understand the reasons why to call the method on the View as the validators are in the view or inside the components. But we break the Bisq2 MVC concept that the View is not doing any logic (beside layout logic) and the controller is responsible for such. 
If we create and reference the validators in the controller and pass it to the view we could do that in the controller. I am aware that this is also not a super nice solution but I would consider it better. Validation is at the end a task for the controller.
But (if you agree) no need to do that change in that PR, could be a follow up PR.

djing-chan (approved):
ACK

See comments for potential improvements which could be impl. in a follow up PR.

desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
namloan: These are validations that are run only when saving (statement and terms length aren't validated on pressing on delete).

desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
namloan: ok will update

desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
namloan: You're bringing up a good point. I need to think more about this. I copied the validation framework from JFoenix as it was suggested in the requirements of https://github.com/bisq-network/bisq2/issues/883. So the validators are coupled to the UI components - not sure if creating them in the controller would be an improvement unless we rethink the whole validation framework. I don't think it's a big problem right now as it's well encapsulated and should be easy to replace by another solution in the future. I'd suggest we give it a thought after we complete the 2.0.0 milestone. Maybe create an issue for that?

desktop/src/main/java/bisq/desktop/main/content/user/user_profile/UserProfileView.java
namloan: actually I had already changed it in https://github.com/bisq-network/bisq2/pull/1335/commits/8d2d61b0709d87cef5049f3e033cd2f945a301bd

Issue #1316, opened 2023-11-01
Improve trade cancellation use case


HenrikJannsen:
If a trade gets cancelled we should display it very clearly to the user (at the moment it is only a "Peer left" chat message).
We also should block the trade protocol to proceed so that the non-cancelling peer does not continue with the trade.

There are different stages for cancelling a trade:
- Abandon/reject trade (before the seller has sent their payment data). Traders can abandon or reject a trade. Abandon is used for the taker and reject used for the maker.
- Cancel trade: after seller has sent payment data. It is a breach of the trade rule to cancel a trade without clear reason or in agreement with the peer.
- Leave trade: after the payment has been completed but the user might have not moved to the complete screen (in case the explorer lookup has failed the user need to manually confirm that trade is completed)

We can add new transitions to the protocol which handle the 3 different events.

Abandon trade (same as reject):
Source states are those before sellers payment data got exchanged.
Target state is a final state `ABANDONED`

Cancel trade:
Source states are those after seller sent payment data but before completed state.
Target state is the final state `CANCELLED`

Leave trade:
After btc confirmed and before completed.
Target state is the final state `PEER_LEFT`.

Not sure if it makes sense to model the leave trade case in the protocol.

In case of the cancel events we should show the state in the left protocol steps as alternative to completed and maybe in a different color (yellow?)

djing-chan:
I can work on that

PR #1336, opened 2023-11-08
Add cancel states to protocol

djing-chan:
Implements https://github.com/bisq-network/bisq2/issues/1316

Warning popup when reject trade is clicked:
Screenshot from 2023-11-08 20-18-22

Warning popup when cancel trade is clicked:
Screenshot from 2023-11-08 20-18-36

Rejected or cancelled trades. The non initiating peer at cancellation gets the report to mediator button shown. The mediator can decide then to report it to the moderator for banning the user due rule violation.

Screenshot from 2023-11-08 19-51-09
Screenshot from 2023-11-08 19-51-06
Screenshot from 2023-11-08 19-51-04

For mediation https://github.com/bisq-network/bisq2/issues/1337 is missing.

The trade state before cancelling the trade is lost as we do not keep the past state before the cancel state got activated.
The trade chat is a protocol for all the trade activities. Maybe we could consider to persist the past state before cancellation to use that for display in the UI (need to ensure the user cannot trigger any events from the old state).

### DIFF
--- a/desktop/src/main/java/bisq/desktop/main/notification/NotificationPanelView.java
+++ b/desktop/src/main/java/bisq/desktop/main/notification/NotificationPanelView.java
@@ -62,9 +62,8 @@ public class NotificationPanelView extends View<BorderPane, NotificationPanelMod
 
         Label separator = new Label("|");
         separator.getStyleClass().add("notification-headline");
-        HBox notificationContent = new HBox(notificationHeadline, separator, goToOpenTradesButton);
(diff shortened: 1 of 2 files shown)


djing-chan:
Screenshot from 2023-11-08 19-51-09
Screenshot from 2023-11-08 19-51-06
Screenshot from 2023-11-08 19-51-04

HenrikJannsen (approved):
ACK

PR #1338, opened 2023-11-08
Add amount validation for Bisq Easy

### LINKED ISSUE #1031: Add validation to BisqEasyTakeOfferRequestHandler
Min/max amounts are currently not checked as if both traders have different market prices the amounts might be different.
Add some tolerance and enable the check again.

mrblue313:
* Adds amount validation with a tolerance of 1%. 1% seems reasonable in the case of BisqEasy, since max amount is quite low.
* Save agreedPriceSpec and marketPrice in BisqEasyContract. AgreedPriceSpec is added since when the taker is the seller, this will change from MarketPrice to whatever the seller sets (fix or floating price). Both of them are needed for future validation and disputes.

Fix #1031


desktop/src/main/java/bisq/desktop/main/content/bisq_easy/trade_wizard/review/TradeWizardReviewModel.java
```diff
@@ -120,4 +122,4 @@ public void reset() {
         showCreateOfferSuccess.set(false);
         showTakeOfferSuccess.set(false);
```
HenrikJannsen, line 123: marketPrice is not reset in reset method

HenrikJannsen (approved):
utACK

desktop/src/main/java/bisq/desktop/main/content/bisq_easy/trade_wizard/review/TradeWizardReviewModel.java
mrblue313: Fixed

trade/src/main/java/bisq/trade/bisq_easy/protocol/messages/BisqEasyTakeOfferRequestHandler.java
```diff
@@ -125,4 +117,43 @@ private void commitToModel(ContractSignatureData takersContractSignatureData, Co
...
+
+        long amount = isBase ? takersContract.getBaseSideAmount() : takersContract.getQuoteSideAmount();
+        String errorMsg = "Market price deviation is too big.";
+        checkArgument(amount >= minAmountWithTolerance, errorMsg);
+        checkArgument(amount <= maxAmountWithTolerance, errorMsg);
+    }
+
+    private Optional<Monetary> getMinAmount(BisqEasyOffer takersOffer, BisqEasyContract takersContract, boolean isBase) {
+        return isBase
+                ? OfferAmountUtil.findBaseSideMinOrFixedAmount(
```
djing-chan: I like the style for the ternary op with the `?` and `:` in the next line, similar like in fluid interfaces the `.` in the next line.
If others agree we could add that to the code guide lines to unify our code style. (Used for longer statements only).

djing-chan (approved):
ACK

HenrikJannsen (approved):
utACK

PR #1346, opened 2023-11-09
Change quote message background colour to differentiate from list view

### LINKED ISSUE #1326: Background of quote message same as list view background
The background color of the quote should be different to the other background.
As it is part of the new message sent maybe the same background as the input textfield box might be an idea.

Screenshot 2023-11-05 at 13 17 41

axpoems:
Add fix for the quote message issue. I keep working on reviewing the Bisq 2 grey shades.

image

Resolve #1326


HenrikJannsen (approved):
utACK

HenrikJannsen:
Do you know why the window header bar is white? On OSX and Linux its grey which fits better to the app style. If you have any idea how to change that please do so. Not much familiar with Windows.

axpoems:
> Do you know why the window header bar is white? On OSX and Linux its grey which fits better to the app style. If you have any idea how to change that please do so. Not much familiar with Windows.

I'm not certain, but I'll look into it.

PR #1347, opened 2023-11-09
Fix remove password bugs

namloan:
Fixes password bugs described in https://github.com/bisq-network/bisq2/pull/1335#issuecomment-1803180368


HenrikJannsen (approved):
utACK

PR #1348, opened 2023-11-09
Add payment confirmation waiting state icon for BisqEasy

axpoems:
image

_*Note that in the image the spinning circle is still in the old green colour (but this is already being updated in different PR #1333)._


HenrikJannsen (approved):
utACK

PR #1349, opened 2023-11-09
Improve final step UI screen in BisqEasy trade

### LINKED ISSUE #1186: Improve completed screen at the trade process
The completed screen could need more love...

axpoems:
After
Screenshot 2023-11-09 184625

Before
Screenshot 2023-11-09 184952

Resolve #1186


HenrikJannsen:
I am not sure if the OK icon takes too much space. I understand it follows the waiting animation style but to me it looks too large for the OK icon and moves the important content further down.

I was more thinking in a summary similar as in the create offer review screen:
Screenshot 2023-11-10 at 08 07 47

But I am aware of the limited space restrictions, so not recommending to apply that screen to here.
But by not using MaterialTextField fields we would be more flexible with design and space.

Issue #1318, opened 2023-11-01
Distributing the market price data in the P2P network


HenrikJannsen:
See https://github.com/bisq-network/bisq2/discussions/1275

djing-chan:
I can work on that

PR #1355, opened 2023-11-10
Distribute market price to p2p network

djing-chan:
Implements https://github.com/bisq-network/bisq2/issues/1318

We persist the market price data so after the first start up price data is immediately available. This data might be outdated and this need to be handled in the client code.
Next to the http requests to price nodes we get market price data provided via the p2p network. The oracle node is requesting from the price nodes and propagate the price to the network. Always the more recent price data will be used.

The UI shows detail information as tooltips:

Screenshot from 2023-11-10 20-04-21
Screenshot from 2023-11-10 20-04-13


HenrikJannsen (approved):
ACK

PR #1359, opened 2023-11-11
Fix path for icon shown in OS notifications on Linux.

djing-chan:
Tested with installation from binary and displayed icon at Ubuntu.


HenrikJannsen (approved):
utACK

PR #1366, opened 2023-11-12
Refactor result object in network module

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1365


HenrikJannsen (approved):
utACK

PR #1367, opened 2023-11-12
Fix missing linebreaks at bisq2 welcome [4]

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1366


HenrikJannsen (approved):
utACK

PR #1422, opened 2023-11-26
Change receive amount description when creating an offer as buyer

### LINKED ISSUE #1342: Amount To Receive is misleading for Buy offers
As the buyer does not defined the trade price the BTC amount is not the amount to receive but just a rough estimate.
We should use a different headline for that. Maybe "Expected amount to receive" or something shorter.
Screenshot from 2023-11-09 12-29-26

mrblue313:
When creating an offer as buyer, at the review screen the amount in BTC stated is just the amount in BTC at the current market price. I.e., not the amount to receive nor the expected amount to receive. Add an additional text description to make this clearer.

Fix #1342

image


HenrikJannsen (approved):
utACK

Issue #1421, opened 2023-11-25
Sometimes most recent chat message is not fully visible


djing-chan:
After scrolling it becomes fully visible. Probably a render timing bug.

namloan:
This actually happens very frequently. @HenrikJannsen please assign me to this one.

PR #1423, opened 2023-11-27
Ensure scrollDown happens after a new message is added

namloan:
Fixes https://github.com/bisq-network/bisq2/issues/1421


HenrikJannsen (approved):
ACK

PR #1433, opened 2023-11-29
[WIP] Clean up translations

hiciefte:
Created a WIP PR for discussion purposes


HenrikJannsen (approved):
ACK

hiciefte:
Thanks for the ACK @HenrikJannsen! @alvasw Could you please allow my first PR workflow so I can see if the PR is mergable? ...and if you could merge it as well would be even better 😉  - so I have a new base to work from for the remaining issue with the Transifex integration.

Issue #1447, opened 2023-12-02
After taking an offer the takers offerbook channel is broken


djing-chan:
When selecting another channel it works, but the channel used when taking the offer is not accessible.

PR #1448, opened 2023-12-02
Fix offerbook channel selection

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1441

Fixes https://github.com/bisq-network/bisq2/issues/1447


HenrikJannsen (approved):
ACK

PR #1449, opened 2023-12-02
Remove visible flag for offerbook channels

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1448


HenrikJannsen (approved):
ACK

PR #1450, opened 2023-12-02
Fix offerbook channel selection at startup

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1449

Fixes https://github.com/bisq-network/bisq2/issues/1446


HenrikJannsen (approved):
ACK

PR #1499, opened 2023-12-08
Let network module do initialisation of default node

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1497


djing-chan:
Unfortunately the changes in main makes it impossible to rebase or merge. I will rework on it based on a new branch. Would be good if coordinate a bit better in future when larger changes happen, as its quite some effort and risk to get the branch recreated.

PR #1523, opened 2023-12-16
Add config for default tor private key

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1522


HenrikJannsen (approved):
ACK

PR #1537, opened 2023-12-19
Remove redundant listener calls in data service

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1534


HenrikJannsen (approved):
ACK

PR #1539, opened 2023-12-19
Unify chat window headers for public channels

axpoems:
Introduce improvements in new design for public channels (BisqEasy offerbook, and domains Discussions, Events and Support).

Offerbook:
image

Events domain:
image

I will follow-up with another PR regarding improvements in private chats and other further design reviews.


HenrikJannsen:
Looks good to me.

Was the change of the green color in the main path buttons at dashboard and bisq easy welcome screen intended?
I find it better with the green and in usability test it worked very well that ppl felt they should follow that suggestion.

For the common chats header I think it would be better to move the icon from the tab to the header and make maybe the header text the same style as in offerbook. 

For the create offer button a bigger text would give it more weight and we could consider to add an icon.

axpoems:
> Was the change of the green color in the main path buttons at dashboard and bisq easy welcome screen intended?
No, this is a regression. Thanks for pointing it out. I will fix this now.

> For the common chats header I think it would be better to move the icon from the tab to the header and make maybe the header text the same style as in offerbook.
I agree. I will add the icons with the PR where I make the improvements in the icons.
As for the header title, I agree as well with aligning both, but I think it should be the offerbook taking after the public chats.

> For the create offer button a bigger text would give it more weight and we could consider to add an icon.
I will review all the text sizes in general so for the moment, some things will appear a bit off.
In this case in particular, I think the surrounding texts have the text size too big. In my PR #1530 I am decreasing the size of the input text, and this will adjust the weights. I will update that PR soon.

> Maybe to move the 2 icons left to the search box would make the search box more aligned with the right side. Alternatively maybe we could find other icons which would fit better to the search box style.
I am working on this. This is part of follow-up PRs that I am intending to make soon.
I will create an issue with the mock-ups that I have made already for this.

axpoems:
@HenrikJannsen 

image

I have aligned both titles now. For the sake of design, I had to disable the channel icon until we have a better one.

HenrikJannsen (approved):
utACK

HenrikJannsen (approved):
ACK

HenrikJannsen:
Small thing: The arrow for the markets looks a bit too bold compared to the headline and the padding might be adjusted a bit.

axpoems:
> The arrow for the markets looks a bit too bold compared to the headline

That's true. I've noticed, and will make a new icon for this.

HenrikJannsen:
> Can you be more specific? I see it centered. "BTC/EUR" is in the middle of the y axis, and then the description continues on the same line. This is by design.

I meant only the arrow down icon for the drop down. It feels to me 1 px too low and maybe 1-2 px too much left. 
The same base line for headline and description looks good to me.

axpoems:
> I meant only the arrow down icon for the drop down. It feels to me 1 px too low and maybe 1-2 px too much left.
Yeah, I did not go pixel perfect there since the arrow (as discussed before) is incorrect. Will fix paddings when I introduce the new arrow in another PR.

axpoems:
> Hi @axpoems, I can't merge this PR due to a merge conflict in `apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/ChatView.java`.

@alvasw - Fixed now. Thanks for the heads up.

HenrikJannsen (approved):
utACK

PR #1545, opened 2023-12-21
Fix bug with sending moderator messages

djing-chan:
Fixes https://github.com/bisq-network/bisq2/issues/1436


HenrikJannsen (approved):
ACK

Issue #1411, opened 2023-11-24
Check if OS notifications work correctly on Windows


HenrikJannsen:
On Windows we use AwtNotifications. We should figure out if there is a native solution which works well on all commonly used Windows versions. If not feasible we need to check if AwtNotifications works sufficiently well.

mrblue313:
I can look into this.

HenrikJannsen:
Depends on https://github.com/bisq-network/bisq2/issues/1383

HenrikJannsen:
@mrblue313 Any update on that task?

mrblue313:
I am working on it. Will give an update in the following days.

mrblue313:
Giving an update on the current status of this.

### Testing the current solution

I have been testing using Clear-net and Windows 11.

The following matrix was considered:
* Private/public channels
* @mentions
* User being online/offline 
* Edit/delete messages
* Notification options (all/@mention/off)
* Windows notification system

These were my findings:
* Clicking on the notification only dismisses it. We could potentially pass an action here and allow navigation. Alternatively, we could improve the navigation path. For instance, when receiving a message in "Conferences", it could say "Events > Conferences". At least when testing it doesn't feel straightforward where the message comes from.

![image](https://github.com/bisq-network/bisq2/assets/144623880/c724eef6-c663-4b40-9d5d-60953ed340b0)

For instance, in the case of the offerbook, this is what it displays (not very indicative):

![image](https://github.com/bisq-network/bisq2/assets/144623880/fdfdcd81-7e80-4845-b442-01333e8f3988)

* "Java(TM) Platform SE binary" appears as the title, since it is the default description of the Java application process running the code. I would like to test this with the binary as well.

![image](https://github.com/bisq-network/bisq2/assets/144623880/385dc827-21ea-4a3e-8353-4bae821bcade)

Same when you click on the three dots icon:

![image](https://github.com/bisq-network/bisq2/assets/144623880/514930c7-cdcf-4d6a-9fc1-24838831aa5a)

And in the system settings:

![image](https://github.com/bisq-network/bisq2/assets/144623880/c3e99e9a-c9ea-47b4-a9d1-3f0fac958876)

* The body/message gets truncated very early. Perhaps an improvement could be to ensure those are shorter for fixed messages like publishing offers.

![image](https://github.com/bisq-network/bisq2/assets/144623880/ea4727d5-0438-477c-9991-63cffa240b6a)

* Notifications do show in the Windows notification panel:

[...]

PR #1549, opened 2023-12-21
Improve notifications

mrblue313:
Related to #1411

### DIFF
--- a/chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
+++ b/chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
@@ -362,14 +362,17 @@ public class ChatNotificationService implements PersistenceClient<ChatNotificati
         } else {
             String channelInfo;
             if (chatMessage instanceof PublicChatMessage) {
-                // For Public messages we show title: `{username} ({channel title})`
-                channelInfo = chatService.findChatChannelService(chatChannel)
+                // For Public messages we show title: `{username} ({channel domain} - {channel title})`
+                String channelTitle = chatService.findChatChannelService(chatChannel)
                         .map(service -> service.getChannelTitle(chatChannel))
                         .orElse(Res.get("data.na"));
+                channelInfo = StringUtils.createNavigationPath(
+                        chatChannel.getChatChannelDomain().getDisplayString(), channelTitle);
             } else {
                 // For Private messages we show title: `{username} ({channel domain} - Private message)`
-                channelInfo = chatChannel.getChatChannelDomain().getDisplayString() + " - " +
-                        Res.get("chat.notifications.privateMessage.headline");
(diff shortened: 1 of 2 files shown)


chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
```diff
@@ -362,14 +362,17 @@ private <M extends ChatMessage> ChatNotification createNotification(String id, C
...
             String channelInfo;
             if (chatMessage instanceof PublicChatMessage) {
-                // For Public messages we show title: `{username} ({channel title})`
-                channelInfo = chatService.findChatChannelService(chatChannel)
+                // For Public messages we show title: `{username} ({channel domain} - {channel title})`
+                String channelTitle = chatService.findChatChannelService(chatChannel)
                         .map(service -> service.getChannelTitle(chatChannel))
                         .orElse(Res.get("data.na"));
+                channelInfo = StringUtils.createNavigationPath(
+                        chatChannel.getChatChannelDomain().getDisplayString(), channelTitle);
```
HenrikJannsen: I think a higher level Util would be better here. One which passed the channel only and the Util method build the full display string). Would need to be in the chat module as it would take ChatChannel as param.

chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
```diff
@@ -367,10 +367,13 @@ private <M extends ChatMessage> ChatNotification createNotification(String id, C
                         .map(service -> service.getChannelTitle(chatChannel))
                         .orElse(Res.get("data.na"));
                 channelInfo = StringUtils.createNavigationPath(
-                        chatChannel.getChatChannelDomain().getDisplayString(), channelTitle);
+                        " - ",
```
HenrikJannsen, line 370: Do we need to customize that delimiter?

HenrikJannsen (approved):
utACK

Nits can be addressed in a follow up PR as well

chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
mrblue313: I added this because I didn't like how the hyphen looked like, therefore the option to change it.
I went ahead and changed it to `>` in #1561. I think the latter is more appropriate for this context, and it looks better in general and in particular in cases where `-` is already in the channel title like `Off-topic`.

chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
mrblue313: I agree. Somehow I understood from [your comment](https://github.com/bisq-network/bisq2/issues/1411#issuecomment-1859412380) that it would be useful in `StringUtil`. I changed this in #1561.

Issue #1538, opened 2023-12-19
Scrolling to bottom needs better solution


djing-chan:
We scroll down to the bottom (with a delay as otherwise it does not get applied) to see the latest messages.
This itself has 2 open issues:
1. The delay causes unpleasant jump
2. The content is sometimes truncated and only after roll-over of messages its rendered correctly

Beside that there is a more serious problem as we apply that scroll to bottom at every new message getting added. Thus, if the user scrolls up to read old messages they would get confused why the app scrolls suddenly down to show the latest message.

Not sure whats the best solution here. One would be to deactivate the auto-scroll if the user has scrolled manually.
Another would be to only autoscroll at activate and show new messages with some UI indicator.

Maybe @axpoems has some idea...

axpoems:
I think this can also be improved from the backend. We do not need to load all the messages in the UI at the beginning. Something like pagination would be a good feature here. Then we can load as needed in a reverse order.

HenrikJannsen:
Hm.. that data is all locally present, so the only thing we would win is that the list view does not get filled with so many items, but if item renderers and list cells are implemented correctly and do not leak then even a very high number of items should not become a problem.
Also if the user would scroll manually and we would keep loading the keeping the  scroll position could be a bit of a challenge.

djing-chan:
I am working on that.

PR #1553, opened 2023-12-27
Fixes https://github.com/bisq-network/bisq2/issues/1538

djing-chan:
Fixes https://github.com/bisq-network/bisq2/issues/1538


axpoems:
Good solution! I just tested it, a few comments:
* The icon is difficult to spot, especially when it overlaps with a message. I think it would be better to have it aligned at the center.
* I would always use the green version, or if you using both, I would use a brighter grey.
* I would also prefer a single arrow, as with the current one it's not as intuitive.

djing-chan:
>     * The icon is difficult to spot, especially when it overlaps with a message. I think it would be better to have it aligned at the center.

I tried to follow loosely the style used in Elements (Matrix). They use a larger icon but they have a large padding right so that the icon does not overlap the messages. We have some buttons on rollover so I needed to make the button small enough to not overlap those buttons of the message (e.g. delete) I agree that its not looking perfect. Maybe you can put your designer hand on it?

> 
>     * I would always use the green version, or if you using both, I would use a brighter grey.
> 
>     * I would also prefer a single arrow, as with the current one it's not as intuitive.

Would you be available to improve that? I am not a designer and I guess my time is better spent on other tasks.

There might be also some bugs/UI glitches (not aware of any open but as I stumbled over so many during dev I would not be surprised if others find some). That relatively small feature took already lot of time, so that's why I wanted to timebox a basic version which solves the critical problem we had before.

There is also some UI bug in the cell renderer which cause sometimes invalid layout. Maybe there is also some issue which causes the unusual long delay after a item gets added until layout processing is completed.

axpoems:
Sure, I will take a look.
I agree with the bugs. I noticed that we still have the problem with the last message not showing properly when scrolling until the end (it only shows completely after a second scroll) - will check that as well.

PR #1554, opened 2023-12-27
Re activate networkload update

djing-chan:


network/network/src/main/java/bisq/network/p2p/node/network_load/NetworkLoadSnapshot.java
```diff
@@ -0,0 +1,52 @@
...
+
+import javax.annotation.Nullable;
+
+@Slf4j
+@Getter
+@ToString
+@EqualsAndHashCode
+public final class NetworkLoadSnapshot {
+    private NetworkLoad currentNetworkLoad;
+    @Nullable
```
namloan: @HenrikJannsen has a preference for using Java's Optional<> instead of the Nullable annotation. See the discussion here https://github.com/bisq-network/bisq2/pull/1106#pullrequestreview-1565988861

What are your thoughts on that?

network/network/src/main/java/bisq/network/p2p/node/network_load/NetworkLoadSnapshot.java
```diff
@@ -0,0 +1,52 @@
...
+import lombok.Getter;
+import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
+
+import javax.annotation.Nullable;
+
+@Slf4j
+@Getter
+@ToString
+@EqualsAndHashCode
```
namloan, line 30: You can replace these by [@Value](https://projectlombok.org/features/Value) which would also allow you to remove the private's and the final's

namloan:
utACK

network/network/src/main/java/bisq/network/p2p/node/network_load/NetworkLoadSnapshot.java
djing-chan: Yes, agree. Will change it. Actually the code had a potential access to a null object. So confirms that using Optional is more safe.

network/network/src/main/java/bisq/network/p2p/node/network_load/NetworkLoadSnapshot.java
djing-chan: I find that omitting final and private reduces clarity. Specially for objects sent over the network having the class defined final is important for security. The @Value annotation would also trigger in IntelliJ a warning that final/private is not required.
I personally also prefer to have it more explicitly visible which annotation are added. In Bisq 1 more of the complex Lombok annotations are used but I find that sub-optimal and error prone(e.g. auto-constructors).

PR #1555, opened 2023-12-27
Fix alignment in main left menu

axpoems:
Change `mainMenuItems` to account for the `selectionMarker` as its children so that they have the same width.

Before
image

After
image

This was also affecting the collapsed menu.


HenrikJannsen (approved):
utACK

PR #1560, opened 2023-12-29
Update icons in package

mrblue313:
Related to #1411


HenrikJannsen:
Those different style icons have been created by @pedromvpg to differentiate a bit from Bisq 1. I have no strong opinion about which version we should use. Maybe @axpoems can add his opinion.

mrblue313:
I changed them since I saw that those are the new ones we are using in the app title. Otherwise it looks inconsistent in the installed binary:

image

HenrikJannsen:
> I changed them since I saw that those are the new ones we are using in the app title. Otherwise it looks inconsistent in the installed binary:

Ok. I leave it to @axpoems for review and ACK. I am ok with both the original icon and the variation (but then I agree it should be more consistent). To have some slight new version of the logo might be not bad to distinguish from Bisq 1.

axpoems:
I gave my view on this in the PR where I introduced them for the app window, see comment [here](https://github.com/bisq-network/bisq2/pull/1296#issue-1957203696).

Moreover, IMO we should go with the ones that @mrblue313 added, because this seems a more mature and recent version of the logo, and its usage as icon differentiates the most with Bisq 1.

> I am ok with both the original icon and the variation (but then I agree it should be more consistent). To have some slight new version of the logo might be not bad to distinguish from Bisq 1.

I also thought about having something new to market it as Bisq2. Not in the logo itself, but rather on the name when it appears next to the logo. Those files have not been updated: https://github.com/bisq-network/bisq2/blob/main/apps/seed-node-app/package/windows/images/WixUIDialogBmp.bmp

Nonetheless, if Bisq2 is intended to reach a wider user population, then maybe it doesn't make sense to rename it as Bisq2 and simply leave it as Bisq as it is now.

PR #1563, opened 2023-12-29
Various chat improvements

axpoems:
image

image


HenrikJannsen:
Is it intended that the chat message box does not fill the available width of the list view?

HenrikJannsen (approved):
utACK

axpoems:
> Is it intended that the chat message box does not fill the available width of the list view?

What do you mean? Are you referring to the fact that there's 2 buttons + input text in one line?

Issue #1491, opened 2023-12-07
Bisq Easy trade UI/UX review


axpoems:
I have performed a UI/UX review of the Bisq Easy trade protocol. I present my findings/highlights here.

![ebisq_trade-steps](https://github.com/bisq-network/bisq2/assets/145597137/f9b19066-1dab-4c94-833c-cbc53b262e2c)

The trading path for Bisq Easy actually consists of seven steps. Nontheless, it shows only five. Steps 2 and 3 are repeated with different and unrelated states, for example step 2 of seller shows first `WAIT UNTIL PAYMENT HAS STARTED` and after `WAIT FOR BITCOIN ADDRESS`. This is confusing for the user who is following the path and suddenly can get lost by the change of name step.

```
     SELLER                             BUYER

1    1 SEND ACCOUNT DATA                1 WAIT FOR ACCOUNT DATA
     |                                  |
2    2 WAIT UNTIL PAYMENT HAS STARTED   2 START PAYMENT
     |                                  |
3    2 WAIT FOR BITCOIN ADDRESS         2 SEND BITCOIN ADDRESS
     |                                  |
4    3 WAIT FOR PAYMENT                 3 WAIT FOR PAYMENT CONFIRMATION
     |                                  |
5    3 SEND BITCOIN                     3 WAIT FOR BITCOIN PAYMENT
     |                                  |
6    4 AWAITING BITCOIN CONFIRMATION    4 AWAITING BITCOIN CONFIRMATION
     |                                  |
7    5 COMPLETED                        5 COMPLETED
```

Moreover, the trading path should be the same for both peers (seller and buyer) and the action messages, as they already are, specific for each one. This way we provide a simple framework to assimilate the Bisq Easy protocol and mental map for either role.

Thus, I propose the following:

```
SELLER / BUYER           SELLER                                        BUYER 

1 PAYMENT ACCOUNT      | Send your payment account data to the buyer | Wait for the seller's payment account data
|
2 PAYMENT TRANSFER     | Wait for the buyer's 'XXX' payment          | Send '100.00 XXX' to the seller's payment account
|
3 BITCOIN ADDRESS      | Wait for the buyer's Bitcoin address        | Fill in your Bitcoin address
|
4 PAYMENT CONFIRMATION | Check if you have received '100.00 XXX'     | Wait for the seller to confirm receipt of payment
|
[...]

PR #1566, opened 2023-12-31
Redesign Bisq Easy Protocol

axpoems:
Towards #1491 

This PR corresponds to the change in the protocol and implements the proposal made in the issue.

A mention on system messages.
I have tested different options for that, and what I present here is what I think is a good solution. However, I would like to discuss this subject further.

I will make follow-up PRs to improve the UI alongside the already discussed changes in different issues.


HenrikJannsen:
Looks great. Feels much better now!

BisqEasyBuyerAsMakerProtocol/BisqEasyBuyerAsTakerProtocol are missing the 2nd option of the order of btc address/fiat payment account data exchange.
It still works as it uses the event queue in the FSM (where the previously received message gets picked up), but I think its better to explicitly layout the valid variants of the protocol flow.

Do you have any feedback or improvement suggestion to the FSM protocol definition? Was it easy for you to change the protocol?
We have duplicated code in the 2 pairs of protocols. I tried to move code to base classes/interfaces to avoid those duplications but it felt less transparent and harder to reason about the protocol flow, so I considered the cod duplication the lesser evil. Do you have any idea for improving that?

What about keeping the UI classes with the trade step numbers synced? SellerState3-> SellerState3a and SellerState4->SellerState3b?

The icon for completed looks a bit like a V. Maybe use a more common checkmark icon or differentiate even more from the 1-2-3 steps?

HenrikJannsen (approved):
ACK

PR #1569, opened 2024-01-02
Bisq Easy protocol improvements

axpoems:
Based on #1566


axpoems:
Also changing the action button's text in my last commit:
image

HenrikJannsen (approved):
utACK

HenrikJannsen (changes requested):
NACK

The use case where the seller sends the account data first triggers for the buyer the change to step 2 thus he cannot send the btc address.

HenrikJannsen (approved):
utACK

I do not have time atm for an in depth testing/review but the PR looks good to me.
When testing all scenarios its important to also test the cases that peers are offline at any state.

PR #1573, opened 2024-01-07
Fix win notifications

### LINKED ISSUE #1411: Check if OS notifications work correctly on Windows

mrblue313:
Fixes #1411 

Towards Windows notifications:
* Fix `trayIcon` not showing icon
* Fix notification title

This is how the notification looks like on Win11 with the changes:
image

---
### Icon
The resolution is not optimal. However, I have tried different Image classes and image formats and resolutions. This is the version that looked best.

### Title
When it comes to the title, as mentioned in #1411, it seemed that what was used as title was the `description` argument. I have looked into this a lot, before opting to change the `description` itself. There is little documentation on this and [this stackoverflow post](https://stackoverflow.com/questions/15832791/change-the-icon-name-in-the-notification-area-in-windows-in-java) seems to confirm that it is indeed `description` what is expected to be used.

Looking into other examples on Windows, which uses notifications, they all seem to have set a description text that matches the name of the app. Therefore, I opt for this solution for Windows as well.


HenrikJannsen (approved):
utACK

PR #1576, opened 2024-01-09
Add Brazilian Portuguese translation

hiciefte:
I also added additional comments to other translations that were missing  and removed brand names of payment methods from the translations only keeping them in the English source file


HenrikJannsen (approved):
utACK

hiciefte:
I think someone with maintainer rights have to force a rebuild for the macOS latest as it got stuck there. Thanks!

PR #1578, opened 2024-01-09
Arrange title and description of support channels

axpoems:


HenrikJannsen (approved):
utACK

hiciefte:
@axpoems As I added the Transifex integration recently in a semi-automated manner (we didn't want to give Transifex full repository access for sync tasks). Someone needs to update the source file after it has been changed by running in that case `tx push -s bisq-2.chatproperties`. As it would need an API key for everyone to update this I think for now the easiest way to go forward would be to ping me and I'll take care of it. Also we need to update translations for this changed keys for the release as well.

axpoems:
> @axpoems As I added the Transifex integration recently in a semi-automated manner (we didn't want to give Transifex full repository access for sync tasks). Someone needs to update the source file after it has been changed by running in that case `tx push -s bisq-2.chatproperties`. As it would need an API key for everyone to update this I think for now the easiest way to go forward would be to ping me and I'll take care of it. Also we need to update translations for this changed keys for the release as well.

We are continuously changing strings, so I don't think pinging you every time there's a modification is a viable solution. I think it would be better if you could run it periodically, for instance once a week. Otherwise, you could also watch all activity in this repo - then you will be able to see if there were changes.
However, IMO the best solution is that you send me the API key and I can run it when applicable. You could document how to do that in the docs so everyone knows how to do it.

PR #1583, opened 2024-01-11
Fix crash if clicking on non-editable TextInputControl with prompt text

mrblue313:
Previous to this fix, if the user tried clicking on the `MaterialTextField` > "Bitcoin payment" > "Waiting for blockchain data..."

image

the program was crashing as editing was not allowed:

image


HenrikJannsen (approved):
utACK

PR #1586, opened 2024-01-13
Various improvements in chat design

### LINKED ISSUE #1542: ChatMessagesComponent - Input text shouldn't be overlapping with send button.
Input text shouldn't be overlapping with send button.

image

axpoems:
Includes:
* Extend list view to full width and set max size to the message cells.
* Change design of chat input text bar
* Fix #1542 
* Misc refactors

image


HenrikJannsen (approved):
utACK

Issue #1541, opened 2023-12-19
Improve design of public and private chat windows


axpoems:
<br>

Here are the mock-ups I've created and am currently implementing. They improve the existing design of public chat tabs, private chats, and the offerbook tab.

The issues addressed here include the headers of the chat window, repositioning the search box, introducing drop-down menus to group options, and resolving UI issues. Further refinements will be made, such as reviewing letter sizes, icons, and addressing small details.

<br>

Private Chats:

![BisqE_pchat_nf_07](https://github.com/bisq-network/bisq2/assets/145597137/b59bd650-4a5d-4265-aac6-cc962998ea84)

![BisqE_pchat_nf_08](https://github.com/bisq-network/bisq2/assets/145597137/e9762d13-9b41-4f81-8703-71c5785cb50b)

<br>

Public Chats:

![Pchannel_nf_07](https://github.com/bisq-network/bisq2/assets/145597137/f3aa9669-f62f-4d39-9811-dc311babca05)

![Pchannel_nf_08](https://github.com/bisq-network/bisq2/assets/145597137/78a8c823-f4da-4c92-835a-c40fb64e40f4)

<br>

Offer Book:

![BisqE_offbook_nf_07](https://github.com/bisq-network/bisq2/assets/145597137/0fadda15-a983-4e75-97d8-013c330d1a1d)

![BisqE_offbook_nf_08](https://github.com/bisq-network/bisq2/assets/145597137/e7199840-3fdd-41ea-93fc-3b8013781b1c)

axpoems:
@HenrikJannsen  - I'm already working on these improvements, so you can assign it to me.

axpoems:
I'll be using this issue to show and discuss in advance the proposed steps towards the improvement of the chat's design.

---

Following up on the discussion from #1530 to apply a max width to the chats as to improve UX, these are two mock-ups showing the options I have explored. I personally like the centered version, although I don't have strong feelings with it and think that the left is a valid option as well. 

Also, included here is the revised input text. The old input text felt overloaded with grey layers and the white button was very bright. This newer version is leaner with a more discreet but noticeable send button, still aligned with the original design.

<br>

![max-width2_center](https://github.com/bisq-network/bisq2/assets/145597137/38aadf10-3fa6-4e1f-8ade-9675f5a6b240)

<br>

![max-width2_left](https://github.com/bisq-network/bisq2/assets/145597137/979c2f6e-7724-4b98-bda4-32aa55240178)

[...]

PR #1589, opened 2024-01-15
More improvements on chat design

axpoems:
Based on #1586 

* Add placeholder in common chats
* Change `SearchBox` style
* Align input text elements and side buttons to the bottom

image

All towards #1541


HenrikJannsen (approved):
utACK

PR #1595, opened 2024-01-18
Fix bug in TradeParty's protobuf encoding

axpoems:


HenrikJannsen (approved):
utACK

Ah good that spotted that bug!

PR #1597, opened 2024-01-18
Set network verison by config [1]

djing-chan:
Make the network version configurable to separate dev networks with production network


HenrikJannsen (approved):
utACK

PR #1598, opened 2024-01-18
Add network messages to whitelist [2]

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1597

PR #1599, opened 2024-01-18
Use optional for address ownership proof [3]

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1598


HenrikJannsen (approved):
utACK

PR #1600, opened 2024-01-18
Use different interface for network and persistable proto messages [4]

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1599

Separating the interfaces makes it easier to spot the critical messages which require validation when we receive it from the network.


HenrikJannsen (approved):
utACK

PR #1601, opened 2024-01-18
Verify network data [5]

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1600

The network message based validation is primarily a protection against dos attacks to set bounds on the size of messages. In some cases we have also specific validation (e.g. amounts to be > 0) but the core validation for domain specific data is up to the domain operating on that data.

We call verify() in all constructors from final classes. Abstract classes do not call validate as it would call the validate method in the extending class which might use fields which are not initialized when the super class's constructor is called.
We could also call the validate from the `fromProto` method, but by using the constructor we verify also the case when the message is not created from network input (e.g. sender user case) so it helps to spot potential bugs in the validation which otherwise would only be thrown at the receiver.

I have not found yet a easy way how to enforce that only NetworkProto instances are used inside network messages. At the moment it is in the responsibility of the developer to check all fields used in a network message. One way to do it would be with reflection, to iterate all nested objects and check for Proto instances and that those are of type NetworkProto. This should be done only once (e.g. use a static flag once validated). But I hope we find a better solution.

We do not validate primitive data as those are bounded anyway. The typical data types we validate are Strings and Collections (lists, maps, arrays, byte arrays).
For common data like public keys or IDs we use the NetworkDataValidation util class.

Encrypted data cannot be validated meaningfully beside a max. data size limit.

### DIFF
--- a/account/src/main/java/bisq/account/accounts/Account.java
+++ b/account/src/main/java/bisq/account/accounts/Account.java
@@ -19,7 +19,7 @@ package bisq.account.accounts;
 
 import bisq.account.payment_method.PaymentMethod;
 import bisq.common.currency.TradeCurrency;
-import bisq.common.proto.Proto;
+import bisq.common.proto.NetworkProto;
 import bisq.common.proto.UnresolvableProtobufMessageException;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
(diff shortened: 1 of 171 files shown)


HenrikJannsen (approved):
utACK

PR #1603, opened 2024-01-19
New revision of grey colours

axpoems:
* Change grey naming convention to a more consistent, readable and flexible format. After working for a while with the Bisq greys introduced previously, although logically named, they were not easy to remember/pronounce nor quantify. This also aligns with the green naming convention we currently have.
* Update two shade of greys, for better cohesion in the application.
* Introduce two new more shades of grey, to be used for the new subheader bar in chats.

Summary of the changes:
bisq-greys_2


djing-chan (approved):
utACK

HenrikJannsen (approved):
utACK

PR #1604, opened 2024-01-20
Improve error handling for trade protocol [6]

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1601

Error handling will be managed as state transition to an error state. The exception which caused the error is stored in the model for access from clients.

In a follow-up PR we will support error messages sent to the other peer in case an error happens in the protocol.


HenrikJannsen (approved):
utACK

common/src/main/java/bisq/common/fsm/Fsm.java
```diff
@@ -124,7 +161,26 @@ private TransitionBuilder(Fsm<M> fsm) {
...
+            }
+            return fromStates(sourceState);
+        }
+
+        public TransitionBuilder<M> fromAny() {
+            from(State.FsmState.ANY);
+            return this;
+        }
+
+        public TransitionBuilder<M> fromStates(State... sourceStates) {
```
axpoems, line 175: Nice improvement!

axpoems:
I tested this and noticed that if there's a bug (UI not checking for max lengths or unallowed values being submitted for instance) this will be considered as an error and thus the trade will close. However, with this error handling perhaps it is less obvious that there was a problem with the app itself (and not something going with the trade per se). Printing the stack trace as implemented in #1618 improves this, but the screaming popup was better in those cases I believe. Do you think that is enough for users to report that? or even better, do we have a system in place to to collect those failures?

djing-chan:
@axpoems Is that bug still happening with my latest PR? 
I fixed several issues there.

To not throw an error at the handle method might not be the best solution. But it would only notify the client if happening from the UI events, not if it was triggered by the messages.
But yes we could at least for those cases throw the exception. I will have a look into that.

axpoems:
Tested again with your lastest PR and I do get the popup now, so should be good now.

image

djing-chan:
That error should not happen anymore with the latest PR. The popup listens on errors from the trade. I think it would be more robust to go back to the initial throwing of exceptions in the handle method. Will make a new PR for that soon.

PR #1609, opened 2024-01-20
Stack reputation, take offer button vertically

DrColver:
The limited horizontal space makes it hard to show the take offer button label nicely when stacked horizontally.

Before
image

After
image

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/components/ChatMessagesListView.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/components/ChatMessagesListView.java
@@ -1012,8 +1012,10 @@ public class ChatMessagesListView {
                                 reputationLabel.getStyleClass().add("bisq-text-7");
 
                                 reputationScoreDisplay.setReputationScore(item.getReputationScore());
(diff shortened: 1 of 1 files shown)


HenrikJannsen:
I think its better to relax the width limitation or use the vertical layout only dynamically once the width gets too small.
I find the horizontal layout better. 

@axpoems What do you think?

DrColver:
If it's acceptable to relax the width that might be better, but keep in mind the button text might be much wider when localized with other languages.

axpoems:
I also agree we should relax the width limitation for now. You can also work out the margins a little bit (especially the right-most margin seems bigger than the left one).

DrColver:
image

The button can now handle the currently longest take offer localization.

axpoems:
The value of `CHAT_MESSAGE_BOX_MAX_WIDTH` is specifically chosen to improve the UX of the chat, and in particular optimise for readability. So this is not really a parameter with flexibility.
I think the best option for now is to relax the width limitation of the button.

DrColver:
Changed to fit button to text

image

image

axpoems (approved):
utACK

PR #1614, opened 2024-01-20
Fix ChatMessageListItem equals

DrColver:
Fix issue with adding duplicates of chat messages due to equals() comparison failing when running contains().


djing-chan (approved):
utACK

PR #1618, opened 2024-01-21
Improve fsm

djing-chan:
Improves the readability of the protocols.
Add support for branching to make multiple possible paths more explicit.
Add restriction that a state transition is only allowed to a higher order state (using ordinal of enum). 
This also enures that no cyclic graphs are permitted.


HenrikJannsen (approved):
ACK

axpoems:
ACK

PR #1621, opened 2024-01-21
Add chat dropdown

axpoems:
Introduce a dropdown menu for the chat containing all the options/actions.

chat-dropdown

Will follow-up with more improvements on this.
Also, the `DropdownMenu` will change to a common control `BisqDropdownMenu` as to unify dropdowns.


djing-chan (approved):
utACK

Issue #1629, opened 2024-01-23
Persist publishUserProfile date


djing-chan:
We are re-publishing the user profile each time we start the app as we do not persist the last publish date.

djing-chan:
I can work on that

PR #1630, opened 2024-01-23
Do not republish user profile at each start

djing-chan:
Implements https://github.com/bisq-network/bisq2/issues/1629


HenrikJannsen (approved):
ACk

PR #1641, opened 2024-01-25
Refactor confidential message listeners

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1633

PR #1642, opened 2024-01-25
Catch errors in listener invocations in network module

djing-chan:
Based on https://github.com/bisq-network/bisq2/pull/1641


HenrikJannsen (approved):
ACK

PR #1649, opened 2024-01-28
Fix zombie private channel

DrColver:
When a user left the last private chat channel no new channel was selected so the lastSelectedPrivateChannel still kept the closed channel alive. This caused strange issues, such as private chat messages ending up in public channels and private chat messages not being delivered to the correct private channel.


axpoems (approved):
ACK

PR #1660, opened 2024-02-05
Improve initial readme

hiciefte:
@HenrikJannsen 
I just added some minor changes to the initial readme. Not sure if it is worth to change - WDYT?


axpoems (approved):
utACK

IMO it looks better now, nice update!👍

PR #1669, opened 2024-02-11
Complete refactor of chat messages for better handling of different message types

axpoems:
Since there multiple types of chat messages that can appear in the chat box (text messages, system messages, offerbook messages, leave chat, and potentially more), I am decoupling the logic so that they are well-defined and other types can be introduced easily.

I will follow-up with another PR decoupling more, especially the logic inside the `Controller` and `Model` of the `ChatMessagesListView`.


djing-chan:
You can edit in IntelliJ settings to add the header to any new class/interface.

djing-chan (approved):
utACK

PR #1670, opened 2024-02-11
UI style improvements

axpoems:
Improve:
* `leaveChat` item menu red tone when hovering
* Remove round corners at the bottom of the chat window
* Remove pixel from highlight selection in private chats list selection


djing-chan (approved):
utACK

PR #1673, opened 2024-02-12
Introduce UI for system messages

### LINKED ISSUE #1639: Define and implement a solution for system messages
During a trade we send system messages in the chat. At the moment these are sent on behalf of the user running the app.
We should find a better solution showing them as system messages.

This needs to be implemented for the mediator messages: https://github.com/bisq-network/bisq2/blob/3ae04e210986d66d2553df4b24b4d59e13089500/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/OpenTradesUtils.java#L87

As well as to the chat messages sent during the trade to notify about the steps (send btc address, etc).

axpoems:
* Introduce UI for system messages.
* Enable system messages when requesting mediation.

Based on #1669 
Resolves #1639

image


djing-chan (approved):
utACK

djing-chan:
Maybe some half transparent line separator or some slightly diff. background could help to distinguish the messages more?

axpoems:
> Maybe some half transparent line separator or some slightly diff. background could help to distinguish the messages more?

I've experimented with various line separator styles but wasn't satisfied with the results. Ultimately, it seemed that having no separator worked best. Your suggestion also works well and looks good:

image-2-1

axpoems:
With the update:

image
I have created an issue to fix the `leave` type #1676

HenrikJannsen (approved):
utACK

axpoems:
> @axpoems There is a merge conflict in the following file: `apps/desktop/desktop/src/main/resources/css/controls.css`.

Fixed now.

PR #1683, opened 2024-02-17
Add dropdown menu with delete option to BisqEasy my offer message

### LINKED ISSUE #1680: Add dropdown menu and confirmation to delete offer
My offer message should have a dropdown menu to delete offer and a double confirmation by using a popup (same as to delete a message).

image

axpoems:
Resolves #1680

* Refactor chat messages: add different classes for BisqEasy offer messages.
* Replace delete offer button with a dropdown menu with the delete option.
* Add popup to confirm deleting an offer.
* Introduce new delete icon.
* Introduce new shades of grey.
* Make `MyMessage` and `PeerMessage` have the same tone of grey (a darker one).

image

image


HenrikJannsen:
Why did you change to same color tone for my messages and peers messages? Typical chat apps like WhatsApp or Signal use the apps base color for my messages. Not arguing to use green, as that would be too heavy, but a slight difference might help to distinguish the messages more.

axpoems:
> Why did you change to same color tone for my messages and peers messages? Typical chat apps like WhatsApp or Signal use the apps base color for my messages. Not arguing to use green, as that would be too heavy, but a slight difference might help to distinguish the messages more.

The main reason was because the peer message was a very light tone and my message was very dark. This kind of contrast was too big and didn't look good with that background.
Nonetheless, I didn't quite achieve the result I wanted, so what I ended up with is sort of a compromise. Since the layout of the messages have the left/right distinction I didn't think it would be an issue to have the same tone (temporarily). However, I agree that a different colour according to the design colours of the app would be preferable. I was exploring some ideas in this field and tried many different options, but wasn't convinced. Therefore, I need to think this more holistically to be able to find the right ones.
I do like the idea of using a green, and I believe a darker green could work but didn't manage to find the ideal one at that moment. Will follow-up with another PR.

HenrikJannsen:
Maybe to use green only as (partial) border color instead of fill color?

axpoems:
> Maybe to use green only as (partial) border color instead of fill color?

I have tried this, and I didn't fully like it. I will create an issue presenting a proposal to tackle this specifically.

PR #1686, opened 2024-02-19
Replace leave chat button with system message

### LINKED ISSUE #1676: Improve solution for leave chat message type
image

We should remove the button "Leave" and instead just add a system message with a link to leave the chat.

axpoems:
Resolves #1676

* Add `LeaveChatMessage` type to display leave messages as system messages with hyperlink to leave chat, and remove the existing button appearing on the right of the input bar.
* Add info popup before leaving a chat.
* Add green leave-chat icon.
* Minor code refactors.

image

image


HenrikJannsen:
Minor thing: Maybe user "cancel" instead "close" and use "leave chat" instead of "Ok"?

axpoems:
Updated now:
image

PR #1687, opened 2024-02-20
Improve implementation to dynamically set paddings to chat messages

axpoems:
Based on #1686

* Fixes paddings not being set correctly for the case of no scrollbar with max width applied.
* Fixes occasional flickering of chat messages.
* Fixes small misalignment with header and input bar.


axpoems:
> @axpoems I can't merge this PR due to merge conflicts.

Fixed now.

djing-chan (approved):
utACK

PR #1688, opened 2024-02-20
Fix chat message rendering issue when updating styles

axpoems:
Based on #1687

Remove the listener from cell and delegate setting max chat box width to the message class so that the latter can be used to keep track of changes and update accordingly.


axpoems:
> @axpoems I can't merge this PR due to merge conflicts.

Fixed now.

djing-chan (approved):
utACK

PR #1689, opened 2024-02-20
Add new info and help icons to drop-menus.

axpoems:
Add new, clean redesigned info and help icons for chats drop-menus.

Towards #1211.


HenrikJannsen (approved):
utACK

PR #1690, opened 2024-02-21
Update navigation icons to match bisq2 green colour

axpoems:
• Update  the main navigation icons to match bisq2 green colour.

Towards #1211.


HenrikJannsen (approved):
utACK

Issue #1697, opened 2024-02-26
IllegalAccessError on javafx VirtualScrollBar when clicking "Open offerbook" on Apple Silicon Mac


cbeams:
Environment:
```
$ ./gradlew --version

------------------------------------------------------------
Gradle 7.6.3
------------------------------------------------------------

Build time:   2023-10-04 15:59:47 UTC
Revision:     1694251d59e0d4752d547e1fd5b5020b798a7e71

Kotlin:       1.7.10
Groovy:       3.0.13
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          17.0.7 (Oracle Corporation 17.0.7+8-LTS-224)
OS:           Mac OS X 13.5.2 aarch64
```

Repro steps:

1. Build and run from latest `main` (`ff4cd1a08` as of this writing): `./gradlew :desktop:desktop-app:run`
2. Click through terms and conditions and icon creation
3. Click "Enter Bisq Easy"
4. Click "Open offerbook"
5. Observe error modal popping up with the following stack trace:

```
Feb-26 12:25:00.222 [JavaFX Application Thread] ERROR b.d.DesktopController: Uncaught exception: java.lang.IllegalAccessError: class bisq.desktop.components.list_view.ListViewUtil (in unnamed module @0x5b8dfcc1) cannot access class com.sun.javafx.scene.control.VirtualScrollBar (in module javafx.controls) because module javafx.controls does not export com.sun.javafx.scene.control to unnamed module @0x5b8dfcc1
	at bisq.desktop.components.list_view.ListViewUtil.findScrollbar(ListViewUtil.java:18)
	at bisq.desktop.components.list_view.ListViewUtil.findScrollbarAsync(ListViewUtil.java:25)
	at bisq.desktop.main.content.components.chatMessages.ChatMessagesListView$View.onViewAttached(ChatMessagesListView.java:700)
	at bisq.desktop.common.view.View.onViewAttachedInternal(View.java:98)
	at bisq.desktop.common.view.View.onViewAttachedPrivate(View.java:93)
	at bisq.desktop.common.view.View.handleSceneChange(View.java:54)
	at bisq.desktop.common.view.View.lambda$new$0(View.java:46)
	at javafx.base/com.sun.javafx.binding.ExpressionHelper$SingleChange.fireValueChangedEvent(ExpressionHelper.java:181)
[...]

Issue #1695, opened 2024-02-26
NoClassDefFoundError on javafx MediaView when clicking "Watch video" on Apple Silicon Mac


cbeams:
Environment:

```
$ ./gradlew --version

------------------------------------------------------------
Gradle 7.6.3
------------------------------------------------------------

Build time:   2023-10-04 15:59:47 UTC
Revision:     1694251d59e0d4752d547e1fd5b5020b798a7e71

Kotlin:       1.7.10
Groovy:       3.0.13
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          17.0.7 (Oracle Corporation 17.0.7+8-LTS-224)
OS:           Mac OS X 13.5.2 aarch64
```

Repro steps:

1. Build and run from latest `main` (`ff4cd1a08` at time of writing): `./gradlew :desktop:desktop-app:run`
2. Click through EUA and icon creation screens
3. Click 'Enter Bisq Easy'
4. Click 'Watch video'
5. Observe error modal popping up with the following stack trace:

```
Feb-26 10:08:08.072 [JavaFX Application Thread] ERROR b.d.DesktopController: Uncaught exception: java.lang.NoClassDefFoundError: javafx/scene/media/MediaView
	at bisq.desktop.main.content.bisq_easy.onboarding.video.BisqEasyVideoController.<init>(BisqEasyVideoController.java:38)
	at bisq.desktop.overlay.OverlayController.createController(OverlayController.java:148)
	at bisq.desktop.common.view.NavigationController.findController(NavigationController.java:137)
	at bisq.desktop.common.view.NavigationController.processNavigationTarget(NavigationController.java:48)
	at bisq.desktop.common.view.Navigation.lambda$navigateTo$1(Navigation.java:66)
	at java.base/java.util.concurrent.CopyOnWriteArrayList.forEach(CopyOnWriteArrayList.java:807)
	at java.base/java.util.concurrent.CopyOnWriteArraySet.forEach(CopyOnWriteArraySet.java:425)
	at bisq.desktop.common.view.Navigation.lambda$navigateTo$2(Navigation.java:66)
	at java.base/java.util.concurrent.ConcurrentHashMap.forEach(ConcurrentHashMap.java:1603)
	at bisq.desktop.common.view.Navigation.navigateTo(Navigation.java:64)
[...]

PR #1698, opened 2024-02-27
Various bug fixes

HenrikJannsen:
Should fix https://github.com/bisq-network/bisq2/issues/1697 and https://github.com/bisq-network/bisq2/issues/1695


cbeams (approved):
I con confirm these changes do fix both #1695 and #1697 on my Mac. Note that I ran against both Oracle JDK 17 and Azul JDK 17, both behave identically.

axpoems (commented):
ACK

Nit. I have seen that occasionally we are getting scrollbar detected when it shouldn't:

image

I have found that a better heuristic to figure out whether the scrollbar is present is to compare list width vs cell width, since list won't change regardless of having scrollbar or not but cell does change. Would need to check if it's possible to apply in this context.

However, this issue is not consistent and therefore not easy to repro. I think it's a minor thing.

axpoems (approved):

PR #1704, opened 2024-02-27
Fix misaligned offerbook messages

axpoems:
In recent changes all the message elements were refactored and now all elements are added in `contentVBox` instead of the component itself. This was missing to update for messages of type offer.


HenrikJannsen (approved):
utACK

PR #1717, opened 2024-02-28
Add offers' filter dropdown menu in BisqEasy Offerbook

### LINKED ISSUE #1548: Add popup to confirm "Ignore user" action
When clicking "Ignore user", all their messages disappear everywhere. This option should be given to the user a bit more carefully.

image

One possible solution is to add a pop-up (with don't show again) explaining what is the result of this action, and how to revert it.

axpoems:
* Add filter offers dropdown menu.

image

* Add warning before applying 'Ignore user' action (Fix #1548)

image

* Misc refactors.


HenrikJannsen (approved):
utACK

Issue #1713, opened 2024-02-28
Make sure that we handle minRequiredTotalReputationScore correctly


djing-chan:
For dev testing we disabled RequiredTotalReputationScore. We need to cross check that it works as expected for release.

PR #1723, opened 2024-02-29
Fix missing reputationscore check

djing-chan:
Fixes https://github.com/bisq-network/bisq2/issues/1713

If reputation score does not match we show those popups:
Screenshot from 2024-02-29 15-26-55
Screenshot from 2024-02-29 15-26-39


djing-chan:
@alvasw Fixed

PR #1726, opened 2024-02-29
Persist filter and sort options in markets [3]

djing-chan:


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/MarketSortType.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package bisq.desktop.main.content.bisq_easy.offerbook;
+
+import lombok.Getter;
+
+import java.util.Comparator;
+
+@Getter
+enum MarketSortType {
```
axpoems, line 25: What do you think about putting all the enums together in one single file? That was my thinking with the `Filters` class, to be a wrapper of all the enums. We could rename it to `FiltersAndSorts` and include this one

Issue #1718, opened 2024-02-28
Offerbook Revision with additional Improvements


axpoems:
Following up on #1572, these were the options considered to place the create offer button in the offerbook:

![ALL-create-offer-button](https://github.com/bisq-network/bisq2/assets/145597137/07da75de-0240-4c79-b9cb-28e14fb5dfba)

After some user testing and gathered feedback, I proceeded with option B.

Further improvements were done for this version, according to the feedback received from the current version that we have in the app.
The changes  are:

* Moved the 'Create Offer' button to the main window header to enhance its visibility and make it more related to the specific selected market.

* Relocated the crypto-fiat pair logo and market title to the main window header, this makes the whole offer book layout more coherent.

* Added the market's crypto-fiat rate to the header for easier reference.

* Renamed the column for fiat markets to 'Markets' for clarity.

* Moved the search function to the sub-header, making it consistent with the Markets column.

* Applied a dimming effect to the fiat logos when they are not selected, reducing the vibrancy of the logo colors.

In chat messages:

* Changed the color of the bubbles: the peer message bubble to a darker grey and my message bubble to a dark desaturated green. This colour differentiation improves user experience, even when message bubbles are positioned on opposite sides.

* Relocated the 'Reputation stars' to be nearer and alongside the user avatar for better association.

* Reduced the 'Take Offer' button's size and moved it closer to the offer text details.

* Decreased the font size and overall size of the message bubble for a more compact and readable chat interface.

<br>

Here is the proposed revision:
![offerbook-revision](https://github.com/bisq-network/bisq2/assets/145597137/e0570466-4169-4604-947d-3d747b755c20)

<br>
@HenrikJannsen what do you think?

HenrikJannsen:
Looks good! Many good ideas implemented!

I am not so sure with the dark green color tone. I would prefer to stick with a grey tone. But not a too strong opinion about that.

Maybe we can drop the hamburger menu right to the create offer button. Opening the channel info could be done by clicking the market icon/headline. The trade guide could be added to the toolbox below or to the tab (right aligned as icon - not sure if that is good...).

axpoems:
@HenrikJannsen  Thanks!

> I am not so sure with the dark green color tone. I would prefer to stick with a grey tone. But not a too strong opinion about that.

[...]

PR #1731, opened 2024-03-01
Revisit styles of Bisq Easy offer messages

axpoems:
Towards #1718

* Relocate `reputation` to be nearer and alongside the user avatar.
* Reduce size of `takeOfferButton` size and moved it closer to the offer text details.
* Decrease font size of the message bubble and increase line spacing.
* Change colour of the my message bubble to a dark desaturated green. 

image


HenrikJannsen (commented):
utACK

HenrikJannsen (approved):
utACK

PR #1732, opened 2024-03-02
Add additional filter in Bisq Easy Offerbook to filter offers by peer reputation

axpoems:
Towards #1572.

* Add filter "by peer reputation".
    image

* Rename "offers filter" to `offerDirectionOrOwnerFilter`, reorder its items and change strings:
    For more clarity, "Buy Bitcoin" filter will show "sell offers" and "Sell Bitcoin" will show "buy offers" (to follow the perspective from the user pov).
    image

* Rename the remaining `robohash` to `cathash`.
* Fix dropdown menu items *sometimes* missing to be resized on first run.
    For this, a `ChangeListener` is wrapped inside a `WakeChangeListener` since the latter applied directly was sometimes missing events.

PR #1734, opened 2024-03-03
Update offerbook headers according to new revision

axpoems:
Based on #1732
Towards #1718

* Update market channel selector and offerbook titles.
* Move create offer button into the offerbook header.
* Move search box to offerbook subheader.

image

Will follow-up with the remaining changes.


HenrikJannsen (approved):
utACK

PR #1738, opened 2024-03-04
Update filtered markets on new offers

DrColver:
To immediately show markets with new offers that previously had no offers the filtered markets need to be updated when new offers appear.

The most obvious use case is to allow a user that creates the first offer for a market to see the market pop up in the filtered list.


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/BisqEasyOfferbookController.java
```diff
@@ -133,6 +134,8 @@ public void onActivate() {
             });
         });
 
+        model.getMarketChannelItems().addListener((ListChangeListener<? super MarketChannelItem>) c -> updateFilteredMarketChannelItems());
```
axpoems: Can you wrap this listener inside a `WeakReference`?

axpoems (commented):
ACK

Good catch! Added one comment.

axpoems (approved):
ACK

DrColver:
@alvasw conflict resolved

PR #1741, opened 2024-03-05
Apply dimming effect to fiat logos and market titles

axpoems:
Towards #1718

* Apply dimming effect to fiat logos and market titles.
* Highlight selected items in sort and filter dropdown menus (including reputation and offer types).
* Remove unused code.

image


HenrikJannsen (approved):
utACK

PR #1748, opened 2024-03-07
Improve mailbox message handling in trade process

djing-chan:


HenrikJannsen (approved):
utACK

network/network/src/main/java/bisq/network/p2p/services/confidential/resend/ResendMessageService.java
```diff
@@ -0,0 +1,177 @@
+package bisq.network.p2p.services.confidential.resend;
```
axpoems, line 18: Nit. A few added files are missing the license.

djing-chan: Ups, thanks. Fixed.

djing-chan:
> ACK
> 
> Nice changes!
> 
> From the UX perspective, I don't see the need for the user to be responsible (or partly responsible) for retrying the messages - especially in the case of `ProtocolLogMessage`s. What about having a cron job that retries these messages? At least for protocol type of message.

I am a bit worried that we spam the mailbox message system by automatically resending them. We do that in that PR at restart which is problematic. I will follow up with an addition to that PR here or in a new PR soon.

I think a viable strategy is to add a timer after resend and check if the state has changed to addedToMailBox. If so we can set a flag to be not resent automatically. Otherwise we can leave it to the UI how to handle it (e.g. show user a resend button with maybe a popup with some context...).  But not clearly decided yet whats the best. Atm I think ensuring that messsages are not lost is highest prio, but not spamming the mailbox system is important as well.

Issue #1739, opened 2024-03-04
Fix TradeWizard for markets with no price feed


DrColver:
TradeWizardReviewController throws an error if market price couldn't be found.

HenrikJannsen:
We support now 2 additional sources for market prices. Oracle nodes distribute prices and users persist prices. The newest price will be used. So the cases where no price is available should hardly happen anymore. Also we filter markets which do not have a market price.

DrColver:
I try to create offers on markets such as EGP or BYN and it fails. I think it's due to lack of market price but possibly it's something else. It would be better to fail gracefully rather than with a stack trace popup even it's won't happen often going forward.

I see you have an oracle node that I'm connected to. Are all the price feeds running as expected at the moment?

HenrikJannsen:
Can you debug into it to see what causes it? If its the missing market price then we should find out why the market is displayed. Maybe it was selected before but not in the list anymore. The changes have been recently added so might be just a glitch from an update (e.g. new users cannot select a market if there is no price feed)

DrColver:
It fails on PriceUtil.java:116

The list of prices in `marketPriceByCurrencyMap` is 49 entries long with all the major currencies. This is also the list shown in the price feed drop down in the top right corner.

When creating offers there is a list of all currencies worldwide it seems. The currency choices are set in TradeWizardMarketController.java:90 with no check for a current price feed that I can see.

HenrikJannsen:
Ah ok. Yes there the filter is missing. Would you like to implement it? Its same behaviour as in the markets list filter.

DrColver:
Sure, I'll do that.

Is there a plan to get the price feeds for those other missing markets? Maybe we should allow creating offers for markets with no price feed using either a fix exchange rate to USD, setting a fix price, or allowing the price to be negotiated in the trade.

HenrikJannsen:
I think we can be happy to get 10% of existing markets with price feeds used by traders... Would postpone such extensions for later once we feel there is realistic chances that those niche markets find users. Adding those to the price feed server would be the easiest way.

PR #1750, opened 2024-03-07
Allow only currencies with price feed when creating offers

DrColver:
Fix https://github.com/bisq-network/bisq2/issues/1739

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/trade_wizard/market/TradeWizardMarketController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/trade_wizard/market/TradeWizardMarketController.java
@@ -17,6 +17,7 @@
 
 package bisq.desktop.main.content.bisq_easy.trade_wizard.market;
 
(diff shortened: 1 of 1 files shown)


axpoems (commented):
I still get the error with stack trace when creating an offer:

image

Steps to repro: Select market `THB` > Create offer > Sell Btc > ... and it fails upon arriving in section `Price` inside the wizard. I think we also need to apply this in `TradeWizardMarketController`, see stack trace:

image

DrColver:
I couldn't reproduce. Sounds like the price feed might be having troubles.

There is a lot of places where a stack trace is shown if there is an issue with the price. I think it would be better to allow creation of offers even when the price feed doesn't work and handle the price issue separately.

axpoems:
> I couldn't reproduce. Sounds like the price feed might be having troubles.
> 
> There is a lot of places where a stack trace is shown if there is an issue with the price. I think it would be better to allow creation of offers even when the price feed doesn't work and handle the price issue separately.

I looked into this more. It fails every time you create an offer and then switch markets and attempt to create a new one there.
The issue was that the price quote was not being reset property. I added the fix for this in #1755.

axpoems (approved):
ACK

PR #1751, opened 2024-03-07
Add display hint for applied market filers

### LINKED ISSUE #1740: Display a hint about filters being applied to offerbook channel list
It would be good to provide a hint in the UX that there are some filters applied to the markets. If the 'With offers' filter is applied and a market without offers is selected, the channel selection list appears empty and it's not straightforward to understand why + it's impossible to switch to a channel without offers unless the user finds out that they need to switch to the 'All' filter.

The solution could be as simple as changing the color of the filter to something that signals that there is a filter applied (e.g., yellow). The 'With offers' filter is currently the default filter - we should probably switch it to 'All' as channels will be ranked by 'Most offers' by default anyways. This way a new user will be able to easily see all the markets available at first sight and we reduce the risk of abandonment because of misleading new users into thinking that the market in which they're interested is not available.

Screenshot 2024-03-05 at 10 03 49
Screenshot 2024-03-05 at 10 04 02

axpoems:
* Add display hint for applied market filers.
    Resolves #1740
* Set show `All` markets by default.
* Adjust the dimming effect, remove  on market titles and reduce it for market logos.

image


HenrikJannsen (approved):
utACK

namloan:
utACK

Thanks a lot!

PR #1754, opened 2024-03-09
Replace nullable with optional

### LINKED ISSUE #1631: Use Optional instead of Nullable in Account module
See TODOs "use Optional instead of Nullable"

dirka1992:
Closes #1631  

* Updated the methods in the PaymentAccountsModel to return optional where the return value could be null.
* Updated the BankAccountPayload by replacing the Nullable annotations with Optional.
* Updated all the classes affected by the changes of adding Optional.

### DIFF
--- a/account/src/main/java/bisq/account/AccountService.java
+++ b/account/src/main/java/bisq/account/AccountService.java
@@ -33,19 +33,17 @@ import bisq.persistence.PersistenceService;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
-import javax.annotation.Nullable;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 
 @Slf4j
+@Getter
 public class AccountService implements PersistenceClient<AccountStore>, Service {
-    @Getter
+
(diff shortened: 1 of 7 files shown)


account/src/main/java/bisq/account/AccountService.java
```diff
@@ -33,19 +33,17 @@
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
-import javax.annotation.Nullable;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 
 @Slf4j
+@Getter
```
dirka1992, line 41: Minor additional change. All instance members were annotated with a Getter, rather moved it to a class level.

axpoems:
@dirka1992 - A few things: You need to sign your commits, also try to follow the Bisq guidelines when making commits and use the style conventions. You can read more about this in the [contributing guide](https://github.com/bisq-network/bisq2/blob/main/docs/dev/contributing.md) as well as the [developer guidelines](https://github.com/bisq-network/bisq2/blob/main/docs/dev/dev-guide.md).

PR #1755, opened 2024-03-09
Fix `TradeWizardPriceController` throwing exception 

axpoems:
Based on #1750

Price controller was throwing an exception upon percentage validation:
https://github.com/bisq-network/bisq2/blob/5ce28113b5710cad4c01bfd36bbfefe7ef0ea229/offer/src/main/java/bisq/offer/price/spec/FloatPriceSpec.java#L48

It worked well while creating offers in one market, but when switching
markets and creating a new offer it failed since it was using the
previous quote price. Thus, this resets the missing fields every time
we use the `TradeWizard`.


djing-chan (approved):
utACK

PR #1777, opened 2024-03-11
Improve user avatars design

### LINKED ISSUE #1765: Improve profile icons

axpoems:
* Increase background diameter.
* Create and add new background overlay patterns.
* Add body layer with two size options.
* Add chest and overlay layers with two size options.
* Correct design bug in ears to distinguish from a background with the same colour.
* Add full colour spectrum for the small and bigger ears.
* Add four new face shapes.
* Create new face overlays and fix existing ones according to new face shapes.
* Adapt CatHash algorithm to the new design scheme.

Resolves #1765

image


HenrikJannsen (approved):
ACK

HenrikJannsen:
Congrats! Looks great!

PR #1781, opened 2024-03-12
Refactor Cathash encoding

axpoems:
Based on #1777 

* Remove unnecessary ordered map.
* Use `Optional` instead of checking for null.


HenrikJannsen (approved):
utACK

PR #1797, opened 2024-03-15
Fix backward compatibility issue with avatar version

HenrikJannsen:
Use integer for version instead of a string and set initial version to 0. This avoids that it breaks backward compatibility as 0 is default value for integers and proto serialization does not change if that value is set or not.

FYI: 
"if a scalar message field is set to its default, the value will not be serialized on the wire" - https://protobuf.dev/programming-guides/proto3/


namloan:
utACK

Issue #1807, opened 2024-03-15
User profile selection not working after profile was added in top left drowdown


HenrikJannsen:

namloan:
Is this the dropdown in User Options -> User Profile? can't see the issue

axpoems:
I am able to repro. I think @HenrikJannsen is referring to the top right dropdown (i.e. the global user profile selection).
Steps:
1. Delete data dir and start up a new user.
2. Go to user options > Create new profile, and create yet another profile for that session.
Result: The global top right profile selection dropdown is static and does not show dropdown when clicking.

It works if you close the app and start again. And it also works fine with all the subsequent users that you create, but not with that first additional one that you create.

namloan:
Thanks, could reproduce now. It only seems to happen in the very first start of the app when there's no data dir.

>  but not with that first additional one that you create.

Subsequent ones didn't work for me either

HenrikJannsen:
I guess its just a incorrect initialization for a observer....

PR #1811, opened 2024-03-16
Fix user profile selection not working after adding new profile

axpoems:
* Add missing listener to `userProfiles`
* Refactors in `UserProfileSelection`

Fix #1807

PR #1812, opened 2024-03-16
Fix user profile selection bug

HenrikJannsen:
Fixes https://github.com/bisq-network/bisq2/issues/1807


HenrikJannsen:
Close as it implemented already by @axpoems in https://github.com/bisq-network/bisq2/pull/1811 in a more complete manner (moving logic from view to controller).

PR #1814, opened 2024-03-16
Fix inconsistent market logo effect

### LINKED ISSUE #1802: Selected market does not match with lit market icon
image

axpoems:
* Fix market logo effect not matching selected row.
  Make `updateItem()` from markets table decide when to lit the market logo depending on whether that row is selected or not. This is also more efficient than updating each `MarketChannelItem` every time a different channel is selected.
  Resolves #1802

* Minor refactors.


HenrikJannsen (approved):
utACK

PR #1816, opened 2024-03-16
Keep market item selected while and after using search

axpoems:
Based on #1814 

Add listener to search predicate to always keep selection visible in table view.
Prior do this, both during search and after search, the selection marker disappears.


HenrikJannsen (approved):
utACK

PR #1817, opened 2024-03-16
Add favourites system to offerbook markets

axpoems:
Based on #1816 

* Add capability to store favourite markets in user settings.
* Use market favourites in offerbook as a way to pin markets.
* Introduce new icons dedicated to favourites.

image

Will follow-up with remaining improvements on this.


HenrikJannsen (approved):
utACK

Issue #1821, opened 2024-03-17
Use "All chat messages" as global default for all chat rooms


HenrikJannsen:
It appears like a bug that no notifications are shown for normal chat rooms as default is "Only if mentioned".

PR #1825, opened 2024-03-17
Use ChatNotificationType.ALL as default

HenrikJannsen:
Implements https://github.com/bisq-network/bisq2/issues/1821


axpoems (approved):
ACK

Issue #1784, opened 2024-03-13
Fix stuck notifications in My Open Trades


axpoems:
![image](https://github.com/bisq-network/bisq2/assets/145597137/32581da4-d6d8-4b8d-87bc-1067b59c64e0)

This happens when a user sends a message in the using the private chat inside My Open Trades. The notification is not marked as consumed and can't be dismissed.

HenrikJannsen:
Related to https://github.com/bisq-network/bisq2/issues/1745 and https://github.com/bisq-network/bisq2/issues/1749

HenrikJannsen:
Probably same issue as https://github.com/bisq-network/bisq2/issues/1749#issuecomment-2002333908

HenrikJannsen:
Actually after a new dev setup I cannot reproduce the bugs anymore with the recent fix.

Issue #1749, opened 2024-03-07
Clear notifications does not clear all notifications


djing-chan:
Sometimes we get incorrect notifications. Clicking the clear all notifications button in the preferences does not clean up all pending notifications.

HenrikJannsen:
It clears all notification in the left nav, but not inside of tabs when running on https://github.com/bisq-network/bisq2/pull/1826

HenrikJannsen:
Actually after a new dev setup I cannot reproduce the bugs anymore with the recent fix.

Issue #1745, opened 2024-03-06
Incorrect notification after trade got closed


djing-chan:
After closing a completed trade on both traders I got that notification on one user.
![Screenshot from 2024-03-06 22-06-07](https://github.com/bisq-network/bisq2/assets/146422224/9ed60ab8-8ed6-44b7-bb49-caa256f3d21b)

HenrikJannsen:
Seems https://github.com/bisq-network/bisq2/pull/1826 has fixed that bug as well. Could not reproduce it anymore. If still reproducable please re-open.

PR #1826, opened 2024-03-17
Fix bug with not removing notification badge when consumed

HenrikJannsen:
Might fix partially https://github.com/bisq-network/bisq2/issues/1745, https://github.com/bisq-network/bisq2/issues/1749, https://github.com/bisq-network/bisq2/issues/1784

Not tested with all the use cases yet, but was a result when implementing https://github.com/bisq-network/bisq2/pull/1825 and found at testing it that the notifications did not work as expected. Fixed at least the bug for that use case (messages from generic public channels).


axpoems (approved):
ACK

PR #1831, opened 2024-03-17
Update Bisq Easy texts according to new protocol

### LINKED ISSUE #1743: Wrong text at cancel trade
If the buyer cancels the trade after sent the BTC address, they get the wrong popup text displayed.
Screenshot from 2024-03-06 21-41-00

Also trade guide text is from old protocol and needs update with the case that both traders can send their data first.

Screenshot from 2024-03-06 21-42-43

Open question if we consider the case when the buyer sent the BTC address as critical as if seller sent the Fiat account data. As BTC address leaks less privacy we could handle it differently. Though harvesting BTC addresses should be prevented as well and added complexity to distinguish the cases suggest to handle them the same way.

axpoems:
Resolves #1743

* Update trade rules:
image

* Update reject trade warning message (now generic for both parties):
image

* Update cancel trade warning message (below for the case of the seller):
image


HenrikJannsen (approved):
utACK

PR #1832, opened 2024-03-17
Add feedback message when user takes an offer

HenrikJannsen:
Fixes https://github.com/bisq-network/bisq2/issues/1789

PR #1846, opened 2024-03-17
Review and update take-offer status feedback overlay

axpoems:
Based on #1832

* Use as waiting icon the Bisq Easy logo.
* Show take-offer state for at least 8 seconds (less than that the user is not able to read and looks confusing).
* Unify overlays so that we don't see interruption between them.

take-offer-fb

I will follow-up with some improvements.


HenrikJannsen (approved):
utACK

PR #1851, opened 2024-03-18
Fix NullPointer exception caused by incorrect call of onViewDetached

HenrikJannsen:
Bug reported by user:

e4

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/common/view/View.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/common/view/View.java
@@ -34,6 +34,7 @@ public abstract class View<R extends Region, M extends Model, C extends Controll
     protected final C controller;
     private final ChangeListener<Scene> sceneChangeListener;
     private ChangeListener<Window> windowChangeListener;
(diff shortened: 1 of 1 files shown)


trille501:
Skærmbillede på 2024-03-29 18-44-49 ny
Hi. I still get this error in v2.0.2 (pre-release) when I try to take an offer.

HenrikJannsen:
Thanks for reporting. Addressed in https://github.com/bisq-network/bisq2/issues/1955

HenrikJannsen:
> Skærmbillede på 2024-03-29 18-44-49 ny Hi. I still get this error in v2.0.2 (pre-release) when I try to take an offer.

Which offer is it you want to take? I cannot reproduce it. Do you get it each time when taking that offer or only at certain cases?

HenrikJannsen:
Ah I found the issue. Its that the quote currency name in Market is "euro" instread of "Euro" because maker used spanish locale. We need to exclude that display strings from EqualsAndHashCode

HenrikJannsen:
Fixed here: https://github.com/bisq-network/bisq2/pull/1960

PR #1868, opened 2024-03-18
Improve style of matching offers in trade wizard

### LINKED ISSUE #1833: Improve style of matching offers in trade wizard
When not selected the background is same as row bg.
Selection style also does not look great.
 
Screenshot 2024-03-17 at 17 43 24
Screenshot 2024-03-17 at 17 43 15

@axpoems Could you have a look into that?

axpoems:
Resolves #1833

* Change matching offers table styles to match the standard style of tables used in the app.
* Add selection marker column.
* Remove unnecessary and confusing column used to select offer.
* Minor refactors.

table-imp


HenrikJannsen (approved):
utACK

PR #1872, opened 2024-03-19
Add "Only favourites" filter and other improvements

axpoems:
* Add "Only favourites" filter.
* Set height correctly for favourite markets and improve paddings.
* Show selected market after applying a new filter to market channel items list.
* Hide horizontal scrollbar in private chats list.

image


HenrikJannsen (approved):
utACK

Issue #1798, opened 2024-03-15
Fix startup issues of Tor


HenrikJannsen:
Tor setup sometimes hangs specially at first time using Bisq2.

lucasbalieiro:
Hi Guys,

it is my first time running the application and I think i get the issue you are presenting here.

Looking at the logs, I get this:


`mar.-17 15:50:59.212 [NetworkService.network-IO-pool-0] ERROR b.d.DesktopApplicationService: Initializing applicationService failed java.util.concurrent.CompletionException: bisq.tor.controller.exceptions.TorBootstrapFailedException: Tor bootstrap timout triggered.
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1770)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: bisq.tor.controller.exceptions.TorBootstrapFailedException: Tor bootstrap timout triggered.
	at bisq.tor.controller.NativeTorController.waitUntilBootstrapped(NativeTorController.java:146)
	at bisq.tor.TorService.initialize(TorService.java:99)
	at bisq.network.p2p.node.transport.TorTransportService.initialize(TorTransportService.java:63)
	at bisq.network.p2p.ServiceNode.getInitializedDefaultNode(ServiceNode.java:229)
	at bisq.network.p2p.ServiceNodesByTransport.lambda$getInitializedDefaultNodeByTransport$1(ServiceNodesByTransport.java:133)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
	... 3 common frames omitted`
	
 Another thing that I notice, every time that I run the application it goes a bit further in the Bootstrap %.

first stoped with 3%, then 6%, 12.50%, 12.75%

Hope that it helps.

PR #1914, opened 2024-03-23
tor: Implement ControlPortFileParser

alvasw:
Presently, we parse Tor's log file to detect when the control port is ready. This is unstable and could break with any Tor update. A better approach is to use Tor's 'ControlPortWriteToFile' config option and parse the control port from that file.

Ref: #1798

### DIFF
--- a/network/tor/tor/src/main/java/bisq/tor/process/ControlPortFileParseFailureException.java
+++ b/network/tor/tor/src/main/java/bisq/tor/process/ControlPortFileParseFailureException.java
@@ -1,0 +1,28 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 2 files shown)


adrelanos:
> A better approach is to use Tor's 'ControlPortWriteToFile' config option and parse the control port from that file.

Did you implement Tor's 'ControlPortWriteToFile' config option yet? I am not saying it's the right solution (https://github.com/bisq-network/bisq2/issues/1924) but I cannot find it in the source code.

adrelanos:
[network/tor/tor/src/main/java/bisq/tor/process/NativeTorProcess.java](https://github.com/bisq-network/bisq2/blob/main/network/tor/tor/src/main/java/bisq/tor/process/NativeTorProcess.java#L82) still contains

>                        Set.of("[notice] Opened Control listener connection (ready) on "),

So parsing of `~/.local/share/Bisq2/tor/debug.log` is still being done.

PR #1916, opened 2024-03-23
Add custom style to take offer button

### LINKED ISSUE #1841: Better offerbook Button instead "Take Offer"
Replace "Take Offer" button (on the Offerbook) to a variable "Buy Bitcoin" or "Sell Bitcoin", this variable button would be per user (taker) perspective.

axpoems:
Resolves #1841

* Use Buy/Sell Bitcoin instead of "Take offer".
* Use different colours for each action.

image


HenrikJannsen:
Hm... I would expect the green color for Buy and a more classic red for sell as its common on most exchanges. I guess we should stick here to the common style used in financial apps. A slight adoption from the colors seems ok to me but these colors are too different to me.

HenrikJannsen:
What about those colors:

    -bisq2-buy: #007800;
    -bisq2-sell: #780000;

Its pure green and red but darkened.

Screenshot 2024-03-24 at 15 24 16

 The green is intentionally not the Bisq green tone to differentiate from the other buttons. Could be even more difference (e.g. . -bisq2-buy: #005a00;)

Screenshot 2024-03-24 at 15 26 16

HenrikJannsen:
I think also we should rename "I want to" to "I offer to" to make clear that its the makers side. By mixing BUY/SELL in the message we get back the old known problem that the BUY/SELL is mirrored between maker and taker and that confused ppl since ever in Bisq1.

axpoems:
@HenrikJannsen - I have changed the message strings to unify all the UI to be from the user perspective, but respecting the "MyMessage" so that everything keeps coherent functionality wise on that side (creating offer, trade process, etc).

I also updated the colours and added a little bit more padding.

image

HenrikJannsen:
I am not sure about the new wording. It confuses me.
What about "Offer Bitcoin for ..." instead of "Buy me ..."?
And
"Offer xxx EUR for Bitcoin" on the buy side?

axpoems:
> I am not sure about the new wording. It confuses me.
> What about "Offer Bitcoin for ..." instead of "Buy me ..."?
> And
> "Offer xxx EUR for Bitcoin" on the buy side?

I was thinking also (world view /user view):

* "BUY Bitcoin for XXX EUR" / "I want to SELL Bitcoin for XXX EUR"
* "SELL me Bitcoin for XXX EUR / I want to BUY Bitcoin for XXX EUR"

HenrikJannsen:
Maybe I am biased as gotten used to the current style but the "BUY Bitcoin for XXX EUR" still is confusing to me.
If we avoid the buy/sell in the world view I think we avoid the confusion. the take-offer button can be then the acting user users view: Buy Bitcoin/ Sell Bitcoin. And the offer could be:
Offer Euro for Bitcoin / Offer Bitcoin for EUR

ItoroD:
@HenrikJannsen I agree. I think buy and sell should be removed from requesters side

> request - I want bitcoin for $40

> offer - Buy bitcoin for $40

something like this

axpoems:
> Maybe I am biased as gotten used to the current style but the "BUY Bitcoin for XXX EUR" still is confusing to me.

I believe you may be biased, as anyone who has thoroughly delved into the problem might be. That's why is still confusing. There's no right or wrong, rather, it's about adopting and maintaining a convention. In this case the user's perspective. You'll notice the use of 'buy' in the text, on the button, and in the filters is now consistently aligned. 'Offer Euro for Bitcoin / Offer Bitcoin for EUR' doesn't seem to enhance clarity and doesn't align with the UI message concept. I understand that 'Buy me...' may sound odd, which is why I amended it to simply 'Buy Bitcoin...'

image

HenrikJannsen:
Sorry but I still have issues with that and I fear we get very language specific issues in translations.
The "Sell me BTC..." is clear as it clearly indicates with the me that its the peers intention that he want to buy (sell me = i buy).
With the "Buy BTC..." though that is gone and I would not be sure if the peer says (I want to) buy BTC or if its is "Buy BTC for ... from ME". with the latter we would be similar as with the Sell me, but as the me is at the end I think it does not read that clearly. 

I doubt that this could be translated clearly in all other languages, specially non Europeans.

My intention with using "offer" was to get rid of the redundant BUY/SELL as it is in the button anyway and putting it into the peers message comes with the problem that we need language tweaks to mirror it (I want to BUY = SELL me).

Maybe we should reduce even more the text and relax the original intention that the message is pure text and move more to a box with data represented in a good readable way. As we change the peers message anyway and now also use the viewers language it is anyway broken that the offer represents a text from the peer - which he has never written as its derived from this offer data.

The Bitcoin part is also not needed as long we do not support other currencies.
Also the currency is redundant as the user has chosen the market and only see EUR or USD currencies.

What about:
Trade amount: 123-234 EUR
Payment method: SEPA
[SELL BITCOIN]

Trade amount: 123-234 EUR
Payment method: SEPA
Price: 5% over marketprice
[BUY BITCOIN]

We could also use "Offered amount" if that adds clarity.

HenrikJannsen:
Ah just saw https://github.com/bisq-network/bisq2/issues/1929 ;-) Wanted to add that we could make all more graphical by using the logo instead of text ;-) but left it out here to not mix up too much...

cbeams:
Agree we should eliminate "Take offer" language, especially in Bisq Easy.

Disagree that differentiating between buy/sell actions with color makes a positive difference; think it may even make a negative difference; don't feel strongly about this aspect either way though.


Here's my suggestion about how to word the button text.


In the following scenario, I am Bob and I am looking at an offer posted by Alice that reads:

> I want to SELL Bitcoin for 264 GBP.

The button that I, Bob, see inside of Alice's offer bubble reads

> [Buy bitcoin from this user]


So:

This user says they want to sell bitcoin and I have a button in front of me that literally says "Buy bitcoin from this user"

That's about as unambiguous as I can imagine it getting.


Flip it around, now, such that Alice's offer reads:

> I want to BUY Bitcoin for 264 GBP

The button, that I, Bob, would see inside of Alice's buy offer bubble would read:

> [Sell bitcoin to this user]

So again:

This user says they want to buy bitcoin and I have a button that literally says "Sell bitcoin to this user".

How could it be as clear or clearer than this with fewer words in the button? I don't know.


Much of grammar boils down to ensuring we have effective ways to tell each other who did—or is going to do—what to whom. The problem with simple "Buy bitcoin" and "Sell bitcoin" buttons is that they don't contain enough grammatical signage to unambiguously indicate who is going to do what to whose bitcoin. And that's a problem color can't solve. The solution is to use more words until the action that's going to be taken by the user becomes utterly intuitive and automatically, unthinkingly, unquestionably clear.

> Buy bitcoin from this user

and 

> Sell bitcoin to this user

are my suggestions about how to achieve that goal.


An added bonus of this approach is the way in which seeing "to this user" and "from this user" over and over again just drives home the p2p nature of what is going on here. You're not selling your bitcoin to a faceless trading engine. You're selling your bitcoin to _this user_. You're buying bitcoin from _this user_. Feels on point with the more human and social ethos we're trying to foster in Bisq 2 / Bisq Easy.

HenrikJannsen:
I think that the colors still make it easier to spot the buy and sell offers. Also maybe the "from this user"/"to this user" could be added in the next line so that the "Buy bitcoin" gets the more weight? Not sure if a line break is a good graphical solution here but to separate the action from the context might be good to give the action more weight.

DrColver:
An interesting discussion. Well needed, it's hard to understand what's going on at the moment.

I think having different colors would be a good differentiation. You press red button, you end up with fewer bitcoin, bad! You press green button, get more bitcoin, good!

The text suggestion "Sell bitcoin to Alice" is great. It makes it utterly clear what I'm doing and to whom. I think the text in the chat bubble isn't as important as long as the action button explains what I'm doing.

DrColver:
As an example from the table view. The color of the take offer button voids the necessity of creator intention. It's clear which offer is which from the buttons. The chat view would look a bit strange without an intention though, but the button color would still distinguish the offer direction.

image

axpoems:
Thanks for all your inputs -I've taken all of them into account.

This is the final result:
image

And this is how the notification looks like now:
image

HenrikJannsen:
Have you tried how it looks with some longer user names?
As the height of the box is higher now, we could maybe put the reputation stars below the user icon to use the spcae more efficiently.

axpoems:
> Have you tried how it looks with some longer user names?

I did check the longest nicknames in prod atm and it works ok. Nonetheless, when testing, I found that info (the user nickname in the title) to be so relevant that I think we should remove it from the offer message and append it later to the title with custom style and hyperlink to user profile. So I will work into adding a max. length and those improvements in anther PR.

The issue with long strings also happens with payment methods and supported languages (the latter does not even have a limit).
I am continuing to work on redesigning the offer message to change payment methods and languages to be more graphical and this will change the layout. So we are walking in this direction, still fine tuning the offer message layout.

> As the height of the box is higher now, we could maybe put the reputation stars below the user icon to use the space more efficiently.

Regarding the reputation, I've maintained it here intentionally to give visibility of the Bisq reputation system at this stage.
My opinion is that the stars should move below the icon everywhere (so when you see an icon, you always see the stars below) - being it stars, or any other representation of the reputation that we choose.
But I leave this change for when we apply this change everywhere.

HenrikJannsen (approved):
utACK

PR #1917, opened 2024-03-23
Improve reputation import account age instructions

### LINKED ISSUE #1834: Enhance instructions for importing account age/signed witness
I saw many reports on the app chat that users fail to import account age/signed account age. It is clearly described how to do it, but ppl are not following it. 
Any ideas how we can make the communication for that better?

@axpoems @pedromvpg @cbeams @pazza83 Any ideas?

axpoems:
Resolves #1834

* Improve instructions flow.
* Update request authorization to get the value from the model instead of the clipboard.

image

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/reputation/accountAge/tab3/AccountAgeTab3Controller.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/reputation/accountAge/tab3/AccountAgeTab3Controller.java
@@ -100,7 +100,6 @@ public class AccountAgeTab3Controller implements Controller {
 
     public void onRequestAuthorization() {
         String signedMessage = String.valueOf(model.getSignedMessage());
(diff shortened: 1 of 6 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/reputation/accountAge/tab3/AccountAgeTab3Controller.java
```diff
@@ -100,7 +100,6 @@ void onClose() {
 
     public void onRequestAuthorization() {
         String signedMessage = String.valueOf(model.getSignedMessage());
-        // Question: do we still need the prefix?
```
HenrikJannsen: As long users use the current Bisq1 version the prefix is included in the copied message. Best to support both the current version and a future update in Bisq1 where the prefix is removed. Applying here a string replace for the prefix with empty string should do it.

HenrikJannsen (approved):
utACK

PR #1930, opened 2024-03-24
Updates in icons

axpoems:
* Use `MaterialTextArea` instead of field in import account age instructions.
* Add new chevron icons (plus missing in white chevron) for the dropdown menus.
* Correct small ellipsis icons.
* Improve help icons.
* Add mini close icons for the display hints to avoid rescaling.


HenrikJannsen (approved):
utACK

PR #1937, opened 2024-03-26
Fix alert banner paddings

axpoems:
* Move alert notification management to a separate service.
* Fix paddings when showing an alert banner (both in the general case and when colliding with another Bisq Easy notification).

not-paddings


axpoems:
> If both orange based panes are displayed the small difference in both looks weird. Maybe better use the same color. With the icon and headline it distinguishes sufficiently from notifications.
> Maybe also make text colors same?

This PR fixes the paddings. The colours can be addressed in another PR.

HenrikJannsen (approved):
utACK

Issue #1955, opened 2024-03-30
Fix NullPointer exception at TakeOfferReviewController


HenrikJannsen:
See https://github.com/bisq-network/bisq2/pull/1851#issuecomment-2027983523

HenrikJannsen:
Ah I found the issue. Its that the quote currency name in Market is "euro" instread of "Euro" because maker used spanish locale. We need to exclude that display strings from EqualsAndHashCode

PR #1960, opened 2024-03-31
Exclude currency names from EqualsAndHashCode in Market

HenrikJannsen:
Resolves https://github.com/bisq-network/bisq2/issues/1955


namloan:
utACK

PR #1963, opened 2024-03-31
Only send system notification for real time messages

### LINKED ISSUE #1546: Improve notification handling on start up for messages in inbox
When starting up the application, the user gets a notification for every single message waiting to be delivered in the inbox, which in turn, needs to be dismissed/wait until dismiss automatically. If the are many messages (more than 4-5?), this does not give a good experience. We should find a better solution for this use case. We could apply some heuristics when delivering these notifications. One possibility could be to notify per channel, e.g. "new notifications in channel X".

axpoems:
While online, for each message we (1) show the badge and (2) send
a system notification. The latter should be avoided when a message was
received before the user went offline. This prevents a flood of system
notifications at start-up.

Resolves #1546


HenrikJannsen (approved):
ACK

PR #1964, opened 2024-03-31
Refactor chat message list view

HenrikJannsen:


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/list_view/ChatMessagesListModel.java
```diff
@@ -0,0 +1,52 @@
+package bisq.desktop.main.content.chat.list_view;
+
+import bisq.chat.ChatChannel;
+import bisq.chat.ChatChannelDomain;
+import bisq.chat.ChatMessage;
+import bisq.user.identity.UserIdentityService;
+import javafx.beans.property.*;
```
namloan, line 7: I know we're just using IntelliJ's default style settings - however would we like to enable single class imports? I always have it enabled.

namloan (approved):

namloan:
ACK

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/list_view/ChatMessagesListModel.java
djing-chan: Yes would be better, we just need to ensure that all devs have the same settings. If you know which config file to add to the .idea folder and commit to GH please do so.

PR #1966, opened 2024-03-31
Improve scroll to bottom behaviour

HenrikJannsen:
Based on https://github.com/bisq-network/bisq2/pull/1965

Could not solve the main problem yet, that the list view gets re-layouted at mouse over and moves the scrollbar from bottom position a bit up. 
It seems that the height of the list view gets recalculated at mouse over and thus change the scroll position.


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/list/ChatMessagesListController.java
```diff
@@ -575,11 +576,7 @@ private <M extends ChatMessage, C extends ChatChannel<M>> Pin bindChatMessages(C
         return channel.getChatMessages().addObserver(new CollectionObserver<>() {
             @Override
             public void add(M chatMessage) {
-                // TODO (low prio) Delaying to the next render frame can cause duplicated items in case we get the channel
-                //  change called 2 times in short interval (should be avoid as well).
-                // @namloan Could you re-test the performance issues with testing if using UIThread.run makes a difference?
-                // There have been many changes in the meantime, so maybe the performance issue was fixed by other changes.
-                UIThread.runOnNextRenderFrame(() -> {
+                UIThread.run(() -> {
```
namloan, line 600: This may have some performance impact https://github.com/bisq-network/bisq2/pull/1310

HenrikJannsen: I think that has been resolved by various changes.
1. CollectionChangeMapper: addAll and setAll not using add in iteration anymore
2. We added a lookup map to return early if the item was already added
3. It would make a difference only when its called from the UI thread. That would only happen at onActivate but there we set all items with setAll before we register the observer. Thus the observer call would return early from the map lookup.
4. We fixed and refactored some cell renderer code which might have been the source for the performance issue. I remember there have been potential leaks with not removed listeners.
5. I tested with live data and could not observe any difference in performance.


It still would be great if @namloan could run his performance test again to see if theory matches reality.

We can leave the PR open a few days to hear back from @namloan or if anyone else has more insight.

namloan: No need to keep this one open then, performance differences would be mainly noticed with very large amount of messages which we probably won't get in the short term and if there are chances that it's fixed anyways then great. I'll run a test this week.

namloan (approved):
utACK

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/list/ChatMessagesListController.java
djing-chan: Thanks @namloan for the feedback. If you discover that `runOnNextRenderFrame` would make a difference please feel free to add a PR. If possible to add the performance tests as well.

PR #1967, opened 2024-03-31
Only send notification if app is not in focus

HenrikJannsen:
Based on https://github.com/bisq-network/bisq2/pull/1963

@axpoems What do you think to add a check for "app is focussed" to decide to send a notification?
If user is using the app the OS notification is not needed IMO.


axpoems (approved):
ACK

I think we should also dismiss the badge in the tab automatically after 5 seconds (or another solution so that the user sees that there's a change but it's handled automatically). This can be done when we work on the next chat iteration.

PR #1975, opened 2024-04-02
Improve auto-scroll-to-bottom behaviour

### LINKED ISSUE #1823: Improve auto-scroll-to-bottom behaviour
The auto scroll to bottom does not work reliable. Sometimes it is not scrolling to the latest message or part of the message is not visible.

axpoems:
Resolves #1823

Use built-in list's scrollbar to achieve consistent scroll to bottom.


HenrikJannsen:
I tested with existing and that PR version but did not see any difference in behavior.

PR #1976, opened 2024-04-02
Fix alignment issue in create offer wizard

axpoems:
Fix alignment of direction box to work well when using a different language other than English.

Before:
image

After:
image


djing-chan (approved):
utACK

PR #1986, opened 2024-04-04
Add filter to reputation table

HenrikJannsen:
Resolves https://github.com/bisq-network/bisq2/issues/1983

Screenshot 2024-04-04 at 20 22 32
Screenshot 2024-04-04 at 20 22 26

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/DropdownMenu.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/DropdownMenu.java
@@ -24,19 +24,27 @@ import javafx.collections.ObservableList;
 import javafx.geometry.Bounds;
 import javafx.geometry.Insets;
 import javafx.geometry.Pos;
-import javafx.scene.control.*;
+import javafx.scene.control.ContextMenu;
+import javafx.scene.control.Label;
+import javafx.scene.control.MenuItem;
+import javafx.scene.control.Tooltip;
 import javafx.scene.image.ImageView;
 import javafx.scene.layout.HBox;
 import javafx.stage.PopupWindow;
 import javafx.stage.WindowEvent;
+import lombok.Getter;
+
+import javax.annotation.Nullable;
+import java.util.Collection;
 
 public class DropdownMenu extends HBox {
     public static final Double INITIAL_WIDTH = 24.0;
-
+    @Getter
     private final Label label = new Label();
     private final ImageView defaultIcon, activeIcon;
     private final ContextMenu contextMenu = new ContextMenu();
     private ImageView buttonIcon;
+    @Nullable
     private ChangeListener<Number> widthPropertyChangeListener;
     private boolean isFirstRun = false;
 
@@ -82,7 +90,11 @@ public class DropdownMenu extends HBox {
         }
     }
 
-    public void addMenuItems(CustomMenuItem... items) {
+    public void addMenuItems(Collection<? extends MenuItem> items) {
+        contextMenu.getItems().addAll(items);
+    }
+
+    public void addMenuItems(MenuItem... items) {
         contextMenu.getItems().addAll(items);
     }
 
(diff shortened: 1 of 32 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/reputation/list/ReputationListController.java
```diff
@@ -0,0 +1,161 @@
...
+                    .forEach(item -> item.applyReputationScore(userProfileId));
+            // Enforce update in view by setting to null first
+            model.getUserProfileIdOfScoreUpdate().set(null);
+            model.getUserProfileIdOfScoreUpdate().set(userProfileId);
+        });
+    }
+
+    private void addFilterMenuItem(ReputationSource reputationSource) {
+        String title = Res.get("user.reputation.source." + reputationSource.name());
+        StandardTable.FilterMenuItem<ReputationListView.ListItem> filterMenuItem = new StandardTable.FilterMenuItem<>(
```
namloan, line 141: Using ```var``` would be worth it here

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/DropdownMenu.java
```diff
@@ -24,19 +24,27 @@
...
 
 public class DropdownMenu extends HBox {
     public static final Double INITIAL_WIDTH = 24.0;
-
+    @Getter
     private final Label label = new Label();
     private final ImageView defaultIcon, activeIcon;
     private final ContextMenu contextMenu = new ContextMenu();
     private ImageView buttonIcon;
+    @Nullable
```
namloan: We were trying to standarize using Optional<> instead of Nullable. Is this still the case?

presentation/src/main/java/bisq/presentation/formatters/TimeFormatter.java
```diff
@@ -59,9 +59,18 @@ public static long getAgeInDays(long date) {
...
+        long ageInDays = days - years * 365;
+        String daysPostFix = ageInDays == 1 ? Res.get("temporal.day") : Res.get("temporal.days");
+        String dayString = ageInDays + " " + daysPostFix;
+        if (years > 0) {
+            String yearsPostFix = years == 1 ? Res.get("temporal.year") : Res.get("temporal.years");
+            String yearString = years + " " + yearsPostFix;
+            return yearString + ", " + dayString;
+        } else {
+            return dayString;
+        }
```
namloan: Better than manually handling plurals, can you use [ChoiceFormat](https://docs.oracle.com/javase/tutorial/i18n/format/choiceFormat.html) here?

namloan (changes requested):

presentation/src/main/java/bisq/presentation/formatters/TimeFormatter.java
HenrikJannsen: Thanks for the pointer to it. I looked into ChoiceFormat but it feels like a cumbersome API to me. Would rather impl. a custom convenience method to the Res class instead... e.g.
Res.maybePlural("temporal.year", years) which maps in the method depending on years to 
Res.get("temporal.year.none", years) 
Res.get("temporal.year.singular", years) 
Res.get("temporal.year.plural", years) 

The 0 value (none) case is mostly not needed in our case. 
I think such an API would be more convenient to use.

HenrikJannsen: See https://github.com/bisq-network/bisq2/pull/1986/commits/d801d1e7a6ad9f5f2915c4c5389b8d9d4cbc52e5

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/reputation/list/ReputationListController.java
HenrikJannsen: I tried it but then `item -> item.getReputationSources()` does not recognize the type.

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/DropdownMenu.java
HenrikJannsen: In the View classes we did not apply that because the JavaFx classes have too much nullables itself.
In that case the field cannot be null as `attachListeners` is called in the constructor. I will remove the annotation therefore.

djing-chan (approved):
ACK

Issue #1972, opened 2024-04-02
Add table view to Bisq Easy offerbook


axpoems:
* Implement table view in Bisq Easy offerbook.
* Add option to switch view mode in offerbook (chat/table).

HenrikJannsen:
There have been an icon for grid/list view for trade protocols. It got removed but the icon should be still in the images folder. If not it should be in the GH history (maybe 6-8 months back).

axpoems:
I have worked on the table view for Bisq Easy. Following are the mock-ups.

* This introduces a concept of collapsible columns that will help with the space management in smaller screens.

![offerbook-upgrade_01](https://github.com/bisq-network/bisq2/assets/145597137/8f85d850-0ebc-42c8-a23a-257944613016)

* The default option would be markets open and list offer closed (just as we have now).

![offerbook-upgrade_02](https://github.com/bisq-network/bisq2/assets/145597137/6fe924df-76a0-4655-a754-142716006bca)

* A new column to the right with the offer list is introduced. This is coupled to the market selected. It can be sorted by user, price and amount and filtered by the rest of the parameters inside an offer.

![offerbook-upgrade_03](https://github.com/bisq-network/bisq2/assets/145597137/4d7e5885-8445-4f66-9ef7-4acac5791b73)


* With markets collapsed.

![offerbook-upgrade_04](https://github.com/bisq-network/bisq2/assets/145597137/3f412cd0-91a2-4f21-9eff-3c2139a22fd4)


* When selecting a row, the chat scrolls to that offer message and the latter is highlighted. Then the offer needs to be taken using the offer message in the chat. This ensures that the user does not miss context around the offer message before taking the offer.

![offerbook-upgrade_05](https://github.com/bisq-network/bisq2/assets/145597137/84ed98cc-bb3b-4563-8552-2d93545cd404)



@HenrikJannsen @DrColver - Any thoughts?

HenrikJannsen:
Great thanks for the work on that!

From a concept POV I think it should be possible to maximize the list view and to have more columns (reputation, payment methods). I think some users just want to have the Bisq 1 like offer book view. I understand that you try to preserve the chat feeling but I am not sure if users will share that.

[...]

PR #1990, opened 2024-04-04
List offers

DrColver:
Add checkbox to show offers as list instead of as chat

It's quite hard to get an overview of the offers in the chats. This draft PR adds an option to display the offers in a compact table that can be sorted on price or reputation for example.

Are there some arguments against showing the offers as a table? It seems like a natural way to get some order, but it does lose some of the community feel of the chat.

image

image

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/components/table/BisqTableView.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/components/table/BisqTableView.java
@@ -66,6 +66,10 @@ public class BisqTableView<T> extends TableView<T> {
         setColumnResizePolicy(TableView.CONSTRAINED_RESIZE_POLICY);
     }
 
+    public void bindSort() {
+        ((SortedList<T>) getItems()).comparatorProperty().bind(comparatorProperty());
+    }
+
     public void setPlaceholderText(String placeHolderText) {
         setPlaceholder(new Label(placeHolderText));
     }
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/BisqEasyOfferbookController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/BisqEasyOfferbookController.java
@@ -71,6 +71,7 @@ public final class BisqEasyOfferbookController extends ChatController<BisqEasyOf
     public BisqEasyOfferbookController(ServiceProvider serviceProvider) {
         super(serviceProvider, ChatChannelDomain.BISQ_EASY_OFFERBOOK, NavigationTarget.BISQ_EASY_OFFERBOOK);
 
+        chatMessageContainerController.setListable(true);
         bisqEasyOfferbookChannelService = chatService.getBisqEasyOfferbookChannelService();
         settingsService = serviceProvider.getSettingsService();
         marketPriceService = serviceProvider.getBondedRolesService().getMarketPriceService();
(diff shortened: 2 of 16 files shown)


HenrikJannsen:
See also https://github.com/bisq-network/bisq2/issues/1972

DrColver:
I wasn't aware of [this issue](https://github.com/bisq-network/bisq2/issues/1972) when I implemented this table view. I was just bothered by the lack of overview. This PR seems to implement the requested feature.

Some points regarding the code:
- The ChatOfferTable is using the same model and controller as the chat view since it's just a different way to display the same data.
- Columns in the table have not been thoroughly considered, just some basic ones that came to mind.
- The price comparison handling for offers is very bothersome to work with. Some convenience class around that is likely needed, both to make code more readable, and to make the usage of prices more uniform within the app. The different options of Percentage above vs Fixed vs Market make sense for the user but will lead to confusion if not clearly annotated if mixed within the order book.
- Binding sorting of a table columns to the associated SordedList seems like it could be done better.

axpoems:
First of all, I appreciate your proactivity. However, as you already are aware, there was an issue open up for discussion regarding this implementation.

Setting aside the specifics of the issue, I'd like to highlight some of the utility of Bisq GH issues. Primarily, it serves as a means to communicate problems or proposed features to both the community and developers and discuss them before development. We also use tagging to prioritise them, signalling what should be worked on first. Additionally, assigning to someone helps prevent overlapping efforts. As you may have noticed, I was assigned to this task and was already working on it.

So in the future, especially when implementing significant UI changes, I'd recommend that you create and/or request assignment of the issue beforehand. And if UI related, use mock-ups (or express with details) upfront to convey the idea. This is important to maintain a coherent UX and UI design in the Bisq 2 app.

Answering your question in the PR description above, yes, there are arguments against just adding a table view just like the one you propose. Having said that, are you able to upload a complete mock-up of your proposal for this feature?

DrColver:
@axpoems I understand your annoyance at this proposal since there was now an issue about it and that you have started working on it. Consider this a discussion basis rather than a code PR to merge. I don't have a mock up, just the code, which has helped me consider the issues concretely.

I saw there was a discussion on the wording and color of take offer buttons so I figured the same should be used here once that's completed. There is the confusion on perspective on sell vs buy. The drop down filter for what I want to do and the offers from the offer creators perspective.

The table content could also include method, ie main net or lightning.

I think the look should be basically the same as other tables, which is what I used here, but UX is not my specialty.

axpoems:
There's no annoyance, I just gave an exposition of the dynamics occurring within Bisq development. As I said, I appreciate the proactivity, and if in a coordinated way, even better. I consider my comments the beginning of the discussion.

I see some issues with this design proposal. I will look into those and to the solution as whole and come back with specifics.

DrColver:
I likely won't have much time to work on this over the coming week or two.

As you've noted, there are issues. If you've already started on an implementation that fits better with the framework, you might just want to use some of the ideas from here.

After using the table view I'm convinced it's needed as a complement to the chat.

axpoems:
Thanks for the notice.
I agree that this will be a useful complement to the chat, but that is exactly the UX problem to solve: how to make the table *complement* the chat (and not replace it completely).

DrColver:
@axpoems I'm looking at this again, and from what I could see you haven't yet implemented the offers list.

I thought a bit more on how to complement the chat rather than having an either or choice of list or chat.

One way could be to do a popup window with the list. That would be more like some traditional trading software where charts and different instruments are shown. The chat then remains the focus and the trade list could focus the particular offer in the chat, while both remains open. Another option is to make the list a sidebar, like the channel info, but I think that might be worse since it couldn't be too wide.

Is there some design policy to not use separate windows within Bisq2? I could see it parting from the current feel but I believe there are benefits as well.

BTW, I really like the improved feel of the chat offers with colored buttons and "sell to" and "buy from" text.

axpoems:
@DrColver 
Thank you for your inputs - I've been working on the mock-up and I've approached the issue as your second suggestion, the sidebar. While it's feasible, I've had to consider the space constraints from the begining. The overall is complete, but I'm still working on the details to give it a good experience.

There's no policy regarding seperate windows, but implementing it on smaller screens would worsen the user experience. Therefore, IMO, it's best to avoid that.

I appreciate your comments on the "sell and buy" issue, further improvements are still to come. The key focus was to adopt a user-centric perspective and maintain consistency throughout the application.

PR #1993, opened 2024-04-04
Improve materialtextfield validation

HenrikJannsen:
Based on https://github.com/bisq-network/bisq2/pull/1992

PR #1994, opened 2024-04-04
Update number validation to be more strict

axpoems:
Based on #1993

* Use a more strict number validation to prevent cases like "123abc" to be flagged as valid.
Before:
image

    After:
    image

* Force a validation after updating the switch for both min. required rep. and diff. factor. This avoids getting into an inconsistent state when switching back to default with a previous errorLabel.
Before:
image

    After:
    image

* Misc. refactors and improvements.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/validator/NumberValidator.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/validator/NumberValidator.java
@@ -23,6 +23,8 @@ import javafx.util.StringConverter;
 import javafx.util.converter.NumberStringConverter;
 import lombok.Getter;
 
+import java.text.NumberFormat;
+import java.text.ParseException;
 import java.util.Optional;
 
 public class NumberValidator extends ValidatorBase {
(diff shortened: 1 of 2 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/validator/NumberValidator.java
```diff
@@ -152,6 +115,10 @@ protected void eval() {
         }
     }
 
+    private String cleanUpFormatting(String inputText) {
+        return inputText.replace(",", "");
```
djing-chan: Unfortunately diff. locales use diff. separator here. I am not a fan of those local number formats and we could just use one default number format independent of users locale.

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/validator/NumberValidator.java
```diff
@@ -152,6 +115,10 @@ protected void eval() {
...
+    }
+
     private boolean isNumber(String inputText) {
         return inputText.matches("^-?\\d+(\\.\\d+)?$");
```
djing-chan: Does that respect any decimal and thousands separator format diff. locals?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/preferences/PreferencesView.java
```diff
@@ -277,6 +276,18 @@ protected void onViewAttached() {
...
+           model.getIgnoreDiffAdjustmentFromSecManager().set(isSelected);
+           difficultyAdjustmentFactor.validate();
+        });
+
+        ignoreMinRequiredReputationScoreFromSecManagerSwitchPin = EasyBind.subscribe(
+                ignoreMinRequiredReputationScoreFromSecManagerSwitch.selectedProperty(), isSelected -> {
+            model.getIgnoreMinRequiredReputationScoreFromSecManager().set(isSelected);
+            minRequiredReputationScore.validate();
+        });
+
```
djing-chan, line 287: Is the update from the model to the switch handled? It was bi-directional binding before not its like a normal binding of switch to model.
I guess we only need it at active to apply the model date to the switch as the model data gets not changed by any other source.

djing-chan (commented):
See open questions

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/validator/NumberValidator.java
axpoems: I think this would be the safest as well. What do you think about using `Locale.US` as default?
I think ideally we only have the number formatted when the input is in read mode, and then remove the formats when writing, but that adds a lot of unnecessary complexity imo.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/preferences/PreferencesView.java
axpoems: Good catch. No need to remove it then, let's just use the listener to trigger validation.

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/validator/NumberValidator.java
axpoems: It does respect any decimals, but not the thousands separator (that's why the clean-up first). But we could have it implicitly if we define the locale.

I still see that inputs like this are accepted (although later we ignore the decimals):

image

Therefore I propose just using `NumberFormat` to validate for now and if we decide on a locale, then we can make it more specific and target those corner cases better. (In another PR)

HenrikJannsen (approved):
utACK

Issue #1899, opened 2024-03-22
Unable to send message in trade chat


strayorigin:
The peer had sent me a message in the trade chat and when replying back and hitting send, I receive the following error message and the message is not sent.  

```
java.lang.NullPointerException: user identity must not be null
	at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:907)
	at bisq.desktop.main.content.components.chatMessages.ChatMessagesComponent$Controller.doSendMessage(ChatMessagesComponent.java:310)
	at bisq.desktop.main.content.components.chatMessages.ChatMessagesComponent$Controller.onSendMessage(ChatMessagesComponent.java:295)
	at bisq.desktop.main.content.components.chatMessages.ChatMessagesComponent$View.lambda$onViewAttached$3(ChatMessagesComponent.java:460)
	at com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:86)
	at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:234)
	at com.sun.javafx.event.EventHandlerManager.disp...
```

HenrikJannsen:
It was caused in `ChatMessagesComponent`:
`UserIdentity userIdentity = checkNotNull(userIdentityService.getSelectedUserIdentity(), "user identity must not be null");`

Seems to be bug that the  `selectedUserIdentity` is not set correctly.

suddenwhipvapor:
Reporting similar issue forwarded to me by a user:

![image](https://github.com/bisq-network/bisq2/assets/124399749/d87c6084-2503-408e-a770-0f293c306442)

![image](https://github.com/bisq-network/bisq2/assets/124399749/13a88f0c-d930-479c-80e6-ad4800582e85)

```
[JavaFX Application Thread] ERROR b.d.DesktopController: Uncaught exception: java.lang.NullPointerException: user identity must not be null
	at [com.google](http://com.google/).common.base.Preconditions.checkNotNull([Preconditions.java:907](http://preconditions.java:907/))
	at bisq.desktop.main.content.components.chatMessages.ChatMessagesComponent$Controller.doSendMessage([ChatMessagesComponent.java:316](http://chatmessagescomponent.java:316/))
	at bisq.desktop.main.content.components.chatMessages.ChatMessagesComponent$Controller.onSendMessage([ChatMessagesComponent.java:301](http://chatmessagescomponent.java:301/))
[...]

PR #1997, opened 2024-04-05
Avoid nullable user identity

HenrikJannsen:
Should resolve https://github.com/bisq-network/bisq2/issues/1899

I could not reproduce and have not seen a valid code path how selected user identity could be null.
Either it might happen at a subscription where a null values is applied to the domain or corrupted persistence files.
In any case we require now that userIdentities must not be empty and in case there is no user identity selected we take the first from the set. Otherwise we ignore the call.


axpoems (approved):
utACK

Issue #1996, opened 2024-04-05
FsmException  in BisqEasyTakeOfferRequestHandler.verifyMessage


threehill:
The peer's trade failed due an error.
Error caused by: NoSuchElementException, FsmException

Stack trace: java.util.NoSuchElementException
    at java.base/java.util.Optional.orElseThrow(Optional.java:377)
    at bisq.trade.bisq_easy.protocol.messages.BisqEasyTakeOfferRequestHandler.verifyMessage(BisqEasyTakeOfferRequestHandler.java:101)
    at bisq.trade.bisq_easy.protocol.messages.BisqEasyTakeOfferRequestHandler.handle(BisqEasyTakeOfferRequestHandler.java:53)
    at bisq.common.fsm.Fsm.handle(Fsm.java:83)
    at bisq.trade.bisq_easy.protocol.BisqEasyProtocol.handle(BisqEasyProtocol.java:49)
   ...

threehill:
Also: "Sending message failed"

HenrikJannsen:
It seems that happened because of network issues. That the offer the taker took was not found on your node. Or that the data in the offer has be altered by recent changes (not aware which one but cannot exclude that). I added more detail logs if it appears again.

PR #1998, opened 2024-04-05
Add more details if takers offer is not found in BisqEasyOfferbookChannel

HenrikJannsen:
Addresses https://github.com/bisq-network/bisq2/issues/1996


axpoems (approved):
utACK

Issue #1949, opened 2024-03-28
Fix concept flaw with lack of buyer's option to set or reject the price 


HenrikJannsen:
Intentionally we did not allow the buyer to define the trade price. The reason for that is that the seller has to invest in reputation, has to pay for miner fees, usually has small trades and should be the helpful guide in the trade process - thus investing more time as it would be the case in Bisq 1. The extra price premium in the Matrix channels seemed to be 5-20%. As Bisq Easy is designed for newbies in the role of Buyers we did not want to force the user to learn and understand that context (they might not know even what miner fees are). If we would allow the buyer to set the price they would likely set a close to market price or even below and for sellers that would not be attractive to take, thus leading to unsatisfied peers on both sides.

We overlooked a conceptual problem in the current trade protocol:
Once a seller as taker has taken a offer from a buyer ,the buyer has no option to reject the take-offer attempts as soon the taker has sent the payment data. If rejecting after that it would be considered a break of contract and could lead to a ban of the buyer.
If the buyer would reject before getting the sellers data it would be ok, but the UI does not support that well and it would be anyway a too narrow frame for options.

I see 2 ways how we can fix that:

1. We add an acceptance stage to the trade protocol. Before the real protocol starts and data gets exchanged there is an explicit "Accept taker-offer intent" screen and only after the buyer has accepted the price of the seller the trade starts. This would also give opportunity to educate the buyer about the context of a potentially higher price premium.

2. We allow to set the price for the buyer but provide sufficiently good UX to avoid the above mentioned problems.
    This could be done for instance in recommended price ranges for the amount the user wants to buy. E.g. small amounts justify a higher premium. Other factors like payment methods or currencies might be a factor as well. The user could then accept the suggested price (could be phrased as: "a price of xxx has high probability to find a seller". If the user change the suggested price then they get a popup with the context. 
This option would also bring the benefit that sellers can be more sure when taking an offer that it will succeed and can get a better picture of the potential profits being made. It also allows for better filtering the offerbook.

I think the 2nd option might be the better.

Are the other solutions how to deal with it?

MwithM:
I don't understand why a seller takes an offer without a price but sends payment data without something as important as price very clear. Seems like forcing buyer to trade.

I prefer to have a price before taking the offer, the doubts about having frustrated buyers because they don't understand why they need to pay such a premium will be reduced gradually, with better UX and seller's competition.

suddenwhipvapor:
[...]

PR #1999, opened 2024-04-07
Add price step when creating a Buy offer

axpoems:
First step towards #1949

* Add price step when creating a Buy offer. This now follows the standard as when creating a Sell offer.
* Remove option to set price when seller takes a Buy offer.
* Changed order of `payment methods` and `amount`, so that the order is `price` - `amount` - `payment methods`.  

I will make a follow-up a PR adding a special view at price step with more context tailored to the buyer.


HenrikJannsen:
I had a invalid import in TakeOfferController.class.
`import bisq.desktop.main.content.bisq_easy.take_offer.price.TakeOfferPriceController;`
Wondering why Ci did not complain about it.

HenrikJannsen:
When you use the trade wizard (at getting started) we should not show the price screen as we filter matching offers and show the offers with the price.

HenrikJannsen:
Seems some css changes have screwed up something here:
Screenshot 2024-04-08 at 11 49 37

axpoems:
> When you use the trade wizard (at getting started) we should not show the price screen as we filter matching offers and show the offers with the price.

I see it was shown for the seller before. Shouldn't they behave the same now? Or show we remove it for both?

HenrikJannsen:
> > When you use the trade wizard (at getting started) we should not show the price screen as we filter matching offers and show the offers with the price.
> 
> I see it was shown for the seller before. Shouldn't they behave the same now? Or show we remove it for both?

Yes I think we should remove it for both as the offer carries the price already.

axpoems:
> I had a invalid import in TakeOfferController.class.

Fixed.

> Here the text is "seller" instead of buyer. I guess we can remove that role now when we allow both sides to set the price.

Good catch. I removed the role, but I consider this temporary until I make the updates in the UI agreed in #1949.

> Not sure about capitalisation here: Should we use just plain grammar rules or capitalize all key words?

The move this discussion for when I make the new update in the offer layout.

> Seems some css changes have screwed up something here:

Fixed.

> Yes I think we should remove it for both as the offer carries the price already.
> 
> Ah we need to support it still for backward compatibility (if user takes an old offer without price). To check if the price is present to show the price screen should be sufficient. But we need to test well with backward comp. cases.

I tested this, and it seems that the start trade wizard is not working in main for either direction. Then I remove it for both. I have created an issue to look more into that issue in particular: #2002

Regarding backwards compatibility, I tested it and I think the best solution to move forward with the fact that the buyer sets the price is to add a check and verify that the price spec in the offer is the same that the one the seller has set. This translates into the following:

* Case 1: Buy offer made in 2.0.2 and taken with 2.0.3.
A buy offer made in 2.0.2 (has market price as default) which is taken with 2.0.3. will be seen as an offer with price market price and can only be taken with that price. No issue here, although the offer won't look very attractive.

* Case 2: Buy offer made in 2.0.3 and taken with 2.0.2.
The buy offer made in 2.0.3. will have been made with a specific price. If this is taken from 2.0.2. there will be 2 issues: (1) the price spec will not be shown in the offer message and (2) the seller will set a price while taking the offer. This ofc will fail since (unless they talk) the price spec won't match. It seems the best option to me to prevent the seller from taking this offer, since now we are saying that the buyer decides. Otherwise, we risk sellers staying in an older version or a malicious actor could just exploit this vulnerability.

Ofc, all this issues are automatically solved when everyone updates.

Later on, I think we should remove completely the `agreedPriceSpec` from the contract and always go with the one in the offer. I created an issue for this too: #2003.

HenrikJannsen:
As in Bisq 1 we should follow the policy that main branch is treated as production version. Merging that now would lead to the mentioned problems if a user runs from source.
I think we should convert that PR to a draft and add on top of that a PR with the backward compatibility solution. Not sure yet whats the best path here but I think it should only get merged with that in place.

HenrikJannsen (approved):
utACK

PR #2000, opened 2024-04-07
Fix encoding issue on Windows

axpoems:
* Use unicode to encode `–` instead of using it directly. This fixes issues on Windows:
    Before:
    image

    After:
    image

* Replace `PauseTransition` with `UIScheduler`.


djing-chan (approved):
utACK

djing-chan (approved):
utACK

PR #2005, opened 2024-04-08
Unify markets table style in trade wizard

axpoems:
* Unify markets table style in trade wizard.
* Use search fiat currency box as column title.

image


HenrikJannsen (approved):
utACK

Issue #1891, opened 2024-03-22
Improve management of failed to send private messages


suddenwhipvapor:
I sent a test private message to brokenmoney, and it reported a failure with a red badge containing an exclamation mark, near a clicable to copy the contents of the message, but a button to "retry sending" was missing.
In the meantime, brokenmoney, who was notified via matrix, had sent me a private message, so the notification badge near discussions reported a `(1)`, I clicked on discussions and then on private chats, and my failed message had disappeared, while the received messages by brokenmoney were in the window.

A user will probably expect to see a small button to retry sending the message, and that the failed message is not lost when the other user messages back

HenrikJannsen:
I have seen that myself with 2.0.1 but I hope that is fixed by the many bugfixes in 2.0.2. I will try to reproduce it. Note that both traders need to have been updated to be sure its not caused by older bugs.

PR #2009, opened 2024-04-09
Fix failing mailbox message delivery

HenrikJannsen:
Resolves https://github.com/bisq-network/bisq2/issues/1891

We had a too low max map size for mailbox messages (100). Thus mailbox message sending did fail in many instances.


axpoems (approved):
utACK

djing-chan (approved):
utACK

djing-chan:
@alvasw Maybe https://github.com/bisq-network/bisq2/pull/2001 is causing the CI failures?

PR #1979, opened 2024-04-03
Translated Bisq Easy resources into Italian

super-e:
Bisq Easy resources were wrongly translated in Portuguese. Completed the translation into Italian.
#1858

PR #2010, opened 2024-04-09
Update translations

hiciefte:
This incorporates changes from #1979, updates from Transifex and missing translations from chatGPT for Czech, German, Brazilian Portuguese & Italian.


djing-chan (approved):
utACK

PR #2013, opened 2024-04-09
Ignore check for meta data until end of May at remove.

HenrikJannsen:
Do not return a failed result in case if it is different (might have further changes in future). The result would only have impact on the broadcast.


djing-chan (approved):
ACK

PR #2017, opened 2024-04-09
Add overlay to accept or reject new seller's price

axpoems:
Towards #1949 

This PR resolves better "Case 2: Buy offer made in 2.0.3 and taken with 2.0.2", as explained in https://github.com/bisq-network/bisq2/pull/1999#issuecomment-2042849307.
Now, instead of throwing when there's a mismatch in price spec, we allow the buyer to decide if he wants to accept or not.

image

Will follow-up with these improvements:
- [ ] Perhaps as it is now, the buyer needs to accept the new price more than once. At the moment, only after he sends the btc address the overlays disappears.
- [ ] Also, if the seller has sent account details, the buyer will receive the warning for violating the contract if cancelling - I will update that too so that it's always a reject for this case.


djing-chan (approved):
utACK

Issue #2014, opened 2024-04-09
Improve reputation score algorithm and display


HenrikJannsen:

strayorigin:
Thanks for the ideas and examples on this.  While I like the idea of the star system and the proposed system above seems great, it seems like it's going to constantly need to be changed as scores could change over time as users becomes inactive, more bsq is invested, etc.  

I like the idea of showing the total score because this solves the problem of needing to reevaluate the points to star ratio.  It's also easy to understand.  If the buyer uses the wizard and gets a list of sellers at the end, they can see actual numbers which I think are more meaningful than stars.  With the star system, if someone has a score of 35k they will have 3.5 stars, vs someone with 59k, who will have 4 stars.  This doesn't seem like that big of a difference, but the person that has the score of 59k has invested almost twice as much into the system and I feel should standout a bit more.  

I also considered the idea of showing the overall ranking but think that if an inactive user returns and pushed people down the list it might confuse buyers.  I've already had a few people message me about stars going down with the current system and think they might notice ranks dropping and think that it's a demerit due to a bad trade.  

At the end of the day, I like the idea of C, above.  It seems like the most straight forward system that users can quickly understand and requires no reevaluations down the road.

devinbileck:
> they might notice ranks dropping and think that it's a demerit due to a bad trade

^ That is my concern with this 5 star-based rating system. It seems to go against common experience amonst other ecommerce sites where filtering for products or sellers that have a certain rating at minimum tends to be a reasonable indicator of product quality or seller reliability as it it based on buyer feedback. Users may be mislead into believing they should avoid any sellers under a particular rating level. Where in fact those at lower levels may be just as trustworthy as those with a higher rating, they just dont have as big a reputation on the line as those that can afford to burn more BSQ.

B.2, C, or D seem like better approaches to me.

HenrikJannsen:
@devinbileck @strayorigin  Good points. 

This in fact puts the star system in question as it carries certain expectations from the user of rating based systems. 
Showing the total score though requires more information and is not intuitive for newbies. 

The ranking would resolve all those issues as its simple and it adjusts dynamically to changes.
We could use the rank order instead of the score as input for the scale. That would avoid distortions by outliers as we just had by one burning 1000 bsq and moving all others 2 stars lower.

[...]

PR #2022, opened 2024-04-10
Change reputation service to use a predefined 5-star system

axpoems:
Towards #2014

* Introduce a table score (defined in https://github.com/bisq-network/bisq2/issues/2014#issuecomment-2045894522) to calculate the stars representation. This new system allows for half a star.
* Add new icons to cover this representation.
* Remove changing star colour in trade wizard, since with the new table style this is not necessary anymore.

image


user/src/main/java/bisq/user/reputation/ReputationService.java
```diff
@@ -148,9 +148,30 @@ private void onUserProfileScoreChanged(String userProfileId) {
     }
 
     @VisibleForTesting
-    static double getRelativeScore(long candidateScore, Collection<Long> scores) {
-        long bestScore = scores.stream().max(Comparator.comparing(Long::longValue)).orElse(0L);
-        return bestScore > 0 ? candidateScore / (double) bestScore : 0;
+    static double getFiveSystemScore(long candidateScore) {
+        if (candidateScore <= 1_199) {
```
djing-chan, line 152: NIT: why not < 1_200 ?

djing-chan (approved):
utACK

user/src/main/java/bisq/user/reputation/ReputationService.java
axpoems: See #2026

Issue #1995, opened 2024-04-05
Add indicator to chat list view about expired messages


HenrikJannsen:
The notifications might signal there are more unseen messages as there are in reality due to expired messages. Also it gives a wrong impression to see channels without any messages without the info that past messages have been removed due expiration date.
We could add a special list entry as first in the list in case there have been messages in the past (or add it by default to all as over time all channels will have had expired messages). This entry could give a basic info that messages have a time to life and gets removed from the network after that time. This would be useful info to users as most will not be aware of that.

HenrikJannsen:
We have no way to know if there have been past messages which have expired as new user just get the current messages without any information about the historical data.

So I think the approach should be that we show that there might have been past messages in a subtle way.
E.g. "Chat messages older than 10 days have been removed from the network."  

That info should be displayed in a subtle way as it is not an key information and should be shown only when user scrolls to the top (of there is no scrollbar).

HenrikJannsen:
@lollerfirst You need to leave a comment here so that I can assign you.

lollerfirst:
Here.

HenrikJannsen:
@lollerfirst Are you working in that?

iMac7:
hey @HenrikJannsen, I'd like to attempt
should it work like this? and styling specifics if needed
(this checks scrollbar position to hide/show the element)

https://github.com/user-attachments/assets/fe1472b1-fdfb-457d-94f9-dfaeade7a615

Also should the label be shown on all chats (even private ones)

HenrikJannsen:
Hi @iMac7 Thanks for the suggestion. I think it would be better to have it inside the list view as a special item rather that the scroll area itself moves down, which looks a bit unexpected. @axpoems  WDYT?

HenrikJannsen:
Any update on that?

axpoems:
Agree with @HenrikJannsen that it should be inside the list view. @iMac7 are you able to work on this?

iMac7:
Sorry got a bit sidetracked here, will get on it

PR #2024, opened 2024-04-11
Add indicator to chat list view about expired messages

lollerfirst:
Changes for issue #1995 
Added a check that prevents expired notifications from being sent
Added a HBox (maybe it can be styled better) with a yellow background that is shown when there are no messages/offers

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/BisqEasyOfferbookController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/BisqEasyOfferbookController.java
@@ -51,7 +51,6 @@ import org.fxmisc.easybind.Subscription;
 import java.lang.ref.WeakReference;
 import java.util.List;
 import java.util.stream.Collectors;
-import java.util.Optional;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/BisqEasyOfferbookController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/BisqEasyOfferbookController.java
@@ -51,6 +51,7 @@ import org.fxmisc.easybind.Subscription;
 import java.lang.ref.WeakReference;
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.Optional;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
@@ -256,10 +257,19 @@ public final class BisqEasyOfferbookController extends ChatController<BisqEasyOf
                 BisqEasyOfferbookChannel channel = (BisqEasyOfferbookChannel) chatChannel;
 
                 // FIXME (low prio): marketChannelItems needs to be a hashmap
-                model.getMarketChannelItems().stream()
+                Optional<MarketChannelItem> marketChannelItem = model.getMarketChannelItems()
+                        .stream()
                         .filter(item -> item.getChannel().equals(channel))
-                        .findAny()
-                        .ifPresent(item -> model.getSelectedMarketChannelItem().set(item));
+                        .findAny();
+
+                marketChannelItem.ifPresent(item -> {
+                        model.getSelectedMarketChannelItem().set(item);
+                        model.getShowsExpiredMessagesIndicator().set(!item.getChannel()
+                                .getChatMessages()
+                                .stream()
+                                .findFirst()
+                                .isPresent());
+                });
 
                 model.getSearchText().set("");
                 resetSelectedChildTarget();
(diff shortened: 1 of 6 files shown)


lollerfirst:
Sorry I forgot to ask for assignment of this task but since it was a small one I hope I didn't cause annoyance.
I will be requesting the assignment from now on :+1:

chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
```diff
@@ -255,6 +255,15 @@ private boolean isConsumed(ChatNotification notification) {
         }
     }
 
+    private boolean isExpired(ChatNotification chatNotification) {
+        if (chatNotification.getChatChannelDomain() == ChatChannelDomain.BISQ_EASY_OFFERBOOK) {
+            return System.currentTimeMillis() - chatNotification.getDate() > BISQ_EASY_OFFERBOOK_MESSAGE_MAX_AGE;
```
HenrikJannsen, line 260: <strike>The TTL handling is already done at the network module level.</strike>
Sorry was too fast with my comment here... Your approach here is correct.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/BisqEasyOfferbookController.java
```diff
@@ -256,10 +257,19 @@ protected void selectedChannelChanged(ChatChannel<? extends ChatMessage> chatCha
...
+                        .findAny();
+
+                marketChannelItem.ifPresent(item -> {
+                        model.getSelectedMarketChannelItem().set(item);
+                        model.getShowsExpiredMessagesIndicator().set(!item.getChannel()
+                                .getChatMessages()
+                                .stream()
+                                .findFirst()
+                                .isPresent());
+                });
```
HenrikJannsen: Why keep a local marketChannelItem variable here?

That code:
```
!item.getChannel()
                                .getChatMessages()
                                .stream()
                                .findFirst()
                                .isPresent()
```
could be replaced by `item.getChannel().getChatMessages().isEmpty()`

HenrikJannsen (changes requested):
NACK

HenrikJannsen:
I will add more details to the issue itself to make more clear which approach to use. The style should be less dominant and maybe be similar to the scroll-to-bottom overlay. Maybe @axpoems have some suggestion.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/BisqEasyOfferbookController.java
a1denvalu3: Sorry for that.

lollerfirst:
So like something that gets displayed whenever the user reaches the top of the scrolling?

chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
```diff
@@ -386,7 +395,9 @@ private <M extends ChatMessage> ChatNotification createNotification(String id, C
     }
 
     private void maybeSendSystemNotification(ChatNotification chatNotification) {
-        if (!isApplicationFocussed && isReceivedAfterStartUp(chatNotification)) {
+        if (!isApplicationFocussed &&
+            isReceivedAfterStartUp(chatNotification) &&
+            !isExpired(chatNotification)) {
```
axpoems, line 400: This will prevent the system notification to be sent. However, we only send system notifications of messages received while using the app so this seems redundant. Have you tested it?
Also, have you tested that the badge is behaving correctly as well? It shouldn't show the badge number for messages which have expired.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/BisqEasyOfferbookView.java
```diff
@@ -72,7 +72,7 @@ public final class BisqEasyOfferbookView extends ChatView<BisqEasyOfferbookView,
     private DropdownTitleMenuItem atLeastTitle;
     private CheckBox hideUserMessagesCheckbox;
     private Label channelHeaderIcon, marketPrice, removeWithOffersFilter, removeFavouritesFilter;
-    private HBox appliedFiltersSection, withOffersDisplayHint, onlyFavouritesDisplayHint;
+    private HBox appliedFiltersSection, withOffersDisplayHint, onlyFavouritesDisplayHint, expiredMessagesIndicator;
```
axpoems, line 75: Why is this only added to the offerbook and not in general to all the chats?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/BisqEasyOfferbookView.java
```diff
@@ -437,10 +439,15 @@ private void addChatBox() {
         subheader.getStyleClass().add("offerbook-subheader");
         subheader.setAlignment(Pos.CENTER);
 
+        Label expiredMessagesIndicatorLabel = new Label(Res.get("bisqEasy.topPane.expiredMessagesIndicator"));
```
axpoems, line 442: I see a gap  when there's no label:
image

axpoems (changes requested):
NACK

I would use the design style for system messages here. This aligns with the suggestion that @HenrikJannsen was making in the issue, which is to add an entry by default at the beginning.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/BisqEasyOfferbookView.java
djing-chan: Note that for private chat messages we do not have an expiry date as its not gossip messages, thus this is not needed. But for common chat messages it should be applied.

HenrikJannsen:
Is this still in progress or can it be closed?

PR #2032, opened 2024-04-12
Add feedback for buyer in price step

axpoems:
Towards #1949 

* Update UI in price step.
* Add feedback section for the buyer which guides the user on what to choose and explains why.

image

image

I will follow-up with the PR that sets by default the recommended price.


HenrikJannsen (approved):
utACK

Issue #2030, opened 2024-04-12
App hangs when user has selected the reputation screen and is bootstrapping


HenrikJannsen:
A new user who selects reputation screen start to consume 100% CPU. It is likely caused by incorrect handling of updates of user profiles, causing some recursive loops.

PR #2039, opened 2024-04-13
Fix performance bug at reputation view

HenrikJannsen:
Resolves https://github.com/bisq-network/bisq2/issues/2030


djing-chan (approved):
ACK

PR #2041, opened 2024-04-13
Prepare for exclude for hash annotation

HenrikJannsen:
This add the `@ExcludeForHash` annotation to provide more flexibility with updates.

We use the serialized data from protobuf data for various use cases.
- Hash for data storage
- Hash for Inventory request
- Payload for Pow  
- Input for signature of AuthenticatedData
- ...

Changing or adding any of that data would break the hash and is a limitation for updates.

With the `@ExcludeForHash` annotation we can mark fields in those classes which should not be included for serialisation. The fields get cleared in the builder and thus are set to the default value.

As we need to traverse the hierarchical data structure we need to pass a flag if the serialisation is triggered from a use case where we want to exclude annotated fields or not.
To accomplish that we add a `getBuilder` method and change all classes implementing `Proto` to support that new framework.

This PR only add the annotation, methods, tests and has some refactorings and cleanups. It does not apply the changes yet.

### DIFF
--- a/account/src/main/java/bisq/account/protocol_type/LoanProtocolType.java
+++ b/account/src/main/java/bisq/account/protocol_type/LoanProtocolType.java
@@ -24,7 +24,7 @@ public enum LoanProtocolType implements ProtocolType {
     REPUTATION;
 
     @Override
-    public bisq.account.protobuf.LoanProtocolType toProto() {
+    public bisq.account.protobuf.LoanProtocolType toProtoEnum() {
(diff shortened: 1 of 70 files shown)


HenrikJannsen:
@alvasw ProtoBufMessageLengthTests fails on Ubuntu. I guess its some OS specific issue...

djing-chan:
Converted to draft as not clear if we want to get that into 2.0.3 or not

djing-chan (approved):
ACK

PR #2042, opened 2024-04-13
Refine Bisq Easy offer message

axpoems:
* Separate system message creation from chat message creation.
* Update title in offer message.
* Format peer nickname and open user profile when clicking.
* Add min width to peer offer message.
* Fix `copyToClipboard` for offer messages.
* Truncate nickname when used in messages.
* Add `code-block` style.

image


djing-chan (approved):
utACK

PR #2049, opened 2024-04-14
Apply changes for exclude for hash annotation support

HenrikJannsen:
Based on https://github.com/bisq-network/bisq2/pull/2041


djing-chan:
Converted to draft as not clear if we want to get that into 2.0.3 or not

djing-chan (approved):
ACK

PR #2050, opened 2024-04-14
Remove DIFFICULTY_ADJUSTMENT_ACTIVATION_DATE check

HenrikJannsen:
DIFFICULTY_ADJUSTMENT_ACTIVATION_DATE was 30.3. As we are past that we can remove that code.


djing-chan (approved):
utACK

PR #2052, opened 2024-04-14
Various fixes and improvements for buyer as maker case

axpoems:
* Add disclaimer that cancelling in the circumstances where seller has changed the price will not be considered a violation.
* Avoid showing pop-up when rejecting/cancelling using the overlay.
* React on trade state changes to decide whether to show the accept/reject seller's price overlay.
* Minor refactors.

image


djing-chan (approved):
utACK

PR #2053, opened 2024-04-14
Fixes in trade wizard's price step

axpoems:
* Close overlay if going back/next.
* Fix overlay not opening up with transition effect on first run.


djing-chan (approved):
utACK

PR #2063, opened 2024-04-15
Fix Tor start-up deadlock

alvasw:
Java's `WatchService` misses filesystem events while a consumer is processing a `WatchKey`.
Java documentation:
> Once the events have been processed the consumer invokes the key's reset method to reset the key which allows the key to be signalled and re-queued with further events.

This behavior caused deadlocks during some Tor start ups.

Changes:
- [Implement ControlPortFilePoller](https://github.com/bisq-network/bisq2/commit/60a7412c2fc35771a803039223c20560161d7687)
- [Let Tor pick control port](https://github.com/bisq-network/bisq2/commit/22238c9a626eed3836bf5ea6caf6fb9d643c6e3d)
  - To select the Tor control port, we bind to a random port, close it and pass the port number to Tor. It can happen that Tor tries to bind to the port before it is closed.  In this case, the Tor startup will fail because Tor will never print a line containing "[notice] Opened Control listener connection (ready) on ". Up until now, Bisq monitored the Tor log file to know when the control port is ready. Now Tor writes the control port to a file and Bisq reads that file.

Ref: https://github.com/bisq-network/bisq2/issues/1798

### DIFF
--- a/network/tor/tor-common/src/main/java/bisq/network/tor/common/torrc/BaseTorrcGenerator.java
+++ b/network/tor/tor-common/src/main/java/bisq/network/tor/common/torrc/BaseTorrcGenerator.java
@@ -25,15 +25,16 @@ import java.util.Map;
 
 @Builder
 public class BaseTorrcGenerator implements TorrcConfigGenerator {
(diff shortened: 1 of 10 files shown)


djing-chan (approved):
utACK

PR #2065, opened 2024-04-15
Improve error handing in connection

HenrikJannsen:
Don't throw a ConnectionClosedException if connection have been shutdown already, and we get a send call


djing-chan (approved):
utACK

PR #2112, opened 2024-04-24
Add activation date for ExcludeForHash annotation.

HenrikJannsen:
Only after that date the annotation will be processed.

Add custom code for ignoring signature in AuthorizedData to be compatible with pre v2.0.4 nodes.


djing-chan (approved):
ACK

Issue #2111, opened 2024-04-24
CPU at max


HenrikJannsen:
I observed new a few times that the CPU was maxed out while doing dev testing with multiple nodes. It might be caused be the oracle node (had a local node). After stopping all nodes and starting them again it was back to normal. It might be related to https://github.com/bisq-network/bisq2/issues/2110 and https://github.com/bisq-network/bisq2/issues/2098.
I suspect that we have some unhealthy dynamics with the network load which creates some form of feedback.

HenrikJannsen:
Likely resolved by https://github.com/bisq-network/bisq2/pull/2113

Issue #2110, opened 2024-04-24
Delay batch broadcast at oracle node


HenrikJannsen:
The oracle node broadcast large amount of messages in a loop which increases network load and with it then the required pow.
We should add delays to the batched broadcast or add a new message type for accumulated messages.

HenrikJannsen:
Likely resolved by https://github.com/bisq-network/bisq2/pull/2113

Issue #2098, opened 2024-04-21
Improve network load calculation


HenrikJannsen:
The calculation it seems to cause too high values. It need better fine tuning.

HenrikJannsen:
Likely resolved by https://github.com/bisq-network/bisq2/pull/2113

HenrikJannsen:
Bug is likely in spentSendTimeImpact

PR #2113, opened 2024-04-24
Throttle sending messages

HenrikJannsen:
Should address: https://github.com/bisq-network/bisq2/issues/2111, https://github.com/bisq-network/bisq2/issues/2110 and https://github.com/bisq-network/bisq2/issues/2098

Need to be observed if those issues are sufficiently resolved by that change.


djing-chan:
@alvasw Could you have a look to that PR?

djing-chan (approved):
ACK

Issue #2033, opened 2024-04-12
Improve price step during create offer (applies to buy direction)


axpoems:
As follow-up from #1949:
* Make offer price recommendation.
* Use colours to give the feedback: red/orange/green.
* Add amount step before price, then this info can be used to adjust the recommended ranges and price recommendation.

HenrikJannsen:
Can that issue be closed?

axpoems:
I intend to work on it, don't close it.

mvp1983:
I can take if it is actual and @axpoems doesn't have time on this

PR #2120, opened 2024-04-25
Show amount step before price in trade wizard

axpoems:
Towards #2033

Show amount step before price in trade wizard, which:
* Relaxes the min/max amounts that are allowed for trading, since price spec is not taken into consideration.
* Allows for trade price recommendation/decision making based on the amount.


djing-chan (approved):
ACK

Issue #1894, opened 2024-03-22
Add Whonix 17 Support


nyxnor:
Hello and thank you for developing Bisq V2.

- Bisq v1.9.14 bootstraps fine.
- Bisq v2.0.1 doesn't bootstrap.

I had to create the dummy tor binary:
```sh
mkdir -p /home/user/.local/share/Bisq2/tor
## Sudo necessary to avoid package overwriting the binary
sudo touch /home/user/.local/share/Bisq2/tor/tor
chmod +x /home/user/.local/share/Bisq2/tor/tor
```

Torrc of Bisqv2 (unmodified by me):
```cfg
HashedControlPassword 16:[REDACTED]
ControlPort 127.0.0.1:34477
Log notice file /home/user/.local/share/Bisq2/tor/debug.log
DataDirectory /home/user/.local/share/Bisq2/tor
SocksPort 44081
DisableNetwork 1
```

Using the command:
```sh
/opt/bisq2/bin/Bisq\ 2 --torControlPort=9051 --torControlPassword=notrequired --socks5ProxyBtcAddress=127.0.0.1:9050 --useTorForBtc=true
```

```
Mar-22 14:58:25.704 [main] INFO  b.d.DesktopAppLauncher: No jar file found. Run default Bisq application with version 2.0.1 
Mar-22 14:58:25.918 [main] INFO  b.a.ApplicationService: 

                    ........                  ......                                                                         
                ..............                ......                                                                         
              .................               ......                                                                         
            ......   ..........   ..          ......                                                                         
           ......      ......   ......        ...............        .....     .........         ..........                  
          .......              ........       ..................     .....   .............     ...............               
          ......               ........       ..........  .......    .....  ......   ...     ........   .......              
         ......                   .....       .......        .....   .....  .....            .....        ......             
         ......    ...        ...             ......         ......  .....   ...........    ......         ......            
         ......   .....      ....             ......         ......  .....    ............  .....          ......            
          ......                               .....         ......  .....         ........ ......         ......            
           ......       ....        ...        ......       ......   .....    ..     ......  ......      ........            
            ........     ..      .......        .................    .....  ..............    ...................            
             ..........       .........           .............      .....   ............       .................            
               ......................                 .....                      ....               ....   ......            
                  ................                                                                         ......            
                        ....                                                                               ......            
                                                                                                           ......            


 
[...]

PR #2122, opened 2024-04-26
Get SOCKS5 port from Tor daemon

alvasw:
We don't know the SOCKS5 port when we're using to a running Tor daemon. This change is needed to support external Tor instances.

Ref: #1894

### DIFF
--- a/network/tor/tor/src/main/java/bisq/tor/controller/NativeTorController.java
+++ b/network/tor/tor/src/main/java/bisq/tor/controller/NativeTorController.java
@@ -139,6 +139,30 @@ public class NativeTorController implements BootstrapEventListener, HsDescUpload
         }
     }
 
+    public Optional<Integer> getSocksPort() {
+        try {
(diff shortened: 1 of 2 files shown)


network/tor/tor/src/main/java/bisq/tor/controller/NativeTorController.java
```diff
@@ -139,6 +139,30 @@ public boolean isHiddenServiceAvailable(String onionUrl) {
...
+            }
+
+            // "127.0.0.1:12345"
+            socksListener = socksListener.replace("\"", "");
+            String portString = socksListener.split(":")[1];
+
+            int port = Integer.parseInt(portString);
+            return Optional.of(port);
+        } catch (IOException e) {
+            return Optional.empty();
```
djing-chan, line 162: Should we log an error or warning?

djing-chan (approved):
utACK

PR #2123, opened 2024-04-26
Let Tor pick SOCKS5 port

alvasw:
To select the Tor SOCKS5 port, we bind to a random port, close it and pass the port number to Tor. It can happen that Tor tries to bind to the port before it is closed. In this case, the Tor startup will fail.

Ref: https://github.com/bisq-network/bisq2/issues/1798


djing-chan (approved):
utACK

PR #2126, opened 2024-04-27
Use system libraries when using system Tor

alvasw:
Only preload shared libraries when using the embedded Tor binary.


djing-chan (approved):
utACK

Issue #2101, opened 2024-04-21
Add import UI to signed witness popup


HenrikJannsen:
Signed witness should have same import UI as account age.

PR #2127, opened 2024-04-28
Add import UI to signed witness popup

HenrikJannsen:
Implements https://github.com/bisq-network/bisq2/issues/2101


djing-chan (approved):
ACK

PR #2128, opened 2024-04-28
Add offer list in Bisq Easy offerbook

axpoems:
Towards #1972

Add offer list in Bisq Easy offerbook to enhance searching offers.

image

I will follow-up with:
  * Making collapsible both markets and offer list columns.
  * Adding more filters and sorting options to offer list.
  * Other small UX and style improvements (chat offer message highlight, icons, saving user preferences, etc).


djing-chan (approved):
ACK

PR #2134, opened 2024-04-30
Improvements in Offer List

axpoems:
* Persist user preference for the offers list filter and collapsible settings.
* Use reputation score and user nickname as default sorting method for offers list.

image


HenrikJannsen:
I think it would be better to invert all the new settings entries to have market list expanded by default (updating user will have value false - not market it collapsed).
`BuyFromOffers` sounds a bit weird in the domain area. I would suggest to use the domain language here: `sellOffer`.

HenrikJannsen (changes requested):
I think it would be better to invert all the new settings entries to have market list expanded by default (updating user will have value false - not market it collapsed).
BuyFromOffers sounds a bit weird in the domain area. I would suggest to use the domain language here: sellOffer.

djing-chan (approved):
utACK

PR #2135, opened 2024-04-30
Exclude reputation score from hash in `OfferMessageItem`

axpoems:
Based on #2134


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/OfferMessageItem.java
```diff
@@ -46,6 +46,7 @@ public class OfferMessageItem {
     private final String userNickname;
     private final Pair<Monetary, Monetary> minMaxAmount;
     private final String minMaxAmountAsString;
+    @EqualsAndHashCode.Exclude
```
HenrikJannsen, line 49: Maybe better to use:
```
@EqualsAndHashCode(onlyExplicitlyIncluded = true)
public class OfferMessageItem {
    @EqualsAndHashCode.Include
    private final BisqEasyOfferbookMessage message;
    ```
as it represents the offer message and all other data are either derived of to be excluded.

HenrikJannsen (changes requested):

djing-chan (approved):
utACK

PR #2140, opened 2024-05-02
Remove legacy external Tor key generation

alvasw:
- [Remove legacy external Tor key generation](https://github.com/bisq-network/bisq2/commit/0a4759850153cbf91f0f6345da2cc01b17a397e5)
  - In the past, we didn't create the Tor onion keys ourselves. Back then,
we let Tor create the keys and persist those in Bisq.
- [OnionServicePublishService: Remove redundant onionAddress argument](https://github.com/bisq-network/bisq2/commit/729cab97d281a2139c493218560a3ccaf716f886)


djing-chan (approved):
utACK

PR #2142, opened 2024-05-03
Fix Local Tor Network Setup with ControlPort

alvasw:
Developer can run a private isolated Tor network to develop Bisq. In a recent change, we let Tor pick the control port and connect to it afterward. The private Tor network setup code was not updated.

Ref: #2063


djing-chan (approved):
utACK

djing-chan:
@alvasw Can you resolve the conflicts?

djing-chan (approved):
utACK

PR #2157, opened 2024-05-11
Add fallback providers to market price request service

HenrikJannsen:
Implements https://github.com/bisq-network/bisq2/issues/2156


djing-chan (approved):
ACK

PR #2166, opened 2024-05-12
Increase number of peers we send in peer exchange.

HenrikJannsen:
Currently the number was limited to 50 reported peers.


djing-chan (approved):
ACK

PR #2184, opened 2024-05-14
Minor improvements translation related

hiciefte:
During my setup of test scripts on bisq.ontestpad, I found some minor issues that I compiled into this PR.


djing-chan (approved):
utACK

PR #2210, opened 2024-05-20
Use embedded binary for local Tor network

alvasw:
Developers can run a local Tor network to test changes and simulate network conditions. This change enforces that all local Tor networks have the same behavior.


djing-chan (approved):
utACK

PR #2211, opened 2024-05-20
Fix connection handshake bannedPeer test

alvasw:
PR #2180 broke this test. Before that change, it wasn't possible to know the ConnectionException reason without parsing its message.


HenrikJannsen:
It was already fixed in commit https://github.com/bisq-network/bisq2/pull/2209/commits/1b0dde9909146f463ddf089f9a2401ecb28b5a24
(same change as you did here)
Sorry for not making an explicit PR...

PR #2219, opened 2024-05-23
Improve status indicators

kciouv:
Adds another state to lastSeenDot (grey), while also changing all of its states to use Bisq colors. This brings the Bisq2 representation of user activity closer to the behavior of similarly real-time networked applications.


HenrikJannsen:
I would prefer to keep the glow effect. Added here a grey one: https://github.com/bisq-network/bisq2/pull/2220

PR #2225, opened 2024-05-25
Upgrade to Java 21

HenrikJannsen:
This is experimental. I could get a Desktop app running with those changes but I have not tested other applications and other gradle tasks like packaging. 
From comparing UI performance its behaves considerable slower (about 2x) than the java 17 version when switching between a heavy UI screen (offerbook with 31 USD offers) and the Getting started screen as fast as rendering allows with useAnimations set to false.
I assume we would need to update other libraries and tweak certain things to fix that performance degradation.


kciouv:
I investigated this a bit. Bump the JFX version to 21.0.3, which is the latest GA release of JFX21. JFX21 had a pretty severe CSS performance regression [that was mitigated in 21.0.3](https://bugs.openjdk.org/browse/JDK-8322795).

I tested this on my machine (https://github.com/HenrikJannsen/bisq2/pull/2), and compared to both 21.0.2 and 17, it seems to come out ahead.

(There's no need to accept the pull request; messing around with GitHub's pull request interface is annoying. Just offering it because it might be slightly less annoying than editing all three config files.)

namloan:
Tested JavaFX 21 back in November, I noticed more lag. See https://github.com/bisq-network/bisq2/issues/948#issuecomment-1792317622.

JavaFX 22 was much more promising in terms of potential performance improvements. I see that it was released in March.

HenrikJannsen:
Can confirm that with 21.0.3 its much faster. Faster as with Java 17.
Also upgraded openjfx-plugin to 0.0.13 (latest).
Will try out Java 22 as well...

HenrikJannsen:
@kciouv  @namloan I added a draft PR for update to Java 22 and the latest javafx plugin and javafx 22.0.1.
https://github.com/bisq-network/bisq2/pull/2227
Could you give it some test runs to see if it stable, faster and there are no issues?
I have not read up about the changed in Java 22 and javafx 22. We should do that also to know more...

kciouv:
@HenrikJannsen Will do. Would be surprised if the performance wasn't better; @namloan is right when pointing out that JFX22 has a lot of compelling performance improvements.

namloan:
Tested - not sure if it's faster (with the current amount of messages / offers) but seems smooth and stable.

kciouv:
I couldn't get it building on Linux, but I was busy today and didn't have time to read the error messages, so it might not be an issue with the code rather my development environment.

namloan:
> Tested - not sure if it's faster (with the current amount of messages / offers) but seems smooth and stable.

This was also meant for the JavaFX 22 PR

PR #2227, opened 2024-05-26
Upgrade to java22

HenrikJannsen:
Follow up from https://github.com/bisq-network/bisq2/pull/2225


kciouv:
I couldn't get it building on Linux, but I was busy today and didn't have time to read the error messages, so it might not be an issue with the code rather my development environment.

namloan:
Tested on OS X Sonoma 14.4 - it's not necessarily faster with the current amount of messages / offers but looks smooth and stable.

djing-chan:
@alvasw It seems update to Java 21 or 22 is technically not a big burden (not checked all build targets like packaging).
What do you think about it?

HenrikJannsen:
Will be replaced soon by a updated PR

PR #2231, opened 2024-05-29
Add info popup when max number of favourites has been reached

axpoems:
Give user feedback when max num. of favourites has been reached.

image

### DIFF
--- a/apps/desktop/desktop-app/src/main/java/bisq/desktop_app/DesktopApplicationService.java
+++ b/apps/desktop/desktop-app/src/main/java/bisq/desktop_app/DesktopApplicationService.java
@@ -37,6 +37,7 @@ import bisq.network.NetworkServiceConfig;
 import bisq.offer.OfferService;
 import bisq.presentation.notifications.SendNotificationService;
 import bisq.security.SecurityService;
+import bisq.settings.FavouriteMarketsService;
 import bisq.settings.SettingsService;
 import bisq.support.SupportService;
 import bisq.trade.TradeService;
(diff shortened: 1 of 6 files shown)


HenrikJannsen (changes requested):
I think the FavouriteMarketsService should not be handled as a static class. I understand the reasons to have convenient access, but it decreases architecture to use global/static classes for such use case. It is ok for utils but service classes should follow the existing pattern to be created by higher level services and passed to the clients (we pass service classes in item renderer classes to make them available there).

axpoems:
> I think the FavouriteMarketsService should not be handled as a static class. I understand the reasons to have convenient access, but it decreases architecture to use global/static classes for such use case. It is ok for utils but service classes should follow the existing pattern to be created by higher level services and passed to the clients (we pass service classes in item renderer classes to make them available there).

I followed the pattern used in [DontShowAgainService](https://github.com/bisq-network/bisq2/blob/main/settings/src/main/java/bisq/settings/DontShowAgainService.java), which is the other service built on top of settings. Should that one be changed as well then?

HenrikJannsen:
> > I think the FavouriteMarketsService should not be handled as a static class. I understand the reasons to have convenient access, but it decreases architecture to use global/static classes for such use case. It is ok for utils but service classes should follow the existing pattern to be created by higher level services and passed to the clients (we pass service classes in item renderer classes to make them available there).
> 
> I followed the pattern used in [DontShowAgainService](https://github.com/bisq-network/bisq2/blob/main/settings/src/main/java/bisq/settings/DontShowAgainService.java), which is the other service built on top of settings. Should that one be changed as well then?

Yes that should be changed as well IMO.

axpoems:
@HenrikJannsen - Updated the PR with the refactor of `FavouriteMarketsService`.

PR #2237, opened 2024-06-01
Refactor `DontShowAgainService`

### LINKED ISSUE #2236: Refactor DontShowAgainService
Avoid static methods in DontShowAgainService

axpoems:
Resolves #2236 
Based on #2231

Change `DontShowAgainService` so that it follows the service pattern instead of util static class.


HenrikJannsen (approved):
utACK

PR #2238, opened 2024-06-01
Clean-up offerbook filters and add remaining observers in table view

axpoems:
Based on #2236 

* Improve offerbook table view to react to changes in reputation and price as percentage (this is only relevant for fixed price).
* Remove chat messages' filters (peer reputation and offer type).
* Refactor `Filters`  class to be specific to market filters.
* Refactor menu item class to be the same for both filter and sortBy items.


djing-chan (approved):
utACK

PR #2240, opened 2024-06-02
Remove option to hide messages in offerbook chat

axpoems:
Based on #2238

Issue #1918, opened 2024-03-24
Add language flag to offer


HenrikJannsen:
With https://github.com/bisq-network/bisq2/pull/1903 we use the users language to display offers. This make the makers used language less visible (only visible when rollover by showing supported languages - if the user has edited that).

We could add a language flag (should not take too much attention from the offer content).
Other options would be that we show a warning when taking an offer with a diff. supported language.
Also when changing the language we should point the user to the supported language setting or auto-set the changed language and remove the default. 

@axpoems Any input on that?

axpoems:
I will propose something for this.

PR #2250, opened 2024-06-03
Add supported languages icon in offer message

axpoems:
Based on #2240 

* Add supported languages icon.
* Replace string with icon and only show languages on hover.

image


djing-chan:
The icon for supported languages is not very clear to me. 
What about adding the country flags for the given language. I know that is not always a clear association but commonly used (for english it could be half UK. half US). It would give the users also a quick filter for the languages they are interested in without need to rollover the icon. If the icon is rather small I think it will not take too much visual attention but still give a bit of color dot to the otherwise grey design.

axpoems:
About the supported languages, I want to call out a few things:
1. Supported languages is not an action button and should therefore, imo, not be in the react toolbar.
2. We need to clear out that space now to fit in the reactions, since that long string with supported languages takes too much space.

So, for (1) I think we have to tackle the UX problem itself, we have an issue for it #1918, we can work on this later. And for (2) I think having this icon with the tooltip is a good transition step. This icon is the standard used for languages in many applications.

And to brush on the supported languages topic, imo, if I am a user trading in a particular market, I would expect all the users there to be proficient with the language(s) in that country. So I think there's no need to indicate supported languages. But, as I said, I think we can postpone this discussion.

Also, I'm not in favour and would avoid to use country flags (or mixed flags) to represent languages. You seem to know this already, so I won't delve into it.

HenrikJannsen:
>  ...if I am a user trading in a particular market, I would expect all the users there to be proficient with the language(s) in that country. 

In EUR market there are many diff. languages and English is not that much use in Latin language countries.

HenrikJannsen (approved):
utACK

Issue #2251, opened 2024-06-03
Unify reaction buttons in chat message


axpoems:
Currently there's a mix of used tooltip styles, fonts, alignments, mouse over effect, etc.
Also, each react button should have its own icon.

HenrikJannsen:
I guess that is complete, no?

PR #2252, opened 2024-06-03
Update dropdown in chat messages

axpoems:
Based on #2250 

* Update dropdown menu on chat messages. Will do the same with the rest of the dropdown menus used in the app as to unify styles.

image

I have created #2251 to improve further the react buttons. Will follow-up with this.


kciouv:
Nit / Concept ACK

Before:
image

After:
image

The alignment to the right edge feels bad. You can move your mouse directly up from the "more options" button, and if you're too far to the right, none of the dropdown options will be within your reach because of the padding/margin. The new dropdowns are aesthetically a large improvement, but this also removes the present way to copy and paste messages.

axpoems:
Thanks for the feedback/suggestions. This PR is part of a broader effort so more changes will come.
The removal of the copy functionality was intended in the dropdown menu. It was meant to be added outside - I have added it now. Thanks for the remainder.

djing-chan (approved):
utACK

djing-chan:
@axpoems Can you resolved the conflicts?

axpoems:
> @axpoems Can you resolved the conflicts?

Fixed now.

Issue #2253, opened 2024-06-03
Loading screen stuck at 24%


SplashRecover:
Shut down PC with bisq 2 still running on it. Upon opening I get this message
![image](https://github.com/bisq-network/bisq2/assets/171462392/87b10aec-e7e5-4b79-9437-7085cf00d2eb)

After that is this

![image](https://github.com/bisq-network/bisq2/assets/171462392/31ef1417-3f86-4629-8532-d2251f9a82eb)

Then after that it asks me if I want to report to github tracker, and loads it still loads in after as if it's about to start working normally, but the loading screen gets stuck at 24%. After 10+ hours it still at 24% where as before it took seconds to get to 24.

It also says start up failed on the loading screen.

![image](https://github.com/bisq-network/bisq2/assets/171462392/b7a0350b-c71f-485e-915e-bcbb365afe13)

Here is the log file

[bisq.log](https://github.com/user-attachments/files/15536548/bisq.log)

HenrikJannsen:
Can you post the tor logs from inside the tor directory?

SplashRecover:
Where can I find them?

MwithM:
Updating the OS solved this in my case. BIsq 2 worked in the previous version, but had this same issue in v2.0.4

wesleyguitars:
I have the same problem, not been able to use it for weeks. Tried updating OS, going back to 2.0.3 version. Same thing :(

m-cf-b:
Still stuck on this as well. when I did a clean home directory it worked... but I need access to  the old one as I just bought some.
[bisq.log](https://github.com/user-attachments/files/15598303/bisq.log)

[debug.log](https://github.com/user-attachments/files/15598297/debug.log)

HenrikJannsen:
@alvasw Can you have a look at the logs?

HenrikJannsen:
Should be fixed with version 2.1.0 (https://github.com/bisq-network/bisq2/releases/tag/v2.1.0)

PR #2288, opened 2024-06-11
Fix Tor startup issues on Windows

### LINKED ISSUE #2239: An error occurred at initializing Bisq: bisq.tor.controller.ControlCommandFailedException: Couldn't connect to control port..
[bisq2 issue.txt](https://github.com/user-attachments/files/15524091/bisq2.issue.txt)

alvasw:
On Windows Tor writes its control port number to a file before the port
is accessible by other applications. Users reported that Bisq throws a
ConnectException (ControlCommandFailedException) at startup. All logs,
I received so far, are from Windows users.

Fixes #2239, #2253


djing-chan (approved):
utACK

PR #2295, opened 2024-06-16
Improvements in user profile details

axpoems:
Based on #2293

* Only show copy icon on mouse over.
* Add copy option for user id.
* Refactors and style improvements.


djing-chan (approved):
utACK

Issue #2291, opened 2024-06-11
Add button for zipping bisq.log + debug.log to error popup as well as show Bisq version in popup


HenrikJannsen:

HenrikJannsen:
@kciouv Can you comment here so I can assign you?

PR #2296, opened 2024-06-17
Add button to error popup to zip log files

kciouv:
Adds a button to error popups that zips debug.log and bisq.log.

Partially solves #2291 

image

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/components/overlay/Overlay.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/components/overlay/Overlay.java
@@ -964,8 +964,8 @@ public abstract class Overlay<T extends Overlay<T>> {
                     if (logPath.toFile().isFile()) {
                         try {
                             Files.copy(logPath, zipfs.getPath(logPath.toFile().getName()), StandardCopyOption.REPLACE_EXISTING);
-                            zipLogLabel.setText(Res.get("popup.reportError.zippedTo", baseDir + "/bisq2-logs.zip"));
-                            OsUtils.open(baseDir);
+                            zipLogLabel.setText(Res.get("popup.reportError.zippedTo", zipDirectory[0] + "/bisq2-logs.zip"));
+                            OsUtils.open(zipDirectory[0]);
                         } catch (IOException e) {
                             throw new RuntimeException(e);
                         }


djing-chan:
Got a 
```
Illegal character in opaque part at index 39: jar:file:/Users/account/Library/Application Support/bisq2_Alice/bisq2-logs.zip
```
exception when trying it out on MAcOS. I guess the space in Application Support cause the issue.

apps/desktop/desktop/src/main/java/bisq/desktop/components/overlay/Overlay.java
```diff
@@ -936,6 +939,21 @@ private void addReportErrorButtons() {
         gridPane.getChildren().add(logButton);
         logButton.setOnAction(event -> OsUtils.open(new File(baseDir, "bisq.log")));
 
+        Button zipLogButton = new Button(Res.get("popup.zipLogs.log"));
+        GridPane.setHalignment(zipLogButton, HPos.LEFT);
+        GridPane.setRowIndex(zipLogButton, gridPane.getRowCount());
+        gridPane.getChildren().add(zipLogButton);
+        zipLogButton.setOnAction(event -> {
+            URI uri = URI.create("jar:file:" + baseDir + "/bisq2-logs.zip");
```
axpoems: This doesn't seem cross-platform compatible. Try this instead:
```
URI uri = URI.create("jar:file:" + Paths.get(baseDir,"bisq2-logs.zip").toUri().getPath());
```

apps/desktop/desktop/src/main/java/bisq/desktop/components/overlay/Overlay.java
```diff
@@ -936,6 +939,21 @@ private void addReportErrorButtons() {
...
+            Map<String, String> env = Map.of("create", "true");
+            try (FileSystem zipfs = FileSystems.newFileSystem(uri, env)) {
+                Files.copy(Path.of(baseDir).resolve("bisq.log"), zipfs.getPath("/bisq.log"), StandardCopyOption.REPLACE_EXISTING);
+                //TODO - when multi-transport support arrives, check which transports are being used and collect all instances of debug.log
+                Files.copy(Path.of(baseDir + "/tor/").resolve("debug.log"), zipfs.getPath("/debug.log"), StandardCopyOption.REPLACE_EXISTING);
```
axpoems: Can you ensure that the files exist before copying them?

i18n/src/main/resources/application.properties
```diff
@@ -236,6 +236,7 @@ popup.headline.invalid=Invalid input
 popup.headline.error=Error
 
 popup.reportError.log=Open log file
+popup.zipLogs.log=Zip log files
```
axpoems: I think it would be better to follow `popup.reportError.zipLogs` as the rest.

djing-chan:
> Perhaps it would be useful to include all available log files, not only bisq.log but also bisq_1.log, etc.

This could become very large... I think the default log should be enough in most cases

djing-chan:
I still get the same exception as I posted above

djing-chan:
> > I still get the same exception as I posted above
> 
> The machine I'm writing this on is a bit too underpowered to make spinning up a new Mac VM convenient. Will do so on a different machine in a day or two and figure out what the proper fix is.

I think you can simulate it by adding a space in the data directory. I think the space in the path is not supported in API you use.

djing-chan:
Yes it works now.

The text shows the `jar:file:` prefix
Screenshot 2024-06-20 at 00 03 00

I think it would be better to open automatically the directory or better let the user decide where to store it. On MacOS the `Application Support` directory is by default invisible and for non tech users its not trivial how to open it. The code for doing a backup in Utilities might provide a good template.

djing-chan:
Yes it opens now the directory. 
If you want to improve it further to let the user decide the location I think it would make it better, but if not its ok as well.

PR #2297, opened 2024-06-17
Update more options menu in peer message

axpoems:
* Add option to open dropdown menu to the right.
* Change more options menu in peer message to open downwards and to the right.


djing-chan (approved):
utACK

Issue #2301, opened 2024-06-17
Not possible to switch to English UI if OS language is not English


hiciefte:
Right now if you have e.g. German as your OS default language, when switching the language in the preferences to English will keep the app after restart in German. It is not possible to switch to English.

I'll push the PR that fixes this issue in a couple of minutes.

PR #2302, opened 2024-06-17
Minor translation improvements

### LINKED ISSUE #2057: TradeWizard: Truncation of Strings
In the confirmation screen the content of one label is too long so it is truncated. But there is no tooltip set to show the full content of the label in German. There is also a grammar error in the translation which is caused by the way how the "Buy/Sell Bitcoin" label is constructed in code.

Bildschirmfoto 2024-04-15 um 11 36 07

hiciefte:
This fixes #2057, #2301 and other minor German translation issues.


djing-chan (approved):
utACK

Issue #2058, opened 2024-04-15
Use users language for trade log messages


HenrikJannsen:
We should use a similar model as with offer to not use the peers language but the users own language.

HenrikJannsen:
We can solve that by sending the resource bundle key and arguments instead of the resolved string.

There are 2 ways how to implement that:
- Create a new message with new fields for the key (text field could be used for that) and the arguments
- Encode the tuple of key and arguments as string and decode it back at the receiver.

The first approach would be more clean but it requires to create a new service class as the message is a generic type of the service. Those changes would cause quite a lot of change effort.

The second approach seems more adequate to the scope of the feature.

To ensure the encoding is not screwed up by some chars used in the key or arguments, we use non-printing chars (https://en.wikipedia.org/wiki/Delimiter#ASCII_delimited_text).
By that we can encode before sending the log message and decode when displaying it without any change in the message infrastructure.

PR #2303, opened 2024-06-17
Use user language for system messages

HenrikJannsen:
Implements https://github.com/bisq-network/bisq2/issues/2058


djing-chan (approved):
ACK

PR #2305, opened 2024-06-17
Implement new design for user profile sidebar

axpoems:
* Add new design for user profile sidebar unifying field style.
* Add icons for ignore, undo ignore, mention, private chat, and report.
* Introduce `MenuItem` control.

image

image


djing-chan (approved):
utACK

PR #2307, opened 2024-06-18
Add version string to error popup

kciouv:
Depends on #2304

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/components/overlay/Overlay.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/components/overlay/Overlay.java
@@ -993,7 +993,7 @@ public abstract class Overlay<T extends Overlay<T>> {
         GridPane.setMargin(buttonBox, new Insets(buttonDistance, 0, 0, 0));
         gridPane.add(buttonBox, 0, gridPane.getRowCount(), 2, 1);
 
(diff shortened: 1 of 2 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/components/overlay/Overlay.java
```diff
@@ -987,10 +988,14 @@ protected void addButtons() {
         }
 
         buttonBox = new HBox(10);
+        buttonBox.setAlignment(Pos.CENTER);
         GridPane.setHalignment(buttonBox, buttonAlignment);
         GridPane.setMargin(buttonBox, new Insets(buttonDistance, 0, 0, 0));
         gridPane.add(buttonBox, 0, gridPane.getRowCount(), 2, 1);
 
+        Label versionLabel = new Label("Version: " + ApplicationVersion.getVersion().toString());
```
djing-chan: The "Version: " string should be in the i18n files.

djing-chan (changes requested):

apps/desktop/desktop/src/main/java/bisq/desktop/components/overlay/Overlay.java
kciouv: @djing-chan solved.

PR #2312, opened 2024-06-18
Peer message and user porfile siderbar improvements

axpoems:
* Add ignore and report icons into the more options message menu.
* Minor improvements for the copy option in user profile sidebar.

image


djing-chan (approved):
utACK

Issue #1958, opened 2024-03-31
Add offer info popup


HenrikJannsen:
To get more detail information of the offer we could add an info icon which opens a offer detail popup.
There is code from a past attempt for a offer details popup in the code base.

wuhualiang:
![Image](https://github.com/user-attachments/assets/e588eb08-90ac-4fd9-ad67-ca59394a7785) 
@HenrikJannsen  Is this the place where the details pop-up window is added?

HenrikJannsen:
The button/icon to open the popup could be put inside the context menu or next to the copy button. @axpoems WDYT?

HenrikJannsen:
We have waited for response from @axpoems . As he did not respond and if @wuhualiang is still interested to work on that, I would suggest to put the button inside the context menu.

I am not sure about the design of the details popup, but is should be similar to the trade details. For optional data we usually don't show the field if its empty.

![Image](https://github.com/user-attachments/assets/1236987f-6fb9-4ec7-bf9b-d8a740d6016f)

wuhualiang:
@mvp1983  You take it.  cc @HenrikJannsen

mvp1983:
As I understand, this task should visualize Offer class'es fields.
I want to create something like Trade Details, suggested by @HenrikJannsen  but without hardcoding of UI and create something like wizard which can show different type of info with additional functionality depending on field type (copy/link/hint/etc...)

HenrikJannsen:
Maybe provide a screenshot of how it should look like.

mvp1983:
This was in code:
![Image](https://github.com/user-attachments/assets/eea73f8f-04ea-491e-8bdc-a2bd373efd98)


I propose something like that
![Image](https://github.com/user-attachments/assets/90889d13-f2b7-4f30-ba02-da2eb19c4532)

maybe, with some effects/changes for some type of field (+ using BitcoinAmountDisplay for BTC).

And build with builder, so, can be used in other places.

HenrikJannsen:
Or we stick to the same as in trade details: Overview, Details
To have some separation is helpful as its lots of fields.

HenrikJannsen:
@axpoems Would you have time to work on that?

PR #2313, opened 2024-06-20
Add an info icon which opens a offer detail popup

skaldjur:
Fix for #1958.


skaldjur:
Added info icon to offer.
icon triggers popup

info icon triggers popup with details.
popup shows more info details on offer

skaldjur:
@HenrikJannsen The more information output in the popup is too verbose. I'll appreciate some guidance on what to take out and what to keep. For now I'm just stringifying the `offerBookMessage` so we can see ALL the data that the model has on the offer.

skaldjur:
Force pushed after signing my commit.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/list/ChatMessagesListController.java
```diff
@@ -307,6 +308,36 @@ public void onTakeOffer(BisqEasyOfferbookMessage bisqEasyOfferbookMessage) {
         Navigation.navigateTo(NavigationTarget.TAKE_OFFER, new TakeOfferController.InitData(bisqEasyOffer));
     }
 
+    public void onMoreInfo(BisqEasyOfferbookMessage bisqEasyOfferbookMessage){
+        Object chatMessageId, chatMessageText, authorProfileID,
```
djing-chan: We use strict typing. Object is only used in rare cases.

djing-chan (changes requested):

djing-chan:
There is class BisqEasyOfferDetailsView which was prepared for offer info. Layout should fit the Bisq style. This version looks very technical.

PR #2334, opened 2024-06-27
Improve error popup

HenrikJannsen:
Based on https://github.com/bisq-network/bisq2/pull/2296
Rebased on Master.

The version at https://github.com/bisq-network/bisq2/pull/2296 had multiple buttons and text display which is not really needed. This version reduce the UX elements to the minimum.
It also improves the display of the error and includes the Bisq Version and OS version to the text which should be used for reporting.

Screenshot 2024-06-27 at 14 19 23

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/components/overlay/Overlay.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/components/overlay/Overlay.java
@@ -953,6 +953,7 @@ public abstract class Overlay<T extends Overlay<T>> {
                         try {
                             Files.copy(logPath, zipfs.getPath(logPath.toFile().getName()), StandardCopyOption.REPLACE_EXISTING);
                             zipLogLabel.setText("Zipped to: " + uri.toString().replace("%20", " "));
+                            OsUtils.open(baseDir);
                         } catch (IOException e) {
                             throw new RuntimeException(e);
                         }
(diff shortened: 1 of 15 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/components/overlay/Overlay.java
```diff
@@ -408,16 +411,29 @@ public T invalid(String message) {
...
+
+        processMessage(Res.get("popup.reportError"));
+
+        String version = ApplicationVersion.getVersion().getVersionAsString();
+        String osInfo = OsUtils.getOsInfo();
+        String errorReport = Res.get("popup.reportBug.report", version, osInfo, message);
+        TextArea errorReportTextArea = new TextArea(errorReport);
+        errorReportTextArea.setPrefWidth(width);
+        errorReportTextArea.setWrapText(true);
+        errorReportTextArea.getStyleClass().add("error-log");
```
axpoems: You could use here `code-block` instead:

image

djing-chan:
> ACK / Nit:
> 
> Technically, the Tor log does contain something sensitive (user's name, if their user account is named after themselves).

Ah true the data directory is logged at some points. We have blinded it in the bisq log but in tor log thats more difficult.
@alvasw Do you have an idea how to filter tor debug.log output to replace user directory with a placeholder to protect privacy of users who name their username to their real name?

apps/desktop/desktop/src/main/java/bisq/desktop/components/overlay/Overlay.java
HenrikJannsen: Impl. with https://github.com/bisq-network/bisq2/pull/2334/commits/0bc93af4f536e92cc8d64ace254bf3b519af615e

HenrikJannsen:
> ACK / Nit:
> 
> Technically, the Tor log does contain something sensitive (user's name, if their user account is named after themselves).

Fixed with https://github.com/bisq-network/bisq2/pull/2334/commits/7ebc9e9de84645caf410eeab1a0e29aa2297c8d8

PR #2340, opened 2024-06-29
Add reactions to my text messages

axpoems:
* Add react menu in my text message.
* Align delete action buttons for both my offer and my text message.
* Fix small alignment regression when using `BisqMenuItems` inside a `DropdownMenu`.


djing-chan (approved):
utACK

PR #2341, opened 2024-06-29
Fix bug with reputation score update.

HenrikJannsen:
The Observable<String> userProfileIdOfUpdatedScore has not been updated if multiple scores have been added and the clients did not get an update event.


djing-chan (approved):
ACK

Issue #2262, opened 2024-06-06
Add txId to ProofOfBurnDto and BondedReputationDto


HenrikJannsen:
If the same amount is burned/bonded in the same block by the same user both data objects are the same and only one will be used, thus the 2nd tx is ignored.
We need to add either the TxId or the index inside the block (we use blocktime which is unique for each block, thus the index would deliver us a unique object).

PR #2344, opened 2024-06-30
Add tx id and block height to proof of burn  and bond data [2a]

HenrikJannsen:
Based on https://github.com/bisq-network/bisq2/pull/2345

Fixes https://github.com/bisq-network/bisq2/issues/2262

Oracle nodes using that code need to run the RestApi code based on https://github.com/bisq-network/bisq/pull/7187 

When we deploy it we need to ensure that one seednode runs on the old version and one on the new version.
The oracle node publishes both a version 0 (old) version and a version 1 (new) version.
The newly added `blockHeight` and `txId` fields are annotated with `@ExcludeForHash(excludeOnlyInVersions = {0})`. This means those fields are ignored in version 0 objects, thus creating the same hash as from the old code base.
The `version` field is annotated with `@ExcludeForHash` - thus it is ignored on all versions.

A old seed node will fail with the new data as the hashes are not matching (new fields not excluded in new version, but as old node don't know those fields they are not present).

A new seed node would provide both the old and new version but as in the inventory response all the data are added the hash is different between the receiver and the sender (new node had included the fields). 

Also as long not many of the new nodes are in the network distribution of the version 1 objects will be poor. To mitigate that we can restart the Oracle node frequently.

For the issue with inventory response we can consider to add `@ExcludeForHash`  to the `inventory` field in `InventoryResponse`. As the response is anyway invalid between old nodes and new seed nodes (or other new nodes) we would not cause more harm by adding the annotation which would alter the hash as well.

### DIFF
--- a/application/src/main/java/bisq/updater/UpdaterService.java
+++ b/application/src/main/java/bisq/updater/UpdaterService.java
@@ -119,14 +119,14 @@ public class UpdaterService implements Service {
             return;
         }
 
-        Version newVersion = releaseNotification.getVersion();
+        Version newVersion = releaseNotification.getReleaseVersion();
         Version installedVersion = ApplicationVersion.getVersion();
         if (newVersion.belowOrEqual(installedVersion)) {
             log.debug("Our installed version is the same or higher as the version of the new releaseNotification.");
             return;
         }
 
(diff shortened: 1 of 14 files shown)


djing-chan (approved):
ACK

PR #2347, opened 2024-06-30
Add version number to user profile [2b]

HenrikJannsen:
Based on https://github.com/bisq-network/bisq2/pull/2345

Implements https://github.com/bisq-network/bisq2/issues/2289

This feature allows us to get information about the update rate at new releases. Currently there is nothing implemented for that but this could easily be added (as done in https://github.com/bisq-network/bisq2/pull/2348).

Similar as in https://github.com/bisq-network/bisq2/pull/2344 we use the newly added support for version parameter in ExcludeForHash to support an added fields in a backward compatible way.
We publish both the version 0 and the version 1 objects so that both old and new versions can operate with the data. The old version will reject the new data.

### DIFF
--- a/application/src/main/java/bisq/updater/UpdaterService.java
+++ b/application/src/main/java/bisq/updater/UpdaterService.java
@@ -119,14 +119,14 @@ public class UpdaterService implements Service {
             return;
         }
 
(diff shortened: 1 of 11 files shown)


djing-chan (approved):
ACK

PR #2348, opened 2024-06-30
Add version distribution info [3b]

HenrikJannsen:
Based on https://github.com/bisq-network/bisq2/pull/2347

Screenshot 2024-06-30 at 22 49 28

"N/A" is for versions which did not support the version field (will be pre 2.0.5 once 2.0.5 is released)


djing-chan (approved):
ACK

Issue #2346, opened 2024-06-30
Make Inventory request more resilient


HenrikJannsen:
In case a hash of an object has changed, the inventory response has a wrong hash and the pow check fails, thus no data is accepted from the response.
We should make the process more resilient.

PR #2349, opened 2024-07-01
Add @ExcludeForHash to inventory in InventoryRequest

HenrikJannsen:
Implements https://github.com/bisq-network/bisq2/issues/2346

When objects get an update we publish both the old version and the new one, where the new one is considered invalid for old nodes. The Invenotry contains all those data and would become itself invalid if we do not exclude the inventory from the hash creation for the pow check. Note that InventoryResponse is a direct message and its only the pow use case where the hash would cause issues.


djing-chan (approved):
ACK

PR #2352, opened 2024-07-01
Add version and applicationVersion to Capability

HenrikJannsen:
Implements https://github.com/bisq-network/bisq2/issues/2350

We use first version 1 when we start the handshake. When connection to an old node it is expected to fail as the pow hash check fails. In that case we retry with version 0.

When we receive a handshake request we check the peers capability version and reply with the same version.

@alvasw 
The changes are not implemented in the new `PeerConnectionsManager` classes as those are not yet used and at time when its used the backward compatibility issue is likely anyway resolved as we dont need to support pre 2.0.5 versions anymore.


djing-chan (approved):
ACK

PR #2356, opened 2024-07-01
Improve message delivery states

### LINKED ISSUE #1756: Revision of message states
Based on the discussion in #1752, let's make an overall revision of how message states are display to the user.

image

axpoems:
Resolves #1756

* Use only three icons to indicate delivery status: success, pending, and failed.
   Tool tip is still more explanatory.
* Add separate button to resend when failure.
* Add new icons for success, pending, failed and try again.


djing-chan (approved):
utACK

PR #2373, opened 2024-07-05
Show mailbox delivery status

axpoems:
* Add new icon to represent mailbox.
* Show 'added to mailbox state' in messages that have delivery status display.


djing-chan (approved):
utACK

PR #2376, opened 2024-07-07
Improve handling of unprocessed reactions

axpoems:
This solves (in a best-effort manner) the cases where a reaction message is processed before the chat message it belongs. This is done by queuing those reactions and then attempting to process them later, when a new message comes in.


djing-chan (approved):
utACK

PR #2383, opened 2024-07-09
Add back Bisq Easy offerbook 'show offers only' setting

axpoems:
* Add back option to show only offers in Bisq Easy offerbook.
* Remove checkbox from the Bisq Easy offerbook chat view (now it's only available from settings).


djing-chan (approved):
utACK

PR #2397, opened 2024-07-11
Dont use MacOS aarch64 tor binaries

HenrikJannsen:
The aarch64 tor binary seems to have issues as tor cannot get started up. We temporarily disable it and use the x86 instead on MacOS with aarch64 (M1,M2,M3)


djing-chan (approved):
ACK

PR #2398, opened 2024-07-11
Fix MacOs aarch64 handling

HenrikJannsen:
Remove MAC_OS_AARCH64 from OS. Check for architecture in PerOsUrlProvider


djing-chan (approved):
utACK

PR #2404, opened 2024-07-13
Improve chat messages' active reactions display

axpoems:
* Add a new menu to display the active reactions that a message has. This now includes the count, and also serves as a toggle to add/remove a reaction.
* Add a new class `ReactionItem` to encapsulate the representation of a reaction.
   Will follow-up aligning the already existing reactions menu to use this same reference.


djing-chan (approved):
utACK

PR #2405, opened 2024-07-14
Update to java 22

HenrikJannsen:


HenrikJannsen:
Could not get it working on Windows and Intel Mac.
On Windows all builds and I can create a exe, but the exe does not start up.

PR #2409, opened 2024-07-14
Align reactions display components

axpoems:
Based on #2408 

* Align react menu to use and listen to reactions items.
* Fix selection marker when user has multiple identities.


djing-chan (approved):
utACK

PR #2416, opened 2024-07-14
Comment out problematic code in PackageFactory

HenrikJannsen:
@alvasw : On Linux that causes a build failure at the packager task. On Windows the exe cannot start up as it seems the running process still has some resources blocked. I reduce the timeout to 2 minutes, that seems to work in my tests


djing-chan (approved):
ACK

Issue #1946, opened 2024-03-28
Show update popup when there is a pre-release available and user toggle preferences 


HenrikJannsen:

HenrikJannsen:
@skaldjur Are you working on that?

PR #2429, opened 2024-07-16
show update popup if prerelease gets enabled in preferences

HenrikJannsen:
Implements https://github.com/bisq-network/bisq2/issues/1946


djing-chan (approved):
ACK

PR #2431, opened 2024-07-16
Implement support for halt trading

HenrikJannsen:
Add support to halt any trade protocol state change when security manager has published an emergency alert with the haltTrading flag set to true.


djing-chan (approved):
ACK

PR #2434, opened 2024-07-17
Add doc about backward compatibility tools

HenrikJannsen:
Implements https://github.com/bisq-network/bisq2/issues/2426


djing-chan (approved):
utACK

PR #2444, opened 2024-07-19
Merge fiat and btc payment methods in one step in trade wizard

axpoems:
Towards #2311 

* Merge fiat payment methods and btc settlement in the same step "Payment Methods" when creating an offer.
   By default, always allow on chain. LN can be added on top of this.

I will follow up with more PRs adding the new design styles for this.


djing-chan (approved):
utACK

PR #2445, opened 2024-07-19
Fine tune reactions' background styles

axpoems:
Based on #2444

This aligns backgrounds of active reactions display menu and the corresponding (my/peer) message.


djing-chan (approved):
utACK

PR #2446, opened 2024-07-19
Tone down reaction icons

axpoems:
Based on #2446

Add colour adjustment to town down default reaction icons.


djing-chan (approved):
utACK

PR #2451, opened 2024-07-22
Improve payment methods step in trade wizard

axpoems:
Towards #2311

* Apply new design style.
* Fix various payment methods handling issues.


djing-chan (approved):
utACK

PR #2458, opened 2024-07-22
MetaData is transient as it will be used indirectly by low level netw…

HenrikJannsen:
…ork classes. Only some low level network classes write the metaData to their protobuf representations.


djing-chan (approved):
ACK

PR #2459, opened 2024-07-22
PpgPublicKeyParser: Warn about uncertified subkeys

alvasw:
The Tor and the Electrum developer's subkeys are signed by their master
keys. The PpgPublicKeyParser enforces this behavior. Unfortunately,
this is not true for all Bitcoin Core developers. One example where this
happens is when we try to verify the key with the fingerprint
`0x57FF9BDBCC301009`.


djing-chan (approved):
utACK

PR #2464, opened 2024-07-22
Fix typo

HenrikJannsen:


HenrikJannsen:
Turned out that leading dot is needed to distinguish between the common package and the network.common package. I guess we should rename the network.common to avoid such confusion

PR #2477, opened 2024-07-25
Fix bug at DropdownMenuItem

HenrikJannsen:
Using DropdownMenuItem.setOnAction leads to duplicate execution of the handler on MacOS 14.5 / M3. Might be a bug in the MacOS specific JavaFX libraries as other devs could not reproduce the issue. Using an eventFilter and consuming the event solves the issue.


djing-chan (approved):
ACK

PR #2481, opened 2024-07-25
Add reacted users popup

axpoems:
Add popup showing the list of users who have reacted. Users are sorted by reaction date time.

image


djing-chan (approved):
utACK

Issue #2485, opened 2024-07-26
Notification of mediator message shown at peers icons


HenrikJannsen:
![Screenshot 2024-07-26 at 16 49 15](https://github.com/user-attachments/assets/9698c7b6-c640-4c10-8243-d826ce1d4413)

We could consider to add the mediator to the table, but space is already very limited.

PR #2486, opened 2024-07-26
Add mediator to trade header if a trade is in mediation.

HenrikJannsen:
Fixes https://github.com/bisq-network/bisq2/issues/2485


djing-chan (approved):
ACk

PR #2491, opened 2024-07-27
Add missing UIThread.run when using addObserver

axpoems:


djing-chan:
@axpoems  Could you resolve the conflicts?

djing-chan (approved):
utACK

Issue #2489, opened 2024-07-27
Offerbook notifications


HenrikJannsen:
Currently we show only notifications for messages to the selected market.
E.g. If user has selected USD market and moves to another view, they will get the notifications when new messages arrive at USD market channel, but not if there are message on any other market.

This behavior might not be intuitive as even @suddenwhipvapor did expect that the badge is for any message in any market.

The easiest way to improve that would be to apply the model we use in private chats here and show badges for each market and show the sum of all notifications of all markets. 

@hiciefte suggested in https://github.com/bisq-network/bisq2/issues/1876 that we should only apply that to favorites as it might become annoying.
An alternative might be to add a "mute" button to the markets (could be used elsewhere as well), so users can mute busy channels they are not interested in.
There are anyone only a handful active channels so to mute those should not be a big issue and might be more simple from UX than to combine it with favorites.
It would have the benefit as well to get some kind of "liveness" indicator of other channels and might motivate some traders to setup multi-currency accounts like Wise to be able to benefit from different markets.

This was discussed here together with another bug:
https://github.com/bisq-network/bisq2/issues/1876

HenrikJannsen:
With 2 digit numbers the space is a bit limited...
![Uploading Screenshot 2024-07-27 at 11.33.01.png…]()

suddenwhipvapor:
From your first post it seems that making all currencies "active" and having users mute some of them (blacklist), is easier to implement than the other way around with a whitelist. Trying to fill the shoes of someone only interested in usd or eur, it would be annoying to go into each ignored market to mute that, while having all markets quiet by default and enabling notifications only for those that you want could be more intuitive. Still it's something that users would only do once, so not a big deal if the coding effort is asymmetrical.

PR #2492, opened 2024-07-27
Show notificationts of all markets

HenrikJannsen:
Implements https://github.com/bisq-network/bisq2/issues/2489


djing-chan (approved):
ACK

PR #2494, opened 2024-07-28
Fixes main branch not compiling

rodvar:
- Fix compilation error on trade module by implementing default case for fromProto method switch
 - Added git ignore for dev .sh scripts
 
 
 Original error:
 
 Task :application:compileJava FAILED []/projects/bisq2/application/src/main/java/bisq/application/ResolverConfig.java:43: error: package bisq.trade.protocol.messages does not exist import bisq.trade.protocol.messages.TradeMessage; ^ 1 error

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ hs_err_pid*
 *.spvchain
 *.wallet
 *.ser
+*.sh
 .DS_Store
 .gradle
 build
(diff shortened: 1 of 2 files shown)


rodvar:
Hey guys! Just a small fix I've done whilst setting my dev environment to get ready to help on Bisq2
I added *.sh to git ignore as well, we can remove it if you see any issue with that. Cheers!

djing-chan (approved):
utACK

djing-chan:
The change in TradeMessage should not have caused compilation errors. But maybe the change invalidated some gradle cache and it forces a recompilation and that fixed the issues. There are sometimes nasty bugs with cache, specially when larger changes like Java 22 update have been applied.

PR #2508, opened 2024-07-28
Update payment methods icons

axpoems:
* Add new icons for all payment methods.
* Add missing icons for some payment methods, which did not have an icon yet.
* Update styles in payment methods page.

image


djing-chan (approved):
utACK

PR #2511, opened 2024-07-28
Allow OnChain, LN or both as settlment methods

axpoems:
* Remove allow LN checkbox.
* Add grid pane for multiple selection of Btc payment methods.

image

I will follow-up with more PRs improving this further.


djing-chan (approved):
utACK

PR #2517, opened 2024-07-29
Re-enable republishing of bonded role data and oracle nodes.

HenrikJannsen:
Fix bug with republishing bonded role data


djing-chan (approved):
ACK

PR #2520, opened 2024-07-30
Fix Bisq2 app logo icons

axpoems:
* Fix size of iOS app logo icons following the Apple guidelines.
* General fine tune of the icons.


djing-chan (approved):
utACK

PR #2525, opened 2024-07-31
Improve payment methods page in take offer

axpoems:
* Align styles in trade wizard and take offer wizard for the payment methods page.
* Fix issue with image naming.
* Small refactors.


djing-chan (approved):
utACK

Issue #1768, opened 2024-03-11
Set Bisq Version Globally and Pass to Submodules


alvasw:
We set the Bisq version in many places:  
- in config files: `desktop.conf`,  `oracle_node.conf`, `rest-api.conf`, and `seed_node.conf`  
- in the packaging plugin: https://github.com/bisq-network/bisq2/blob/5ce28113b5710cad4c01bfd36bbfefe7ef0ea229/build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/PackagingPlugin.kt#L25  
- in the desktop module (for the fatJar):  https://github.com/bisq-network/bisq2/blob/ea32c88d0ec77042b18734b837b87a17af204fc7/apps/desktop/desktop-app/build.gradle.kts#L13
- in the DesktopAppLauncher: https://github.com/bisq-network/bisq2/blob/5ce28113b5710cad4c01bfd36bbfefe7ef0ea229/apps/desktop/desktop-app-launcher/src/main/java/bisq/desktop_app_launcher/DesktopAppLauncher.java#L64  

Release Managers must update the version number in all of these files. We could provide an bump version number script (like in Bisq 1) for now. I'd recommend to set the version globally and pass is to all sub-modules long-term.

Relates to #1764.

HenrikJannsen:
Found it also in `gradle.properties`: `version=2.0.0`

HenrikJannsen:
@alvasw Whats the easiest way for gradle scripts to read in the version number? Would a text file `version` with only the version string in it be good? Or better a property file?
For the config we could read in that file at startup and remove it from the config files.
We would though apply one global version to all applications (e.g. if no changes happened in oracle it still would get a new version number).

rodvar:
@alvasw Happy to tackle this one. My approach would be to leave the application version definition ONLY on `gradle.properties`  and then using gradle task and a new static class link it to the java world and make it globally accessible updating all the current usages.

rodvar:
started work on this one first

rodvar:
@alvasw @HenrikJannsen hey guys! have a look at this when you have a moment! :)

 - https://github.com/bisq-network/bisq2/pull/2526
 
 I think we could get rid of version.txt (at least in common) and its usages but we can always do it in a cleanup PR, let me know your thoughts, cheers!

rodvar:
I believe this one can be closed now :)

Issue #2522, opened 2024-07-30
Add ProGuard to decrease Bisq file size and potentially improve performance


HenrikJannsen:
Any dev experienced with ProGuard and interested to add it to our gradle build?

rodvar:
@HenrikJannsen starting work on this one

HenrikJannsen:
The work on that has been conserved in https://github.com/bisq-network/bisq2/tree/Proguard-support but it got dropped so far as the reduction was not as big as hoped and did not justify the added complexity and risks.

PR #2526, opened 2024-07-31
Global versioning

rodvar:
- Implementation of https://github.com/bisq-network/bisq2/issues/1768
 - SPOF of version in gradle.properties
 - common module generates and compiles a BuildVersion static class
 - adapted usages to this new mechanism


rodvar:
Just realized I've missed a replacement that could be done on WebcamAppPlugin.kt
But I see that code has also its own version.txt file.
Let's decide what to do there after the first review, or of we create a different PR for it

common/src/main/java/bisq/common/application/ApplicationVersion.java
```diff
@@ -17,20 +17,16 @@
...
 
-import java.io.IOException;
-
 public class ApplicationVersion {
     private static Version version;
 
     public static Version getVersion() {
         if (version == null) {
             try {
-                String versionString = FileUtils.readStringFromResource("version.txt");
```
djing-chan, line 31: We should also remove the version.txt file

djing-chan (changes requested):
Looks all good but for some reasons the generateInstallers task does not produce a binary at MacOS. No error messages...

HenrikJannsen:
> Just realized I've missed a replacement that could be done on WebcamAppPlugin.kt But I see that code has also its own version.txt file. Let's decide what to do there after the first review, or of we create a different PR for it

WebcamApp has its own version and is best kept as it is. It will not be updated in sync with Desktop.

common/src/main/java/bisq/common/application/ApplicationVersion.java
rodvar: Thanks @djing-chan, I'll fix this and remove the version.txt in common in a tick!

Regarding the binary issue on MacOS, that sounds unrelated to this change.. What it can happen is a cache issue, if you are using Intellij to test this make sure you invalidate caches and restart. On terminal, a gradlew clean should do. 

I have a macbook here I'll give it a test just in case in a few moments, thanks!

rodvar:
@djing-chan Just did a quick test on my macbook pro (M1), fresh install from terminal run `./gradlew clean build --info` and succeded :)

To make sure the binary gets generated and also runs I did `./gradlew :apps:desktop:desktop-app-launcher:run` afterwards to see the desktop app running in prod mode :)

If after checking the potential cache issue I've commented above you still have issues, can you try running an install on terminal maybe with --debug and share the logs?

IMHO this PR is ready to go now :)

common/src/main/java/bisq/common/application/ApplicationVersion.java
djing-chan: I tried again with invalidating cache but still the generateInstallers dont create the binaries. When switching back to main it works again. Maybe there is some issue with the source sourceSets?
Have you tried it yourself if you can build a binary?

djing-chan (changes requested):

rodvar:
I did and it would run the app (MacBook Pro M1). Which macOS version and hw
are you using specifically?

Let's ask Henrik to build it too he's got an M3

djing-chan:
I debugged the version in desktop-app and the verison in the jar files manifest is not defined:
```
Manifest-Version: 1.0
Implementation-Title: desktop-app
Implementation-Version: unspecified
Main-Class: bisq.desktop_app.DesktopApp
```

I think the assignment of `version = project.version` does not work with the give multi layered project structure. 
I am not that experienced with gradle to be able to give advice how to fix that.

Also in dekstop-app-launcher:
```
packaging {
    name.set("Bisq2")
    version.set(project.version.toString())
}
```
seems to be the problem. 
When I change to:
```
packaging {
    name.set("Bisq2")
    version.set("2.1.1")
}
```

I get the binary built.

Can you debug to see if you get the version set?

`./gradlew :apps:desktop:desktop-app-launcher:generateInstallers` is the task for the binary installer

rodvar:
> I debugged the version in desktop-app and the verison in the jar files manifest is not defined:
> 
> ```
> Manifest-Version: 1.0
> Implementation-Title: desktop-app
> Implementation-Version: unspecified
> Main-Class: bisq.desktop_app.DesktopApp
> ```
> 
> I think the assignment of `version = project.version` does not work with the give multi layered project structure. I am not that experienced with gradle to be able to give advice how to fix that.
> 
> Also in dekstop-app-launcher:
> 
> ```
> packaging {
>     name.set("Bisq2")
>     version.set(project.version.toString())
> }
> ```
> 
> seems to be the problem. When I change to:
> 
> ```
> packaging {
>     name.set("Bisq2")
>     version.set("2.1.1")
> }
> ```
> 
> I get the binary built.
> 
> Can you debug to see if you get the version set?
> 
> `./gradlew :apps:desktop:desktop-app-launcher:generateInstallers` is the task for the binary installer

I see what you mean now, thank you @djing-chan I'm working on it and will push a fix soon!

djing-chan:
Does `./gradlew :apps:desktop:desktop-app-launcher:generateInstallers` work on main for you? If not what are the error messages, if none run with some of the debug options to get more info.

Yes one cannot reference to the real root project by `rootproject` or `parent`, the `apps` seems to be the root there.
I guess one option might be to read out the value as property and reference to it using the directory structure, but not that nice... maybe there is a better approach. @alvasw Do you have a suggestion?

rodvar:
> Does `./gradlew :apps:desktop:desktop-app-launcher:generateInstallers` work on main for you? If not what are the error messages, if none run with some of the debug options to get more info.
> 
> Yes one cannot reference to the real root project by `rootproject` or `parent`, the `apps` seems to be the root there. I guess one option might be to read out the value as property and reference to it using the directory structure, but not that nice... maybe there is a better approach. @alvasw Do you have a suggestion?

that's correct, I have a fix almost finished where there is shared code to read the gradle.properties in the root from this includedBuild sub projects, I'll send it in a bit.

Answering your request above, if I run `./gradlew :apps:desktop:desktop-app-launcher:generateInstallers` on main branch I get

```
FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':apps:desktop:desktop-app-launcher:generateInstallers'.
> Failed to query the value of task ':apps:desktop:desktop-app-launcher:generateInstallers' property 'runtimeImageDirectory'.
   > Cannot find a Java installation on your machine matching this tasks requirements: {languageVersion=22, vendor=AZUL, implementation=vendor-specific} for LINUX on x86_64.
      > No locally installed toolchains match and toolchain download repositories have not been configured.

```

rodvar:
@djing-chan I spent a significant amount of time trying to modularise the function I've added in 2 different files to do the same thing (see last push). This allows us to read the gradle.properties from a composite-built submodule. 

I couldn't get it to work without copying the same function in both files, it all boils down to the use of includeBuild(), I guess it's not meant to share versioning or anything really from the root. 

In my opinion, it isn't worth changing the overall gradle approach which is a big change just to share a version, so I'm providing this code redundancy that does the job and achieves what we want which is having a single point of failure for the versioning (now you truly only need to change versioning on gradle.properties which was the whole purpose of this)

If anybody finds a better way I'm all ears :)
Let me know your thoughts!

djing-chan:
> Answering your request above, if I run `./gradlew :apps:desktop:desktop-app-launcher:generateInstallers` on main branch I get
> 
> ```
> FAILURE: Build failed with an exception.
> 
> * What went wrong:
> Could not determine the dependencies of task ':apps:desktop:desktop-app-launcher:generateInstallers'.
> > Failed to query the value of task ':apps:desktop:desktop-app-launcher:generateInstallers' property 'runtimeImageDirectory'.
>    > Cannot find a Java installation on your machine matching this tasks requirements: {languageVersion=22, vendor=AZUL, implementation=vendor-specific} for LINUX on x86_64.
>       > No locally installed toolchains match and toolchain download repositories have not been configured.
> ```

I had that recently myself after updating from a pre java22 project. after a second run it worked. i think it was just not finished downloading the sdk or other dependeincies. If you still get the issue, try to clean not only the root project but also the projects in the apps as those are not cleaned/built when calling from root. If that still does not work maybe delete the gradle cache dir at your OS user directory, but then things will get slow afterwards until cache is filled again. But at least once I needed to do that to tame gradle...

djing-chan (approved):
ACK

rodvar:
> > Answering your request above, if I run `./gradlew :apps:desktop:desktop-app-launcher:generateInstallers` on main branch I get
> > ```
> > FAILURE: Build failed with an exception.
> > 
> > * What went wrong:
> > Could not determine the dependencies of task ':apps:desktop:desktop-app-launcher:generateInstallers'.
> > > Failed to query the value of task ':apps:desktop:desktop-app-launcher:generateInstallers' property 'runtimeImageDirectory'.
> >    > Cannot find a Java installation on your machine matching this tasks requirements: {languageVersion=22, vendor=AZUL, implementation=vendor-specific} for LINUX on x86_64.
> >       > No locally installed toolchains match and toolchain download repositories have not been configured.
> > ```
> 
> I had that recently myself after updating from a pre java22 project. after a second run it worked. i think it was just not finished downloading the sdk or other dependeincies. If you still get the issue, try to clean not only the root project but also the projects in the apps as those are not cleaned/built when calling from root. If that still does not work maybe delete the gradle cache dir at your OS user directory, but then things will get slow afterwards until cache is filled again. But at least once I needed to do that to tame gradle...

Thank you @djing-chan I'll try this sometime later today, and thanks for reviewing my PRs and all your useful comments!

rodvar:
> > > Answering your request above, if I run `./gradlew :apps:desktop:desktop-app-launcher:generateInstallers` on main branch I get
> > > ```
> > > FAILURE: Build failed with an exception.
> > > 
> > > * What went wrong:
> > > Could not determine the dependencies of task ':apps:desktop:desktop-app-launcher:generateInstallers'.
> > > > Failed to query the value of task ':apps:desktop:desktop-app-launcher:generateInstallers' property 'runtimeImageDirectory'.
> > >    > Cannot find a Java installation on your machine matching this tasks requirements: {languageVersion=22, vendor=AZUL, implementation=vendor-specific} for LINUX on x86_64.
> > >       > No locally installed toolchains match and toolchain download repositories have not been configured.
> > > ```
> > 
> > 
> > I had that recently myself after updating from a pre java22 project. after a second run it worked. i think it was just not finished downloading the sdk or other dependeincies. If you still get the issue, try to clean not only the root project but also the projects in the apps as those are not cleaned/built when calling from root. If that still does not work maybe delete the gradle cache dir at your OS user directory, but then things will get slow afterwards until cache is filled again. But at least once I needed to do that to tame gradle...
> 
> Thank you @djing-chan I'll try this sometime later today, and thanks for reviewing my PRs and all your useful comments!

Found it, whilst working on https://github.com/bisq-network/bisq2/issues/2522 - the codebase has a validation to use only Azul OpenJDK 22.x for the installer binaries so I've just downloaded it and it works :)

would be good to add this to the wiki eventually

HenrikJannsen:
Can you make a PR with adding it to the dev md file?

Issue #1358, opened 2023-11-11
Remove usage of MultiStyleLabelPane and use TextFlow instead


djing-chan:

rodvar:
starting work on this one

rodvar:
I believe this one can be closed now :)

PR #2539, opened 2024-08-01
Replace custom Multi Pane class with JavaFX TextFlow

rodvar:
- Refactor requested in https://github.com/bisq-network/bisq2/issues/1358
 - Replaced the custom multi label pane with TextFlow and an utilitary class to reuse the setup
 - apply changes on both usages
 - Allow trades to go through without checking reputation in DevMode


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/take_offer/review/TakeOfferReviewView.java
```diff
@@ -189,6 +189,10 @@ protected void onViewAttached() {
         takeOfferStatusPin = EasyBind.subscribe(model.getTakeOfferStatus(), this::showTakeOfferStatusFeedback);
     }
 
+    private void updatePriceTextFlow(String priceText) {
```
djing-chan: Why extract this to a method (single use)?

apps/desktop/desktop/src/main/java/bisq/desktop/util/TextFlowUtils.java
```diff
@@ -0,0 +1,47 @@
+package bisq.desktop.util;
```
djing-chan: I think the`bisq.desktop.components.controls` package would be a better place. Maybe a util subpackage there. We have also `bisq.desktop.common.utils` but that is more for not UI controls specific stuff.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/take_offer/review/TakeOfferReviewView.java
```diff
@@ -105,47 +105,47 @@ class TakeOfferReviewView extends View<StackPane, TakeOfferReviewModel, TakeOffe
 
         rowIndex++;
         Label priceDescription = new Label(Res.get("bisqEasy.takeOffer.review.price.price"));
-        priceDescription.getStyleClass().add(descriptionStyle);
+        priceDescription.getStyleClass().add(DESCRIPTION_STYLE);
         content.add(priceDescription, 0, rowIndex);
 
-        price = new MultiStyleLabelPane();
-        price.getStyleClass().add(valueStyle);
+        price = new TextFlow();
```
djing-chan, line 111: Actually I don't see a reason why TextFlow is needed here. We have a rather static layout without expected linebreaks.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/trade_wizard/review/TradeWizardReviewView.java
```diff
@@ -124,7 +125,7 @@ class TradeWizardReviewView extends View<StackPane, TradeWizardReviewModel, Trad
         priceDescription.getStyleClass().add(descriptionStyle);
         gridPane.add(priceDescription, 0, rowIndex);
 
-        price = new MultiStyleLabelPane();
+        price = new TextFlow();
```
djing-chan, line 128: Actually I don't see a reason why TextFlow is needed here. We have a rather static layout without expected linebreaks.

djing-chan (changes requested):

djing-chan (approved):
ACK, beside small nits

apps/desktop/desktop/src/main/java/bisq/desktop/util/TextFlowUtils.java
rodvar: I agree, I couldn't find a better place when I was doing it I'll refactor it now

rodvar:
applied review suggestions

djing-chan (approved):
utACK

Issue #2538, opened 2024-08-01
Handle shutdown when a sent message is still in a pending state


HenrikJannsen:
We should make clear to the user that when sending a message the app should not be closed quickly afterwards as the message sending might fail. It takes a while until the message gets stored in mailbox (after direct connection fails, that can take up to 2 minutes, but usually its 5-20 sec).
We could use that popup to explain the message delivery icons which are present in the chats and trade protocol messages.
There the user can see if the message is still in connection-establishing phase, message sending phase, arrived at peer or stored in mailbox. This should help users to know when its safe to close the app.

See https://github.com/bisq-network/bisq2/discussions/2537 for more context details

HenrikJannsen:
A better solution is to make sure pending messages are stored as mailbox at shutdown.

PR #2555, opened 2024-08-03
Handle pending messages at shutdown

HenrikJannsen:
Implements https://github.com/bisq-network/bisq2/issues/2538

At shutdown we store all pending messages in mailbox. This prevent the issue when a user quickly close the app after sending a message. Before only after restart the message would be resent. Now the peer receives the message as mailbox message immediately.


djing-chan (approved):
utACK

PR #2559, opened 2024-08-03
Introduce payment columns in offerbook listview

### LINKED ISSUE #2316: Add payment method column in offer list view on right column of offerbook
User suggested having
| peer profile | price | usd amount | payment method |

axpoems:
* Add icons for custom methods.
* Add fiat payment methods column in offerbook listview.
* Add Btc settlement methods column in offerbook listview.

image

Resolves #2316


djing-chan (approved):
utACK

Issue #1474, opened 2023-12-04
Validation of transaction ID


hiciefte:
When filling in a transaction ID it should be validated to reduce unnecessary mediation cases because of typos.

e.g.
https://github.com/bisq-network/bisq2/blob/225b4db2265616049f38dc07aa5704bfc89adf46/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3.java#L122

djing-chan:
We should permit non-BTC transactions for more flexibility (e.g. traders use LN or other currencies). But we should show a warning popup if we detect an invalid BTC transactions.

rodvar:
@djing-chan same for this one, happy to take it using the approach I've commented on https://github.com/bisq-network/bisq2/issues/1473

rodvar:
Starting work on this one

Issue #1473, opened 2023-12-04
Validation of Bitcoin address


hiciefte:
When filling in a Bitcoin address it should be validated to reduce unnecessary mediation cases because of typos.

e.g.
https://github.com/bisq-network/bisq2/blob/225b4db2265616049f38dc07aa5704bfc89adf46/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/BuyerState2.java#L126

rodvar:
@HenrikJannsen @djing-chan I'll love to look into this one. My approach would be

 1. creating an util component that can recognize if a string is a valid BTC address or not
 2. Using that on every place I can find in the UI where we ask for BTC address
 3.  Show a warning/pop up when a string is not a BTC address (using the right i18n) -> I'm sure there is some reusable component for this right? 
 
 LMK! Cheers!

djing-chan:
We should use the validation of the MaterialFextField. But we should be tolerant to allow users to go with non BTC address/tx as they might use another currency or L2 (liquid, RSK). 
We should not import any new dependency for the check from any btc library as we are not decided yet which bitcoin lib we will use. I guess some regex should do it...

rodvar:
@djing-chan yep, we are on the same page - - I'll get into this one after the ones I'm starting to work on now, thanks!

rodvar:
Starting work on this one

PR #2564, opened 2024-08-05
Feature/btc address tx hints

rodvar:
**Related issues**

 - Implementation of https://github.com/bisq-network/bisq2/issues/1473
 - Implementation of https://github.com/bisq-network/bisq2/issues/1474

**Approach**

 - Pattern based validator for BTC addresses and txIds. Unit test provided
 - Warning pop up on every user input for these strings, allowing to proceed if the user wants to dismiss the warning.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/BuyerState1a.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/BuyerState1a.java
@@ -21,6 +21,7 @@ import bisq.account.payment_method.BitcoinPaymentRail;
 import bisq.bisq_easy.NavigationTarget;
 import bisq.chat.bisqeasy.open_trades.BisqEasyOpenTradeChannel;
 import bisq.common.observable.Pin;
+import bisq.common.validation.BitcoinDataValidation;
 import bisq.common.util.ExceptionUtil;
 import bisq.desktop.ServiceProvider;
 import bisq.desktop.common.threading.UIScheduler;
@@ -32,7 +33,10 @@ import bisq.desktop.components.controls.BisqIconButton;
 import bisq.desktop.components.controls.BisqTooltip;
 import bisq.desktop.components.controls.MaterialTextField;
 import bisq.desktop.components.controls.WrappingText;
+import bisq.desktop.components.controls.validator.BitcoinDataValidator;
 import bisq.desktop.components.controls.validator.SettableErrorValidator;
+import bisq.desktop.components.controls.validator.ValidatorBase;
+import bisq.desktop.components.overlay.Popup;
 import bisq.desktop.webcam.WebcamAppModel;
 import bisq.desktop.webcam.WebcamAppService;
 import bisq.i18n.Res;
@@ -138,12 +142,23 @@ public class BuyerState1a extends BaseState {
         }
 
         private void onSend() {
-            String name = model.getBisqEasyTrade().getContract().getBaseSidePaymentMethodSpec().getPaymentMethod().getPaymentRail().name();
-            String key = "bisqEasy.tradeState.info.buyer.phase1a.tradeLogMessage." + name;
-            sendTradeLogMessage(Res.encode(key, model.getChannel().getMyUserIdentity().getUserName(), model.getBitcoinPaymentData().get()));
-            bisqEasyTradeService.buyerSendBitcoinPaymentData(model.getBisqEasyTrade(), model.getBitcoinPaymentData().get());
+            String btcAddress = model.getBitcoinPaymentData().get();
+            if (BitcoinDataValidation.validateWalletAddressHash(btcAddress)) {
+                send(btcAddress);
+            } else {
+                new Popup().warning(Res.get("bisqEasy.takeOffer.btcAddress.warning"))
+                        .actionButtonText(Res.get("bisqEasy.takeOffer.btcAddress.warning.proceed"))
+                        .onAction(() -> send(btcAddress))
+                        .show();
+            }
         }
 
+        private void send(String btcAddress) {
+            String name = model.getBisqEasyTrade().getContract().getBaseSidePaymentMethodSpec().getPaymentMethod().getPaymentRail().name();
+            String key = "bisqEasy.tradeState.info.buyer.phase1a.tradeLogMessage." + name;
+            sendTradeLogMessage(Res.encode(key, model.getChannel().getMyUserIdentity().getUserName(), btcAddress));
+            bisqEasyTradeService.buyerSendBitcoinPaymentData(model.getBisqEasyTrade(), btcAddress);
+        }
         void onOpenWalletHelp() {
             Navigation.navigateTo(NavigationTarget.WALLET_GUIDE);
         }
@@ -290,6 +305,7 @@ public class BuyerState1a extends BaseState {
         private final Timeline webcamStateAnimationTimeline;
         private final Region webcamStateMarkerLine;
         private final SettableErrorValidator webcamStateValidator;
+        private final BitcoinDataValidator walletAddressValidator;
         private Subscription qrCodeDetectedFromWebcamPin, webcamStateVisiblePin;
 
         private View(Model model, Controller controller) {
@@ -315,6 +331,17 @@ public class BuyerState1a extends BaseState {
 
             webcamStateMaterialTextField = new MaterialTextField(Res.get("bisqEasy.tradeState.info.buyer.phase1a.scanQrCode.webcamState.description"));
             webcamStateValidator = new SettableErrorValidator();
+            walletAddressValidator = new BitcoinDataValidator() {
+                @Override
+                protected String getData() {
+                    return model.getBitcoinPaymentData().get();
+                }
+
+                @Override
+                protected BitcoinDataType getType() {
+                    return BitcoinDataType.WALLET_ADDRESS;
+                }
+            };
             webcamStateMaterialTextField.setValidators(webcamStateValidator);
             webcamStateMaterialTextField.getTextInputControl().setEditable(true);
             webcamStateMaterialTextField.showIcon();
@@ -357,6 +384,7 @@ public class BuyerState1a extends BaseState {
             bitcoinPayment.setDescription(model.getBitcoinPaymentDescription());
             bitcoinPayment.setPromptText(model.getBitcoinPaymentPrompt());
             bitcoinPayment.setHelpText(model.getBitcoinPaymentHelp());
+            bitcoinPayment.setValidators(walletAddressValidator);
 
             bitcoinPayment.textProperty().bindBidirectional(model.getBitcoinPaymentData());
             scanQrCodeButton.visibleProperty().bind(model.getScanQrCodeButtonVisible());
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
@@ -20,6 +20,7 @@ package bisq.desktop.main.content.bisq_easy.open_trades.trade_state.states;
 import bisq.account.payment_method.BitcoinPaymentRail;
 import bisq.chat.bisqeasy.open_trades.BisqEasyOpenTradeChannel;
 import bisq.common.data.Pair;
+import bisq.common.validation.BitcoinDataValidation;
 import bisq.common.util.MathUtils;
 import bisq.desktop.CssConfig;
 import bisq.desktop.ServiceProvider;
@@ -31,6 +32,9 @@ import bisq.desktop.common.utils.KeyHandlerUtil;
 import bisq.desktop.components.controls.BisqTooltip;
 import bisq.desktop.components.controls.MaterialTextField;
 import bisq.desktop.components.controls.WrappingText;
+import bisq.desktop.components.controls.validator.BitcoinDataValidator;
+import bisq.desktop.components.controls.validator.ValidatorBase;
+import bisq.desktop.components.overlay.Popup;
 import bisq.desktop.overlay.OverlayController;
 import bisq.i18n.Res;
 import bisq.settings.SettingsService;
@@ -154,11 +158,26 @@ public class SellerState3a extends BaseState {
             String name = model.getBisqEasyTrade().getContract().getBaseSidePaymentMethodSpec().getPaymentMethod().getPaymentRail().name();
             String proof = Res.get("bisqEasy.tradeState.info.seller.phase3a.tradeLogMessage.paymentProof." + name);
             String userName = model.getChannel().getMyUserIdentity().getUserName();
-            if (model.getPaymentProof().get() == null) {
+            String txId = model.getPaymentProof().get();
+            if (txId == null) {
+                sellerConfirmedBtcSent(null, userName, proof);
+            } else {
+                if (BitcoinDataValidation.validateTransactionId(txId)) {
+                    sellerConfirmedBtcSent(txId, userName, proof);
+                } else {
+                    new Popup().warning(Res.get("bisqEasy.tradeState.info.seller.phase3a.txId.warning"))
+                            .actionButtonText(Res.get("bisqEasy.tradeState.info.seller.phase3a.txId.warning.proceed"))
+                            .onAction(() -> { sellerConfirmedBtcSent(txId, userName, proof); }).show();
+                }
+            }
+        }
+
+        private void sellerConfirmedBtcSent(String txId, String userName, String proof) {
+            if (txId == null) {
                 sendTradeLogMessage(Res.encode("bisqEasy.tradeState.info.seller.phase3a.tradeLogMessage.noProofProvided", userName));
             } else {
                 sendTradeLogMessage(Res.encode("bisqEasy.tradeState.info.seller.phase3a.tradeLogMessage",
-                        userName, proof, model.getPaymentProof().get()));
+                        userName, proof, txId));
             }
             bisqEasyTradeService.sellerConfirmBtcSent(model.getBisqEasyTrade(), Optional.ofNullable(model.getPaymentProof().get()));
         }
@@ -236,6 +255,7 @@ public class SellerState3a extends BaseState {
     public static class View extends BaseState.View<Model, Controller> {
         private static final Interpolator INTERPOLATOR = Interpolator.SPLINE(0.25, 0.1, 0.25, 1);
 
+        private final BitcoinDataValidator txIdValidator;
         private final Button sentButton;
         private final MaterialTextField paymentProof;
         private final WrappingText sendBtcHeadline, fiatReceiptConfirmed;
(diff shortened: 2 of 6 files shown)


rodvar:
@hiciefte @djing-chan @HenrikJannsen I've created a draft with the basics of the solution.

It's being applied only for the BTC input of the taker in BisqEasy trade process.

Before going ahead and applying it everywhere else and completing the i18n I would like your input on the solution and also the texts used.

Thanks!

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/BuyerState1a.java
```diff
@@ -138,10 +140,21 @@ private void unBindQrCodePins() {
         }
 
         private void onSend() {
-            String name = model.getBisqEasyTrade().getContract().getBaseSidePaymentMethodSpec().getPaymentMethod().getPaymentRail().name();
-            String key = "bisqEasy.tradeState.info.buyer.phase1a.tradeLogMessage." + name;
-            sendTradeLogMessage(Res.encode(key, model.getChannel().getMyUserIdentity().getUserName(), model.getBitcoinPaymentData().get()));
-            bisqEasyTradeService.buyerSendBitcoinPaymentData(model.getBisqEasyTrade(), model.getBitcoinPaymentData().get());
+            String btcAddress = model.getBitcoinPaymentData().get();
+            Runnable proceed = () -> {
```
djing-chan: Why you wrap that code in a Runnable?

i18n/src/main/resources/bisq_easy.properties
```diff
@@ -314,6 +314,8 @@ bisqEasy.takeOffer.tradeLogMessage={0} has sent a message for taking {1}''s offe
 
 bisqEasy.takeOffer.noMediatorAvailable.warning=There is no mediator available. You have to use the support chat instead.
 bisqEasy.takeOffer.makerBanned.warning=The maker of this offer is banned. Please try to use a different offer.
+bisqEasy.takeOffer.btcAddress.warning=The input Bitcoin address does not look like a valid Bitcoin address. You can still proceed, if unsure please click "Close" and verify the Bitcoin Address you have entered
```
djing-chan: I would phrase it like: 
The Bitcoin address you entered appears to be invalid. If you are sure that the address is valid you can ignore that warning and proceed.

i18n/src/main/resources/bisq_easy.properties
```diff
@@ -314,6 +314,8 @@ bisqEasy.takeOffer.tradeLogMessage={0} has sent a message for taking {1}''s offe
...
+bisqEasy.takeOffer.btcAddress.warning.proceed=Proceed anyways
```
djing-chan: I would use: Ignore warning

common/src/main/java/bisq/common/util/BitcoinUtils.java
```diff
@@ -0,0 +1,58 @@
...
+    private static final Pattern BECH32_PATTERN = Pattern.compile("^(bc1|tb1)[a-z0-9]{25,39}$");
+    private static final Pattern TXID_PATTERN = Pattern.compile("^[a-fA-F0-9]{64}$");
+
+    /**
+     * Checks if the given string is a Bitcoin address.
+     *
+     * @param address The string to be checked.
+     * @return True if it is a Bitcoin address, false otherwise.
+     */
+    public static boolean isWalletAddress(@NotNull String address) {
```
djing-chan: I would call it `isValid` and I would split the class in a BitcoinAddressValidator and BitcoinTxValidator

common/src/main/java/bisq/common/util/BitcoinUtils.java
```diff
@@ -0,0 +1,58 @@
+package bisq.common.util;
```
djing-chan: I think it should be moved to validator package

common/src/main/java/bisq/common/util/BitcoinUtils.java
djing-chan: Instead of returning a boolean we could throw an exception with information what went wrong.

NetworkDataValidation uses that approach:

   public static void validateBtcTxId(String txId) {
        checkArgument(txId.length() == 64,
                "BTC txId must be 64 characters. txId=" + txId);
    }

    public static void validateBtcAddress(String address) {
        checkArgument(address.length() >= 26 && address.length() <= 90,
                "BTC address not in expected size. address=" + address);
    }

Not sure if that is better in that case... just adding it for consideration

djing-chan:
If the btc address is invalid we should marke the text field red. The MaterialTextField has support for Validators.

Not sure if we should also do a validation at the seller side at state 3 when we display the btc address. We could show a warning popup and mark the textfield with the error style. 

At the tx input there is no validation yet, right.

djing-chan (changes requested):

rodvar:
> If the btc address is invalid we should marke the text field red. The MaterialTextField has support for Validators.
> 
> Not sure if we should also do a validation at the seller side at state 3 when we display the btc address. We could show a warning popup and mark the textfield with the error style.
> 
> At the tx input there is no validation yet, right?

correct, the validation for tx is done but I haven't given any example linked to the UI
I've been looking into the validators implementation. Would it support a warning marking the field yellow or red like you said, without blocking the user from continuing?

common/src/main/java/bisq/common/util/BitcoinUtils.java
rodvar: Thanks @djing-chan , I'll have a look if this is worth when applying in the rest of the use cases

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/BuyerState1a.java
rodvar: quick way of reusing it twice in the same method, I'll move it to a private method in my next commit

common/src/main/java/bisq/common/util/BitcoinUtils.java
rodvar: Not sure about this one, I thought about it as a generic way to identifying wallet address and tx id to be reusable. If I were to implement a validator, the validator would be calling those methods as well.

 Doesn't it make more sense as is, as it could be used in many areas to make sure a wallet address / transaction is valid?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/BuyerState1a.java
djing-chan: Ah, have not seen the second usage... but yes I guess a private method is better.

djing-chan:
> > If the btc address is invalid we should marke the text field red. The MaterialTextField has support for Validators.
> > Not sure if we should also do a validation at the seller side at state 3 when we display the btc address. We could show a warning popup and mark the textfield with the error style.
> > At the tx input there is no validation yet, right?
> 
> correct, the validation for tx is done but I haven't given any example linked to the UI I've been looking into the validators implementation. Would it support a warning marking the field yellow or red like you said, without blocking the user from continuing?

Currently there is only error state in red. 
I think we should use the input validation style, and when clicking the action button we show the waring popup and allow the user to proceed. I guess red is ok as we detect invalid inputs. If user ignores that bc they use for instance another currency, its their "override" but our validation for btc is still correctly failing.
The validation is not triggered at input but at button click.

rodvar:
> > > If the btc address is invalid we should marke the text field red. The MaterialTextField has support for Validators.
> > > Not sure if we should also do a validation at the seller side at state 3 when we display the btc address. We could show a warning popup and mark the textfield with the error style.
> > > At the tx input there is no validation yet, right?
> > 
> > 
> > correct, the validation for tx is done but I haven't given any example linked to the UI I've been looking into the validators implementation. Would it support a warning marking the field yellow or red like you said, without blocking the user from continuing?
> 
> Currently there is only error state in red. I think we should use the input validation style, and when clicking the action button we show the waring popup and allow the user to proceed. I guess red is ok as we detect invalid inputs. If user ignores that bc they use for instance another currency, its their "override" but our validation for btc is still correctly failing. The validation is not triggered at input but at button click.

Cool then, I'll add in also the validator to make the input read pre-warning the user. Then if the user insists the current warning I added will show up, and they will still have the posibility to click "Ignore warning" and proceed.

After I finish with this is just a matter of identifying all the use cases (txIds and places where btc address is used as input). if you have it at the top of your head please shout them out otherwise I'll search for them. Thanks for all the help as usual @djing-chan !!!

djing-chan:
Its just used in the trade process when btc is selected.
for LN we should add also validation if it makes sense (invoices have many options but it guess there is some structure which can be validated).  But lets do that in a new PR...

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
```diff
@@ -154,12 +157,27 @@ private void onBtcSent() {
             String name = model.getBisqEasyTrade().getContract().getBaseSidePaymentMethodSpec().getPaymentMethod().getPaymentRail().name();
             String proof = Res.get("bisqEasy.tradeState.info.seller.phase3a.tradeLogMessage.paymentProof." + name);
             String userName = model.getChannel().getMyUserIdentity().getUserName();
-            if (model.getPaymentProof().get() == null) {
+            String txId = model.getPaymentProof().get();
+            if (txId == null) {
                 sendTradeLogMessage(Res.encode("bisqEasy.tradeState.info.seller.phase3a.tradeLogMessage.noProofProvided", userName));
```
djing-chan: sendTradeLogMessage should be moved into sellerConfirmedBtcSent.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
```diff
@@ -154,12 +157,27 @@ private void onBtcSent() {
...
+                sellerConfirmedBtcSent();
             } else {
-                sendTradeLogMessage(Res.encode("bisqEasy.tradeState.info.seller.phase3a.tradeLogMessage",
-                        userName, proof, model.getPaymentProof().get()));
+                if (BitcoinUtils.isTransactionId(txId)) {
+                    sendTradeLogMessage(Res.encode("bisqEasy.tradeState.info.seller.phase3a.tradeLogMessage",
```
djing-chan: Here the sellerConfirmedBtcSent call is missing

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
```diff
@@ -319,6 +337,13 @@ protected void onViewAttached() {
             qrCodeImageView.setImage(model.getSmallQrCodeImage());
 
             paymentProof.textProperty().bindBidirectional(model.getPaymentProof());
+            paymentProof.setValidators(new ValidatorBase() {
```
djing-chan: Better to create a custom Validator class for reuse.

apps/desktop/desktop/src/main/java/bisq/desktop/webcam/WebcamProcessLauncher.java
```diff
@@ -45,7 +45,12 @@ public WebcamProcessLauncher(String baseDir) {
     public CompletableFuture<Process> start(int port) {
         return CompletableFuture.supplyAsync(() -> {
             try {
-                String version = FileUtils.readStringFromResource("webcam-app/version.txt");
+                String version = "1.0.0";
+                try {
+                    version = FileUtils.readStringFromResource("webcam-app/version.txt");
```
djing-chan: The version file is expected to exist (in dev mode it requires a gradle task to get called but in binary it must exist). so failing the whole code block seems correct to me instead of only logging an error.

i18n/src/main/resources/bisq_easy_de.properties
```diff
@@ -279,7 +279,8 @@ bisqEasy.takeOffer.tradeLogMessage={0} hat eine Nachricht gesendet, um das Angeb
 
 bisqEasy.takeOffer.noMediatorAvailable.warning=Es steht kein Mediator zur Verfügung. Sie müssen stattdessen den Support-Chat verwenden.
```
djing-chan: `Ã¼` seems you file encoding is not set to UTF8. Also at other following changes.
Btw. it is not needed that devs do the translation (though nice to have). The translation team should pick up (delayed) missing translations. 
Devs usually only set the english versions.

common/src/main/java/bisq/common/util/BitcoinUtils.java
djing-chan: I still would prefer a rename here. isWalletAddress has for devs coming from a BitcoinJ context a certain meaning (a add contained in the wallet). As we specifically check validity i would use isValid or the like...

common/src/main/java/bisq/common/util/BitcoinUtils.java
djing-chan: As we have a validation package and the class is doing validation it would be for me logical to look up for that class there.
In Utils we usually dont have validation code but util-specific stuff. Validation is a general concept worth to be grouped apart from rather unspecific utils.

djing-chan (changes requested):

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
rodvar: to be addressed in the next amend

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
rodvar: to be addressed in the next amend - thanks!

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
rodvar: I'll try to make one that encapsulates both usages then

apps/desktop/desktop/src/main/java/bisq/desktop/webcam/WebcamProcessLauncher.java
rodvar: I was not supposed to push that change, it's always throwing exceptions when I use the camera locally...
I'll revert this in the next amend

i18n/src/main/resources/bisq_easy_de.properties
rodvar: that's cool, sure I'll revert this in the next amend

common/src/main/java/bisq/common/util/BitcoinUtils.java
rodvar: I see, I'll think about it for the next amend

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
rodvar: done, please have a look at the latest push

djing-chan: > I'll try to make one that encapsulates both usages then

It is better to have one validator for one thing. UI controls can take multiple validators. BitcoinAddressValidator, BitcoinTransactionValidator,...

apps/desktop/desktop/src/main/java/bisq/desktop/webcam/WebcamProcessLauncher.java
djing-chan: You need to run the task described in the readme in the webcam module if u want to try it out

common/src/main/java/bisq/common/util/BitcoinUtils.java
djing-chan: There have been a refacotring PR in the meantime for common to clean up the utils package. Be sure to have rebased on main

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
```diff
@@ -250,6 +270,18 @@ public static class View extends BaseState.View<Model, Controller> {
         private View(Model model, Controller controller) {
             super(model, controller);
 
+            txIdValidator = new BitcoinDataValidator() {
```
djing-chan, line 273: I think this just pollutes the code here. 
`private final BitcoinTxValidator bitcoinTxValidator = new BitcoinTxValidator();`
would be much more concise.

Of course such things are not hard science and diff. ppl will have diff. opinions... just posting mine. Not a showstopper for ACK...

djing-chan (changes requested):
I forgot in the previous reviews to mention that we must check if bitcoin or LN is used and only apply the validator then. You could add a dummy validator for LN already which just always succeeds, so the client code will not require changes later.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
rodvar: @djing-chan thanks for the review. I agree that this last fine tuning just depends on devs preferences.
The `pollution` you are talking about is quite used in the current Bisq2 codebase so I thought its something generally accepted so I think we are good to go.

However I see that you requested changes, is this related to Lightning ? Does Bisq-Easy support lightning invoices? I thought you mentioned in a previous comment that we should leave that for another PR. If the current implementation doesn't break anything then I think we should go as is and create a separate issue?

Let me know your thoughts!

HenrikJannsen: Yes we need at least only do the BTC checks if LN is not used. And yes we support LN now. Have you done some test trades?

rodvar: @HenrikJannsen I've done test trades with bitcoin only (bitcoin addresses and txIds)  - I can try do some lightning to see how it works in a few mins

HenrikJannsen: But I can do that as well in a follow up PR... Will merge so that it makes it into the release...

HenrikJannsen (approved):
utACK

djing-chan (approved):
utACK

PR #2572, opened 2024-08-06
Fix recursive call of findScrollbar

HenrikJannsen:
The correct scrollbar visibility and layout for the open trades list when switching between 3 and 4 trades is still not working correctly. It will be addressed in a separate commit


djing-chan (approved):
ACK

PR #2578, opened 2024-08-07
Refactoring: Rename classes, packages and move classes.

HenrikJannsen:
The util package was over-crowded and filled with non-util classes. Now we keep the util class to a relevant package if it exists.


djing-chan (approved):
utACK

PR #2583, opened 2024-08-07
Make Bisq Easy offer message more visual

### LINKED ISSUE #1929: Use payment methods logos instead of text in Bisq Easy offer message
image

For instance, in this case, instead of "Wise", we should show the Wise icon.
Alongside with this, we could add a filter for payment methods.

axpoems:
* Add new design for Bisq Easy offer messages.
* Add new interchangeable icon.

Resolves #1929 and #1918.


HenrikJannsen (approved):
utACK

PR #2585, opened 2024-08-08
Finalize validation for btc addresses/tx and LN invoice/preimage.

HenrikJannsen:
LN invoice/preimage is just a skeleton without real implementation.


rodvar (approved):
utACK - All my changes are in, with the optional suggestions @djing-chan made

PR #2621, opened 2024-08-12
Ensure single newline when Shift+Enter

### LINKED ISSUE #2315: Shift-Enter in chat will sometimes jump more than one newline
This is a pet-peeve since I do not use Shift-Enter that often anyway, and this only happens "after a while" that Bisq2 has been running, but maybe it's just a simple overlook that is very easy to fix?
I couldn't report any specific event that makes it so, except that it won't happen right after starting Bisq2, but when it does happen, now for example, Bisq2 has been running on linux for a while, and instead of jumping to the next line it will jump 4 times, and I have to backspace. Once I've seen it jump 11 lines.
In all these cases, restarting Bisq2 solves the issue.

namloan:
Fixes #2315.

The event key handler wasn’t being removed when the view was detached. As a result, every time a new chat window was attached, a new key event handler was registered without removing the previous ones. Consequently, Shift+Enter was being processed by each handler, leading to unwanted blank lines being added.


HenrikJannsen (approved):
utACK

Good find! I guess we have more `addEventFilter` with missing remove...

PR #2623, opened 2024-08-15
Update notification badge when text chat message is removed

### LINKED ISSUE #2577: Market items show a badge when there are no offers
It seems that market items show a badge for expired offers. I only see this issue when scrolling down.

image

namloan:
This was only working when the chat message was an offer. For text-only chat messages, the number in the notifications badge of the items in the channel selection wasn't updated when the text message was removed.

Fixes #2577

### DIFF
--- a/chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
+++ b/chat/src/main/java/bisq/chat/notifications/ChatNotificationService.java
@@ -261,6 +261,7 @@ public class ChatNotificationService implements PersistenceClient<ChatNotificati
             wasRemoved = candidate.map(notification -> {
                         boolean result = persistableStore.getNotifications().remove(notification);
                         if (result) {
+                            changedNotification.set(null);
                             changedNotification.set(notification);
                         }
                         return result;


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/MarketChannelItem.java
```diff
@@ -119,12 +119,13 @@ private void setUpColorAdjustments() {
...
     }
 
-    private void updateNumOffers() {
+    private void updateNumbers() {
         UIThread.run(() -> {
             int numOffers = (int) channel.getChatMessages().stream()
                     .filter(BisqEasyOfferbookMessage::hasBisqEasyOffer)
                     .count();
             getNumOffers().set(numOffers);
+            activateNotifications();
```
HenrikJannsen: I think a `applyNotification(notification)` should be enough. To unbind and rebind the observer seems to be wasteful and signals a different intention.

HenrikJannsen (changes requested):

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/MarketChannelItem.java
namloan: I assumed you may have faced troubles with this and that was the quick fix you had found out as the same method is called for similar updates across BisqEasyOfferbookController (```updateFilteredMarketChannelItems``` and ```updateFavouriteMarketChannelItems```). So I took the 'safe approach' of following the same pattern. I'll change and test those code paths as well so that we remain consistent.

HenrikJannsen: Oh yes there its also not the best approach. Better to make the `applyNotification` method public and call that.

HenrikJannsen (approved):
utACK

PR #2628, opened 2024-08-16
Only bind servers to all interfaces on Whonix

alvasw:
The Bisq P2P ports are exposed to the local network even when Tor is the only selected network. This is only needed on Whonix.


HenrikJannsen (approved):
utACK

PR #2629, opened 2024-08-16
packaging: Add Windows MSI support

alvasw:
Users have to manually update Bisq and to support auto-updates we need to ship msi files. We can't use the existing exe file because it requires user interaction during installation or updating.


HenrikJannsen (approved):
utACK

PR #2642, opened 2024-08-18
Improve chat notifications efficiency

namloan:
Removes all bindings from MarketplaceChannelItem, that were also being re-created per each notification, to create a single binding in BisqEasyOfferbookController. Now each notification is processed only once to update the numbers of the channel where it was produced instead of refreshing all channels.


HenrikJannsen (approved):
utACK

PR #2644, opened 2024-08-18
Improve mediation buttons

### LINKED ISSUE #2478: Improve open mediation buttons
Currently we show a hyperlink with text "Report to mediator" and only after confirmation of payment it switches to a green outlined button with text "Request mediation".

This switch might be confusing and the point of the switch in the protocol might be set a bit late. When the buyer sends the Fiat seems the be the better moment to switch.

Screenshot 2024-07-25 at 14 38 28
Screenshot 2024-07-25 at 14 44 32

Goal was to provide the ability to report suspicious behavior of a trade peer in the stage of the protocol where no funds are at risk.
At that stage we don't want to give the mediation option too much attention.
Later once its more "serious" there should be a clearly visible option for "requesting support".

I think the current approach requires a bit of fine-tuning.

axpoems:
Resolves #2478

* Show request mediation button after Fiat has been sent.
* Align trade options menu using `BisqMenuItem`.
* Add missing wallet icon in white.


HenrikJannsen (approved):
utACK

PR #2651, opened 2024-08-19
Offerbook fixes

axpoems:
* Add a different string for positive and negative float prices.
* Increase price column width to allow the minus sign.


HenrikJannsen (approved):
utACK

PR #2652, opened 2024-08-19
Add effect on offer message highlight

axpoems:
Based on #2651

* Use a drop shadow instead of a solid border to highlight the offer message.
* Add an effect to remove the highlight after a few seconds.


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/list/message_box/BubbleMessageBox.java
```diff
@@ -360,4 +359,31 @@ protected static void onCopyMessage(ChatMessage chatMessage) {
...
+
+        showHighlightedPin = EasyBind.subscribe(item.getShowHighlighted(), showHighlighted -> {
+            if (showHighlighted) {
+                dropShadow.setRadius(15);
+                dropShadow.setSpread(0.3);
+                messageBgHBox.setEffect(dropShadow);
+                timeline.play();
+            } else {
+                timeline.stop();
+                messageBgHBox.setEffect(null);
```
HenrikJannsen, line 393: Is that also applied at deactivate? If user switch screen while timeline is running and come back later we need to ensure there is no effect.

HenrikJannsen (approved):
utACK

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/list/message_box/BubbleMessageBox.java
HenrikJannsen: Looked into src code more closely:
```
timeline.stop();
messageBgHBox.setEffect(null);
```
need to be added to the `dispose` method. timeline needs to be a fields and null checks are required in dispose....

axpoems: Added the changes now.

HenrikJannsen (approved):
utACK

Issue #828, opened 2023-05-05
Gradle: Pass Binary Version Number to Java Code


alvasw:
The Electrum/Tor version is set in the `build.gradle` file, and we can pass it to the Java code.

HenrikJannsen:
@rodvar 
Would you want to look into that?
The tor version number could be logged next to the Bisq version at startup and displayed at the network view where we also display version and commit hash.

Electrum version is not needed yet as undecided which wallet will be used.

rodvar:
@HenrikJannsen for sure!

rodvar:
@HenrikJannsen I believe we can close this one

PR #2654, opened 2024-08-20
Expose Tor Version from gradle.properties

rodvar:
- Implementation of https://github.com/bisq-network/bisq2/issues/828
 - move tor version definition to root gradle.properties
 - do necessary changes to read from root in current tor version usages
 - expose tor version in generated BuildVersion Java class
 - print tor version along with bisq version on startup logs
 - add label to network info view with tor version


application/src/main/java/bisq/application/ApplicationService.java
```diff
@@ -156,6 +156,8 @@ public ApplicationService(String configFileName, String[] args) {
         log.info(AsciiLogo.getAsciiLogo());
         log.info("Data directory: {}", config.getBaseDir());
         log.info("Version: v{} / Commit hash: {}", ApplicationVersion.getVersion().getVersionAsString(), ApplicationVersion.getBuildCommitShortHash());
+        log.info("Tor Version: v{}", ApplicationVersion.getTorVersionString());
```
HenrikJannsen, line 159: Maybe just add it to the previous log?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/network/NetworkInfoView.java
```diff
@@ -54,9 +54,11 @@ public NetworkInfoView(NetworkInfoModel model, NetworkInfoController controller,
         Label myVersionAndCommitHash = new Label(Res.get("settings.network.myVersionAndCommitHash",
                 ApplicationVersion.getVersion().getVersionAsString(),
                 ApplicationVersion.getBuildCommitShortHash()));
+        Label torVersion = new Label(Res.get("settings.network.torVersion",
```
HenrikJannsen: Better add it to the `settings.network.myVersionAndCommitHash` string. As it is now, the vertical distance is different between the first 2 lines and tor version as first 2 use a linebreak in the string.

HenrikJannsen (changes requested):
Some small issues. The layout is not ok as it is in version display. As mentioned best to append the tor version with a linebreak to existing label

application/src/main/java/bisq/application/ApplicationService.java
rodvar: I think is clearer this way

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/network/NetworkInfoView.java
rodvar: the space was left on purpose, but we discussed on matrix and agree to try to put them together which I'm gonna do now

HenrikJannsen (approved):
utACK

Issue #2530, opened 2024-08-01
Resolve warning about Kotlin Gradle plugin


HenrikJannsen:
`The Kotlin Gradle plugin was loaded multiple times in different subprojects, which is not supported and may break the build. 
This might happen in subprojects that apply the Kotlin plugins with the Gradle 'plugins { ... }' DSL if they specify explicit versions, even if the versions are equal.
Please add the Kotlin plugin to the common parent project or the root project, then remove the versions in the subprojects.
If the parent project does not need the plugin, add 'apply false' to the plugin line.`

PR #2672, opened 2024-08-24
Fix Kotlin DSL warning

namloan:
Fixes https://github.com/bisq-network/bisq2/issues/2530

https://youtrack.jetbrains.com/issue/KT-64210/False-positive-warning-that-Kotlin-Gradle-plugin-was-loaded-multiple-times-in-different-subprojects-is-displayed


HenrikJannsen (approved):
utACK

PR #2676, opened 2024-08-24
Add payments filter in offer list

axpoems:
Towards #2675

Add capability for filtering by payment methods in offer list. (Will follow-up with additional improvements to this.)


HenrikJannsen (approved):
utACK

PR #2677, opened 2024-08-25
Proguard integration for jars optimization

rodvar:
## Reason

 - This is an implementation of issue  https://github.com/bisq-network/bisq2/issues/2522

## Approach

Significant effort has been made to be able to seamlessly integrate Proguard into our build project. This implementation followed the following guidelines:

 - no modifications needed on the current way of building the project (e.g. git clean build works as it used to), and same goes for all the important commands to generate apps packaging.
 - enable as much optimizations as possible (currently, all 4 stages or proguard are being used: preverify, shrink, obfuscate, optimize)
 - Tests with proguarded jars passing as well as the current unproguarded state, no skip tests was needed (after hours of work to set it up right)

## Output

 - jar sizes reduction of average ~7%
 - `installDist` and `generateInstallers` use the integration and proguarded jars are used in the packaging
 - unobfuscated jars are keept in the corresponding build directories on a subdir called `unobfuscated`
 - a handy `listJars` task is provided that allows to quickly compare jars and their size

## Known issues / future lines of improvement

 - the current implementation is not playing well with gradle parallelism. IMHE most of the times is fine, but sometimes you'll get a proguardTask failing because it tried to use a jar that was being proguarded simultaneously. To have a deterministic result when testing just use the `--max-workers=1` parameter. E.g. `./gradlew clean build --max-workers=1`
 - Even thought obfuscation is enabled, a significant portion (specially in common) has been configured (proguard-rules.pro ) to be kept as not doing so would cause different issues either in tests or in the apps itself (specially related to security and persistence modules).
 - Optimization passes have been left to the current proguard stable version default (1). Trying to elevate this number resulted in builds, this could be a future project.
 - After nailing the basics, a refactor was done into a reusable custom gradle plugin for future integration
 - webcam project has been left aside in the context of this first integration, should be easily integrated thanks to the usage of gradle custom plugins.
 - Final proguard-rules.pro config file might have more rules than what is actually needed, this will be clean up in future optimization PRs after this first integration merge.
 - listJars has a bug where in certain modules it will print the jars more than once :-P
 - `gradlew apps:desktop:desktop-app-launcher:generateInstallers` only works for me in MacOS (M1). This is `not` an issue of this PR, as I have the same problem in `main`. So users with Linux and Windows systems that already have this working would be really appreacited to test the installers. In my tests installer works great in MacOS, and I even did real life trades in prod whilst testing the proguarded installed app.

## How to test this PR

 - once branch downloaded and checked out, first do `./gradlew clean build` . If everything is fine you should be able to do `gradlew listJars` and see 2 jars per proguarded module. 
 - then you can test the apps in the different ways you normally work. Run the seed node, run the oracle node, run the desktop and with 2 users and trade. Run the app in prod mode, etc.
 - generate installers, install and do a trade in prod.

### Dev's commit logs:

 - define global proguard config file in root
 - define global setup of proguard in root build.gradle.kts, excluding special sub project platform
 - add task to list generated jar files
 - add useful listJars gradle task to print involved jars with size
 - clean and build logic for wallets included submodule
 - configuration of proguard linked to root on leaf projects only, using the main proguard rules file from root
 - persistance module generates proguard jar
 - generic java app rules for proguard, fix proguard not finding java classpaths
 - avoid shas of non proguard jars
 - move proguard logic into custom gradle plugin for reuse in composite modules
 - added proguard to daemon module
 - added handy cleanAll buildAll root tasks to completely wipe jars and regenerate
 - allow configure relative path to proguard config per module
 - include into installDist the proguardTask needs to run after jar instead of build
 - better inclusion of proguard in build cycle by executing right after jar and keeping jar orginal name fix plugin sometimes not finding jar files, copying jar into unobfuscated subdir
 - avoid configuring already configured subprojects/modules
 - added docs to explain how to use proguard in the dev docs
 - fix dependencies to avoid loss of protobuf classes on proguarded builds (tests pass)
 - upgraded protobuf and protobuf gradle plugin to latest stable versions
 - cleanup and docs
 - add python scripts to gitignore

Thanks for taking time to help testing this PR !

### DIFF
--- a/build-logic/commons/src/main/kotlin/bisq.java-conventions.gradle.kts
+++ b/build-logic/commons/src/main/kotlin/bisq.java-conventions.gradle.kts
@@ -38,6 +38,8 @@ dependencies {
         testAnnotationProcessor(it)
         testCompileOnly(it)
     }
+    // needed for proguard
+    implementation("org.projectlombok:lombok:1.18.34")
 
     versionCatalog.findLibrary("slf4j-api").ifPresent {
         implementation(it)
--- a/proguard-rules.pro
+++ b/proguard-rules.pro
@@ -1,0 +1,321 @@
+# Filter manifirst
+#-injars  my.jar(!META-INF/MANIFEST.MF)
+
+# careful when using dontwarn..
+#-dontwarn javax.**
+#-dontwarn java.**
+#-dontwarn org.**
+#-dontwarn bisq.**
+#-dontwarn com.google.protobuf.**
+#-dontwarn bisq.common.proto.**
+
+#----------------------------------------------------------------
+#Warning: bisq.common.jvm.JvmUtils: can't find referenced class java.lang.management.ManagementFactory
+#Warning: bisq.common.jvm.JvmUtils: can't find referenced class java.lang.management.ManagementFactory
+#Warning: bisq.common.jvm.JvmUtils: can't find referenced class java.lang.management.RuntimeMXBean
+#Warning: bisq.common.jvm.JvmUtils: can't find referenced class java.lang.management.RuntimeMXBean
+# found these the only acceptable warnings to just skip
+-dontwarn java.lang.management.**
+#----------------------------------------------------------------
+
+#-printmapping build/libs/mapping.txt
+#-printseeds build/libs/out.seeds
+
+#-addconfigurationdebugging
+#-verbose
+
+-adaptresourcefilenames **.properties,**.gif,**.jpg,**.svg,**.png
+-adaptresourcefilecontents **.properties,META-INF/MANIFEST.MF
+
+## GENERAL RULES
+-verbose
+-allowaccessmodification
+-dontusemixedcaseclassnames
+-dontskipnonpubliclibraryclasses
+-keepclasseswithmembernames class * {
+    native <methods>;
+}
+-keepclasseswithmembers public class * {
+    public static void main(java.lang.String[]);
+}
+-keep public class * {
+    public protected *;
+}
+
+-keepattributes *Annotation*
+-keepattributes EnclosingMethod
+-keepparameternames
+-renamesourcefileattribute SourceFile
+-keepattributes Signature,Exceptions,*Annotation*,
+                InnerClasses,PermittedSubclasses,EnclosingMethod,
+                Deprecated,SourceFile,LineNumberTable
+
+-keepclasseswithmembernames,includedescriptorclasses class * {
+    native <methods>;
+}
+
+-keepclassmembers,allowoptimization enum * {
+    public static **[] values();
+    public static ** valueOf(java.lang.String);
+}
+
+-keepclassmembers class * implements java.io.Serializable {
+    static final long serialVersionUID;
+    private static final java.io.ObjectStreamField[] serialPersistentFields;
+    private void writeObject(java.io.ObjectOutputStream);
+    private void readObject(java.io.ObjectInputStream);
+    java.lang.Object writeReplace();
+    java.lang.Object readResolve();
+}
+
+## **IMPORTANT** During the first addition of proguard, Optimization and preverification failed due to not finding the
+## protobuf classes used as parent of our classes ()
+## removing preverification helped to avoid but problems came down the line when testing final builds.
+## uncomment the following lines only if you know what and why are you doing it.
+## More information: https://community.guardsquare.com/t/facing-an-error-with-cant-find-common-super-class-of-while-obfuscating-a-maven-based-java-application/1437/3
+
+# Controls for debugging
+
+#-dontpreverify
+#-dontshrink
+#-dontoptimize
+## Decision has been made to do not use obfuscation for security and deterministic size reasons.
+## Keep rules will still be kept below but doesn't really serve much purpose until this gets commented out
+-dontobfuscate
+
+# 1 is the default - On the first attempt to setup proguard effort was invested in raising this number unsuccessfully
+# as it would cause cause problems on the generated desktop app (with javafx)
+-optimizationpasses 1
+#-optimizations !code/allocation/variable
+
+## BISQ MODULES
+# PERSISTANCE
+# As a general rule, we don't wont to obfuscate bisq. modules code. This could be improved but on the first attempt it proved to be very challenging
+# If this were to be removed, its recommended to leave the specific keeps below in this file.
+#-keep class bisq.** { *; }
+
+# specific keeps on common
+-keep class bisq.common.util.** { *; }
+-keep class bisq.common.util.StringUtils { *; }
+-keep class bisq.common.util.ClassUtils { *; }
+-keep class bisq.common.logging.LogMaskConverter {
+    public <methods>;
+    *;
+}
+
+# General rules to keep Java core classes
+-keep class java.lang.** { *; }
+-keep class java.util.** { *; }
+-keep class java.io.** { *; }
+-keep class java.nio.** { *; }
+-keep class java.lang.management.ManagementFactory { *; }
+-keep class java.lang.management.RuntimeMXBean { *; }
+# Add specific rules for referenced classes
+-keep class bisq.user.reputation.requests.AuthorizeTimestampRequest { *; }
+
+-keepclassmembers class bisq.common.util.ClassUtils {
+    public static ** getEnclosingClass();
+}
+
+#Protobuf
+-keepclassmembers class * {
+    ** ADAPTER; ** Schema; ** Protobuf;
+}
+-keepclasseswithmembers class * {
+    ** ADAPTER; ** Schema; ** Protobuf;
+}
+-keep class com.google.protobuf.** { *; }
+-keep class com.google.protobuf.GeneratedMessageLite { *; }
+-keepclassmembers class com.google.protobuf.GeneratedMessageLite$Builder { *; }
+-keepclassmembers class com.google.protobuf.GeneratedMessageLite {
+    public static <fields>;
+}
+-keepclassmembers class ** {
+    public static void registerAllExtensions(com.google.protobuf.ExtensionRegistryLite);
+}
+-keepclassmembers class **$Builder {
+    public ** build();
+}
+-keep class bisq.common.proto.** { *; }
+# MacOS persistence tests would fail without the following specific keep
+-keep class bisq.common.protobuf.** { *; }
+-keep class bisq.common.protobuf.StringLongPair { *; }
+-keep class bisq.common.protobuf.StringLongPair$1 { *; }
+-keep class bisq.common.protobuf.StringLongPair$Builder { *; }
+
+# Keep the interface and its methods
+-keep interface bisq.common.proto.PersistableProto { *; }
+# Keep all implementations of PersistableProto
+-keep class * implements bisq.common.proto.PersistableProto { *; }
+# Ensure that ProGuard does not strip any fields or methods of classes implementing PersistableProto
+-keepclassmembers class * implements bisq.common.proto.PersistableProto {
+    <fields>;
+    <methods>;
+}
+
+-keep interface bisq.common.proto.Proto { *; }
+-keep class * implements bisq.common.proto.Proto { *; }
+-keep class bisq.persistence.TimestampStore { *; }
+-keep class bisq.persistence.PersistableStoreReaderWriter { *; }
+-keep class bisq.persistence.PersistableStoreReaderWriterTests { *; }
+-keep public class * extends com.google.protobuf.GeneratedMessageLite { *; }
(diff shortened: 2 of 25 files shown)


rodvar:
Task :build-logic:packaging:compileKotlin FAILED
Kotlin does not yet support 22 JDK target, falling back to Kotlin JVM_21 JVM target
w: Inconsistent JVM-target compatibility detected for tasks 'compileJava' (22) and 'compileKotlin' (21).
This will become an error in Gradle 8.0.
Consider using JVM Toolchain: https://kotl.in/gradle/jvm/toolchain
Learn more about JVM-target validation: https://kotl.in/gradle/jvm/target-validation 

e: file:///home/runner/work/bisq2/bisq2/build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/ProGuardPlugin.kt:7:28 Unresolved reference: cc
e: file:///home/runner/work/bisq2/bisq2/build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/ProGuardPlugin.kt:70:13 Unresolved reference: logger
e: file:///home/runner/work/bisq2/bisq2/build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/ProGuardPlugin.kt:[192](https://github.com/bisq-network/bisq2/actions/runs/10543607079/job/29211734820?pr=2677#step:4:196):17 Unresolved reference: logger
e: file:///home/runner/work/bisq2/bisq2/build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/ProGuardPlugin.kt:230:13 Unresolved reference: logger
e: file:///home/runner/work/bisq2/bisq2/build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/ProGuardPlugin.kt:271:17 Unresolved reference: logger

will have to replace logger with printlns

HenrikJannsen:
I tried to create a binary on MacOS (M3) but get that error: `Caused by: java.util.zip.ZipException: zip file is empty`. But it could still create the dmg file, but that is larger than the 2.1.0 release dmg file (303 MB vs 286 MB), I have not build from main for comparison but I am not aware of any code change which would explain a 20 MB increase.

So my main surprise is that the reduction in jar files is only about 7%. 

With the obfuscate feature we probably lose the property of deterministic builds. So I think we should remove the obfuscations stage and also test if the resulting jar file of repeated builds and on different machines for the same platform is the same.

Do you have any idea what could be the reason why the reduction is not bigger (or even the opposite in the dmg file)?
If we do not get a significant reduction I am not sure if the added complexity and probably build time increase is worth it. Would be a pity if that is the result of that hard work to get it working but still then we at least know that there is not much headroom for further optimization on that level. 
There might be other options like optimizing images in the resources, stripping down the packaged jre to the minimum,...

HenrikJannsen:
I just did a binary build with main and the dmg is 300MB. Still 3MB smaller, which is not a significant difference but surprising that Proguard version is larger instead of smaller.

rodvar:
@HenrikJannsen 

The mentioned 7% is average, some cases are much significant. 
I'm not sure why we are having just a small reduction in size yet, but I have some ideas to have a look trying to optimize further:

 1. it looks like proguard jar is included in the jar's libs (we should avoid this from happening as it's not needed in run time)
 2. we don't have resources shrinking on (in android R8 engine is used, I need to have a look if we can do something similar for our use case)
 3. proguard adds some metadata to the proguarded jar file, need to check how big this is.
 
 if we need deterministic sizing, there are some techniques to accomplish that
 
  - make sure the input jars for proguard are always sorted in the same order (I don't think a priori this is a problem, because each subproject inputs 1 jar only..)
  - save mappings on each build and re-fed the mappings to proguard so that it uses the same obfuscation consistently -> this should make it without turning off obfuscation
 
 I'm planning to spend a little bit of time today to check why some of the build servers failed to build (see above), I'll also work a bit on this and put my comments below.

rodvar:
on a matrix discussion with Henrik we decided to disable obfuscation and put some effort into increasing optimization to reduce build size (this includes giving a first attempt to incorporate proguard to the web-cam module build)

rodvar:
timeboxed effort: include webcam-app as well
result: too many issues to deal with, should be done in a separate future PR

rodvar:
@HenrikJannsen I think some build servers are failing because they do a build using caches like this one for example `Build Bisq 2 / Build apps module on macOS-latest / Build apps module on macOS-latest (pull_request) Failing after 2m
` (see screenshot)

can we modify them to do a `gradle clean build --scan` instead? I'm not sure if I have permissions to do this myself ?

I just pushed a way to disable proguard from running linked to gradle.properties flag with the flag in false and still some failed because of this.

Screenshot from 2024-08-27 11-25-56

build-logic/commons/src/main/kotlin/bisq.java-conventions.gradle.kts
```diff
@@ -1,3 +1,5 @@
+import gradle.kotlin.dsl.accessors._a0d7d695c10041dcc5b883f9bcd31010.implementation
```
HenrikJannsen: Is this a standard package path? The '_a0d7d695c10041dcc5b883f9bcd31010' looks like its generated dynamically...

proguard-rules.pro
```diff
@@ -0,0 +1,315 @@
...
+-dontwarn java.lang.management.**
+#----------------------------------------------------------------
+
+#-printmapping build/libs/mapping.txt
+#-printseeds build/libs/out.seeds
+
+#-addconfigurationdebugging
+#-verbose
+
+-adaptresourcefilenames **.properties,**.gif,**.jpg,**.svg
```
HenrikJannsen: Shouldn't that contain `.png` as well?

proguard-rules.pro
```diff
@@ -0,0 +1,315 @@
...
+}
+-keep class javax.annotation.**
+-dontwarn javax.annotation.**
+
+-keepattributes *Annotation*, Signature, EnclosingMethod
+-keep class * extends java.lang.annotation.Annotation { *; }
+-keep class * implements java.lang.reflect.InvocationHandler { *; }
+
+-keep class org.springframework.beans.factory.annotation.**
+-dontwarn org.springframework.beans.factory.annotation.**
```
HenrikJannsen: In our code base there is no springframework dependency. I guess its from sub depenendencies. Do you know from where?

HenrikJannsen (changes requested):
Open questions and I did not get it working on Mac OS/M3. 
With Gradle 8.9 I got that error:

`Caused by: java.util.zip.ZipException: zip END header not found`

I tried then to use Gradle 8.8 (as I remembered to had issues as well before with latest Gradle verson). But get more issues: `Caused by: java.io.IOException: Can't process class [bisq/contract/Contract$1.class] (null)` and as well `Caused by: java.util.zip.ZipException: zip END header not found`.

I has though created a dmg, but when starting the app I get a runtime excpetion:
```
java.lang.NullPointerException: Cannot invoke "javafx.scene.Scene.isDepthBuffer()" because the return value of "javafx.scene.Node.getScene()" is null
	at com.sun.javafx.scene.input.PickResultChooser.processOffer(PickResultChooser.java:185)
	at com.sun.javafx.scene.input.PickResultChooser.offer(PickResultChooser.java:143)
	at javafx.scene.Node.intersects(Node.java:5227)
```
I guess some JavaFx classes are missing.

The file size of the dmg or app file is nearly exact the same. 

 A build (after a previous build) on desktop takes 4 seconds for me. With proguard: 20 sec.

Minor issue: When I run `cleanAll` I get that error:
```
Execution failed for task ':cleanAll'.
> A problem occurred starting process 'command 'gradlew''

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.
BUILD FAILED in 1s
45 actionable tasks: 9 executed, 36 up-to-date
```

The cleanAll and runAll are very handy!

HenrikJannsen:
As the reduction in files size in jars is < 10% (rather around 5%), I think its not justified to add that complexity.
The build time increase from 4 sec. to 20 sec. is another drawback. If we would keep it turned off while in dev and only turn it on for release, we run a high risk that things break just when we want to deploy and that bug caused by proguard are not covered as testers and devs have turned it off. As those are mostly runtime bugs and likely edge cases they might be hard to cover in smoke tests at release.  

Kudos for what you achieved, was for sure lot of work! But I think to merge that even if it is deactivated, we add mental load to anyone who touches the already complex Gradle tasks and if it will not be used longer there is high risk anyway that it will not work when turned on due lack on maintenance.

I would suggest to either create a branch in the Bisq Repo to keep the work accessible if we once decide to use it, or at your private GH repo.

Some improvements like the `cleanAll`, or `runAll` tasks could be added in an isolated PR.

HenrikJannsen:
I created the https://github.com/bisq-network/bisq2/tree/Proguard-support branch where we could park the efforts.

proguard-rules.pro
rodvar: I added this as part of the "std java project rules" first rule generation, but I guess we can safely remove it. I'll do a quick test and clean it up

proguard-rules.pro
rodvar: indeed, I'll add it in

build-logic/commons/src/main/kotlin/bisq.java-conventions.gradle.kts
rodvar: that was auto-added by Intllij... doesn't sound like its really needed I'll erase this as well after a quick test

rodvar:
> I did not get it working on Mac OS v14.6.1 / M3. With Gradle 8.9 I got that error:
> 
> `Caused by: java.util.zip.ZipException: zip END header not found`
> 
> I tried then to use Gradle 8.8 (as I remembered to had issues as well before with latest Gradle verson). But get more issues: `Caused by: java.io.IOException: Can't process class [bisq/contract/Contract$1.class] (null)` and as well `Caused by: java.util.zip.ZipException: zip END header not found`.
> 
> I has though created a dmg, but when starting the app I get a runtime excpetion:
> 
> ```
> java.lang.NullPointerException: Cannot invoke "javafx.scene.Scene.isDepthBuffer()" because the return value of "javafx.scene.Node.getScene()" is null
> 	at com.sun.javafx.scene.input.PickResultChooser.processOffer(PickResultChooser.java:185)
> 	at com.sun.javafx.scene.input.PickResultChooser.offer(PickResultChooser.java:143)
> 	at javafx.scene.Node.intersects(Node.java:5227)
> ```
> 
> I guess some JavaFx classes are missing.
> 
> The file size of the dmg or app file is nearly exact the same.
> 
> A build (after a previous build) on desktop takes 4 seconds for me. With proguard: 20 sec.
> 
> Minor issue: When I run `cleanAll` I get that error:
> 
> ```
> Execution failed for task ':cleanAll'.
> > A problem occurred starting process 'command 'gradlew''
> 
> * Try:
> > Run with --stacktrace option to get the stack trace.
> > Run with --info or --debug option to get more log output.
> > Run with --scan to get full insights.
> > Get more help at https://help.gradle.org.
> BUILD FAILED in 1s
> 45 actionable tasks: 9 executed, 36 up-to-date
> ```
> 
> The cleanAll and runAll are very handy!

Those zip-related kind of errors generally occur when the machine you are using to build is running out of space, is your HD edging its limit?  (I got them as well in my MacBook had to delete some stuff to be able to continue testing)

I can see though that the dmg is broken with the latest changes I've done yesterday, I'll fix that in a bit

rodvar:
@HenrikJannsen I'll updated the PR and pointed to the bisq2 branch.

I'll make a separate PR to merge the gradle improvements like cleanAll buildAll to main later today. Also I upgraded protobuf versions in this PR, the versions we are currently using have known security reports. Shall I include that upgrade as well?

Thanks for all the help

HenrikJannsen:
> @HenrikJannsen I'll updated the PR and pointed to the bisq2 branch.
> 
> I'll make a separate PR to merge the gradle improvements like cleanAll buildAll to main later today. Also I upgraded protobuf versions in this PR, the versions we are currently using have known security reports. Shall I include that upgrade as well?
> 
> Thanks for all the help

I did a lot of updates for vulnerable libraries. If you rebase on main it should be it.

Regarding diskspace: I have lot of space free.

HenrikJannsen:
I would say lets wrap it up. The size reduction (its rather an increase) is just not justifying the overall complexity and efforts. Lets park it in the new branch an move on...

rodvar:
> I would say lets wrap it up. The size reduction (its rather an increase) is just not justifying the overall complexity and efforts. Lets park it in the new branch an move on...

I agree, this is a "risky" change for a very small final improvement -  the PR is pointing to the "parking branch" so feel free to merge it there and let's move on.

I'll send a different PR to main with the gradle improvements :)

Issue #2667, opened 2024-08-22
Add more filters to Bisq Easy offerlist


axpoems:
Add more filters to Bisq Easy offerlist, such as:
- Payment methods
- Settlement methods
- My offers only

PR #2691, opened 2024-08-25
Further improvements in offer list payments filter

axpoems:
Towards #2667

* Add new `DropdownMenuItem` decoupled from `BisqMenuItem` for greater flexibility.
* Show payment methods icons.
* Add highlight to the menu when filter is active.


HenrikJannsen (approved):
utACK

PR #2698, opened 2024-08-26
Persist payment filters by market in offerlist

axpoems:
Towards #2667

* Add "clear filters" option.
* Persist selection of payment filters by market.


HenrikJannsen (approved):
utACK

PR #2722, opened 2024-08-28
Update icons and languages ui in message offer

axpoems:
Based on #2721 

- update network icon
- update languages icon
- unify all user input text in white


HenrikJannsen (approved):
utACK

PR #2741, opened 2024-09-03
wallets: Don't wait for subprocess shutdown after response

alvasw:
We send a `stop` RPC call and wait for the subprocess after receiving a `success/ok` message. This is unnecessary because we received a `success` reply already.


HenrikJannsen (approved):
utACK

PR #2742, opened 2024-09-03
wallets: Fix bitcoind deadlock

alvasw:
We parse stdout and wait for the "init message: Done loading" line to
detect when bitcoind is ready. After seeing this line, we stop reading
stdout causing the stdout buffer to fill up. When the stdout buffer
fills up, bitcoind waits until its parent process reads from it but this
unfortunately never happens leading to a deadlock.


HenrikJannsen (approved):
utACK

PR #2745, opened 2024-09-03
bitcoind: Fix blockHash ZMQ test

alvasw:
The test mines blocks and waits until it sees the block hashes of the
mined blocks over ZMQ. Unfortunately, ZMQ publishers are stateless and
can't re-transmit lost messages [1]. Consequently, the test was wrong.
The fixed test subscribes to the publisher, starts mining blocks, and
waits until it observes a new block hash.


HenrikJannsen (approved):
utACK

PR #2754, opened 2024-09-03
Update translations for v2.1.1

hiciefte:
Besides translations I also fixed a truncation issue in languages with longer strings than in the English source

Before
Bildschirmfoto 2024-09-03 um 11 07 07

After
Bildschirmfoto 2024-09-03 um 11 05 33

Unfortunately it wasn't possible to solve this by using Hbox growth priorities.


hiciefte:
In this update I also went through all warnings on Transifex, so I hope any inconsistencies should be fixed now 🤞

HenrikJannsen:
Did you use the latest commit version? There have been recent changes. Mostly in reputation and user files.
Regarding the market list truncation: Have you tried to reduce the right padding? I guess that can be reduces to avoid truncation. 
An alternative would be to use either `TextFlowUtils` to use a smaller font (maybe even a darker color) for the number of offers text, or to make a custom HBox with TextFlow or just 2 labels.
`bisqEasy.tradeWizard.review.price={0} <{1} style=trade-wizard-review-code>`

hiciefte:
> Did you use the latest commit version? There have been recent changes. Mostly in reputation and user files. Regarding the market list truncation: Have you tried to reduce the right padding? I guess that can be reduces to avoid truncation. An alternative would be to use either `TextFlowUtils`[1] to use a smaller font (maybe even a darker color) for the number of offers text, or to make a custom HBox with TextFlow or just 2 labels.
> 
> [1] Example: `bisqEasy.tradeWizard.review.price={0} <{1} style=trade-wizard-review-code>`

Yes, I pulled main yesterday before I pushed the PR. I'll de-couple the fix from the translations and will look at the different options to solve this issue

HenrikJannsen (approved):
utACK

PR #2769, opened 2024-09-05
Cleanup trade state controller

axpoems:
* Remove visual elements from `TradeStateController` and move them to the view.
* Update "Request mediation" button to grey.


HenrikJannsen (approved):
utACK

Issue #2681, opened 2024-08-25
Use JRE for binaries instead of JDK


HenrikJannsen:
Currently we package the same JDK version used for the build into the binaries. We should use the JRE (of same java  and vendor version) instead to safe about 60 MB on file size.

The `BisqToolchainResolver` would need an extension to be used for the packager. The download path is exact the same just `jdk` will be `jre`.

The `getJPackageJdkDirectory` method in `PackagingPlugin` delivers the path for the runtime. This would require changes to support the JRE.

HenrikJannsen:
@alvasw Do you see any issues with that?
@rodvar Would you be interested to work on that?

rodvar:
starting work on this one

rodvar:
@HenrikJannsen just did some initial tests, it looks like this is as easier as changing the urls to use jre instead of jdk.
zulu doesn't have any specific subdirs for jre, so no need to change the Packaging plugin.

The download size is significant, but the final binaries doesn't get much of an improvement... shall I send a PR to discuss anyways?

rodvar:
> @HenrikJannsen just did some initial tests, it looks like this is as easier as changing the urls to use jre instead of jdk. zulu doesn't have any specific subdirs for jre, so no need to change the Packaging plugin.
> 
> The download size is significant, but the final binaries doesn't get much of an improvement... shall I send a PR to discuss anyways?

clarified the above on a matrix chat.
We need to keep the JDK resolver as is, but the idea is to build a new JRE resolver and set it up in the jpackager to be used for binaries (as opposed to let it use the jdk bins as it is now

alejandrogarcia83:
I have to revert PR #2773 because the changes broke the release tasks. The release has been already delayed by more than a week and I don't want to delay the release further.

rodvar:
Hi @alejandrogarcia83 , I'm surprised and sorry to hear that, saw you comment as well in https://github.com/bisq-network/bisq2/pull/2773#issuecomment-2407996717

Could you please be more specific, which release tasks were broken? Do you have any logs? I'm happy to give you a hand to fix this so we can include this substantial size improvement for the next release hopefully.

mvp1983:
I can take if @alvasw doesn't have time on this

rodvar:
it would be great to get another couple of eyes on this @mvp1983 thanks for offering
on my end is a green light, @HenrikJannsen ?

HenrikJannsen:
[...]

PR #2773, opened 2024-09-06
Use JRE instead of JDK Desktop Binary

rodvar:
## Purpose

 - Implementation for [JDK/JRE Build Optimization Issue](https://github.com/bisq-network/bisq2/issues/2681)
 
## Approach

 - ~~Use Gradle Java Toolchain~~  --> _Couldn't get it to pick up the JRE_
 - ~~Download JRE manually and use it for runtime~~  --> _Too complicated, increase build times_
 - Build a skimmed JRE out of our dev JDK  --> **Success ! :)**
 
 ## Results
 
  - Reduction on final binary are around **~160MB (30% less)**
  - **new** results after tweaking some unnecessary mods during review are **~245MB (46% less)**
  - This PR is fully  tested in MacOS (M1). Also tested that it will run in Windows and Linux, but for those environments I need help as the binary won't generate in my envs even from main branch (Eventually I need to work why is this happening and submit a PR to fix it  or fix whatever is wrong in my envs) - Please help test this PR!
  
## How to test this PR
  

1.  `$ ./gradlew apps:desktop:desktop-app-launcher:clean  apps:desktop:desktop-app-launcher:generateInstallers --info`  
2. you'll find in in `apps/desktop/desktop-app-launcher/build` folder the `custom-jre` folder with the gerated skimmed JRE, and in `packages/jpackages/package/Bisq2...` the installer corresponding to your env (e.g. Bisq2....dmg) . Install it and run the app. Also check the size of the binary. Please comment your results with details on your environment in this PR.
  
## Known Issues / Future lines of improvement

 - Left TODOs in the code to reference this

## Dev Commit logs

 - approach using jlink to generate a slim jre from our dev jdk
 - generateInstallers depends on jlink task now
 - work out the list of mods needed for the app to function
 - optimize jlink command to trim as much unneded stuff as possible
 - adjust code for windows env
 - cleanup code, leave TODOs for future lines of improvement/optimization

### DIFF
--- a/build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/JLinkTask.kt
+++ b/build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/JLinkTask.kt
@@ -3,14 +3,18 @@ package bisq.gradle.packaging
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.RegularFileProperty
-import org.gradle.api.tasks.InputDirectory
-import org.gradle.api.tasks.InputFile
-import org.gradle.api.tasks.Optional
-import org.gradle.api.tasks.OutputDirectory
-import org.gradle.api.tasks.TaskAction
-import java.util.concurrent.TimeUnit
+import org.gradle.api.tasks.*
+import java.io.BufferedReader
+import java.io.InputStreamReader
+
+
+//import java.util.regex.Pattern
 
 abstract class JLinkTask : DefaultTask() {
+
+//    companion object {
+//        val pattern: Pattern = Pattern.compile(".*\\b(java\\..*|javax\\..*)\\b.*\$")
+//    }
 
     @get:InputDirectory
     abstract val jdkDirectory: DirectoryProperty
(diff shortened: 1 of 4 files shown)


build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/JLinkTask.kt
```diff
@@ -27,46 +31,103 @@ abstract class JLinkTask : DefaultTask() {
...
+            }
+
+            logger.lifecycle("Preparing process builder..")
+            processBuilder.inheritIO()
+
+            logger.lifecycle("Starting process builder..")
+            val process = processBuilder.start()
+            val reader =
+                BufferedReader(InputStreamReader(process.inputStream))
+            while ((reader.readLine()) != null) {
```
HenrikJannsen: Maybe wrap it into a if(isWindows()) branch?

build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/JLinkTask.kt
```diff
@@ -27,46 +31,103 @@ abstract class JLinkTask : DefaultTask() {
...
+            "java.se",
+            "java.security.jgss",
+            // javafx
+            "javafx.base",
+            "javafx.controls",
+            "javafx.fxml",
+            "javafx.graphics",
+            "javafx.media",
+            "javafx.swing",
+            "javafx.web",
```
HenrikJannsen: I am not aware that we have any dependency to javafx.web

HenrikJannsen (approved):
utACK

See Nits

Would be good if @alvasw could review as well

build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/JLinkTask.kt
rodvar: it doesn't really affect the other platforms, but sure I'll have a look when testing removing javafx.web ;)

build-logic/packaging/src/main/kotlin/bisq/gradle/packaging/JLinkTask.kt
rodvar: great, I'll do a quick test to see if we can safely remove that - this will reduce the size of the final build even more :)

rodvar: thanks so much @HenrikJannsen , indeed the binary works perfectly without javafx.web mod and we save an impressive 100MB more being the total reduction ~46.14% now :dagger: :smile:

HenrikJannsen: Could you try how much difference removal of `javafx.swing` cause? We cannot remove it now as its used but not in major use cases. If reduction is large enough it would justify investigating how to replace the current usage (SystemTray, AwtImage,....).

rodvar: @HenrikJannsen not much of a difference, just a few kbs

HenrikJannsen: Ok, thanks. Then we leave it

HenrikJannsen (approved):
utACK

alejandrogarcia83:
I have to revert this PR unfortunately because the changes broke the release tasks. The release has been already delayed by more than a week and I don't want to delay the release further.

PR #2777, opened 2024-09-07
Reviewed ES translations for chat properties

counterweightoperator:
This PR makes improvements on the Spanish translations for the chat properties file.

There were a few mistakes in place, such as dangling English strings, awkward machine translations and inconsistencies around language usage.

PR #2778, opened 2024-09-07
Reviewed ES translations for chat properties

counterweightoperator:
This PR makes improvements on the Spanish translations for the chat properties file.

There were a few mistakes in place, such as dangling English strings, awkward machine translations and inconsistencies around language usage.

Replaces #2777, which had the same contents but commits were not signed.


HenrikJannsen (approved):
utACK

PR #2785, opened 2024-09-08
Reviewed ES translations for network properties

counterweightoperator:
This PR makes improvements on the Spanish translations for the default properties file.

There were a few mistakes in place, such as dangling English strings, awkward machine translations and inconsistencies around language usage.


HenrikJannsen (approved):
utACK

PR #2790, opened 2024-09-08
 Add new icons for guides and report in trade process

axpoems:
* Design and add new icons for trade process options.
* Update hover colour of mediation button.


HenrikJannsen (approved):
utACK

PR #2792, opened 2024-09-08
Reviewed ES translations for wallet properties

counterweightoperator:
Just a tiny improvement in one string for this section.


HenrikJannsen (approved):
utACK

PR #2794, opened 2024-09-08
Update notification bell and fix reputation icons

axpoems:
* Introduce new bell icon set
* Fix reputation green icons of left menu


HenrikJannsen (approved):
utACK

PR #2796, opened 2024-09-09
Improves check for TX token

hiciefte:
The current check works, but it didn't properly echoed the String. Additionally I exited the build at that point.
I've checked it on my remote repo and everything should work now as expected.


HenrikJannsen (approved):
utACK

PR #2803, opened 2024-09-10
wallets: Remove dependency to Bisq 2 Coin Class

alvasw:
We can't depend on Bisq 2's Coin class to be able to reuse the wallets
library in Bisq 1.


HenrikJannsen (approved):
utACK

PR #2816, opened 2024-09-11
Move Scanners to bisq.wallets.regtest module

alvasw:
We can't depend on Bisq 2's common module to be able to reuse the
wallets library in Bisq 1.


HenrikJannsen (approved):
utACK

PR #2841, opened 2024-09-19
More improvements on user selection

axpoems:
Based on #2840 

* Fix user profile selection styles in build reputation section.
* Fix user profile selection tooltips (user and reputation details).


HenrikJannsen (approved):
utACK

PR #2843, opened 2024-09-20
Create new icons for open and copy links

axpoems:
* Add new icons for open and copy block explorer link.
* Add missing strings for buyer and seller trade protocol phase3b.


HenrikJannsen (approved):
utACK

PR #2858, opened 2024-09-27
Correct description when creating account

makhni-dev:
Corrects the desciption text in CreatePaymentAccountView, that previously displayed the account name as description for the account data field. This can be seen in the images below, where the text "Payment account name" was duplicated previously:
**Before:**
before
**After:**
after


HenrikJannsen (approved):
utACK

PR #2912, opened 2024-10-08
packaging: Remove dependency to Bisq 2 commons module

alvasw:


rodvar (changes requested):
Hey Alva @alvasw , just reviewed this one and was wondering if the duplication of code approach is justified by the elimination of the common dependency? I guess it would be clarified if you comment on the benefits and/or motivations behind this one? thanks

HenrikJannsen:
> Hey Alva @alvasw , just reviewed this one and was wondering if the duplication of code approach is justified by the elimination of the common dependency? I guess it would be clarified if you comment on the benefits and/or motivations behind this one? thanks

From what I suspect its because the build is re-used for Bisq 1 and certain dependencies would not work there (Bisq 1 use Java 17). I assume the main motivation is for testing the BitcoinJ update with a integration tests using bitcoind as there is already a lot of infrastructure available in Bisq 2 for that. But @alvasw can explain for sure in more detail...

PR #2913, opened 2024-10-08
bitcoind-build-logic: Remove unused protobuf dependency

alvasw:


rodvar:
removed from wallets module

utACK

PR #2914, opened 2024-10-09
bitcoind: Remove dependency to root build-logic

alvasw:


rodvar:
removed from wallets module

utACK

PR #2919, opened 2024-10-09
BitcoindRegtestProcess: Expose P2P port number

alvasw:
BitcoinJ connects to Bitcoin Core's P2P port. Up to now, we selected a
random port without saving the p2p port number.


rodvar:
utACK

PR #2940, opened 2024-10-21
Update translations

hiciefte:
I just updated the translations for the changed sources. Finally I managed to setup a working local translation pipeline leveraging the chatGPT API. This should make it also more easily to handle new locales.


HenrikJannsen (approved):
utACK

PR #2952, opened 2024-10-24
Update setting translations

hiciefte:
Translate new keys in settings section


hiciefte:
A couple of translation sections have been added in my absence. Support was the last one missing (hopefully)

Issue #2568, opened 2024-08-05
Add trade details popup


HenrikJannsen:
The popup should display all data of the trade which can have some relevance to the user. Best to check the domain data objects (offer, trade, contract, userProfile) and see what might be of interest. Bisq 1 trade popup might be a guidance as well. We should not display pure technical data, but all data which a user might be interested in. This includes all the data which is already shown in the UI. As it might be quite some data fields we should use some groupings, the previous version (as in 2.1.0 but changed in main for 2.1.1) of the settings UI might be an option.

PR #2957, opened 2024-10-25
Create bisq easy trade details view

makhni-dev:
Closes bisq-network/bisq2#2568
This commit creates a pop up screen that contains trade details of a bisq easy trade. The design is similar to that used in trade wizard review offer page.
The open button looks like this: 
Screenshot from 2024-10-25 14-02-21
This opens the following popup:
Trade details

**Benefits**
This view can act as a central point for displaying important trade data, which is currently displayed in several separate locations, meaning this might increase clarity for users. It currently does not display anything that is not already visible in some other way, but this can easily be extended if wanted. 

**Future work**
Several more fields can be added here that can be of use, for example:
-  time since last user activity from peer
-  transaction ID
-  blockchain confirmation
**I suggest that I write an issue where we all can input what more fields that would be nice to have when this is merged.**

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_details/TradeDetailsController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_details/TradeDetailsController.java
@@ -1,0 +1,199 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.desktop.main.content.bisq_easy.open_trades.trade_details;
+
+import bisq.bisq_easy.NavigationTarget;
+import bisq.chat.bisqeasy.open_trades.BisqEasyOpenTradeChannel;
+import bisq.common.monetary.Coin;
+import bisq.common.monetary.Fiat;
+import bisq.common.monetary.Monetary;
+import bisq.common.monetary.PriceQuote;
+import bisq.common.network.AddressByTransportTypeMap;
+import bisq.common.util.StringUtils;
+import bisq.contract.bisq_easy.BisqEasyContract;
+import bisq.desktop.ServiceProvider;
+import bisq.desktop.common.view.Controller;
+import bisq.desktop.common.view.InitWithDataController;
+import bisq.desktop.common.view.NavigationController;
+import bisq.desktop.overlay.OverlayController;
+import bisq.i18n.Res;
+import bisq.network.identity.NetworkId;
+import bisq.offer.Direction;
+import bisq.offer.price.spec.FixPriceSpec;
+import bisq.offer.price.spec.FloatPriceSpec;
+import bisq.offer.price.spec.MarketPriceSpec;
+import bisq.offer.price.spec.PriceSpec;
+import bisq.presentation.formatters.AmountFormatter;
+import bisq.presentation.formatters.DateFormatter;
+import bisq.presentation.formatters.PercentageFormatter;
+import bisq.presentation.formatters.PriceFormatter;
+import bisq.trade.bisq_easy.BisqEasyTrade;
+import bisq.trade.bisq_easy.BisqEasyTradeFormatter;
+import bisq.trade.bisq_easy.BisqEasyTradeUtils;
+import bisq.user.identity.UserIdentityService;
+import bisq.user.profile.UserProfile;
+import com.google.common.base.Joiner;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+@Slf4j
+public class TradeDetailsController extends NavigationController
+        implements InitWithDataController<TradeDetailsController.InitData> {
+    protected final UserIdentityService userIdentityService;
+    @Getter
+    private final TradeDetailsView view;
+
+    @Getter
+    private final TradeDetailsModel model;
+
+    public TradeDetailsController(ServiceProvider serviceProvider) {
+        super(NavigationTarget.BISQ_EASY_TRADE_DETAILS);
+
+        userIdentityService = serviceProvider.getUserService().getUserIdentityService();
+        model = new TradeDetailsModel();
+        view = new TradeDetailsView(model, this);
+    }
+
+    private static String getPriceSpec(BisqEasyContract contract) {
+        PriceSpec priceSpec = contract.getAgreedPriceSpec();
+        String priceSpecStr = "";
+        switch (priceSpec) {
+            case MarketPriceSpec marketPriceSpec ->
+                    priceSpecStr = Res.get("bisqEasy.openTrades.tradeDetails.marketPrice");
+            case FloatPriceSpec floatPriceSpec -> {
+                String absPercent = PercentageFormatter.formatToPercentWithSymbol(Math.abs(floatPriceSpec.getPercentage()));
+                priceSpecStr = Res.get(floatPriceSpec.getPercentage() >= 0
+                        ? "bisqEasy.openTrades.tradeDetails.aboveMarketPrice"
+                        : "bisqEasy.openTrades.tradeDetails.belowMarketPrice", absPercent);
+            }
+            case FixPriceSpec fixPriceSpec -> priceSpecStr = Res.get("bisqEasy.openTrades.tradeDetails.fixedPrice");
+            case null, default -> {
+                // this should not happen unless a new PriceSpec is added
+            }
+        }
+        return priceSpecStr;
+    }
+
+    private static String formatNetworkAddresses(AddressByTransportTypeMap addressMap) {
+        return Joiner.on(", ").join(addressMap.entrySet().stream()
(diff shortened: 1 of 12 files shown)


HenrikJannsen:
Thanks for the PR! Here some comments.
In minimal app size the popup's height should not exceed the app height.
Screenshot 2024-10-26 at 13 02 19

I understand that you wanted to avoid a vertical scrollbar, which I agree is preferred if possible, but here I think we either minimize further the spacing of the rows or accept scrolling. We have usually design with lot of spacing, so I think a scrollbar here and following design from other screens is better here. Maybe @axpoems can add his input here or suggest a design?

My role: missing buyer/seller role
Market price should be Trade price, we can/should add the information if market price, float price or fixed price was used.
Amount in Fiat: Better use as well a textfield to allow copy

Fields which have no provided data yet: we could also just hide them. Or maybe better add another section of process-related data (e.g. data which will be set during the trade process, there will be more like blockchain confirmation, tx id,...)
In that case we should show all fields.

My username/My network address: I think we can drop those
Offer taken time stamp: We use the term date -> Trade date

Further important data:
Selected mediator (can be empty but is not in live network)

You could use instead of text fields the `CopyOnClickLabel`. Then the layout could be a bit condensed.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_details/TradeDetailsController.java
```diff
@@ -0,0 +1,150 @@
...
+        userIdentityService = serviceProvider.getUserService().getUserIdentityService();
+        model = new TradeDetailsModel();
+        view = new TradeDetailsView(model, this);
+    }
+
+    @Override
+    public void initWithData(InitData initData) {
+        BisqEasyTrade trade = initData.getBisqEasyTrade();
+        BisqEasyOpenTradeChannel channel = initData.channel;
+        model.setTradeId(new SimpleStringProperty(trade.getId()));
```
HenrikJannsen: We prefer to use final fields in the model for properties. Not a big problem to do it like that but better to stick with the established style.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_details/TradeDetailsController.java
```diff
@@ -0,0 +1,150 @@
...
+        Monetary quoteAmount = Fiat.from(quoteSideAmount, trade.getOffer().getMarket().getQuoteCurrencyCode());
+
+        NetworkId peerNetworkId = trade.getPeer().getNetworkId();
+        String peerAddress = formatNetworkAddresses(peerNetworkId.getAddressByTransportTypeMap());
+        model.setPeerNetworkAddress(new SimpleStringProperty(peerAddress));
+        NetworkId myNetworkId = trade.getMyself().getNetworkId();
+        String myAddress = formatNetworkAddresses(myNetworkId.getAddressByTransportTypeMap());
+        model.setMyNetworkAddress(new SimpleStringProperty(myAddress));
+        
+        String amountInFiat = AmountFormatter.formatAmount(quoteAmount);
```
HenrikJannsen, line 128: Better use `AmountFormatter.formatAmountWithCode`

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_details/TradeDetailsController.java
```diff
@@ -0,0 +1,150 @@
...
+        String paymentAccountData = trade.getPaymentAccountData().get();
+        model.setPaymentAccountData(new SimpleStringProperty(paymentAccountData));
+
+        String myRole = BisqEasyTradeFormatter.getMakerTakerRole(trade);
+        model.setMyRole(new SimpleStringProperty(myRole));
+    }
+
+    public String formatNetworkAddresses(AddressByTransportTypeMap addressMap) {
+        return Joiner.on("; ").join(addressMap.entrySet().stream()
+                .map(e -> StringUtils.truncate(e.getValue().getFullAddress(), Integer.MAX_VALUE))
```
HenrikJannsen: Why use truncate when setting Integer.MAX_VALUE?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_details/TradeDetailsModel.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Getter
+    private boolean paymentAccountDataProvided, bitcoinPaymentAddressProvided;
+
+    @Override
+    public NavigationTarget getDefaultNavigationTarget() {
+        return NavigationTarget.BISQ_EASY_TRADE_DETAILS;
+    }
+
+    public SimpleStringProperty getBitcoinPaymentAddress() {
```
HenrikJannsen: We use models as pure value objects, any logic even simple should be done in the controller.

HenrikJannsen: In that case I would just use the dataNotYetProvided as the default value in the property and set the property with a value once its gets set.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_details/TradeDetailsView.java
```diff
@@ -0,0 +1,223 @@
...
+import lombok.extern.slf4j.Slf4j;
+import javafx.scene.control.ScrollPane;
+
+@Slf4j
+public class TradeDetailsView extends NavigationView<AnchorPane, TradeDetailsModel, TradeDetailsController> {
+
+    private final Button closeButton;
+    private final ScrollPane scrollPane;
+
+    private final VBox allInfoContainer, informationContainer, detailsContainer;
```
HenrikJannsen: Can be local as well as scrollPane. IntelliJ shows warnings for such issues.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_details/TradeDetailsView.java
```diff
@@ -0,0 +1,223 @@
...
+        root.setPrefWidth(OverlayModel.WIDTH);
+        root.setPrefHeight(OverlayModel.HEIGHT + 200);
+        int rowSpacing = 2;
+
+        allInfoContainer = new VBox(25);
+        scrollPane = new ScrollPane(allInfoContainer);
+        scrollPane.setFitToWidth(true);
+        scrollPane.setPadding(new Insets(15, 15, 15, 15));
+        allInfoContainer.setPadding(new Insets(30, 30, 30, 30));
+        allInfoContainer.prefHeightProperty().bind(root.heightProperty());
```
HenrikJannsen: bindings must be set in onViewAttached. Also follow established style to use onViewAttached and onViewDetached after constructor as those are the most important methods.

All bindings, event listeners, event handlers must be deactivated in onViewDetached. See the dev docs as well or other code.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_details/TradeDetailsView.java
```diff
@@ -0,0 +1,223 @@
...
+        closeButton.setOnAction(e -> controller.onClose());
+    }
+
+    @Override
+    protected void onViewDetached() {
+        closeButton.setOnAction(null);
+    }
+
+    private void setupColorChangeEffects() {
+        bitcoinPaymentAddress.textField.textProperty().addListener(new ChangeListener<String>() {
```
HenrikJannsen: listeners need to be stored in a field and removed at onViewDetached.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/TradeStateView.java
```diff
@@ -186,6 +194,7 @@ protected void onViewAttached() {
         });
 
         cancelButton.setOnAction(e -> controller.onInterruptTrade());
+        tradeDetailsButton.setOnAction(e -> controller.onViewTradeDetails());
```
HenrikJannsen, line 196: Missing the `tradeDetailsButton.setOnAction(null)` in onViewDetached

HenrikJannsen (changes requested):

makhni-dev:
@HenrikJannsen Thank you for the review! I will work on the requested changes tomorrow and if I dont have time to finish, continue on friday.

makhni-dev:
@HenrikJannsen I have fixed the majority of the requests on my computer but have two questions:
> Fields which have no provided data yet: we could also just hide them. Or maybe better add another section of process-related data (e.g. data which will be set during the trade process, there will be more like blockchain confirmation, tx id,...) In that case we should show all fields.

This is a great idea, do you want the blockchain confirmation and tx id to be filled in in this pull request or should it be done separately, possibly in another issue?



> Further important data: Selected mediator (can be empty but is not in live network)

I have searched but not found a good way to test mediator assignment on localhost, is there some way to assign a mediator through config or JVM options somehow?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_details/TradeDetailsView.java
```diff
@@ -0,0 +1,202 @@
...
+
+        closeButton.setOnAction(null);
+    }
+
+    private VBox createSectionHeader(String headlineText) {
+        VBox section = new VBox(ROW_SPACING);
+        Label headline = new Label(headlineText);
+        headline.getStyleClass().add("bisq-text-headline-2");
+        Region separator = Layout.hLine();
+        separator.setStyle("-fx-background-color: green;");
```
axpoems: is this needed?

makhni-dev:
@axpoems and @HenrikJannsen, I have now updated the design to be much more in line with that used in TradeWizard review offer page. Please have a look and tell me what you think! I have updated the pull request description as well.

HenrikJannsen (approved):
utACK

Issue #2951, opened 2024-10-24
Publish Bisq 2 libraries for re-use in Bisq-mobile


HenrikJannsen:
We need to publish the Bisq 2 modules used in Bisq-mobile as jar libraries.

- Publish as jar to local maven repository
- As there are many modules, avoid duplication of the Gradle tasks (maybe create plugin?)
- Include protobuf files and resources 
- Use those libraries as dependencies in Bisq-mobile
- Add `publish-all` convenience task

We could use jitpack instead/additional to the local maven repository, but this will require verification.
I think local maven is at least during development easier and faster.

We could also consider to create a shadow jar to pack all into one jar. But I think having individual jars is probably better. If we package all we must avoid the non-android compatible modules like `java-se`, os-specific`, `apps`,...

When I used that approach in my POC branch I had issues with plugin `id("bisq.java-library")` as on android that caused a missing dependency to 'bisq::platform' which is not a java library and I did not manage to publish that. I guess it has to be recreated on the android project. By replacing the `id("bisq.java-library")` plugin with:
```
java
id("bisq.java-conventions")
```
I could resolve that issue.

At the network module when adding the publish tasks (at `network-identity`) it caused problems on the network sub-projects (like tor). I did not manage to resolve those issues. Some Gradle experts will know for sure how to deal with that....

HenrikJannsen:
@rodvar Can I assign that to you?

rodvar:
@HenrikJannsen sure thing!

rodvar:
starting work on this one

rodvar:
Current status:

 - can merge from main, all the work of ongoing adaptations has already been merged
 - code is fully J17 compatible now
 - Ideally (TBD) we should set that in the gradle setup (J17 compilation), but still target Java 22
 - local publishing to maven has not been merged, part of this issue tasks is to put it in place fully working for all the modules we need
 
 Modules to work out:
 
 ```
 - bisq-easy
 - bonded-roles
 - chat
 - common
 - contract (?)
 - daemon (?)
 - i18n (?)
 - identity
 - network
 - offer
 - persistence (?)
 - security
 - settings (?)
 - support (?)
 - trade
 - user
 - wallets (?)
 ```

PR #2965, opened 2024-10-29
Gradle task to publish dependencies for bisq mobile using mavenLocal

rodvar:
- Implementation for https://github.com/bisq-network/bisq2/issues/2951
 - create custom plugin to apply to each specific module where we need to publish jar to maven local
 - applied to all the necesary modules we need for AndroidNode
 - created handy publishAll task in root
 - adaptations for recursive submodules like network


HenrikJannsen:
Those modules need to be added as well:
`account`
`application`
`presentation`

The tor project is another project inside network with sub-projects. That is not published yet. 

When using in Android there is an error with missing dependency to bisq::platform, which is part of the pom.xml file
```
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>bisq</groupId>
        <artifactId>platform</artifactId>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
```

I think the protobuf and resources need to be added to sources for the jars.

rodvar:
> Those modules need to be added as well: `account` `application` `presentation`
> 
> The tor project is another project inside network with sub-projects. That is not published yet.
> 
> When using in Android there is an error with missing dependency to bisq::platform, which is part of the pom.xml file
> 
> ```
>   <dependencyManagement>
>     <dependencies>
>       <dependency>
>         <groupId>bisq</groupId>
>         <artifactId>platform</artifactId>
>         <type>pom</type>
>         <scope>import</scope>
>       </dependency>
>     </dependencies>
>   </dependencyManagement>
> ```
> 
> I think the protobuf and resources need to be added to sources for the jars.

cool I'll work on this next 💪

rodvar:
this is proving to be quite complex, because of the complexities of gradle composite builds we use in the project the generated default pom files have significant issues (corrupted files with no version in dependencies, or wrong reference name). I'm dealing with each one by one.
For the platform issue, I needed to publish a fake BOM dependency to maven local and it solved the issue.

rodvar:
this feature forces us to define requirement of dependencies properly (always specifying a version) cause
 - pom dependency without  versioning is a corrupted file (already fixed)
 - dependency references to be able to work in m2 therefore, always need to specify the version appart from the group and artifact
 
 So we need to fix the following now:
 
 ```
 Cause 1: org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find tor:tor:.
Required by:
    project :app > bisq:network:2.1.2
Cause 2: org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find network:network:.
Required by:
    project :app > bisq:identity:2.1.2
    project :app > bisq:account:2.1.2
    project :app > bisq:settings:2.1.2
    project :app > bisq:bonded-roles:2.1.2
    project :app > bisq:user:2.1.2
    project :app > bisq:contract:2.1.2
    project :app > bisq:offer:2.1.2
    project :app > bisq:trade:2.1.2
    project :app > bisq:support:2.1.2
    project :app > bisq:application:2.1.2
    project :app > bisq:chat:2.1.2
Cause 3: org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find network:network-identity:.
Required by:
    project :app > bisq:identity:2.1.2
    project :app > bisq:bonded-roles:2.1.2
    project :app > bisq:user:2.1.2
    project :app > bisq:contract:2.1.2
    project :app > bisq:offer:2.1.2
    project :app > bisq:trade:2.1.2
    project :app > bisq:support:2.1.2
    project :app > bisq:chat:2.1.2
```

**Comment by rodvar** [comment]
agreed with henrik we can ignore bisq-easy dependency as its not really needed and forces us to export wallet and bitcoind which we wold like to avoid.

**Comment by rodvar** [comment]
removed bisq-easy, only issue pending for the POC to build is solving protobuf duplicated defs:

```
Type bisq.account.protobuf.Account$Builder-IA is defined multiple times: andorid_only_poc/app/build/intermediates/project_dex_archive/release/dexBuilderRelease/out/bisq/account/protobuf/Account$Builder.dex, 
andorid_only_poc/app/build/intermediates/external_libs_dex/release/mergeExtDexRelease/classes2.dex
```

**Comment by rodvar** [comment]
Managed to get rid of those issues by using a different package for the proto generated in the mobile POC (some libs jave the jar generation including the proto sources and everything in bisq2 depends on that so cleaning it up is very complex)

now I'm into a different deal of problems, looks like code is still being compiled using Java22

```
import bisq.user.cathash.CatHashService;
                        ^
  bad class file: .m2/repository/bisq/user/2.1.2/user-2.1.2.jar(/bisq/user/cathash/CatHashService.class)
    class file has wrong version 66.0, should be 61.0
```

need to use this PR to make sure Java 17 is used for compilation

**Comment by rodvar** [comment]
worth to mention that it builds now, the error above is in runtime :)

```
BUILD SUCCESSFUL in 1m 6s
117 actionable tasks: 117 executed
```

**Comment by rodvar** [comment]
@HenrikJannsen this version works. You can test it by running `./gradlew clean build publishAll` and then run the POC you can find here https://github.com/bisq-network/bisq-mobile/tree/main/pocs/android_only_m2_j17jars

Next lines of work after this is merged:

 - some protobuf errors of classes not recognised (see logs when running the POC)
 - optimizations on the generated jars: debug apk size is ~75MB, we could trim the generated jars removing all of the protobuf stuff for example, which we decided its gonna be client based
  - allow publishing on internet maven repository (this will be very useful for our CI setup)


Let me know your thoughts!

**Comment by rodvar** [comment]
@HenrikJannsen happy to revert the changes on wallets/ module as we ended up not needing them, let me know

**Comment by HenrikJannsen** [comment]
@rodvar Regaring the setting of the Java version 17: We still would like to use Java 22 for Bisq 2, so it would be good to have the version configurable so that the publish tasks use Java 17 but normal build/deployment tasks use Java 22.
If that adds lot of effort we can delegate it to a dedicated PR as well.

**Comment by rodvar** [comment]
> @rodvar Regaring the setting of the Java version 17: We still would like to use Java 22 for Bisq 2, so it would be good to have the version configurable so that the publish tasks use Java 17 but normal build/deployment tasks use Java 22. If that adds lot of effort we can delegate it to a dedicated PR as well.

I'll give this a go today, if its not trivial I agree we can do it in a subsequent PR

**Comment by rodvar** [comment]
it's all broken after rebase, even switching to main or downloading a fresh new copy of bisq I get


```
* What went wrong:
Error resolving plugin [id: 'bisq.gradle.toolchain_resolver.ToolchainResolverPlugin']
> Included build '/projects/bisq_new/wallets/bitcoind/build-logic' does not exist.


org.gradle.api.GradleException: Error resolving plugin [id: 'bisq.gradle.toolchain_resolver.ToolchainResolverPlugin']


Caused by: org.gradle.api.InvalidUserDataException: Included build '/home/rok/projects/bisq_new/wallets/bitcoind/build-logic' does not exist.

```

HenrikJannsen:
I will make a PR with additional build instructions for git subproject in a minute.


 **Setup bitcoind git submodule:**
   At project setup run first:
   ```bash
   git submodule init
   git submodule update
   ```

   In case the submodule has changed after a project update, run:
   ```bash
   git submodule update
   ```

rodvar:
Project won't build until this gets merged to wallets/bitcoind submodule

 - https://github.com/bisq-network/bitcoind/pull/1

HenrikJannsen:
This PR need some follow-up but to not block the work on Bisq mobile we merge it as it is.
Open tasks for future PR:
- We use Java 17 but we would like to use Java 22 for the Bisq 2 app build/deployment. For publishing for mobile though we need java 17. It should be made default Java 22 and the publish task change it for the time it is running to Java 17. We should also rename the publishAll task to specify the special purpose for Bisq mobile (requiring Java 17).
- Jitpack not used yet. Need to be figured out if we want to use it. For now we use MavenLocal
- The Bisq verison in `libs.versions.toml` should be dynamically set from the root version

rodvar:
> This PR need some follow-up but to not block the work on Bisq mobile we merge it as it is. Open tasks for future PR:
> 
> * We use Java 17 but we would like to use Java 22 for the Bisq 2 app build/deployment. For publishing for mobile though we need java 17. It should be made default Java 22 and the publish task change it for the time it is running to Java 17. We should also rename the publishAll task to specify the special purpose for Bisq mobile (requiring Java 17).
> * Jitpack not used yet. Need to be figured out if we want to use it. For now we use MavenLocal
> * The Bisq verison in `libs.versions.toml` should be dynamically set from the root version

Thanks Henrik, I've created an [issue](https://github.com/bisq-network/bisq2/issues/2989) for this please have a read and contribute with anything you think appropiate

PR #2976, opened 2024-11-04
migration: Don't keep MigrationService in memory

alvasw:
The MigrationService isn't needed anymore after the data directory has
been migrated. This change removes the ApplicationService's reference to
the MigrationService after it has run.


HenrikJannsen (approved):
utACK

Issue #2634, opened 2024-08-17
Simple Html frontend for Bisq 2 monitor


HenrikJannsen:
With https://github.com/bisq-network/bisq2/pull/2633 we can request the storage data and other node data from seed nodes and oracle nodes.

We use the Rest api app. It accepts the address (or list of addresses) of the node we are requesting the report and we get a json object with the relevant data back. 

At the moment we just display the json in the browser. It would be good to get a more user friendly interface where multiple nodes are shown so that it is easy to spot the differences.

It should stay simple and low-effort. If we want to put more effort in it we should integrate it with our Grafana monitor. This could be a follow up task.

The json data:
![Screenshot 2024-08-17 at 18 35 44](https://github.com/user-attachments/assets/c600f3a4-8d5f-4708-b9ff-80547bc92198)

bsc7:
I'll take that

PR #2978, opened 2024-11-05
Simple HTML frontend for Bisq 2 monitor

bsc7:
Accessible via http://localhost:8082/node-monitor/index.html

The address list is automatically loaded via a REST interface (get-address-list). Seed node addresses are loaded from the configuration, while two oracle nodes are hardcoded.

The address list and port filter can be set in the settings as a list, separated by newlines or commas, and will be saved in a cookie. Comments (lines beginning with #) and empty lines are allowed.

Origin: https://github.com/bisq-network/bisq2/issues/2634

### DIFF
--- a/apps/rest-api-app/src/main/java/bisq/rest_api/endpoints/ReportApi.java
+++ b/apps/rest-api-app/src/main/java/bisq/rest_api/endpoints/ReportApi.java
@@ -19,6 +19,7 @@ package bisq.rest_api.endpoints;
 
 import bisq.common.network.Address;
 import bisq.common.util.CollectionUtil;
+import bisq.common.util.CompletableFutureUtils;
 import bisq.network.NetworkService;
 import bisq.rest_api.JaxRsApplication;
 import bisq.rest_api.RestApiApplicationService;
@@ -82,7 +83,9 @@ public class ReportApi {
     )
     @GET
     @Path("get-report/{address}")
-    public ReportDto getReport(@Parameter(description = "address from which we request the report") @PathParam("address") String address) {
+    public ReportDto getReport(
+            @Parameter(description = "address from which we request the report")
+            @PathParam("address") String address) {
         try {
             return fetchReportForAddress(address).join();
         } catch (Exception e) {
@@ -111,17 +114,13 @@ public class ReportApi {
             throw new RuntimeException("Failed to parse addresses from CSV input: " + addresses);
         }
 
-
         List<CompletableFuture<ReportDto>> futures = addressList.stream()
                 .map(this::fetchReportForAddress)
                 .toList();
 
-        CompletableFuture<Void> allFutures = CompletableFuture.allOf(futures.toArray(new CompletableFuture[0]));
+        CompletableFuture<List<ReportDto>> allFutures = CompletableFutureUtils.allOf(futures);
 
-        return allFutures.thenApply(v -> futures.stream()
-                        .map(CompletableFuture::join)
-                        .toList())
-                .join();
+        return allFutures.join();
     }
 
     private CompletableFuture<ReportDto> fetchReportForAddress(String addressString) {
--- a/apps/rest-api-app/src/main/java/bisq/rest_api/endpoints/ReportApi.java
+++ b/apps/rest-api-app/src/main/java/bisq/rest_api/endpoints/ReportApi.java
@@ -17,11 +17,9 @@
 
 package bisq.rest_api.endpoints;
 
-import bisq.common.util.CollectionUtil;
-import bisq.common.util.CompletableFutureUtils;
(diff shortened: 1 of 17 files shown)


apps/rest-api-app/src/main/java/bisq/rest_api/endpoints/ReportApi.java
```diff
@@ -86,16 +100,47 @@ public ReportDto getReport(@Parameter(description = "address from which we reque
...
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to parse addresses from CSV input: " + addresses);
+        }
+
+
+        List<CompletableFuture<ReportDto>> futures = addressList.stream()
+                .map(this::fetchReportForAddress)
                 .toList();
+
+        CompletableFuture<Void> allFutures = CompletableFuture.allOf(futures.toArray(new CompletableFuture[0]));
```
HenrikJannsen: We have a util method `CompletableFutureUtils.allOf` with an improved API (see comments in the method about the differences to CompletableFuture.allOf).

HenrikJannsen (approved):
ACK
Small Nit see in comment...

HenrikJannsen:
I tested it locally now. Looks great! A few small suggestions:
- Can we remove `version`, `serializedSize` and `excludedFields`? Those are properties from the `Proto` class and not relevant here. 
- Could we display the operators user name (or if "N/A" the bond user name) and node type (Seed, Oracle) next to the address? Like: "Seed node operator: HenrikJannsen; Address: hj2kj43oyq4mhd5gx4vokalnci3vlbwzclv7usocfwuj5f5iks3eivqd.onion:1000" or the like...
- Could we add a "Show diff" button and display all data which from the `Authorized Data Per Class Name`, `Authenticated Data Per Class Name` and `Mailbox Data Per Class Name` groups which do not have a 100% match among all nodes? This would help to see data which is not well distributed. It is expected due the nature of the P2P network that data is not always 100% the same but should be minimal if the network is healthy. As that might be a bigger change, feel free to do it in a new PR if you wish.

There was an inactive seed node still in the network. I banned it now and added below the code for filtering banned nodes.
Feel free to add it, then I don't need to do a new PR on top of yours...

```
public List<String> getAddressList() {
        Set<Address> bannedAddresses = bondedRolesService.getAuthorizedBondedRolesService().getBondedRoles().stream()
                .filter(BondedRole::isBanned)
                .map(BondedRole::getAuthorizedBondedRole)
                .map(AuthorizedBondedRole::getAddressByTransportTypeMap)
                .filter(Optional::isPresent)
                .map(Optional::get)
                .flatMap(map -> map.values().stream())
                .collect(Collectors.toSet());
        Map<TransportType, Set<Address>> seedAddressesByTransport = networkService.getSeedAddressesByTransportFromConfig();
        Set<TransportType> supportedTransportTypes = networkService.getSupportedTransportTypes();
        List<String> addresslist = seedAddressesByTransport.entrySet().stream()
                .filter(entry -> supportedTransportTypes.contains(entry.getKey()))
                .flatMap(entry -> entry.getValue().stream())
                .filter(address -> !bannedAddresses.contains(address))
                .map(Address::toString)
                .collect(Collectors.toList());

        // Oracle Nodes
        addresslist.add("kr4yvzlhwt5binpw7js2tsfqv6mjd4klmslmcxw3c5izsaqh5vvsp6ad.onion:36185");
        addresslist.add("s2yxxqvyofzud32mxliya3dihj5rdlowagkblqqtntxhi7cbdaufqkid.onion:54467");

        return addresslist;
    }
```

HenrikJannsen:
> > * Could we display the operators user name...
> 
> I'll need to consider the best way to implement this, but overall, it shouldn't be a big issue.

Here a code snippet how to get the data:
```
Set<String> nickNamesOrBondNames = bondedRolesService.getAuthorizedBondedRolesService().getBondedRoles().stream()
                .map(BondedRole::getAuthorizedBondedRole)
                .map(authorizedBondedRole -> userService.getUserProfileService().findUserProfile(authorizedBondedRole.getProfileId())
                        .map(UserProfile::getNickName)
                        .orElse(authorizedBondedRole.getBondUserName()))
                .collect(Collectors.toSet());
```
> 
> > * Could we add a "Show diff" button...
> 
> This will require a bit more thought on my end :) but it should also be doable.
> 
> I'm still unsure whether I'll make a separate PR for these two points; it depends on how the timing works out.

Yes feel free to do it as you wish. I also assume the diff feature might be a bit more work and would justify an additional PR.

HenrikJannsen (approved):
utACK

HenrikJannsen:
Maybe better to merge the PR. There are a few other larger PRs and risk for merge conflict rises....

Issue #2712, opened 2024-08-27
Update apache httpclient to httpclient5


HenrikJannsen:
`org.apache.httpcomponents:httpclient:4.5.14` has a transitive vulnerable dependency to `commons-codec:commons-codec:1.11`.

The next version probably without vulnerable dependencies is `org.apache.httpcomponents.client5:httpclient5:5.3.1` but it comes with many API changes. The creation of the `PoolingHttpClientConnectionManager` in `TorHttpClient` has changes APIs and it would require more investigation into their API changes to figure out what the correct way is now to create the PoolingHttpClientConnectionManager.

HenrikJannsen:
@namloan Would you be interested to look into that?

namloan:
@HenrikJannsen sure

namloan:
@HenrikJannsen how did you detect / how are you seeing that this is a vulnerable dependency?

HenrikJannsen:
in IDE select bisq root and right click: analyse/vulnerable dependencies

namloan:
This doesn't look like a vulnerability anymore. I don't know what triggered this re-evaluation. 

<img width="1015" alt="Screenshot 2024-10-13 at 8 25 13 AM" src="https://github.com/user-attachments/assets/22c10762-e585-4400-aaee-f23bdcde52cd">

The [Httpclient 4 package in Maven](https://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient/4.5.14) still references a vulnerability ([CVE-2020-15250](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-15250)) that looks very minor. Still worth upgrading to httpclient 5 to get all the improvements, but not high priority from the security perspective.

namloan:
@HenrikJannsen this issue can be closed.

PR #2990, opened 2024-11-09
Migrate Apache Http libraries from 4.x to 5.x

namloan:
I couldn't find a simpler way to avoid the DNS lookup for .onion addresses. I'll submit the feature request to Apache.

https://github.com/bisq-network/bisq2/issues/2712


HenrikJannsen (approved):
utACK

PR #3000, opened 2024-11-13
Feature/2634/simple html node monitor

bsc7:
NodeMonitor: FE: +ReportDiffs, +NodeType/UserName
* ReportDiffs: Compares reports and highlights deviations based on thresholds.
* NodeType/UserName: Displays node type and username alongside the address.
* Unified BE/FE naming conventions and API endpoints for consistency.

---
ReportDiffs enables comparison across all reports and provides average values. It supports three configurable thresholds, which can be adjusted in the settings. Threshold violations are visually highlighted, hover shows the specific average value and the amount by which it was exceeded. For subtables, the background of the button that toggles the detail view is marked based on the highest exceeded threshold.

Origin: https://github.com/bisq-network/bisq2/issues/2634, Previous PR: https://github.com/bisq-network/bisq2/pull/2978

### DIFF
--- a/apps/rest-api-app/src/main/resources/node-monitor/js/controllers/AppController.js
+++ b/apps/rest-api-app/src/main/resources/node-monitor/js/controllers/AppController.js
@@ -39,6 +39,7 @@ export class AppController {
 
     async initApp() {
         try {
+            this.#initSettings();
             await this.#updateAddressesAndLoadAllReports();
             this.#setInitialUIState();
             await this.#loadAddressDetails();
@@ -61,6 +62,11 @@ export class AppController {
     //////////////////////
     // PRIVATE METHODS
     //////////////////////
+
+    #initSettings() {
+        const savedThresholds = this.storageService.getDeviationThresholds() || Constants.DEFAULT_DEVIATION_THRESHOLDS;
+        Constants.initialize({ DEVIATION_THRESHOLDS: savedThresholds });
+    }
 
     #setInitialUIState() {
         const hasAddresses = this.addressList && this.addressList.length > 0;


bsc7:
Bildschirmfoto vom 2024-11-13 15-12-50
Bildschirmfoto vom 2024-11-13 15-13-30

HenrikJannsen:
Looks great!

Could we change the algorithm for coloring for certain fields? 
- Num Connections: Don't apply color if >10. Nodes can configure target number but should be min. 12
- Memory Used: Don't apply if < 500
- Num Threads: Don't apply if < 70
- Node Load: Don't apply if < 0.4

Let me know if you can do that in this PR or prefer a new one.

bsc7:
I'll take care of it in this PR; it shouldn't be a big deal. I just don't know if I'll get around to it in the next few days.

bsc7:
Bildschirmfoto vom 2024-11-18 18-30-03

apps/rest-api-app/build.gradle.kts
```diff
@@ -1,5 +1,6 @@
 plugins {
     id("bisq.java-library")
+    id("bisq.protobuf")
```
HenrikJannsen: Why is that needed? We don't have protobuf files here.

HenrikJannsen: Ah saw now you added a protobuf file for AddressDetails.

What is the intention? We use protobuf only or p2p network communication. Here it seems it is just a dto for the response. If so, I would suggest to remove the protobuf part and move the file into report package.

HenrikJannsen (changes requested):
See comment for AddressDetails. Beside that it looks good!

apps/rest-api-app/src/main/java/bisq/rest_api/proto/AddressDetails.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public void verify() {
+        checkArgument(address != null && address.length() > 0);
+        checkArgument(bondedRoleType != null && bondedRoleType.length() > 0);
+        checkArgument(nickNameOrBondUserName != null && nickNameOrBondUserName.length() > 0);
+    }
+
+    @Override
+    public bisq.rest_api.protobuf.AddressDetails toProto(boolean serializeForHash) {
```
HenrikJannsen: Best to use import to avoid the long name space.

HenrikJannsen (changes requested):

apps/rest-api-app/build.gradle.kts
bsc7: Ah, okay. I wasn't aware that Protobufs are primarily for P2P communication. I added them because you mentioned we should avoid using DTOs as much as possible. Since I needed one, and Protobufs seemed like the "state of the art" solution to me as a beginner, I implemented it that way.

But regarding the solution:

> I would suggest to remove the protobuf part ...

You mean I should remove the line id("bisq.protobuf") from build.gradle.kts, which means Protobuf won't be generated, and as a result, AddressDetails would need to be converted to AddressDetailsDto.
Or should Protobuf still be used? If so, I'm not very familiar with the build process: where exactly should I then define the usage of Protobuf(id("bisq.protobuf"))?

> I would suggest to ... move the file into report package.

You mean placing the transport class (whether Protobuf or DTO) alongside ReportRestApi, or using a separate package for transport classes alongside ReportRestApi? I find the latter option more elegant.

apps/rest-api-app/src/main/java/bisq/rest_api/proto/AddressDetails.java
bsc7: It would result in a name collision since the classes have the same name, which makes the import impossible. Unless, of course, I misunderstood what you meant. :)

apps/rest-api-app/build.gradle.kts
HenrikJannsen: Yes the id("bisq.protobuf") from gradle and `rest_api.proto` file can be removed as well as the `implements NetworkProto` and the 3 PB related methods in the `AddressDetails` class. The `void verify()` can be kept if you like, but its also mainly for network data to protect against ddos. 
And move that class to `bisq.rest_api.report`.

apps/rest-api-app/src/main/java/bisq/rest_api/proto/AddressDetails.java
HenrikJannsen: I just tried on your branch with `import bisq.rest_api.proto.AddressDetails;` and it worked.

bsc7: I guess the truth lies somewhere in the middle :))

HenrikJannsen (approved):
utACK

PR #3005, opened 2024-11-16
Refactor rest api

HenrikJannsen:
This PR explores the feasibility to avoid dto for the api but to use the java domain models. For the objects needed for UserProfile it worked well so far.
It also moves the API class to the domain service.


HenrikJannsen:
This PR explores the feasibility to avoid dto for the api but to use the java domain models. For the objects needed for UserProfile it worked well so far.
It also moves the API class to the domain service.

Issue #1924, opened 2024-03-24
follow established conventions for Tor integration


adrelanos:
Bisq seems to take quite a unique approach to Tor integration, not following any established conventions. Therefore making things difficult down the line and incompatible with Tor based operating systems such as Whonix  and perhaps Tails.

issues:

* It is currently not possible to use Bisq with an already installed system Tor?
* Bisq cannot be with a Tor that is already running on a different computer (or virtual machine)?
* https://github.com/bisq-network/bisq2/issues/1894

old issues fixed in git:

* Bisq even [used to](https://github.com/bisq-network/bisq2/pull/1914) parse Tor's log file. Just mentioning this to show that invented solutions without looking at the exiting conventions and talking to the Tor community, complex things will be implemented, incompatible, causing later issues, requiring refactoring.

Notes:

* Tor over Tor should be prevented. (reasons: adversarial for anonymity, bad for performance, see [Refrain from "Tor over Tor" Scenarios](https://www.whonix.org/wiki/DoNot#Refrain_from_.22Tor_over_Tor.22_Scenarios))

What are the conventions? See:

* Tor: https://gitlab.torproject.org/legacy/trac/-/wikis/doc/Tor_friendly_applications_best_practices
* Whonix: https://www.whonix.org/wiki/Dev/Project_friendly_applications_best_practices

Full disclosure: I am a developer of Whonix.

sk757a:
With the current implementation, it is impossible to connect to the network from countries with restrictions. I need to be able to use my own Tor configuration or a standalone Tor instance.

PR #3046, opened 2024-12-07
Add support for system tor

HenrikJannsen:
Adds support for using external tor process.

We provide now the `external_tor.config` config file in the tor directory.
It has a flag `UseExternalTor` to enable the usage of external Tor.

Alternatively One can set the environment variable `TOR_SKIP_LAUNCH`, or use the the JVM config [1] `-Dapplication.network.configByTransportType.tor.useExternalTor`.

If Bisq runs on a Whonix system we use the Whonix system Tor process (The `external_tor.config` can be used for customizing the config). 

If cookie authentication is enabled at the external Tor torrc file, the user need to edit the `external_tor.config` config file to mirror the `CookieAuthentication` and `CookieAuthFile` value from torrc.

`ControlPort` is set to the default value `127.0.0.1:9051`. 
The SOCKS port will be dynamically requested from the Tor control server.

In case the connection to the external Tor fails, Bisq falls back to the embedded Tor (except for Whonix, in which case we fail). 

This PR addresses issues raised in https://github.com/bisq-network/bisq2/issues/1924. In case @adrelanos would have time for a review, it would be highly appreciated.

Open issue:
It would be also a good feature to allow the embedded Tor process to stay running after Bisq gets closed. This would align better with Tor network usage and would speed up next Bisq start.
See https://github.com/bisq-network/bisq2/issues/3048

[1] See https://github.com/bisq-network/bisq2/blob/main/docs/dev/build.md how to pass JVM args to the Bisq application.


HenrikJannsen:
Found some issues...

HenrikJannsen:
Changed the initial comment to reflect the changes after the initial commit.

HenrikJannsen:
Screenshot 2024-12-09 at 21 59 56

PR #3094, opened 2025-01-12
Move go to offer logic to controller

axpoems:


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/offers/ProfileCardOffersController.java
```diff
@@ -72,4 +77,10 @@ public void updateUserProfileData(UserProfile userProfile) {
     public String getNumberOffers() {
         return String.valueOf((long) model.getListItems().size());
     }
+
+    void goToOfferbookMessage(BisqEasyOfferbookMessage bisqEasyOfferbookMessage) {
```
HenrikJannsen: We use the `on` prefix for calls from the view by convention.

HenrikJannsen (approved):
utACK

PR #3103, opened 2025-01-16
Improve user profile card handling

HenrikJannsen:


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/components/UserProfileDisplay.java
```diff
@@ -93,10 +91,10 @@ public void setUserProfile(@Nullable UserProfile userProfile) {
         if (userProfile != null) {
             userName.setText(userProfile.getUserName());
         }
+        configureOpenProfileCard(userProfile);
```
axpoems, line 94: I don't think adding this by default makes a great UX.
In particular, when the user profile display or icon are used in tables or dropdowns, clicking on the user icon will do two things: open the popup and select that row/menu. If the user's intention was to the latter, then it is annoying and confusing. Therefore, I would stick to adding it in a case by case scenario. See: https://github.com/bisq-network/bisq2/pull/3041#issue-2721016569

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/details/ProfileCardDetailsView.java
```diff
@@ -71,47 +83,59 @@ public ProfileCardDetailsView(ProfileCardDetailsModel model,
         lastUserActivityLabel = new Label();
         HBox lastUserActivityBox = createAndGetTitleAndDetailsBox("user.profileCard.details.lastUserActivity", lastUserActivityLabel);
 
-        // Version
-        versionLabel = new Label();
```
axpoems, line 75: I think this was useful to have, especially for moderators/mediators.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/terms/ProfileCardTermsView.java
```diff
@@ -15,34 +15,30 @@
...
+    private final Label tradeTermsLabel;
 
-    public ProfileCardOverviewView(ProfileCardOverviewModel model,
-                                   ProfileCardOverviewController controller) {
+    public ProfileCardTermsView(ProfileCardTermsModel model,
+                                ProfileCardTermsController controller) {
         super(new VBox(), model, controller);
 
-        // Statement
-        statementLabel = new Label();
```
axpoems: I would leave this field here instead of details, and show this as first page if any of the two is set.
Then we can revert to the original rectangular dimensions that are more characteristic of a "card", hence the profile card name.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/components/UserProfileDisplay.java
HenrikJannsen: Hm... i got confused that in some cases i could not open it like in offer list and some other places. 
We could invert the handling by using a method to not open a popup if not wanted, that would reduce the cases where we need to deal with it. But not too strong opinion on that. We can also revert it, but IMO it was missing in some places.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/details/ProfileCardDetailsView.java
HenrikJannsen: OK. As discussed in matrix, we can change the overview/details content. i will make a commit soon

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/components/UserProfileDisplay.java
HenrikJannsen: Which cases do you have in mind to not apply it?

PR #3162, opened 2025-02-02
Improve warning when trying to take buy offer

axpoems:
Add action button to navigate to reputation when seller cannot take offer.


HenrikJannsen (approved):
utACK

PR #3186, opened 2025-02-07
Network rest api service responds to ping requests

rodvar:
- ping/pong implementation as part of network api service

### DIFF
--- a/http-api/src/main/java/bisq/http_api/HttpApiService.java
+++ b/http-api/src/main/java/bisq/http_api/HttpApiService.java
@@ -25,6 +25,7 @@ import bisq.http_api.rest_api.RestApiResourceConfig;
 import bisq.http_api.rest_api.RestApiService;
 import bisq.http_api.rest_api.domain.explorer.ExplorerRestApi;
 import bisq.http_api.rest_api.domain.market_price.MarketPriceRestApi;
+import bisq.http_api.rest_api.domain.network.NetworkRestApi;
 import bisq.http_api.rest_api.domain.offers.OfferbookRestApi;
 import bisq.http_api.rest_api.domain.settings.SettingsRestApi;
 import bisq.http_api.rest_api.domain.trades.TradeRestApi;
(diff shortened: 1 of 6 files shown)


rodvar:
@HenrikJannsen please review 🙏  (see referencing bisq-mobile issue)

HenrikJannsen:
See my comments at https://github.com/bisq-network/bisq-mobile/issues/130#issuecomment-2642396610

Beside that, we should use a NetworkApi for such a feature. UserIdentityRestApi is limited to aspects concerning user identity. Also for such types if handshake we use a nonce in the ping which will be used in the pong to make sure the ping/pong comes from the correct request.

rodvar:
> See my comments at [bisq-network/bisq-mobile#130 (comment)](https://github.com/bisq-network/bisq-mobile/issues/130#issuecomment-2642396610)
> 
> Beside that, we should use a NetworkApi for such a feature. UserIdentityRestApi is limited to aspects concerning user identity. Also for such types if handshake we use a nonce in the ping which will be used in the pong to make sure the ping/pong comes from the correct request.

I'll refactor it into a NetworkApi.

Regarding the nonce, I though that was validated elsewhere from what I saw in the implementation with a uuid ? ... let me double check that and if needed I'll add that mechanism in the ping/pong itself.

rodvar:
> > I'll refactor it into a NetworkApi.
> > Regarding the nonce, I though that was validated elsewhere from what I saw in the implementation with a uuid ? ... let me double check that and if needed I'll add that mechanism in the ping/pong itself.
> 
> I think the whole feature is not needed as for node it is visible in the network service and for client the websocket connection triggers a disconnect if connection is broken.

The idea is to grab that into a unique component that every component can use to check connectivity. Specifically intended to be used by the semaphore solution to show the user the connectivity status next to the avatar.

This particular PR is needed to achieve that cause its used to check if connectivity is slow, makes sense?

HenrikJannsen:
> The idea is to grab that into a unique component that every component can use to check connectivity. Specifically intended to be used by the semaphore solution to show the user the connectivity status next to the avatar.

A service can expose that to any component as we do with all the other services. I can do that once I am back on mobile again. 

> This particular PR is needed to achieve that cause, it's used to check if connectivity is slow, makes sense?

As discussed on the other issue at mobile, this can be achieved much easier.

rodvar:
Let's do the measuring solution for all apps then for now.

PR #3194, opened 2025-02-09
Add go to message functionality in Profile Card message

axpoems:
* Sort public messages in Profile Card by date.
* Add new "go to message" option.


HenrikJannsen (approved):
ACK

PR #3198, opened 2025-02-10
expose connected nodes on network service

rodvar:


network/network/src/main/java/bisq/network/NetworkService.java
```diff
@@ -183,6 +183,12 @@ public void onPersistedApplied(NetworkServiceStore persisted) {
     // Service
     /* --------------------------------------------------------------------- */
 
+    public Integer getNumberOfConnectedNodes() {
+        NetworkId defaultNetworkId = networkIdService.getOrCreateDefaultNetworkId();
+        Map<TransportType, CompletableFuture<Node>> map = serviceNodesByTransport.getInitializedDefaultNodeByTransport(defaultNetworkId);
+        return map.size();
```
HenrikJannsen, line 189: This would return 1 in most cases as its the map of transports.
See bisq.desktop.main.left.NetworkInfo how to get the number of connections. There will be in future transport types (tor+i2p), thats why most sub services are in a map by TransportType.

HenrikJannsen (changes requested):
NACK

network/network/src/main/java/bisq/network/NetworkService.java
rodvar: oh I thought that was the nodes by transport connected, but it makes sense why the variable is called "services by transport" now.. 😛

rodvar:
needs more work

PR #3201, opened 2025-02-11
rest-api websocket default setup for LAN

rodvar:
- default websocket http rest api setup to allow LAN connectivity (to be used as trusted node for bisq-mobile with real devices)

### DIFF
--- a/apps/http-api-app/src/main/resources/http_api_app.conf
+++ b/apps/http-api-app/src/main/resources/http_api_app.conf
@@ -28,10 +28,10 @@ application {
         includeRestApi = true
         server = {
             protocol = "http://"
-            host = "localhost"
+            host = "0.0.0.0"
             port = 8090
         }
-        localhostOnly = true    // Accessible only via localhost
+        localhostOnly = false    // Accessible only via localhost, default allow LAN
         whiteListEndPoints = []
         blackListEndPoints = []
         supportedAuth = []   // supported auth schemes. If empty no authentication is required


apps/http-api-app/src/main/resources/http_api_app.conf
```diff
@@ -28,10 +28,10 @@ application {
         includeRestApi = true
         server = {
             protocol = "http://"
-            host = "localhost"
+            host = "0.0.0.0"
```
HenrikJannsen, line 31: This would allow any IP address. I don't think that is a good default value. For home network its usually 192.168.0.0 but then lcoalhost would not work. I think its best to pass the value as JVM argument for the non-localhost use cases. Localhost can be considered as a safe default and I think its best to leave it as that.

We need to get anyway Tor working and then we need some way to share the onion address.

apps/http-api-app/src/main/resources/http_api_app.conf
```diff
@@ -28,10 +28,10 @@ application {
...
             port = 8090
         }
-        localhostOnly = true    // Accessible only via localhost
+        localhostOnly = false    // Accessible only via localhost, default allow LAN
```
HenrikJannsen, line 34: Also here I would leave it to a JVM argument.

Users who run a headless node will be able to pass that argument.

HenrikJannsen (commented):
NACK

rodvar:
@HenrikJannsen Fair enough, just noting that without changing that config you can't connect from a real device to your same LAN rest-http-api instance. I'll let the team know that they need to manually change it then. We should also communicate it somehow to the operators of trusted nodes that want to pair it with Bisq Connect (created [pre-release task for that matter](https://github.com/bisq-network/bisq-mobile/issues/203)).

HenrikJannsen:
> @HenrikJannsen Fair enough, just noting that without changing that config you can't connect from a real device to your same LAN rest-http-api instance. I'll let the team know that they need to manually change it then. We should also communicate it somehow to the operators of trusted nodes that want to pair it with Bisq Connect (created [pre-release task for that matter](https://github.com/bisq-network/bisq-mobile/issues/203)).

You need to add the jvm argument for that setting. all properties can be set as jvm arg as well.

Issue #3187, opened 2025-02-07
Linebreake not working in edit message


HenrikJannsen:
When trying to add a line break by `shift+new line` in the middle of the text, it adds the new line at the end of the text.

fargp:
possible fix in https://github.com/bisq-network/bisq2/pull/3204

PR #3204, opened 2025-02-11
bug fix for Line break not working in edit message #3187

fargp:
updated processEnterKey function to fix the break line bug

Fixes https://github.com/bisq-network/bisq2/issues/3187


HenrikJannsen (approved):
utACK

PR #3211, opened 2025-02-12
Improve Profile Card messages

axpoems:
* Move logic to controller.
* Use constructor in `ChannelMessageBox` instead of setter.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/messages/ChannelMessagesDisplayList.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/messages/ChannelMessagesDisplayList.java
@@ -365,11 +365,6 @@ public class ChannelMessagesDisplayList<M extends PublicChatMessage> {
             setFillHeight(true);
             setPadding(new Insets(0, 50, 0, 50));
             getChildren().add(messageBg);
-
-            initialize();
-        }
-
-        private void initialize() {
         }
 
         private void dispose() {


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/messages/ChannelMessagesDisplayList.java
```diff
@@ -299,38 +298,46 @@ private Region getLine() {
...
+                    citationMessage.setText(citation.getText());
+                }
+            });
+
             VBox textMessageVBox = new VBox(10, citationMessageVBox, textMessageLabel);
             HBox.setMargin(textMessageVBox, new Insets(0, 0, 0, -10));
 
             catHashImageView = new ImageView();
+            catHashImageView.setImage(CatHash.getImage(channelMessageItem.getSenderUserProfile(),
+                    catHashImageView.getFitWidth()));
```
HenrikJannsen: catHashImageView.setFitWidth(30); is set after the catHashImageView.getFitWidth().
I think better put the image in the constructor and use a variable for the size:

```
int size = 30;
catHashImageView = new ImageView(CatHash.getImage(channelMessageItem.getSenderUserProfile(), size));
catHashImageView.setFitWidth(size);
catHashImageView.setFitHeight(size);
```

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/messages/ChannelMessagesDisplayList.java
```diff
@@ -299,38 +298,46 @@ private Region getLine() {
         }
     }
 
-    private static class ChannelMessageBox extends HBox {
-        private final Label dateTimeLabel, textMessageLabel, citationMessage, citationAuthor;
-        private final VBox citationMessageVBox;
+    private class ChannelMessageBox extends HBox {
```
HenrikJannsen: Better keep the static for inner classes and pass the controller as contr. argument:
`private ChannelMessageBox(ChannelMessageItem channelMessageItem, ChannelMessagesDisplayList<?>.Controller  controller) {`

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/messages/ChannelMessagesDisplayList.java
```diff
@@ -358,21 +366,11 @@ private ChannelMessageBox() {
...
-                    citationMessageVBox.setVisible(true);
-                    citationMessageVBox.setManaged(true);
-                    citationMessage.setText(citation.getText());
-                    citationAuthor.setText(citationAuthorName);
-                }
-            });
-            catHashImageView.setImage(CatHash.getImage(channelMessageItem.getSenderUserProfile(),
-                    catHashImageView.getFitWidth()));
-            textMessageLabel.setText(channelMessageItem.getMessage());
+        private void initialize() {
```
HenrikJannsen: As its empty lets remove it

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/messages/ChannelMessagesDisplayList.java
```diff
@@ -270,21 +275,15 @@ protected void onViewDetached() {
...
+                messageListVBox.getChildren().add(new ChannelMessageBox(item));
             });
         }
 
         private void clearMessageListVBox() {
-            messageListVBox.getChildren().forEach(item -> {
-                if (item instanceof ChannelMessageBox channelMessageBox) {
-                    channelMessageBox.dispose();
+            messageListVBox.getChildren().forEach(node -> {
+                if (node.getClass().equals(ChannelMessageBox.class)) {
```
HenrikJannsen: Why not use `if (node instanceof ChannelMessageBox)` - without the class being static it was not possible, maybe that was the reason?

HenrikJannsen (changes requested):

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/messages/ChannelMessagesDisplayList.java
axpoems: If I recall correctly, in previous conversations we agreed to always add initialize and dispose methods, regardless of usage. I guess it's fine to remove it as long as we remember to use the pattern.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/messages/ChannelMessagesDisplayList.java
axpoems: Yes, that was the reason. I have changed it again now with the usage of static.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/messages/ChannelMessagesDisplayList.java
HenrikJannsen: I guess that was in the context of list items, no? If we use the initialize/dispose pattern then we should move the relevant code to the initialize method (e.g. adding observers, event handlers,...).

HenrikJannsen (approved):
utACK

Issue #3206, opened 2025-02-12
Increase a bit visibility of version number next to bisq logo


HenrikJannsen:

fargp:
attempted fix in https://github.com/bisq-network/bisq2/pull/3213

PR #3213, opened 2025-02-13
Increase a bit visibility of version number next to bisq logo #3206

fargp:
updated version font size and Y position, and since the connections on the bottom shared the same class, they also got slightly bigger. 

image


fargp:
done, changes now only affect the version number. Initially I added the network also because on my (linux) it was showing so small that the font would look unreadable, image bellow 

image

PR #3219, opened 2025-02-21
fix crash on Samsung mobiles

rodvar:
- fix for https://github.com/bisq-network/bisq-mobile/issues/220 makes bisq core jars crash and user gets stuck on splash screen
 - use J8+ compatible API for path access to avoid samsung J11 ART issue

PR #3220, opened 2025-02-21
fix crash on Samsung mobiles

nostrbuddha:
* fix for #220 makes bisq core jars crash and user gets stuck on splash screen
 * use J8+ compatible API for path access to avoid samsung J11 ART issue

Extending few fixes over https://github.com/bisq-network/bisq2/pull/3219


HenrikJannsen (approved):
utACK

Do you have a link to the background info about the Samsung version support? I guess there will be more issues.

nostrbuddha:
@HenrikJannsen Wrong issue got tagged in the description. Updated the PR description to point to the right issue here at https://github.com/bisq-network/bisq-mobile/issues/220.
This has more info. This was happening in my new android device. Let me know if you need any more device info, I can get you.

PR #3231, opened 2025-02-24
Add bonded role badge for mediators and moderators

### LINKED ISSUE #2449: Add badge for user profiles who have an authorized roles (mediators, moderators)
With regards to how scammers like to impersonate support team members on matrix to steal users' seeds, it's a good idea to make the official roles on Bisq2 stand out from the rest, so a suddenwhipvapor impersonator other than having a different icon, will also miss a badge certifying him as the real thing.
This is a very botched draft I thought about:

image

axpoems:
Resolves #2449


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/components/BondedRoleBadge.java
```diff
@@ -0,0 +1,69 @@
...
+        boolean hasBondedRole = !userBondedRoles.isEmpty();
+        setVisible(hasBondedRole);
+        setManaged(hasBondedRole);
+        String bondedRoleBadgeTooltip =
+                userBondedRoles.contains(BondedRoleType.MEDIATOR) && userBondedRoles.contains(BondedRoleType.MODERATOR)
+                        ? Res.get("user.profileCard.bondedRoleBadge.MediatorAndModerator")
+                        : userBondedRoles.contains(BondedRoleType.MODERATOR) ? Res.get("user.profileCard.bondedRoleBadge.Moderator")
+                        : userBondedRoles.contains(BondedRoleType.MEDIATOR) ? Res.get("user.profileCard.bondedRoleBadge.Mediator")
+                        : "";
+        tooltip.setText(hasBondedRole ? bondedRoleBadgeTooltip : "");
```
HenrikJannsen: The hasBondedRole can be omitted as bondedRoleBadgeTooltip would be empty string anyway if hasBondedRole is true.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/components/BondedRoleBadge.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+        getChildren().add(badgeIcon);
+
+        setVisible(false);
+        setManaged(false);
+
+        Tooltip.install(this, tooltip);
+    }
+
+    public void setUserProfileBondedRoles(List<BondedRoleType> userBondedRoles) {
```
HenrikJannsen: Better use as Set as we don't require order but expect no duplication.

We don't have a strict guideline for such naming conventions but I think its better to call it applyBondedRoleTypes instead of setUserProfileBondedRoles as it is not a pure setter (also the name does not refer to the value - BondedRoleTypes). With apply prefix it gives a hint that more happens as setting a property (which is actually not happening here).

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/components/BondedRoleBadge.java
```diff
@@ -0,0 +1,69 @@
...
+        } else {
+            badgeIcon = ImageUtil.getImageViewById("moderator-badge");
+            setPadding(new Insets(0.5, -6, 0, 0));
+        }
+        getChildren().add(badgeIcon);
+
+        setVisible(false);
+        setManaged(false);
+
+        Tooltip.install(this, tooltip);
```
HenrikJannsen: Tooltip.install could be moved to setUserProfileBondedRoles

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/ProfileCardModel.java
```diff
@@ -18,21 +18,22 @@
...
+
 @Getter
 public class ProfileCardModel extends TabModel {
     @Setter
     private UserProfile userProfile;
     @Setter
-    private boolean shouldShowBondedRoleBadge;
-    @Setter
-    private String bondedRoleBadgeTooltip;
+    private List<BondedRoleType> userProfileBondedRoles;
```
HenrikJannsen: I know it gets set before we access it, but it would be more safe against potential NPE if we set a final empty list and fill it at the initWithData.

HenrikJannsen (approved):
ACK

Some NITS...

PR #3245, opened 2025-03-01
Make test agnostic of running OS

mrblue313:
Add `isWindowsPath` flag as input parameter. Otherwise, since `Path` is used, this test can fail depending on which OS it is run from.

### DIFF
--- a/persistence/src/main/java/bisq/persistence/backup/BackupService.java
+++ b/persistence/src/main/java/bisq/persistence/backup/BackupService.java
@@ -246,7 +246,8 @@ public class BackupService {
 
     @VisibleForTesting
     static Path resolveDirPath(Path dataDir, Path storeFilePath) {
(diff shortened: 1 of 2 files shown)


HenrikJannsen:
This only moves the isWindowsPath as parameter instead if having it inside the method. Or did I miss something?

mrblue313:
> This only moves the isWindowsPath as parameter instead if having it inside the method. Or did I miss something?

Yes, because if the OS running this test is Windows, this flag inside the method will always be true (Path is encoded using Windows format), and hence all tests except for the Windows ones will fail:

```
        String normalizedRelativePath = normalizedFilePath.substring(normalizedDataDir.length());
        if (isWindowsPath) {
            return normalizedRelativePath.replace("/", "\\");
        } else {
            return normalizedRelativePath;
        }
```

Another solution could be to use Strings as input param instead of Paths.

HenrikJannsen (approved):
utACK

PR #3247, opened 2025-03-02
Make 'reason for payment' text part of the protocol

### LINKED ISSUE #3238: Change chat rule and text for usage of Trade ID as reason for payment
To protect against triangular scams we should change the recommendation to use the Trade ID as reason for payment rather that to leave it blank as it is currently recommended.

It has the trade off to make it easier for certain banks to block accounts with Bitcoin related activities, but as we have seen that even with the required reputation for sellers in 2.1.6 did no stop those scams we have to add more protection. There has been at least one case where the scammer tricked the victim to trade outside Bisq thus the 2.1.6 protection did not work as the scammer acted as seller to the victim outside of Bisq.

axpoems:
Resolves #3238


HenrikJannsen (approved):
utACK

PR #3249, opened 2025-03-02
Fix notification banner visibility

axpoems:
Fix notification banner showing up again when changing screens even if previously dismissed.

### DIFF
--- a/bisq-easy/src/main/java/bisq/bisq_easy/BisqEasyNotificationsService.java
+++ b/bisq-easy/src/main/java/bisq/bisq_easy/BisqEasyNotificationsService.java
@@ -94,7 +94,6 @@ public class BisqEasyNotificationsService implements Service {
     }
 
     public CompletableFuture<Boolean> shutdown() {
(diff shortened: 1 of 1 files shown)


bisq-easy/src/main/java/bisq/bisq_easy/BisqEasyNotificationsService.java
```diff
@@ -89,17 +87,15 @@ private void handleNotification(ChatNotification notification) {
...
                 .collect(Collectors.toSet()));
-        // Reset dismissed state if we get a new notification
-        if (!tradeIdsOfNotifications.isEmpty()) {
-            isNotificationPanelDismissed.set(false);
-        }
         updateNotificationVisibilityState();
     }
 
     public CompletableFuture<Boolean> shutdown() {
+        chatNotificationService.getChangedNotification().removeObserver(this::handleNotification);
```
HenrikJannsen: We use the result from `chatNotificationService.getChangedNotification().addObserver` to unbind...

HenrikJannsen (approved):
utACK

Some small NIT

Issue #3150, opened 2025-01-28
Clean up outdated locally stored user profiles


HenrikJannsen:
We keep locally user profiles which have expired their TTL of 15 days.

The profile age data has a TTL of 30 days, and the other reputation data of 100 days.

Thus if the locally stored user profile is older than 30 days we show N/A as profile age. If it is older than 100 days, the reputation data will get removed and the reputation score will drop.

We should consider to either clean up user profiles which have expired, or mark them in the UI (visible mainly in the reputation score list) as expired if we want to keep them.

For privacy reasons I guess purging them is the preferred way.

HenrikJannsen:
Reputation data is republished by oracle nodes at startup, thus the 100 day TTL will not be affected by the users activity. Though we should consider also to take that into account to avoid keeping dead data in the network forever.

PR #3252, opened 2025-03-03
Remove persisted expired user profiles at start-up

mrblue313:
Implements #3150


HenrikJannsen:
I need to look closer into all involved areas and test it. As I am busy atm with 2.1.7 preparations and that issue is not required to get into 2.1.7 I might need a bit for review.

HenrikJannsen:
Just an update: I am a bit buried with other work, and as that is not urgent I did not had time to look closer into it. Please have patience...

PR #3261, opened 2025-03-05
Add jump to quoted message functionality

axpoems:
Based on #3258


chat/src/main/java/bisq/chat/Citation.java
```diff
@@ -29,12 +29,12 @@
 public final class Citation implements NetworkProto {
     public static final int MAX_TEXT_LENGTH = 1000;
 
-    private final String authorUserProfileId;
-    private final String text;
+    private final String authorUserProfileId, text, chatMessageId;
```
HenrikJannsen: Better not use one liner declarations for classes which have a protobuf representation as we often need to handle fields extra for backward compatibility and they are better visible as distinct fields.

Here we need to exclude the new field for the hash as otherwise old nodes would treat the message as a different object, leading to duplicated messages. I have not tested it with an old node to see if that is the case, but I highly expect it is that way.

```
    private final String authorUserProfileId;
    private final String text;
    
    // Added with v2.1.7
    @EqualsAndHashCode.Exclude
    @ExcludeForHash
    private final String chatMessageId;
```

chat/src/main/java/bisq/chat/ChatMessageType.java
```diff
@@ -25,7 +25,7 @@ public enum ChatMessageType implements ProtoEnum {
     LEAVE,
     TAKE_BISQ_EASY_OFFER,
     PROTOCOL_LOG_MESSAGE,
-    EXCHANGE_INFO_WARNING;
+    CHAT_RULES_WARNING;
```
HenrikJannsen, line 28: Missing the addition to the protobuf file.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/priv/PrivateChatsController.java
```diff
@@ -162,7 +162,7 @@ private void maybeSelectFirst() {
         if (selectionService.getSelectedChannel().get() == null &&
                 !channelService.getChannels().isEmpty() &&
                 !model.getSortedList().isEmpty()) {
-            selectionService.selectChannel(model.getSortedList().get(0).getChannel());
+            selectionService.selectChannel(model.getSortedList().getFirst().getChannel());
```
HenrikJannsen: We need to keep java 17 language compatibility for mobile. Thus use `get(0)`

chat/src/main/java/bisq/chat/Citation.java
```diff
@@ -59,11 +60,12 @@ public bisq.chat.protobuf.Citation toProto(boolean serializeForHash) {
...
-                proto.getText());
+                proto.getText(), proto.getChatMessageId());
     }
 
     public boolean isValid() {
-        return authorUserProfileId != null && !authorUserProfileId.isEmpty() &&
-                text != null && !text.isEmpty();
+        return authorUserProfileId != null && !authorUserProfileId.isEmpty()
+                && text != null && !text.isEmpty()
+                && chatMessageId != null && !chatMessageId.isEmpty();
```
HenrikJannsen: chatMessageId is empty for data from old nodes.

i18n/src/main/resources/chat.properties
```diff
@@ -158,12 +159,13 @@ chat.chatRules.content=\
   - No advertising: Avoid promoting commercial products, services, or posting promotional links.\
   - No trolling: Disruptive behavior and intentional provocation are not welcome.\
   - No impersonation: Do not use nicknames that mislead others into thinking you are a well-known person.\
+  - No fraud or scamming: Fraudulent activities are strictly forbidden.\
```
HenrikJannsen: Maybe better:

No scamming—zero tolerance: Any attempt at fraud will result in an immediate ban from the network.

i18n/src/main/resources/chat.properties
```diff
@@ -158,12 +159,13 @@ chat.chatRules.content=\
...
+  - No luring users out of the Bisq platform: Trading should be done using the Bisq protocol and network. \
```
HenrikJannsen: Maybe better:

Keep trades within Bisq: Bisq is built for secure trading. Asking peers to trade outside the platform increases the risk of fraud and will result in a network ban.

support/src/main/java/bisq/support/moderator/ModeratorService.java
```diff
@@ -180,7 +180,7 @@ public CompletableFuture<SendMessageResult> contactUser(UserProfile userProfile,
 
                     if (channel.getChatMessages().isEmpty() && isReportingUser) {
                         return twoPartyPrivateChatChannelService.sendTextMessage(Res.get("authorizedRole.moderator.replyMsg"),
-                                citationMessage.map(msg -> new Citation(userProfile.getId(), msg)),
+                                citationMessage.map(msg -> new Citation(userProfile.getId(), msg, "")),
```
HenrikJannsen: As we do not use the chatMessageId here, it would be better to make chatMessageId optional to reflect that it is a valid case that it is not set.

HenrikJannsen (changes requested):

chat/src/main/java/bisq/chat/Citation.java
HenrikJannsen: Also in case of moderator contacting requester the quote would not be shown bc of that

chat/src/main/java/bisq/chat/ChatMessageType.java
axpoems: This message is not supposed to be sent through the network. It's strictly for UI and only added as an Item. We could have it in protobuf for consistency, but it shouldn't be necessary.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/priv/PrivateChatsController.java
axpoems: I thought for the desktop app this doesn't matter, since it's not published as jar for mobile?

chat/src/main/java/bisq/chat/Citation.java
axpoems: Yes, I was going to fix this.

chat/src/main/java/bisq/chat/ChatMessageType.java
HenrikJannsen: Yes I saw that later in the review/feature but for consistency I think its still better to add. Also if we add other types later we want to keep the ordinals in sync.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/priv/PrivateChatsController.java
HenrikJannsen: Ah yes. But still better to stick with it IMO for consistency

chat/src/main/java/bisq/chat/Citation.java
```diff
@@ -17,24 +17,33 @@
...
     public static final int MAX_TEXT_LENGTH = 1000;
 
-    private final String authorUserProfileId, text, chatMessageId;
+    private final String authorUserProfileId;
+    private final String text;
+
+    // Added with v2.1.7
+    @EqualsAndHashCode.Exclude
+    @ExcludeForHash
+    private final String chatMessageId;
```
HenrikJannsen: This should be optional as well to reflect that the field can be left empty. Using empty strings or null are less explicit.

chat/src/main/java/bisq/chat/Citation.java
```diff
@@ -60,12 +69,12 @@ public bisq.chat.protobuf.Citation toProto(boolean serializeForHash) {
 
     public static Citation fromProto(bisq.chat.protobuf.Citation proto) {
         return new Citation(proto.getAuthorUserProfileId(),
-                proto.getText(), proto.getChatMessageId());
+                proto.getText(), Optional.of(proto.getChatMessageId()));
```
HenrikJannsen: The protobuf field should be set optional as well and we should use hasChatMessageId to check its presence.

HenrikJannsen (changes requested):

chat/src/main/java/bisq/chat/Citation.java
axpoems: It wasn't the intention to allow this field to be empty in the first place, but I guess the usage of `Citation` in moderation requires this.

HenrikJannsen (approved):
utACK

PR #3266, opened 2025-03-06
Add null check for isNotificationPanelVisible

axpoems:


apps/desktop/desktop/src/main/java/bisq/desktop/main/alert/AlertBannerController.java
```diff
@@ -49,8 +49,8 @@ public AlertBannerController(ServiceProvider serviceProvider) {
     @Override
     public void onActivate() {
         unconsumedAlertsPin = alertNotificationsService.getUnconsumedAlerts().addObserver(this::showAlertBanner);
-        isBisqEasyNotificationVisiblePin = bisqEasyNotificationsService.getIsNotificationPanelVisible().addObserver(
-                this::updateIsBisqEasyNotificationVisible);
+        isBisqEasyNotificationVisiblePin = bisqEasyNotificationsService.getIsNotificationPanelVisible().addObserver(isVisible ->
+                updateIsBisqEasyNotificationVisible(isVisible != null ? isVisible : false));
```
HenrikJannsen: NIT
In case we consider null as valid value (default) we usually do not apply the method call (e.g. return if null).
In case the default value should be false, its better to set it in the observable.

HenrikJannsen (approved):
NIT

HenrikJannsen (approved):
utACK

PR #3281, opened 2025-03-09
Add splitpane in offerbook

axpoems:
Add split pane with chat and offer list.
I will follow up with small improvements.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListController.java
@@ -176,6 +176,10 @@ public class OfferbookListController implements bisq.desktop.common.view.Control
         });
     }
 
+    boolean getUseAnimations() {
+        return settingsService.getUseAnimations().get();
+    }
+
     void toggleOfferList() {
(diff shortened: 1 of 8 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListController.java
```diff
@@ -176,6 +176,10 @@ public void clear() {
         });
     }
 
+    boolean getUseAnimations() {
```
HenrikJannsen: In our MVC pattern the view does only read data from model not from the controller. You can move the method to the model.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListView.java
```diff
@@ -188,6 +190,14 @@ protected void onViewAttached() {
             }
         });
 
+        widthPropertyListener = EasyBind.subscribe(root.widthProperty(), widthProperty -> {
```
HenrikJannsen: We use 'Pin' or  'subscription' as postfix for subscriptions.

HenrikJannsen (approved):
ACK

Some NITs

HenrikJannsen (approved):
utACK

PR #3282, opened 2025-03-09
Kickstart Bisq Musig UI

axpoems:


HenrikJannsen:
I think its best to move that to a branch (on bisq GH) or to add a flag to the config to hide the MS content, but maybe a separate branch is more safe. We need to keep that branch always up to date with frequent rebase...

HenrikJannsen (approved):
ACK

PR #3293, opened 2025-03-12
Fallback strategy for market prices resilience

rodvar:
- main provider was not returning a JSON for a while, found the issue was the tor get never returning and therefore the fallbacks where not being used
 - add 30 sec timeouts also for tor get
 - force exception logic when the json parsing returns no market prices

### DIFF
--- a/network/network/src/main/java/bisq/network/http/TorHttpClient.java
+++ b/network/network/src/main/java/bisq/network/http/TorHttpClient.java
@@ -25,10 +25,12 @@ import bisq.network.http.utils.Socks5ProxyProvider;
 import com.runjva.sourceforge.jsocks.protocol.Socks5Proxy;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.hc.client5.http.classic.methods.HttpGet;
+import org.apache.hc.client5.http.config.RequestConfig;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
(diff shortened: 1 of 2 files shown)


network/network/src/main/java/bisq/network/http/TorHttpClient.java
```diff
@@ -99,10 +101,15 @@ protected String doRequest(String param,
         var cm = new PoolingTorHttpClientConnectionManager();
         cm.setDefaultSocketConfig(SocketConfig.custom()
                 .setSocksProxyAddress(socksAddress)
+                .setSoTimeout(Timeout.ofSeconds(30))
                 .build());
         try {
             closeableHttpClient = checkNotNull(HttpClients.custom()
-                    .setConnectionManager(cm).build());
+                    .setConnectionManager(cm)
+                    .setDefaultRequestConfig(RequestConfig.custom()
```
HenrikJannsen, line 109: Isn't the timeout already set by the connectionManager?

rodvar: looks like its was not working. You can quickly test it by using one of the fallback tor nodes that doesn't work (or by making the runBTC call without the `/getAllMarketPrices` which would be perhaps more accurate to reproduce the same issue)

Also feel free to edit the hardcoded timeouts (I've used `30` seconds) before merging if you think another quantity is better than that 🙏)

HenrikJannsen (approved):
utACK

PR #3296, opened 2025-03-13
Reduce TTL for remove messages

HenrikJannsen:
Should not get merged into 2.1.7


HenrikJannsen:
We still require fill TTL as otherwise a node who have been offline while the remove message was sent might stay on a stale state if the the remove have been pruned by TTL in the meantime.

e.g. assume TTL is 1 day. 
We get an Add with seq nr 1 and then later a Remove with seq nr 2. State: Remove.
Another node was online while the Add message arrived, thus have state: Add(1)
When the that node goes online after the expiry for Remove triggered it will not receive the Remove message at inventory request, thus stick on it's Add state.
If TTL is the same, then the previous add state has been removed as well.

PR #3299, opened 2025-03-15
Add more improvements to the offerbook list splitpane

axpoems:
- Fix margins for a smoother transition
- Reduce min width of offerbook chatbox
- Hide splitpane horizontal resize cursor when offerlist is collapsed


HenrikJannsen (approved):
utACK

PR #3304, opened 2025-03-18
Improve discoverability of expandable offerlist

### LINKED ISSUE #2635: Make right column of offerbook more prominent
I knew there is a right column in the offerbook that allows ordering existing offers by several metrics:

image

but I didn't see this column showing, so I wondered how to enable it. Missing any other prominent feature, I clicked on the top right area of the panel, and in fact it showed as in the picture above, but I did that only because I already knew it was a feature.
I believe a new user will not find that column nor will think much of the small active area on top to enable it, so I think it is a good idea to make it open by default, and let users learn that is a thing, or make the activation area more visible/intuitive

axpoems:
Resolves #2635


HenrikJannsen (approved):
ACK

PR #3305, opened 2025-03-20
fix crash on generating swagger docs

rodvar:
- fix crash when generating swagger docs for http-api, and code docs default port
 
 
Screenshot 2025-03-20 at 15 32 11

### DIFF
--- a/apps/http-api-app/src/main/java/bisq/http_api_node/HttpApiApp.java
+++ b/apps/http-api-app/src/main/java/bisq/http_api_node/HttpApiApp.java
@@ -23,7 +23,7 @@ import lombok.extern.slf4j.Slf4j;
 
 /**
  * JAX-RS application for the Bisq REST API
- * Swagger docs at: http://localhost:8082/doc/v1/index.html
+ * Swagger docs at: http://localhost:8090/doc/v1/index.html
  */
 @Slf4j
 public class HttpApiApp extends Executable<HttpApiApplicationService> {
--- a/apps/http-api-app/src/main/java/bisq/http_api_node/HttpApiApp.java
+++ b/apps/http-api-app/src/main/java/bisq/http_api_node/HttpApiApp.java
@@ -23,7 +23,8 @@ import lombok.extern.slf4j.Slf4j;
 
 /**
  * JAX-RS application for the Bisq REST API
- * Swagger docs at: http://localhost:8090/doc/v1/index.html
+ * Swagger docs at: http://localhost:8090/doc/v1/index.html or http://localhost:8082/doc/v1/index.html in case RestAPI
+ * is used without websockets
  */
 @Slf4j
 public class HttpApiApp extends Executable<HttpApiApplicationService> {
--- a/http-api/src/main/java/bisq/http_api/web_socket/util/GrizzlySwaggerHttpHandler.java
+++ b/http-api/src/main/java/bisq/http_api/web_socket/util/GrizzlySwaggerHttpHandler.java
@@ -25,6 +25,7 @@ import org.glassfish.grizzly.http.server.Response;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import java.net.URLConnection;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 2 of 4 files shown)


http-api/src/main/java/bisq/http_api/web_socket/util/GrizzlySwaggerHttpHandler.java
```diff
@@ -51,16 +51,41 @@ public void service(Request request, Response response) throws Exception {
...
             response.setStatus(200);
-            URL resourceUrl = classLoader.getResource(resourceName);
-            if (resourceUrl != null) {
-                Path path = Paths.get(resourceUrl.toURI());
-                response.setContentLengthLong(Files.size(path));
-            }
-            inputStream.transferTo(response.getOutputStream());
+            
+            // Read all bytes to determine content length instead of using Paths which causes a crash in most cases
+            byte[] data = inputStream.readAllBytes();
```
HenrikJannsen, line 57: inputStream.readAllBytes() loads the entire resource into memory as a byte[]

Better:

```
byte[] buffer = new byte[8192];
    int bytesRead;
    while ((bytesRead = inputStream.read(buffer)) != -1) {
        outputStream.write(buffer, 0, bytesRead);
    }
    ```

File: http-api/src/main/java/bisq/http_api/web_socket/util/GrizzlySwaggerHttpHandler.java [hunk 2]
```diff
@@ -51,16 +51,41 @@ public void service(Request request, Response response) throws Exception {
...
+            response.setContentLengthLong(data.length);
+            setContentType(response, resourceName);
```
HenrikJannsen, line 58: Better use provided methods to avoid missing some types like `.gif`.

```
String contentType = URLConnection.guessContentTypeFromName(resourceName);
    if (contentType == null) {
        contentType = "application/octet-stream"; // Default binary
    }
```

apps/http-api-app/src/main/java/bisq/http_api_node/HttpApiApp.java
```diff
@@ -23,7 +23,7 @@
 
 /**
  * JAX-RS application for the Bisq REST API
- * Swagger docs at: http://localhost:8082/doc/v1/index.html
+ * Swagger docs at: http://localhost:8090/doc/v1/index.html
```
HenrikJannsen: Rest API use 8082. if Websocket is used with includeRestApi enabled we use 8090. Maybe use:

Swagger docs at: http://localhost:8090/doc/v1/index.html or http://localhost:8082/doc/v1/index.html in case RestAPI is used without websockets

HenrikJannsen (changes requested):

http-api/src/main/java/bisq/http_api/web_socket/util/GrizzlySwaggerHttpHandler.java
rodvar: excellent, I'll change that now

http-api/src/main/java/bisq/http_api/web_socket/util/GrizzlySwaggerHttpHandler.java
rodvar: that's correct, but the memory needed for this would be quite small if I'm not mistaken so I thought the convenience and readability of `readAllBytes` was better in this case.

If you still think it needs to be changed feel free to do it before merge 🙏

apps/http-api-app/src/main/java/bisq/http_api_node/HttpApiApp.java
rodvar: thanks for pointing this out, I'll fix it now

http-api/src/main/java/bisq/http_api/web_socket/util/GrizzlySwaggerHttpHandler.java
HenrikJannsen: Yes, we can leave it.

HenrikJannsen (approved):
utACK

PR #3324, opened 2025-03-24
Feature/api expose profile age

rodvar:
- for connect apps to solve https://github.com/bisq-network/bisq-mobile/issues/215


HenrikJannsen (commented):
NACK

Profile age is part of the reputation domain and not user profile.
I will provide a solution after I have done the chat integration.

PR #3325, opened 2025-03-25
http-api-bisq-version API

rodvar:
- trusted node work for https://github.com/bisq-network/bisq-mobile/issues/136
 - expose api version in settings/version path


HenrikJannsen (approved):
utACK

Issue #2949, opened 2024-10-23
Replace usage of deprecated `com.sun.nio.file.SensitivityWatchEventModifier`


HenrikJannsen:

HenrikJannsen:
@namloan Would you be available for that task?

namloan:
@HenrikJannsen yes, please assign it to me.

wuhualiang:
https://github.com/bisq-network/bisq2/pull/3327

PR #3327, opened 2025-03-26
remove usage of deprecated SensitivityWatchEventModifier

wuhualiang:
Replace usage of deprecated com.sun.nio.file.SensitivityWatchEventModifier

https://github.com/bisq-network/bisq2/issues/2949

### DIFF
--- a/java-se/src/main/java/bisq/java_se/utils/FileCreationWatcher.java
+++ b/java-se/src/main/java/bisq/java_se/utils/FileCreationWatcher.java
@@ -72,7 +72,7 @@ public class FileCreationWatcher {
                     }
                 }
                 if(!watchKey.reset()){
-                    throw new IllegalStateException("File watcher is no longer valid.");
+                    log.warn("File watcher is no longer valid.");
                 }
             }
         } catch (IOException | InterruptedException e) {
--- a/java-se/src/main/java/bisq/java_se/utils/FileCreationWatcher.java
+++ b/java-se/src/main/java/bisq/java_se/utils/FileCreationWatcher.java
@@ -47,9 +47,12 @@ public class FileCreationWatcher {
     private Path waitForNewFile(Optional<Path> optionalPath) {
         try (WatchService watchService = FileSystems.getDefault().newWatchService()) {
             directoryToWatch.register(watchService,
-                    new WatchEvent.Kind[]{StandardWatchEventKinds.ENTRY_CREATE});
+                    StandardWatchEventKinds.ENTRY_CREATE);
             while (true) {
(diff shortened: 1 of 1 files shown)


java-se/src/main/java/bisq/java_se/utils/FileCreationWatcher.java
```diff
@@ -48,8 +48,7 @@ public Future<Path> waitForFile(Path path) {
     private Path waitForNewFile(Optional<Path> optionalPath) {
         try (WatchService watchService = FileSystems.getDefault().newWatchService()) {
             directoryToWatch.register(watchService,
-                    new WatchEvent.Kind[]{StandardWatchEventKinds.ENTRY_CREATE},
-                    SensitivityWatchEventModifier.HIGH);
+                    new WatchEvent.Kind[]{StandardWatchEventKinds.ENTRY_CREATE});
```
HenrikJannsen: Can be  shortened to:`directoryToWatch.register(watchService, StandardWatchEventKinds.ENTRY_CREATE);`

HenrikJannsen (approved):
utACK

@alvasw Would you like to have a look as well?

From java docs:
```
Deprecated
The sensitivity levels were historically used by polling-based WatchService implementations to configure the polling interval. They are are no longer used. The WatchService implementations in the JDK ignore these WatchEvent modifiers if they are specified when registering a directory to be watched.
```

java-se/src/main/java/bisq/java_se/utils/FileCreationWatcher.java
wuhualiang: done

java-se/src/main/java/bisq/java_se/utils/FileCreationWatcher.java
```diff
@@ -67,7 +70,9 @@ private Path waitForNewFile(Optional<Path> optionalPath) {
                     } else if (optionalPath.get().equals(newFilePath)) {
                         return newFilePath;
                     }
-                    watchKey.reset();
+                }
+                if(!watchKey.reset()){
+                    throw new IllegalStateException("File watcher is no longer valid.");
```
HenrikJannsen: Maybe a warning log would be better here. Not sure if there is a case where that can happen, but it feels its not a fatal issue to justify an exception which is likely not handled gracefully upstream. What would be a graceful handling in that case?

HenrikJannsen (approved):
utACK with questions

java-se/src/main/java/bisq/java_se/utils/FileCreationWatcher.java
wuhualiang: OK change to record warning log；the exception suggest from chartgpt

Issue #3318, opened 2025-03-22
Investigate issue with PeerExchangeResponse receiving invalid nonce


HenrikJannsen:
I saw some times warn log:
`Received a PeerExchangeResponse from [6002] with an invalid nonce. response.nonce()=-746010233, nonce=-1647025841`.
Investigate what causes that

wuhualiang:
@HenrikJannsen There should be two possibilities：
1. Link maintenance (extendPeerGroupAsync) The last request response arrives
2. It may be that the previous response under the retry mechanism (retryPeerExchangeAsync) has arrived
It seems to be related to the network environment. Are there many such logs?

wuhualiang:
Yes no effect. Ignore this response. Let me change the description and level of this log.

PR #3337, opened 2025-03-31
Investigate issue with PeerExchangeResponse receiving invalid nonce #3318

wuhualiang:
PeerExchangeResponse does not match with requesthandler ， ignore it


HenrikJannsen (approved):
utACK

PR #3350, opened 2025-04-01
FileCreationWatcher: Fix timeout deadlock

alvasw:
PR https://github.com/bisq-network/bisq2/pull/3327 introduced a deadlock. We handle the timeout explicitly now.

Other changes:
- [FileCreationWatcher: Format code](https://github.com/bisq-network/bisq2/commit/acad7aed4f8955117ad981df9a1bb7650946cc9c)


HenrikJannsen (approved):
utACK

Issue #3347, opened 2025-04-01
STUCK ON LOADING AND WON'T OPEN ON WINDOWS 11 ENTERPRISE


BeatleRider:
I'm running Bisq2 on Windows 11 enterprise and New Version 2.1.6 & 2.1.7 always stuck on loading and keeps freezing on startup. I've tried all the troubleshoot I could ever think of, from uninstalling & reinstalling to deleting the tor directory folder, but Bisq2 won't startup still

BeatleRider:
@HenrikJannsen After troubleshooting once more, I successfully launched Bisq2 by moving the entire Bisq2 directory, uninstalling and reinstalling the program, and creating a new Bisq2 folder. However, I'm concerned that I'll lose my profile settings with this new installation. Is there a workaround to transfer my existing profile to the new setup?

BeatleRider:
@HenrikJannsen I finally got Bisq2 working! I did a lot of troubleshooting to get it right though, had to to delete all backup folders to get it running… Thanks for your reply and support, really do appreciate it

wuhualiang:
@HenrikJannsen  seems it's tuck with OOM 
Apr-01 04:27:40.155 [PeriodicScheduler.JvmMemoryReportService.logReport] INFO  b.j.j.JvmMemoryReportService: 
************************************************************************************************************************
Total memory: 882 MB; Used memory: 851.887 MB; Free memory: 30.113 MB; Max memory: 882 MB; No. of threads: 41
************************************************************************************************************************

Apr-01 04:33:11.258 [Network.IO-2.PeerExchangeAttempt.request] ERROR b.d.DesktopExecutable: Uncaught exception: java.lang.OutOfMemoryError: Java heap space

Apr-01 04:33:11.258 [CompletableFutureDelayScheduler] ERROR b.d.DesktopApplicationService: Initializing applicationService failed java.util.concurrent.TimeoutException: null
	at java.base/java.util.concurrent.CompletableFuture$Timeout.run(CompletableFuture.java:2920)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
[...]

PR #3353, opened 2025-04-03
heap size config

wuhualiang:
max heap size from 2G to 6G and the heap size can be adjusted

### DIFF
--- a/apps/desktop/desktop-app/build.gradle.kts
+++ b/apps/desktop/desktop-app/build.gradle.kts
@@ -28,6 +28,7 @@ java {
 
 application {
     mainClass.set("bisq.desktop_app.DesktopApp")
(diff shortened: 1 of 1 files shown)


apps/desktop/desktop-app/build.gradle.kts
```diff
@@ -28,6 +28,7 @@ java {
 
 application {
     mainClass.set("bisq.desktop_app.DesktopApp")
+    applicationDefaultJvmArgs = listOf("-XX:MaxRAM=8g", "-XX:MaxHeapFreeRatio=10", "-XX:MinHeapFreeRatio=5")
```
54corbin, line 31: Out of quriosity: why do we need this change? I may lost some context.

Thank you.

wuhualiang: https://github.com/bisq-network/bisq2/issues/3347  here oom

PR #3357, opened 2025-04-04
Added BtcSatsText component

mvp1983:
Added BtcSatsText component and used it in ProfileCardOverviewView, ReviewDataDisplay, TopPanelView.

Inspired by [this](https://github.com/bisq-network/bisq-mobile/blob/main/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/components/atoms/BtcSatsText.kt) component

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced Bitcoin display components that present amounts in a clean, formatted style for transaction views, trade screens, profiles, and balance panels.
  - Implemented dynamic updates to reflect Bitcoin-specific transactions, improving clarity when sending and receiving funds.

- **Style**
  - Rolled out updated visual styling and formatting rules to ensure consistent presentation of Bitcoin figures, including integer parts, leading zeros, and significant digits.

- **Tests**
  - Added comprehensive tests verifying the correctness and locale-sensitive formatting of the Bitcoin display component.


HenrikJannsen:
Please use the 

`devModeReputationScore` jvm argument as described in the dev docs.


### Setting devMode and devModeReputationScore
A seller requires reputation for trading. This can be achieved in dev environment with setting up an oracle node and Bisq 1, though this setup is a bit advanced.
To make it easier for devs, we added a `devModeReputationScore` field to the config. It also requires that the `devMode` flag is set to true.
Set the desired reputation score to `devModeReputationScore` and apply that to all your trade apps. This value will not 
be applied per user profile but overrides the reputation score lookup globally.
For proper release testing a correct setup should be used, as this workaround does not cover all the use cases for reputation.
```sh
JAVA_OPTS="-Dapplication.appName=bisq2_Alice_clear \
    -Dapplication.network.supportedTransportTypes.0=CLEAR \
    -Dapplication.network.seedAddressByTransportType.clear.0=127.0.0.1:8000 \
    -Dapplication.network.seedAddressByTransportType.clear.1=127.0.0.1:8001
    -Dapplication.devMode=true \
    -Dapplication.devModeReputationScore=50000" \
    apps/desktop/desktop-app/build/install/desktop-app/bin/desktop-app
```

If you still can replicate the stack overflow post the stack trace. I guess it was caused by those other dev changes.

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BtcSatsText.java
```diff
@@ -0,0 +1,197 @@
...
+    @Getter
+    private final Text significantDigits = new Text();
+
+    @Getter
+    private final Text btcCode = new Text();
+
+    public BtcSatsText(String amount) {
+
+        setSpacing(0);
+        setPadding(new Insets(0));
```
HenrikJannsen: Not needed as its the default.

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BtcSatsText.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @Getter
+    private final Text significantDigits = new Text();
+
+    @Getter
+    private final Text btcCode = new Text();
+
+    public BtcSatsText(String amount) {
+
+        setSpacing(0);
```
HenrikJannsen: Not needed as its the default.

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BtcSatsText.java
```diff
@@ -0,0 +1,197 @@
...
+        while (i < formattedFractional.length() &&
+                (formattedFractional.charAt(i) == '0' || formattedFractional.charAt(i) == ' ')) {
+            leadingZeros.append(formattedFractional.charAt(i));
+            i++;
+        }
+
+        String significantDigits = formattedFractional.substring(i);
+
+        this.integerPart.setText(integerPart + ".");
+        this.integerPart.setFill(Integer.parseInt(integerPart) > 0 ? Color.WHITE : Color.GRAY);
```
HenrikJannsen: We should use our css or at least the color codes defined in the base.css

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BtcSatsText.java
```diff
@@ -0,0 +1,197 @@
...
+
+        this.leadingZeros.setText(leadingZeros.toString());
+        this.leadingZeros.setFill(Color.GRAY);
+
+        this.significantDigits.setText(significantDigits);
+        this.significantDigits.setFill(Color.WHITE);
+
+        if (showCode) {
+            this.btcCode.setText(" BTC");
+            this.btcCode.setFill(Color.GRAY);
```
HenrikJannsen: Can be set a style in Text as it never changes dynamically

HenrikJannsen:
Got below exception on app with German language selected (in German `,` is used as comma separator).

```
java.lang.NumberFormatException: For input string: "0,00000000"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
	at java.base/java.lang.Integer.parseInt(Integer.java:662)
	at java.base/java.lang.Integer.parseInt(Integer.java:778)
	at bisq.desktop.components.controls.BtcSatsText.formatBtcAmount(BtcSatsText.java:120)
	at bisq.desktop.components.controls.BtcSatsText.updateDisplay(BtcSatsText.java:85)
	at bisq.desktop.components.controls.BtcSatsText.lambda$new$0(BtcSatsText.java:69)
```

HenrikJannsen:
In the below cases its not yet implemented. I know you mentioned that in the item renderer you got problems and we can leave it to me or @axpoems to tackle that. But the other cases would be good to get added as well. 

Screenshot 2025-04-05 at 14 45 13
Screenshot 2025-04-05 at 14 45 04
Screenshot 2025-04-05 at 14 44 47
Screenshot 2025-04-05 at 14 44 34
Screenshot 2025-04-05 at 14 40 51

apps/desktop/desktop/src/main/resources/css/controls.css
```diff
@@ -1075,3 +1075,36 @@
...
 }
+
+/*******************************************************************************
+ *                                                                             *
+ * BtcSatsText                                                                 *
+ *                                                                             *
+ ******************************************************************************/
+
+.btc-sats-text {
+    -fx-font-family: "IBM Plex Sans Light"
```
HenrikJannsen: Missing `;`

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BtcSatsText.java
```diff
@@ -0,0 +1,197 @@
...
+import javafx.geometry.Pos;
+import javafx.scene.layout.HBox;
+import javafx.scene.paint.Color;
+import javafx.scene.text.Font;
+import javafx.scene.text.Text;
+import javafx.scene.text.TextAlignment;
+import javafx.scene.text.TextFlow;
+import lombok.Getter;
+
+public class BtcSatsText extends HBox {
```
HenrikJannsen: Maybe better component name would be `BitcoinAmountDisplay`

HenrikJannsen (changes requested):

mvp1983:
> Got below exception on app with German language selected (in German `,` is used as comma separator).
> 
> ```
> java.lang.NumberFormatException: For input string: "0,00000000"
> 	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
> 	at java.base/java.lang.Integer.parseInt(Integer.java:662)
> 	at java.base/java.lang.Integer.parseInt(Integer.java:778)
> 	at bisq.desktop.components.controls.BtcSatsText.formatBtcAmount(BtcSatsText.java:120)
> 	at bisq.desktop.components.controls.BtcSatsText.updateDisplay(BtcSatsText.java:85)
> 	at bisq.desktop.components.controls.BtcSatsText.lambda$new$0(BtcSatsText.java:69)
> ```

yep, my fault. will be fixed in the next commit

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BtcSatsText.java
mvp1983: fixed

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BtcSatsText.java
mvp1983: fixed

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BtcSatsText.java
mvp1983: fixed

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BtcSatsText.java
mvp1983: fixed

apps/desktop/desktop/src/main/resources/css/controls.css
mvp1983: fixed

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BtcSatsText.java
mvp1983: done

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BitcoinAmountDisplay.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    public void setFontSize(double fontSize) {
+        integerPart.setFont(new Font(integerPart.getFont().getName(), fontSize));
+        leadingZeros.setFont(new Font(leadingZeros.getFont().getName(), fontSize));
+        significantDigits.setFont(new Font(significantDigits.getFont().getName(), fontSize));
+    }
+
+    public void setBtcCodeFontSize(double fontSize) {
+        Text btcCodeText = getBtcCode();
```
HenrikJannsen: No need for getter here

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BitcoinAmountDisplay.java
```diff
@@ -0,0 +1,216 @@
...
+        setAlignment(Pos.CENTER);
+        setSpacing(0);
+        valueTextFlow.setLineSpacing(0);
+    }
+
+    public void setTextAlignment(TextAlignment alignment) {
+        valueTextFlow.setTextAlignment(alignment);
+    }
+
+    public void setHeightConstraints(double minHeight, double maxHeight) {
```
HenrikJannsen: When we apply min and max height/width we use `setFixedHeight` and pass only one parameter. Otherwise if different params are expected we don't need that method as the setMinHeight and setMaxHeight can be used directly.

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BitcoinAmountDisplay.java
```diff
@@ -0,0 +1,216 @@
...
+                    "btc-leading-zeros-empty", "btc-leading-zeros-dimmed");
+        } else {
+            setExclusiveStyle(this.leadingZeros, "btc-leading-zeros-dimmed",
+                    "btc-leading-zeros-empty");
+        }
+
+        this.significantDigits.setText(significantDigits);
+
+        if (showCode) {
+            this.btcCode.setText(" BTC");
```
HenrikJannsen: This can be set in the constructor as it never changes. `Text btcCode = new Text(" BTC");`

HenrikJannsen: And then you can use `btcCode.setVisible(showCode)`

HenrikJannsen: And the `this.` is not required here. To use the same name for the text and the component (like with `significantDigits`) might be a bit harder to read. I think its better to use different var names like significantDigitsValue (significantDigitsText is here not so great as Text is ambiguous for the component and the text property).

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BitcoinAmountDisplay.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    private void setExclusiveStyle(Text textNode, String styleToAdd, String styleToRemove) {
+        textNode.getStyleClass().remove(styleToRemove);
+        if (!textNode.getStyleClass().contains(styleToAdd)) {
+            textNode.getStyleClass().add(styleToAdd);
+        }
+    }
+
+    private void formatBtcAmount(String amount, boolean showCode) {
```
HenrikJannsen: I think it would be good to make that method testable and add unit tests. There are many possible inputs which can be tested if the result is as expected. No need to test the styles in the grouping, just the formatted output.

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/MaterialBtcTextField.java
```diff
@@ -0,0 +1,139 @@
...
+
+import bisq.common.util.StringUtils;
+import javafx.beans.value.ChangeListener;
+import javafx.beans.value.ObservableValue;
+import javafx.beans.value.WeakChangeListener;
+import lombok.Getter;
+
+import javax.annotation.Nullable;
+
+public class MaterialBtcTextField extends MaterialTextField {
```
HenrikJannsen: Should derive its name from BitcoinAmountDisplay -> e.g. `MaterialBitcoinAmountDisplay`

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BitcoinAmountDisplay.java
```diff
@@ -0,0 +1,216 @@
...
+        integerPart.getStyleClass().add("btc-integer-part");
+        leadingZeros.getStyleClass().add("btc-leading-zeros-empty");
+        significantDigits.getStyleClass().add("btc-significant-digits");
+        btcCode.getStyleClass().add("btc-code");
+
+        valueTextFlow.getChildren().addAll(integerPart, leadingZeros, significantDigits, btcCode);
+
+        btcAmount.set(amount);
+
+        btcAmount.addListener((obs, old, newVal) -> updateDisplay());
```
HenrikJannsen: Better to use WeakChangeListener

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/components/ReviewDataDisplay.java
```diff
@@ -145,12 +156,23 @@ private View(Model model, Controller controller) {
...
+            configureBitcoinAmountDisplay(toSendBtcAmount);
+            configureBitcoinAmountDisplay(toReceiveBtcAmount);
+
             direction = getElements(Res.get("bisqEasy.tradeState.header.direction"));
             toSend = getAmountElements();
             toReceive = getAmountElements();
             paymentMethod = getElements();
         }
 
+        private void configureBitcoinAmountDisplay(BitcoinAmountDisplay btcText) {
```
HenrikJannsen: We arrange the methods so that private methods are after public and protected ones, specially the onViewAttached and onViewDetached as those are the most relevant ones.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/FormUtils.java
```diff
@@ -104,6 +105,17 @@ public static MaterialTextField getTextField(String description, String value, b
         return field;
     }
 
+    public static MaterialBtcTextField getBtcTextField(String description, String value, boolean isEditable) {
```
HenrikJannsen: Should be `getMaterialBtcTextField` or after rename `getMaterialBitcoinAmountDisplay`

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
```diff
@@ -264,12 +262,21 @@ protected Model(BisqEasyTrade bisqEasyTrade, BisqEasyOpenTradeChannel channel) {
...
+            BASE_AMOUNT
+        }
+
         private static final Interpolator INTERPOLATOR = Interpolator.SPLINE(0.25, 0.1, 0.25, 1);
 
         private final Button sentButton;
         private final MaterialTextField paymentProof;
-        private final WrappingText sendBtcHeadline, fiatReceiptConfirmed;
-        private final MaterialTextField baseAmount;
+        @SuppressWarnings("FieldCanBeLocal")
```
HenrikJannsen: Why suppress and not make them local?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
```diff
@@ -362,6 +381,12 @@ protected void onViewAttached() {
             baseAmount.getIconButton().setOnAction(e -> ClipboardUtil.copyToClipboard(model.getBaseAmount()));
             qrCodeVBox.setOnMouseClicked(e -> controller.onShowQrCodeDisplay());
             sentButton.setOnAction(e -> controller.onConfirmedBtcSent());
+            sendBtcHeadline.setFocusTraversable(true);
+            root.layoutBoundsProperty().addListener((obs, old, newVal) -> {
```
HenrikJannsen: Why is that needed? In case we need a slight delay because of some render order issues we can use `UIThread.runOnNextRenderFrame();` . If the listener is needed it must be kept as field and removed in `onViewDetached`

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/TradeDataHeader.java
```diff
@@ -245,32 +404,47 @@ private Triple<Text, Text, VBox> getElements(@Nullable String description) {
...
             return new Triple<>(descriptionLabel, valueLabel, vBox);
         }
 
-        private Triple<Triple<Text, Text, Text>, HBox, VBox> getAmountElements() {
+        private Triple<Triple<Text, Node, Text>, HBox, VBox> getAmountElements() {
             Text descriptionLabel = new Text();
             descriptionLabel.getStyleClass().add("bisq-easy-open-trades-header-description");
-            Text amount = new Text();
-            amount.getStyleClass().add("bisq-easy-open-trades-header-value");
+            BitcoinAmountDisplay btcAmountText = new BitcoinAmountDisplay("0");
```
HenrikJannsen: Use `BitcoinAmountDisplay bitcoinAmountDisplay`. When renaming a class IntelliJ asks you if you want to rename fields with the matching name.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/TradeDataHeader.java
```diff
@@ -195,44 +205,193 @@ private View(Model model, Controller controller) {
...
+            );
+
             userProfilePin = EasyBind.subscribe(model.getPeersUserProfile(), peersUserProfileDisplay::setUserProfile);
             reputationScorePin = EasyBind.subscribe(model.getReputationScore(), peersUserProfileDisplay::setReputationScore);
+
+            layoutBoundsListener =
+                    (obs, old, newVal) -> refreshBitcoinAmountDisplayComponents();
+            root.layoutBoundsProperty().addListener(layoutBoundsListener);
+
+            refreshBitcoinAmountDisplayComponents();
```
HenrikJannsen: We use mostly `EasyBind.subscribe` instead of JavaFx listeners as with `EasyBind.subscribe` the handler code is called when subscribing thus avoiding the additional call of the handler method and in simple cases does not require to extract the handler code as method.

Only when we explicitly don't want that behavior or for listchanges we use JavaFx listeners.

This applies to the other cases where you use JavaFx listeners as well

apps/desktop/desktop/src/main/resources/css/controls.css
```diff
@@ -1075,3 +1075,48 @@
     -fx-border-insets: 2 0 2 0;
     -fx-padding: -2 2 -2 2;
 }
+
+/*******************************************************************************
+ *                                                                             *
+ * BtcSatsText                                                                 *
```
HenrikJannsen: Apply renamed version

i18n/src/main/resources/bisq_easy_af_ZA.properties
```diff
@@ -765,7 +765,8 @@ bisqEasy.tradeState.info.seller.phase2b.fiatReceivedButton=Bevestig ontvangs van
 bisqEasy.tradeState.info.seller.phase2b.tradeLogMessage={0} het die ontvangs van {1} bevestig
 
 bisqEasy.tradeState.info.seller.phase3a.fiatPaymentReceivedCheckBox=Ek het bevestig dat ek {0} ontvang het
-bisqEasy.tradeState.info.seller.phase3a.sendBtc=Stuur {0} na die koper
+bisqEasy.tradeState.info.seller.phase3a.sendBtc.prefix=Stuur
```
HenrikJannsen: No need to edit translated strings. This will be automatically done before release. Only default i18n file need to be edited.

HenrikJannsen (changes requested):
Good progress! I am aware that this task is quite advanced and not all details can be known to new devs. Just pointed out what I spotted. 

Please add tests so we can be sure all edge cases are correctly handled.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
mvp1983: My fault. My thoughts were to keep reference on component, but actually HBox (sendBtcHeadline) keeps reference on them, so you are right - both can be local.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerState3a.java
mvp1983: Actually, it was a fight with JavaFx (and solution works), however, it is not very good solution, I agree. 
I've remade with css

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BitcoinAmountDisplay.java
mvp1983: Tests are added

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/TradeDataHeader.java
mvp1983: layoutBoundsListener was removed after css-changes, but I used EasyBind for other listeners in this file (and some others)

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BitcoinAmountDisplay.java
```diff
@@ -65,12 +70,69 @@ public BitcoinAmountDisplay(String amount) {
 
         btcAmount.set(amount);
 
-        btcAmount.addListener((obs, old, newVal) -> updateDisplay());
-        showBtcCode.addListener((obs, old, newVal) -> updateDisplay());
+        btcAmount.addListener(new WeakChangeListener<>(amountChangeListener));
+
+        btcCode.setText(" BTC");
+        getStyleClass().add("btc-sats-text");
```
HenrikJannsen: css should reflect component name (`bitcoin-amount-display`)

Applies also to above like: "btc-integer-part". As those are global css in control.css we need to make clear which component they belong to.

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BitcoinAmountDisplay.java
```diff
@@ -65,12 +70,69 @@ public BitcoinAmountDisplay(String amount) {
 
         btcAmount.set(amount);
 
-        btcAmount.addListener((obs, old, newVal) -> updateDisplay());
-        showBtcCode.addListener((obs, old, newVal) -> updateDisplay());
+        btcAmount.addListener(new WeakChangeListener<>(amountChangeListener));
+
+        btcCode.setText(" BTC");
```
HenrikJannsen: Should be in constructor: `Text btcCode = new Text(" BTC")`

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BitcoinAmountDisplay.java
```diff
@@ -65,12 +70,69 @@ public BitcoinAmountDisplay(String amount) {
...
+        getStyleClass().add("btc-sats-text");
 
         updateDisplay();
```
HenrikJannsen, line 81: Remove trailing empty line after updateDisplay();

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BitcoinAmountDisplay.java
```diff
@@ -65,12 +70,69 @@ public BitcoinAmountDisplay(String amount) {
...
+        setAlignment(Pos.CENTER);
+        setSpacing(0);
+        valueTextFlow.setLineSpacing(0);
+    }
+
+    public void setTextAlignment(TextAlignment alignment) {
+        valueTextFlow.setTextAlignment(alignment);
+    }
+
+    public void setFixedHeight(double maxHeight) {
```
HenrikJannsen: Param should be height not maxHeight

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/MaterialBitcoinAmountDisplay.java
```diff
@@ -25,44 +25,52 @@
...
 
-    public MaterialBtcTextField(@Nullable String description,
-                                @Nullable String prompt,
-                                @Nullable String help) {
+    public MaterialBitcoinAmountDisplay(@Nullable String description,
+                                        @Nullable String prompt,
+                                        @Nullable String help) {
         super(description, prompt, help);
 
         bitcoinAmountDisplay = new BitcoinAmountDisplay("0");
```
HenrikJannsen: Better add a no arg constructor: new BitcoinAmountDisplay()

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/components/amount_selection/BaseAmountBox.java
```diff
@@ -155,23 +169,36 @@ private View(Model model, Controller controller) {
...
             root.getStyleClass().add("base-amount-box");
 
             amountListener = this::onAmountChanged;
+            isBtcListener = this::onIsBtcChanged;
+            codeListener = this::onCodeChanged;
         }
 
         @Override
         protected void onViewAttached() {
+            boolean isBtc = model.isBtc.get();
```
HenrikJannsen: isBtc is used only once in the recent version. No need to extract it as local var.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/components/amount_selection/BaseAmountBox.java
```diff
@@ -130,9 +130,7 @@ void reset() {
     }
 
     private static class View extends bisq.desktop.common.view.View<HBox, Model, Controller> {
-        private final ChangeListener<Monetary> amountListener;
-        private final ChangeListener<Boolean> isBtcListener;
-        private final ChangeListener<String> codeListener;
+        private Subscription amountListener, isBtcListener, codeListener;
```
HenrikJannsen: We use `Pin` by convention for subscriptions.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/TradeDataHeader.java
```diff
@@ -225,29 +223,80 @@ private View(Model model, Controller controller) {
...
         }
 
-        private void configureBitcoinAmountDisplay(BitcoinAmountDisplay btcText) {
-            btcText.setFontSize(18);
-            btcText.setBtcCodeFontSize(15);
-            btcText.setBaselineAlignment();
-            btcText.setHeightConstraints(28, 28);
-        }
-
         public void refreshBitcoinAmountDisplayComponents() {
```
HenrikJannsen: Should be private and moved after onViewDetached.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/TradeDataHeader.java
```diff
@@ -178,9 +177,8 @@ private static class View extends bisq.desktop.common.view.View<HBox, Model, Con
         private final Label peerDescription;
         private final Triple<Triple<Text, Node, Text>, HBox, VBox> leftAmount, rightAmount;
         private Subscription userProfilePin, reputationScorePin;
-        private ChangeListener<Object> layoutBoundsListener;
-        private final BitcoinAmountDisplay leftBtcText, rightBtcText;
-        private final List<ChangeListener<String>> amountAndCodeListeners = new ArrayList<>();
+        private final BitcoinAmountDisplay leftBitcoinAmountDisplay, rightBitcoinAmountDisplay;
+        private final List<Subscription> amountAndCodeSubscriptions = new ArrayList<>();
```
HenrikJannsen: Better use HashSet and use `amountAndCodePins`

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/TradeDataHeader.java
```diff
@@ -225,29 +223,80 @@ private View(Model model, Controller controller) {
...
+
+            peersUserProfileDisplay.dispose();
+        }
+
+        private void configureBitcoinAmountDisplay(BitcoinAmountDisplay btcText) {
+            btcText.getIntegerPart().getStyleClass().add("bisq-easy-open-trades-header-value-btc-integer");
+            btcText.getLeadingZeros().getStyleClass().add("bisq-easy-open-trades-header-value-btc-leading-zeros");
+            btcText.getBtcCode().getStyleClass().add("bisq-easy-open-trades-header-code-btc");
+            btcText
+                    .getSignificantDigits()
```
HenrikJannsen: Why linebreak?

apps/desktop/desktop/src/test/java/bisq/desktop/components/controls/BitcoinAmountDisplayTest.java
```diff
@@ -0,0 +1,374 @@
...
+        assertEquals("0" + decimalSeparator, bitcoinAmountDisplay.getIntegerPart().getText());
+        assertEquals("00 000 00", bitcoinAmountDisplay.getLeadingZeros().getText());
+        assertEquals("1", bitcoinAmountDisplay.getSignificantDigits().getText());
+
+        assertTrue(bitcoinAmountDisplay.getLeadingZeros().getStyleClass().contains("btc-leading-zeros-dimmed"));
+    }
+
+    @Test
+    void testVerySmallBtcAmount() {
+        bitcoinAmountDisplay.setBtcAmount("0.000000124");
```
HenrikJannsen, line 108: Bitcoin has only max 8 digits.
1 sat = 0.00 000 001 BTC

apps/desktop/desktop/src/test/java/bisq/desktop/components/controls/BitcoinAmountDisplayTest.java
```diff
@@ -0,0 +1,374 @@
...
+        assertEquals("24", bitcoinAmountDisplay.getSignificantDigits().getText());
+
+        assertTrue(bitcoinAmountDisplay.getIntegerPart().getStyleClass().contains("btc-integer-part-dimmed"));
+        assertTrue(bitcoinAmountDisplay.getLeadingZeros().getStyleClass().contains("btc-leading-zeros-dimmed"));
+    }
+
+    @Test
+    void testBtcAmountWithOneLeadingZeroAndMultipleSignificantDigits() {
+        bitcoinAmountDisplay.setBtcAmount("0.120011");
+        //0.120 011
```
HenrikJannsen: Should be:
//0.12 001 100

HenrikJannsen (changes requested):

HenrikJannsen:
The component should always use the same pattern as in list views it would be less readable if mixed:

0-> 0.0
0.x -> 0.x0 000 000
0.0000000x -> 0.00 000 00x
x0> x.0

we could even argue to always show all trailing 0s. Full bitcoin values are anyway not realistically used in Bisq Easy.

nostrbuddha:
Hi,
Happy to this format getting into bisq 2 :-D
In mobile, we **always** use the full length display like `0.00 000 000`, The trailing zeroes are always shown.

Though `0.123 456` and `0.12 345 600` are the same, it will be confusing for user, if we show in different formats at different places through the app.

[Bitcoin design guide](https://bitcoin.design/guide/designing-products/units-and-symbols/) talks about `Satcomma standard`, which can be either 
```
 * 0.000 250 00 bitcoin
 * 0.00 025 000 bitcoin
```
There is no consensus on which one is better

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/TradeDataHeader.java
mvp1983: image
Can't find right now, but during my initial exploring of bisq-guides I found statement that the first vertical margin in IDEA is for comments, the second one - is for code. Thats why I used linebreak here.

Where our right margin for code should be and when we should use linebreak?

HenrikJannsen:
I think `0.00 025 000 bitcoin` is better as the first 2 with the left of separator forms also a group of 3.
The case that there are >9 BTC where that would not be the case is at least for us not realistic.

So lets stick with that and have always trailing zeros.
How do we display 0? `0.00 000 000 bitcoin` or `0.0`. I guess better to use the same format for all.

mvp1983:
> I think `0.00 025 000 bitcoin` is better as the first 2 with the left of separator forms also a group of 3. The case that there are >9 BTC where that would not be the case is at least for us not realistic.
> 
> So lets stick with that and have always trailing zeros. How do we display 0? `0.00 000 000 bitcoin` or `0.0`. I guess better to use the same format for all.

it should be 
0.12 300 000 bitcoin
or
0.12 300 000 BTC
?

mvp1983:
image
image
image
image
image
image
image

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/components/amount_selection/BaseAmountBox.java
```diff
@@ -130,7 +130,7 @@ void reset() {
     }
 
     private static class View extends bisq.desktop.common.view.View<HBox, Model, Controller> {
-        private Subscription amountListener, isBtcListener, codeListener;
+        private Subscription amountListenerPin, isBtcListenerPin, codeListenerPin;
```
HenrikJannsen, line 133: Should be `amountPin` - you can see that pattern used everywhere in our UI code.

HenrikJannsen (approved):
ACK

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/components/amount_selection/BaseAmountBox.java
HenrikJannsen: I will fix it myself in a upcoming PR

PR #3366, opened 2025-04-09
Fix Bitcoin Payment Confirmation Handling

mvp1983:
The issue can be in UserProfileService.getUserName() which can return different getUserName for the same inputs for different users  (UserProfileStore contains different information)

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced the reliability of transaction confirmations: the interface now more accurately displays Bitcoin receipt updates by accommodating variations in username formats.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerStateLightning3b.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerStateLightning3b.java
@@ -33,7 +33,6 @@ import bisq.desktop.main.content.bisq_easy.components.WaitingAnimation;
 import bisq.desktop.main.content.bisq_easy.components.WaitingState;
 import bisq.i18n.Res;
 import bisq.trade.bisq_easy.BisqEasyTrade;
-import bisq.user.profile.UserProfileService;
 import javafx.beans.property.SimpleStringProperty;
 import javafx.beans.property.StringProperty;
 import javafx.geometry.Insets;
@@ -103,12 +102,17 @@ public class SellerStateLightning3b extends BaseState {
                                 String expectedEncoded = Res.encode("bisqEasy.tradeState.info.buyer.phase3b.tradeLogMessage.ln", peersUserName);
                                 boolean shouldUpdateUI = encodedLogMessage.equals(expectedEncoded);
                                 if (!shouldUpdateUI) {
-                                    String clearedPeersUserName = UserProfileService.getClearedUserName(peersUserName);
-                                    if (!clearedPeersUserName.equals(peersUserName)) {
-                                        String simplifiedExpectedEncoded = Res.encode("bisqEasy.tradeState.info.buyer.phase3b.tradeLogMessage.ln",
-                                                clearedPeersUserName);
-                                        shouldUpdateUI = encodedLogMessage.equals(simplifiedExpectedEncoded);
-                                    }
+                                    /*
+                                     * userName formats may differ between sender and receiver. A userName appears as
+                                     * simple "nickname" if unique in user's storage, or as "nickname [nym-id]" if
+                                     * multiple users share that nickname. Due to persistence state differences, one
+                                     * peer might use the simple format while the other uses the extended format.
+                                     * We therefore compare the encoded message against both the userName and the
+                                     * nickname alone to ensure reliable message matching.
+                                     */
+                                    String simplifiedExpectedEncoded = Res.encode("bisqEasy.tradeState.info.buyer.phase3b.tradeLogMessage.ln",
+                                            bisqEasyOpenTradeChannel.getPeer().getNickName());
+                                    shouldUpdateUI = encodedLogMessage.equals(simplifiedExpectedEncoded);
                                 }
 
                                 if (shouldUpdateUI) {
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerStateLightning3b.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerStateLightning3b.java
@@ -99,23 +99,21 @@ public class SellerStateLightning3b extends BaseState {
                         public void add(BisqEasyOpenTradeMessage message) {
                             if (message.getChatMessageType() == ChatMessageType.PROTOCOL_LOG_MESSAGE && message.getText().isPresent()) {
                                 String encodedLogMessage = message.getText().get();
-                                String expectedEncoded = Res.encode("bisqEasy.tradeState.info.buyer.phase3b.tradeLogMessage.ln", peersUserName);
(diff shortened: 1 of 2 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerStateLightning3b.java
```diff
@@ -100,8 +101,19 @@ public void add(BisqEasyOpenTradeMessage message) {
                             if (message.getChatMessageType() == ChatMessageType.PROTOCOL_LOG_MESSAGE && message.getText().isPresent()) {
                                 String encodedLogMessage = message.getText().get();
                                 String expectedEncoded = Res.encode("bisqEasy.tradeState.info.buyer.phase3b.tradeLogMessage.ln", peersUserName);
-                                if (encodedLogMessage.equals(expectedEncoded)) {
-                                    UIThread.run(() -> model.getBuyerHasConfirmedBitcoinReceipt().set(Res.get("bisqEasy.tradeState.info.seller.phase3b.receiptConfirmed.ln")));
+                                boolean shouldUpdateUI = encodedLogMessage.equals(expectedEncoded);
+                                if (!shouldUpdateUI) {
+                                    String clearedPeersUserName = UserProfileService.getClearedUserName(peersUserName);
```
HenrikJannsen: No need to extract the nickname as it is a property in UserProfile:
`String peersNickName = bisqEasyOpenTradeChannel.getPeer().getNickName();`

We show the nickname (e.g. `Alice`) if there is only one such nickname in the network.
In case there are to multiple users with `Alice` as their nickname we append the nym (bot ID) to avoid impersonation attacks.

As you correctly observed, it can be that one trader has an additional user with the nickname it his network view and thus append the nym (e.g. `Alice [Fast-Scintillating-Storefront-502]` and the other user has that additional user not present. This is expected as the network state is dynamic and inactive user profiles get removed from the network but likely its because of a known issues how we persist the user profiles, thus users who are longer time online already have more inactive user profiles and by that more liklyhood to show the appended nym.

With that background (which is not much documented, thus its difficult for new devs to know the details about it), it would be better to rewrite the handling to make it more clear why we compare with both nickname and userName. Also add a comment for context.

user/src/main/java/bisq/user/profile/UserProfileService.java
```diff
@@ -156,6 +157,17 @@ public String getUserName(String nym, String nickName) {
         }
     }
 
+    public static String getClearedUserName(String userName) {
```
HenrikJannsen: Not needed. nickname provides what you extract here.

HenrikJannsen (changes requested):

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerStateLightning3b.java
mvp1983: Just for clarifying for myself:
In theory, both places of using of getName (for sender and receiver) can be replaces by using of getNickName() only. 
But we have production versions of the app and the next issues will arise (suppose, fixed version will be 2.1.8):

1) sender has version < 2.1.8 => current issue (not showing message) can happen ("Alice[[Fast-Scintillating-Storefront-502]] => Alice] );
2) receiver has version < 2.1.8 => also can happen ("Alice"=>["Alice[[Fast-Scintillating-Storefront-502]]).

So, that why we need to check getName and getNickName.

Is it correct?

HenrikJannsen: We have 3 properties we want to deal with:
```
String encodedLogMessage = message.getText().get();
String expectedEncodedWithUserName = Res.encode("bisqEasy.tradeState.info.buyer.phase3b.tradeLogMessage.ln", peersUserName);
String expectedEncodedWithNickName = Res.encode("bisqEasy.tradeState.info.buyer.phase3b.tradeLogMessage.ln", peersNickName);
```      

The `encodedLogMessage` can contain either the username or the nickname also when both users have the new version. 

The factor which matters here is if there is another same nickname on one of the peers network view. There is not guarantee that both have the same network view, thus we need to check for both.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerStateLightning3b.java
```diff
@@ -103,12 +102,17 @@ public void add(BisqEasyOpenTradeMessage message) {
                                 String expectedEncoded = Res.encode("bisqEasy.tradeState.info.buyer.phase3b.tradeLogMessage.ln", peersUserName);
                                 boolean shouldUpdateUI = encodedLogMessage.equals(expectedEncoded);
                                 if (!shouldUpdateUI) {
-                                    String clearedPeersUserName = UserProfileService.getClearedUserName(peersUserName);
-                                    if (!clearedPeersUserName.equals(peersUserName)) {
-                                        String simplifiedExpectedEncoded = Res.encode("bisqEasy.tradeState.info.buyer.phase3b.tradeLogMessage.ln",
-                                                clearedPeersUserName);
-                                        shouldUpdateUI = encodedLogMessage.equals(simplifiedExpectedEncoded);
-                                    }
+                                    /*
```
HenrikJannsen: The comment is not describing the context adequately. "multiple users share that nickname" - it is just that user can choose nicknames freely thus multiple could come up with the same nickname. "sharing" is misleading.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/SellerStateLightning3b.java
```diff
@@ -103,12 +102,17 @@ public void add(BisqEasyOpenTradeMessage message) {
...
+                                     * userName formats may differ between sender and receiver. A userName appears as
+                                     * simple "nickname" if unique in user's storage, or as "nickname [nym-id]" if
+                                     * multiple users share that nickname. Due to persistence state differences, one
+                                     * peer might use the simple format while the other uses the extended format.
+                                     * We therefore compare the encoded message against both the userName and the
+                                     * nickname alone to ensure reliable message matching.
+                                     */
+                                    String simplifiedExpectedEncoded = Res.encode("bisqEasy.tradeState.info.buyer.phase3b.tradeLogMessage.ln",
+                                            bisqEasyOpenTradeChannel.getPeer().getNickName());
+                                    shouldUpdateUI = encodedLogMessage.equals(simplifiedExpectedEncoded);
```
HenrikJannsen: I think a more clean way is to just have an `or` expression with checks the nickname version if the username version fails. The Res.encode call is super cheap thus code readability is here preferred.

HenrikJannsen (changes requested):

HenrikJannsen (approved):
utACK

PR #3368, opened 2025-04-09
Fix closing of QrCodeWindow

mvp1983:
QrCodeWindow closing is depended from model state. 
Before fix: if we leave SellerState3a (click on other item of the table, for instance), QrCodeWindow leaves as is and from this moment - can't be closed (via 'x' or 'close' button).

After: if we leave SellerState3a QrCodeWindow closes


HenrikJannsen (approved):
ACK

Good spot!

Issue #1459, opened 2023-12-03
For 'Learn more' external links show URL on mouse over


darawhelan:
Bisq 2 has lots of 'Learn more' external links that send the user to an external website (they open in user's default browser).

![links](https://github.com/bisq-network/bisq2/assets/121062323/9365f9cb-9b75-4b54-8e79-4740af39f4bb)

It would be better for usability if the URL destination of the links would be displayed on mouse over. Similar to how it is on some browsers. Right clicking enabled to copy the link would also be good if this was a possibly. 

Reasons:

- User might not want to open the link on their default browser
- User might be hesitant to click a link with no final destination showing

HenrikJannsen:
Good point!

54corbin:
@HenrikJannsen I would like to take this one, if we still need this feature. Could you assigne it to me? Thanks.

PR #3369, opened 2025-04-10
Feat: For 'Learn more' external links show URL on mouse over

54corbin:
# As discussed in the issue #1459, this pr is for showing external URLs on mouse hovering.
image

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/academy/AcademyBaseView.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/academy/AcademyBaseView.java
@@ -22,6 +22,7 @@ import bisq.desktop.common.utils.ImageUtil;
 import bisq.desktop.common.view.Controller;
 import bisq.desktop.common.view.Model;
 import bisq.desktop.common.view.View;
+import bisq.desktop.components.controls.BisqTooltip;
 import bisq.i18n.Res;
 import javafx.geometry.Insets;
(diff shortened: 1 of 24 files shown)


HenrikJannsen:
Thanks for your contribution!

As the pattern is repeating (hyperlink has tooltip which shows the url which gets opened when clicked) I think a `BisqHyperlink` component which has an `url` parameter and the click handler would reduce duplication and boilerplate code (clickhandlers in presenters). I gues in 90% of the cases that can be applied, otherwise the current approach can still be used.

HenrikJannsen:
Another small suggestion: Use branches for any feature work (using descriptive branch name not issue #), not main.

54corbin:
> Thanks for your contribution!
> 
> As the pattern is repeating (hyperlink has tooltip which shows the url which gets opened when clicked) I think a `BisqHyperlink` component which has an `url` parameter and the click handler would reduce duplication and boilerplate code (clickhandlers in presenters). I gues in 90% of the cases that can be applied, otherwise the current approach can still be used.

Thanks for your prompt review and I like your proposal(BisqHyperlink). Should I merge the current pr and try to work on 'BisqHyperlink' component?

54corbin:
> Another small suggestion: Use branches for any feature work (using descriptive branch name not issue #), not main.

Thanks for the suggestion. Will do in the future PRs.

HenrikJannsen:
> `Merge branch 'bisq-network:main' into main`

Better to use rebase instead of merge

PR #3373, opened 2025-04-11
remove first before add to ranking list

wuhualiang:
user profile updated will add it to reputation ranking list, remove it first

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved user profile management by automatically removing outdated profiles before adding a refreshed or new one, ensuring that only the most current information is maintained.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/reputation/ranking/ReputationRankingController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/reputation/ranking/ReputationRankingController.java
@@ -88,6 +88,7 @@ public class ReputationRankingController implements Controller {
                             ReputationRankingController.this,
                             model.getFilterMenuItemToggleGroup(),
                             userProfileService);
+                    remove(key);
                     model.getListItems().add(listItem);
                 });
             }
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/reputation/ranking/ReputationRankingController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/reputation/ranking/ReputationRankingController.java
@@ -88,7 +88,6 @@ public class ReputationRankingController implements Controller {
                             ReputationRankingController.this,
                             model.getFilterMenuItemToggleGroup(),
                             userProfileService);
-                    remove(key);
                     model.getListItems().add(listItem);
                 });
             }
--- a/user/src/main/java/bisq/user/profile/UserProfileService.java
+++ b/user/src/main/java/bisq/user/profile/UserProfileService.java
@@ -170,6 +170,7 @@ public class UserProfileService implements PersistenceClient<UserProfileStore>, 
                 ObservableHashMap<String, UserProfile> userProfileById = getUserProfileById();
                 synchronized (persistableStore) {
                     addNymToNickNameHashMap(userProfile.getNym(), userProfile.getNickName());
+                    existingUserProfile.ifPresent(e -> userProfileById.remove(e.getId()));
                     userProfileById.put(userProfile.getId(), userProfile);
                 }
                 numUserProfiles.set(userProfileById.size());


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/reputation/ranking/ReputationRankingController.java
```diff
@@ -88,6 +88,7 @@ public void put(String key, UserProfile userProfile) {
                             ReputationRankingController.this,
                             model.getFilterMenuItemToggleGroup(),
                             userProfileService);
+                    remove(key);
```
HenrikJannsen: This would add the iteration with the search at each add operation which is not great to for performance (even if we do not have that many entries).

The better approach is to see why we get called multiple times from the observable map.
In `UserProfileService.processUserProfileAddedOrRefreshed` we replace the user profile in case the profile or the applicationVersion has changed. 
If we use there a remove in that case and then the `userProfileById.put(userProfile.getId(), userProfile);` we would trigger the remove just for those rare cases and then get added the profile.

Can you test that if that solves the issue? You can reproduce it by changing the terms or statement of an existing profile. this should trigger a duplicated entry as it is now. With the suggested fix that should not happen anymore.

HenrikJannsen (changes requested):

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/reputation/ranking/ReputationRankingController.java
wuhualiang: Oh yes, duplication may only occur when refresh

HenrikJannsen: As far I see it only occurs when there are changed userProfiles. But not looked that deep into it. But at least that seems to be an expected case not handled yet. 

When user profiles are refreshed/republished it should be the same userProfile hashcode (`publishDate` is not included in equals and hashcode), thus not triggering an update. 

I suggest to reproduce the first expected case first, fix that and then look if the issue can still be reproduced/observed (usually after longer runtime).

wuhualiang: 1、delete local db file user_profile_store
2、starting app open ranking list  search 111 and then easy see  duplicate profile ，reopen ranking list duplicate will disappear（I found that the version that was retained was 2.1.2. The duplicate version 2.1.6. this seems unusual. I am observing it.）
3、use this branch to repeat the above steps and will not see any duplication.

HenrikJannsen: > search 111 and then easy see duplicate profile

If you refer to live data and the 2 profiles with nickname 111, that is valid as those are different profiles (see their diff. nym/botId).

HenrikJannsen (approved):
utACK

HenrikJannsen:
The code change looks good to me. Can you ensure to reproduce the current issues and check if its fixed with that change?
Furthermore if thats the case, can you make a long running test (several minutes) to see if you get any duplicates still with live data?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/reputation/ranking/ReputationRankingController.java
wuhualiang: > > search 111 and then easy see duplicate profile
> 
> If you refer to live data and the 2 profiles with nickname 111, that is valid as those are different profiles (see their diff. nym/botId).


I compared the userid and found no duplicates. I ran it for a while.
For this account, I found that the software version was 2.1.6, which should be the latest profile, but the local version was finally saved as 2.1.2. It doesn't seem reasonable. When saving and updating locally, should we compare the new and old versions instead of replacing them if they are not equal? ​​This way, we can get the latest version faster. But here, the main reason is that this user is not active, otherwise his data should be refreshed all the time.

wuhualiang:
> The code change looks good to me. Can you ensure to reproduce the current issues and check if its fixed with that change? Furthermore if thats the case, can you make a long running test (several minutes) to see if you get any duplicates still with live data?

Confirm that it has been fixed

PR #3377, opened 2025-04-12
Shutdown gracefully

mvp1983:
System.exit(...) terminates the JVM.
Platform.exit() signals the JavaFX Toolkit to shut down. So, some job(s) can be finished


application/build.gradle.kts
```diff
@@ -1,5 +1,11 @@
 plugins {
     id("bisq.java-library")
+    alias(libs.plugins.openjfx)
```
HenrikJannsen: the application module is used in headless appliactions (seed node,...) and we do not want to add a JavaFx dependency to those.

application/src/main/java/bisq/application/Executable.java
```diff
@@ -69,7 +69,11 @@ protected void notifyAboutShutdown() {
 
     protected void exitJvm() {
         log.info("Exiting JVM");
-        System.exit(PlatformUtils.EXIT_SUCCESS);
+        if (Platform.isFxApplicationThread()) {
```
HenrikJannsen: DesktopExecutable is a subclass and the only one useing JavaFx, there we override the exitJvm method and use the Platform.exit if we have already initiated the Platform.

HenrikJannsen (commented):
NACK

See comments

application/src/main/java/bisq/application/Executable.java
mvp1983: I see and understand now better.

Actually, System.exit() causes error sometimes and if we cannot remove it, maybe adding of some delay can solve this issue? (added)

HenrikJannsen: Which errors?

Issue #1813, opened 2024-03-16
[Minor UI issue] Provide system tray icon with transparend background


RequestPrivacy:
bisq2 v2.0.1
Ubuntu flavor (Kubuntu 22.04 LTS)

Currently the system tray icon has a white background which looks not well integrated into the desktop if e.g. used in a dark mode environment. A fix would be to use an icon which has a transparent one so that the underlying color can shine through.

![image](https://github.com/bisq-network/bisq2/assets/72392544/c7cf6c64-0357-45e7-a911-0a8691040663)

qxotk:
What is required?
Assuming I should get a dev environment set up, then find the icon in question, add a second version with appropriate name so UI devs can choose the version they want (transparent bg or not) ... then update the one used for system tray to be the transparent background one.

Does that sound right?

HenrikJannsen:
Yes, its mainly to figure out which icon is used there. As far I am aware we only have transparent images, so might be needed to dig deeper to find out why the image is maybe converted is bitmap or the like. Maybe we need another format instead of png like svg. Feel free to give it a try. The other devs are all busy with other tasks.

zabilal:
@HenrikJannsen Assign this to me

puchis0323:
Can I take?

PR #3378, opened 2025-04-13
Fix system tray icon bg

zabilal:
- Fixed background issue for the system tray icon
- Edited all other Application window icons to be transparent as they will have the same issue as the system tray

## Summary by CodeRabbit

- **Chores**
  - Updated the `.gitignore` file to exclude Visual Studio Code settings, ensuring a cleaner repository for collaboration.


HenrikJannsen (commented):
NACK

The https://github.com/bisq-network/bisq2/issues/1813 is about Linux. We want to change only the icon used for that particular case. The icon with the grey surrounding is the app icon. Also the changed icon has not good quality. We have somewhere that shape in high quality without artefacts.

Also squash commits to that dev accident commits are removed for reviewers.

HenrikJannsen:
I will close that as it does not address the issue correctly.

Issue #1785, opened 2024-03-13
Change display of My Profile in private chats


axpoems:
My profile should be shown next to the input text message bar (as in the rest of the chats) but without the option to switch profiles, just showing my profile for that chat.

![image](https://github.com/bisq-network/bisq2/assets/145597137/a5fab104-d77c-4d25-b865-41d3395eeaf5)

54corbin:
I can take this one.

54corbin:
@axpoems Is this what you mean?

![Image](https://github.com/user-attachments/assets/3107fd5a-24c2-4ed9-ab03-5ef5ea4dd325)

54corbin:
@HenrikJannsen Could you please assign this one to me ? Thank you.

54corbin:
#3379

PR #3379, opened 2025-04-13
Feat: Move myprofile down to the left of the input textArea

54corbin:
Issue id: #1785


54corbin:
> While testing locally I got a NPE:
> 
> java.lang.NullPointerException: Cannot invoke "bisq.desktop.main.content.components.UserProfileDisplay.setUserProfile(bisq.user.profile.UserProfile)" because "this.chatMyUserProfileDisplay" is null
> 
> Apart from this, a few more comments:
> 
> * The size of the input bar should not change.
> * "My profile" string is redundant.
> 
> The least expected for this task was to replicate what is already done for the public chats.


1,Addressed the NPE
2,Removed the "My profile" string 
3, The size of the input bar remains the same as it's in public chat
image

HenrikJannsen:
> I'm not quite understand what do you mean by "replicate what is already done for the public chats." as there's no nither PeerProfile on the top nor MyProfile next to the input bar, in the public chat.

I guess the confusion comes from the fact that the profile is only show in the public chat if there are multiple user profiles set up. 
Screenshot 2025-04-14 at 10 18 47

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/priv/PrivateChatsView.java
```diff
@@ -284,14 +280,39 @@ private void createHeaderVBox(boolean hasPeerToDisplay) {
     }
 
     private VBox createUserProfileVBox(UserProfileDisplay userProfileDisplay, String descriptionKey) {
```
axpoems: Do you think it's still necessary to use this function? it adds very little value if no description label is used.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/priv/PrivateChatsView.java
```diff
@@ -284,14 +280,39 @@ private void createHeaderVBox(boolean hasPeerToDisplay) {
...
         return vbox;
     }
 
+    private void updateAndShowMyProfile(UserProfile userProfile, PrivateChatsModel model) {
+        if (null == chatMyUserProfileDisplay) {
+            chatMyUserProfileDisplay = new UserProfileDisplay(25);
+        }
+        chatMyUserProfileDisplay.setUserProfile(userProfile);
+        chatMyUserProfileDisplay.setReputationScore(model.getMyUserReputationScore());
+        HBox sendMessageBox = (HBox) chatMessagesComponent.lookup("#chat-messages-bottom-bar-container > #chat-messages-bottom-bar");
```
axpoems: We try to avoid using `lookup` as much as possible, unless there's a major reason for it.
I don't think this case justifies it.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/priv/PrivateChatsView.java
54corbin: addresed. PTAL Thanks

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/priv/PrivateChatsView.java
54corbin: addresed. PTAL Thanks

axpoems: Your code update does the same as if you were using the `lookup`.
I followed up with a PR moving the implementation to `ChatMessageContainerView`, see #3396.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/priv/PrivateChatsView.java
```diff
@@ -292,6 +287,28 @@ private VBox createUserProfileVBox(UserProfileDisplay userProfileDisplay, String
...
 
+    private void updateAndShowMyProfile(UserProfile userProfile, PrivateChatsModel model) {
+        if (null == chatMyUserProfileDisplay) {
+            chatMyUserProfileDisplay = new UserProfileDisplay(25);
+        }
+        chatMyUserProfileDisplay.setReputationScore(model.getMyUserReputationScore());
+        chatMyUserProfileDisplay.setUserProfile(userProfile);
+        chatMyUserProfileDisplay.setReputationScore(model.getMyUserReputationScore());
+        chatMessagesComponent.getChildren().stream()
+                .filter(node -> "chat-messages-bottom-bar-container".equals(node.getId()))
```
HenrikJannsen, line 298: We do not access UI components by ID but use their reference instead. I have not looked into that particular case but I am pretty sure there is a more robust way to do it.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/priv/PrivateChatsView.java
```diff
@@ -292,6 +287,28 @@ private VBox createUserProfileVBox(UserProfileDisplay userProfileDisplay, String
         return vbox;
     }
 
+    private void updateAndShowMyProfile(UserProfile userProfile, PrivateChatsModel model) {
+        if (null == chatMyUserProfileDisplay) {
+            chatMyUserProfileDisplay = new UserProfileDisplay(25);
+        }
+        chatMyUserProfileDisplay.setReputationScore(model.getMyUserReputationScore());
+        chatMyUserProfileDisplay.setUserProfile(userProfile);
+        chatMyUserProfileDisplay.setReputationScore(model.getMyUserReputationScore());
```
HenrikJannsen, line 296: Duplicated line.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/priv/PrivateChatsView.java
54corbin: Really appriciate your review and comment though this PR has been closed.

PR #3383, opened 2025-04-14
Improve shutdown handling

HenrikJannsen:
We had a `Platform.isFxApplicationThread()` check, but usually we got called from the application service, thus a non UI thread and the `Platform.exit` case was never reached but instead the `exitJvm()` which was a non-graceful shutdown of the JavaFx runtime.

This should fix the error logs like below, which can be caused if there are pending `Platform.runLater` calls or new ones get fired after the rendering subsystem is already torn down (by threads, schedulers,...). With mapping our own `exitJavaFXPlatform` call in a `runLater`, it should give time to process the pending tasks (the runLater calls are executed in the order as they got called).
There might be still some edge cases left. We ignore the exception in the `uncaughtExceptionHandler` in case its caused from the `GraphicsPipeline`.

```
Feb-14 16:23:30.876 [ForkJoinPool.commonPool-worker-21] INFO b.a.Executable: Exiting JVM
Feb-14 16:23:31.075 [JavaFX Application Thread] ERROR b.d.DesktopExecutable: Uncaught exception: java.lang.NullPointerException: Cannot invoke "com.sun.prism.GraphicsPipeline.is3DSupported()" because the return value of "com.sun.prism.GraphicsPipeline.getPipeline()" is null
at [com.sun.javafx.tk](http://com.sun.javafx.tk/).quantum.QuantumToolkit.isSupported([QuantumToolkit.java:1230](http://quantumtoolkit.java:1230/))
at com.sun.javafx.application.PlatformImpl.isSupportedImpl([PlatformImpl.java:995](http://platformimpl.java:995/))
at com.sun.javafx.application.PlatformImpl.isSupported([PlatformImpl.java:655](http://platformimpl.java:655/))
at javafx.application.Platform.isSupported([Platform.java:271](http://platform.java:271/))
```

Credit to @mvp1983 who provided the logs and pushed for a fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced shutdown error reporting to ensure accurate messaging during failure scenarios.
  - Improved exception handling during application exit, leading to a smoother and more robust shutdown process.

### DIFF
--- a/apps/desktop/desktop-app/src/main/java/bisq/desktop_app/DesktopApplicationService.java
+++ b/apps/desktop/desktop-app/src/main/java/bisq/desktop_app/DesktopApplicationService.java
@@ -352,7 +352,7 @@ public class DesktopApplicationService extends JavaSeApplicationService {
                             log.info("ApplicationService shutdown completed");
                             return true;
                         } else {
-                            startupErrorMessage.set("Shutdown applicationService failed with result=false");
+                            shutDownErrorMessage.set("Shutdown applicationService failed with result=false");
                             log.error(shutDownErrorMessage.get());
                         }
                     } else {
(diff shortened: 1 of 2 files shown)


mvp1983:
ACK

PR #3391, opened 2025-04-17
feat: Allow selecting and copying text from chat message bubbles

54corbin:
## Description:
**Problem**: Users could not select text inside chat message bubbles, making it difficult to copy parts of a message.

**Solution**
Modified the chat message view components (e.g., replacing Labels with read-only TextAreas or similar) to enable standard text selection via mouse drag

**Impact**
Users can now easily select and copy text content (using Ctrl+C/Cmd+C) from individual chat messages.
image

**Performance Testing**
A performance test was conducted by rendering a large number of fake message bubbles (using the code on branch: [performance-test-allow-direct-text-selection-inside-chat-bubbles](https://github.com/54corbin/bisq2/commit/77241c7fc7d1e48eac97cb6ff76c29b2cdcc3480) to compare SelectableLabel against the standard Label. No noticeable difference in UI responsiveness or scroll performance was observed.

## Summary by CodeRabbit

- **New Features**
  - Enhanced chat message boxes to support selectable text, allowing users to copy message content.
  - Added the ability to control whether chat messages are editable or read-only.

- **Improvements**
  - Improved consistency between displayed and internal text values in chat message components.
  - Refined label dimension handling for better text display and interaction in chat messages.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/SelectableLabel.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/SelectableLabel.java
@@ -21,6 +21,7 @@ import bisq.desktop.common.threading.UIThread;
 import javafx.beans.value.ChangeListener;
 import javafx.collections.ListChangeListener;
 import javafx.collections.WeakListChangeListener;
+import javafx.geometry.Bounds;
 import javafx.scene.Cursor;
 import javafx.scene.control.ContextMenu;
 import javafx.scene.control.Label;
@@ -42,7 +43,8 @@ import java.util.List;
 public class SelectableLabel extends StackPane {
     private final Label label = new Label();
     private final TextArea textArea = new TextArea();
-    boolean isInSelectionMode;
+    boolean isSwapped;
+
     @SuppressWarnings("FieldCanBeLocal") // Need to keep a reference as used in WeakChangeListener
     private final ListChangeListener<String> styleClassListener = c -> {
         c.next();
@@ -56,14 +58,10 @@ public class SelectableLabel extends StackPane {
             textArea.getStyleClass().removeAll(removedList);
         }
     };
-    private final ChangeListener<Number> labelHeightListener = (observable, oldValue, newValue) -> {
-        if (label.getHeight() > 0) {
-            if (label.getWidth() > 0) {
-                swap();
-            } else {
-                // Not expected as height is set before width
-                UIThread.runOnNextRenderFrame(this::swap);
-            }
+
+    private final ChangeListener<Bounds> labelBoundsListener = (observable, oldBounds, newBounds) -> {
+        if (!isSwapped && newBounds.getWidth() > 0 && newBounds.getHeight() > 0) {
+            UIThread.runOnNextRenderFrame(this::swap);
         }
     };
 
@@ -78,16 +76,27 @@ public class SelectableLabel extends StackPane {
         textArea.setContextMenu(new ContextMenu());
         textArea.setText(text);
         textArea.getStyleClass().addAll("selectable-label", "hide-vertical-scrollbar");
-        label.heightProperty().addListener(labelHeightListener);
+        label.boundsInParentProperty().addListener(labelBoundsListener);
         getStyleClass().addListener(new WeakListChangeListener<>(styleClassListener));
     }
 
+    public void setEditable(boolean editable){
+        this.textArea.setEditable(editable);
+    }
+
+    public String getText() {
+        return this.textArea.getText();
+    }
+
     private void swap() {
-        label.heightProperty().removeListener(labelHeightListener);
-        textArea.setMinWidth(label.getWidth());
-        textArea.setMaxWidth(textArea.getMinWidth());
-        textArea.setMinHeight(label.getHeight());
-        textArea.setMaxHeight(textArea.getMinHeight());
+        if (isSwapped) {
+            return;
+        }
+
+        isSwapped = true;
+        label.boundsInParentProperty().removeListener(labelBoundsListener);
+        textArea.setPrefWidth(label.getWidth());
+        textArea.setPrefHeight(label.getHeight());
         getChildren().remove(label);
         getChildren().add(textArea);
     }
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/SelectableLabel.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/SelectableLabel.java
@@ -75,13 +75,10 @@ public class SelectableLabel extends StackPane {
         label.setCursor(Cursor.TEXT);
         textArea.setContextMenu(new ContextMenu());
         textArea.setText(text);
+        textArea.setEditable(false);
         textArea.getStyleClass().addAll("selectable-label", "hide-vertical-scrollbar");
         label.boundsInParentProperty().addListener(labelBoundsListener);
         getStyleClass().addListener(new WeakListChangeListener<>(styleClassListener));
-    }
-
-    public void setEditable(boolean editable){
-        this.textArea.setEditable(editable);
     }
 
     public String getText() {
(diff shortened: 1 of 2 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/SelectableLabel.java
```diff
@@ -43,6 +44,8 @@ public class SelectableLabel extends StackPane {
     private final Label label = new Label();
     private final TextArea textArea = new TextArea();
     boolean isInSelectionMode;
```
54corbin: Should we remove `isInSelectionMode` as there's no usage for it.

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/SelectableLabel.java
```diff
@@ -43,6 +44,8 @@ public class SelectableLabel extends StackPane {
...
+    @Getter
+    private String text = "";
```
54corbin: Use cases:
image

HenrikJannsen:
When adding a new message in the discussion chat I see that: 
Screenshot 2025-04-20 at 13 51 10

Only after click in the message list I gets displayed correctly.

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/SelectableLabel.java
HenrikJannsen: It is better to avoid a duplicated storage of the text property as it is already present in Label. Use a getter method which delegates via`label.getText()`.

HenrikJannsen (changes requested):

HenrikJannsen:
I looked into the issue and it seems the labelHeightListener is called too early in SelectableLabel.
With 
` UIThread.runOnNextRenderFrame(() -> label.heightProperty().addListener(labelHeightListener));` i could not get that UI glitch anymore.

54corbin:
> I looked into the issue and it seems the labelHeightListener is called too early in SelectableLabel. With ` UIThread.runOnNextRenderFrame(() -> label.heightProperty().addListener(labelHeightListener));` i could not get that UI glitch anymore.

Hey @HenrikJannsen ,
Thanks for digging into that UI glitch!
Your idea of delaying the listener with runOnNextRenderFrame definitely seemed right, but unfortunately, I was still seeing the glitch sometimes.
It looks like the core issue is a bit tricky – even with the delay, the heightProperty listener sometimes fires before the label's final width is actually calculated. So, when the swap() happened, it was occasionally still grabbing a width of 0, messing up the TextArea size.
Just wanted to share my findings! It seems like one of those subtle JavaFX layout timing things.

HenrikJannsen:
> > I looked into the issue and it seems the labelHeightListener is called too early in SelectableLabel. With ` UIThread.runOnNextRenderFrame(() -> label.heightProperty().addListener(labelHeightListener));` i could not get that UI glitch anymore.
> 
> Hey @HenrikJannsen , Thanks for digging into that UI glitch! Your idea of delaying the listener with runOnNextRenderFrame definitely seemed right, but unfortunately, I was still seeing the glitch sometimes. It looks like the core issue is a bit tricky – even with the delay, the heightProperty listener sometimes fires before the label's final width is actually calculated. So, when the swap() happened, it was occasionally still grabbing a width of 0, messing up the TextArea size. Just wanted to share my findings! It seems like one of those subtle JavaFX layout timing things.

Oh, damn. can you try out some addition strategies like checking for the width?

54corbin:
> > > I looked into the issue and it seems the labelHeightListener is called too early in SelectableLabel. With ` UIThread.runOnNextRenderFrame(() -> label.heightProperty().addListener(labelHeightListener));` i could not get that UI glitch anymore.
> > 
> > 
> > Hey @HenrikJannsen , Thanks for digging into that UI glitch! Your idea of delaying the listener with runOnNextRenderFrame definitely seemed right, but unfortunately, I was still seeing the glitch sometimes. It looks like the core issue is a bit tricky – even with the delay, the heightProperty listener sometimes fires before the label's final width is actually calculated. So, when the swap() happened, it was occasionally still grabbing a width of 0, messing up the TextArea size. Just wanted to share my findings! It seems like one of those subtle JavaFX layout timing things.
> 
> Oh, damn. can you try out some addition strategies like checking for the width or putting the addListener into a layout method?

Sorry forgot to let you know, that I hava solved this UI issue with the latest commit. Could you please take a look and test on your end when you got a moment? Thanks

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/SelectableLabel.java
```diff
@@ -78,16 +77,31 @@ public SelectableLabel(String text) {
...
+        return this.textArea.getText();
+    }
+
     private void swap() {
-        label.heightProperty().removeListener(labelHeightListener);
-        textArea.setMinWidth(label.getWidth());
-        textArea.setMaxWidth(textArea.getMinWidth());
-        textArea.setMinHeight(label.getHeight());
-        textArea.setMaxHeight(textArea.getMinHeight());
+        if (isInSelectionMode) {
```
HenrikJannsen: Maybe rename better to `isSwapped`

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/SelectableLabel.java
```diff
@@ -78,16 +77,31 @@ public SelectableLabel(String text) {
...
+            return;
+        }
+        // double-check dimensions here, just in case runLater was called prematurely
```
HenrikJannsen: "runLater was called prematurely" - i dont think that can be the case. Only reasons I can imagine is that bounds are not 100% aligend with label.width/height or that in the delay the w/h get reset to 0 (dont expect). I think both cases are close to 0 likely. If you cannot reproduce the issue with out that extra check I would suggest to remove it, as it seems not needed.

HenrikJannsen (changes requested):

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/SelectableLabel.java
54corbin: Addressed.
PTAL

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/SelectableLabel.java
54corbin: Addressed
PTAL

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/list/message_box/BubbleMessageBox.java
```diff
@@ -316,13 +317,13 @@ private void handleQuoteMessageBox() {
...
-    private Label createAndGetMessage() {
-        Label label = new Label();
+    private SelectableLabel createAndGetMessage() {
+        SelectableLabel label = new SelectableLabel(item.getMessage());
         label.maxWidthProperty().unbind();
         label.setWrapText(true);
         label.setPadding(new Insets(10));
         label.getStyleClass().addAll("text-fill-white", "medium-text", "font-default");
-        label.setText(item.getMessage());
+        label.setEditable(false);
```
HenrikJannsen: I think we should set that in the SelectableLabel by default to false.

HenrikJannsen (changes requested):

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/list/message_box/BubbleMessageBox.java
HenrikJannsen: Beside that small part all looks good

54corbin: Addressed
PTAL
Thanks

HenrikJannsen (approved):
utACK

PR #3392, opened 2025-04-18
Feat: Shift+Enter Newline Insertion at Caret in MyMessage Edit Text Area

54corbin:
**Problem:**

Pressing Shift+Enter while editing a chat message currently appends a newline at the end of the text, ignoring the cursor position. This results in unexpected behavior where the new line is added only at the end, and text after the cursor is not pushed down.

**Solution:**

Modify the edit text area's key event handling so that when Shift+Enter is pressed, a newline character is inserted exactly at the caret (cursor) position instead of appended to the end. This ensures the text after the cursor moves to the next line, matching common multiline editing expectations.
image

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of Shift+Enter in the message edit input so that newlines are inserted at the current cursor position, making multi-line message editing more intuitive.


54corbin:
@HenrikJannsen Could you please take a look when you got a moment. Thanks

HenrikJannsen (approved):
ACK

PR #3404, opened 2025-04-21
fixed cast to class exception

wuhualiang:
fixed take offer request handler cast to class exception 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling when verifying trade offers to ensure users are notified if a trade does not exist.

### DIFF
--- a/trade/src/main/java/bisq/trade/bisq_easy/protocol/messages/BisqEasyTakeOfferRequestHandler.java
+++ b/trade/src/main/java/bisq/trade/bisq_easy/protocol/messages/BisqEasyTakeOfferRequestHandler.java
@@ -33,6 +33,7 @@ import bisq.offer.bisq_easy.BisqEasyOffer;
 import bisq.offer.price.PriceUtil;
 import bisq.trade.ServiceProvider;
 import bisq.trade.Trade;
+import bisq.trade.TradeService;
 import bisq.trade.bisq_easy.BisqEasyTrade;
 import bisq.trade.bisq_easy.BisqEasyTradeService;
 import bisq.trade.protocol.events.TradeMessageHandler;
@@ -111,7 +112,7 @@ public class BisqEasyTakeOfferRequestHandler extends TradeMessageHandler<BisqEas
                 .filter(offer -> offer.equals(takersOffer))
                 .findAny();
         if (matchingOfferInChannel.isEmpty()) {
-            BisqEasyTradeService bisqEasyTradeService = (BisqEasyTradeService) serviceProvider;
+            BisqEasyTradeService bisqEasyTradeService = ((TradeService) serviceProvider).getBisqEasyTradeService();
             // After TRADE_ID_V1_ACTIVATION_DATE we might have trades in the open trades list which have an id created
             // with the createId_V0 method, thus we need to check for both.
             String v0_tradeId = Trade.createId(takersOffer.getId(), takersContract.getTaker().getNetworkId().getId());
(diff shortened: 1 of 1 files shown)


trade/src/main/java/bisq/trade/bisq_easy/protocol/messages/BisqEasyTakeOfferRequestHandler.java
```diff
@@ -111,7 +112,7 @@ protected void verifyMessage(BisqEasyTakeOfferRequest message) {
                 .filter(offer -> offer.equals(takersOffer))
                 .findAny();
         if (matchingOfferInChannel.isEmpty()) {
-            BisqEasyTradeService bisqEasyTradeService = (BisqEasyTradeService) serviceProvider;
+            BisqEasyTradeService bisqEasyTradeService = ((TradeService) serviceProvider).getBisqEasyTradeService();
```
HenrikJannsen, line 115: The naming of TradeService and BisqEasyTradeService suggests a inheritance relationship, though thats not the case. TradeService is the root level service in the trade module, following the convention used in all domain modules. BisqEasyTradeService is the service for the BE protocol.  I don't have a good suggestion how to mitigate that confusing name relationship.

Instead of the cast we should add the `BisqEasyTradeService getBisqEasyTradeService();` getter to the ServiceProvider, thus not requiring the confusing cast.

trade/src/main/java/bisq/trade/bisq_easy/protocol/messages/BisqEasyTakeOfferRequestHandler.java
```diff
@@ -119,8 +119,8 @@ protected void verifyMessage(BisqEasyTakeOfferRequest message) {
             String v1_tradeId = Trade.createId(takersOffer.getId(), takersContract.getTaker().getNetworkId().getId(), takersContract.getTakeOfferDate());
             boolean hasTradeWithSameTradeId = bisqEasyTradeService.getTrades().stream().anyMatch(trade ->
                     trade.getId().equals(v0_tradeId) || trade.getId().equals(v1_tradeId));
-            if (hasTradeWithSameTradeId) {
-                String errorMessage = String.format("A trade with the same tradeId already exist.\n" +
+            if (!hasTradeWithSameTradeId) {
```
HenrikJannsen, line 122: This is wrong. We want to check of we have already an trade with the same ID and if so we would throw an exception.

HenrikJannsen (changes requested):

trade/src/main/java/bisq/trade/bisq_easy/protocol/messages/BisqEasyTakeOfferRequestHandler.java
wuhualiang: The case I went through was that after deleting the offer, another user took the offer before the offer disappeared. This scenario can be reproduced, but when it gets to this point, it will report an exception that the trade exists. In fact, there is no duplication. From BisqEasyTradeService::handleBisqEasyTakeOfferMessage, it can be seen that it will first initialize the trade (and determine whether it is a duplicate) and then handle the message, so it must exist. Did I understand it wrong?

HenrikJannsen: I added a comment to make it more clear: https://github.com/bisq-network/bisq2/pull/3407

HenrikJannsen: The extra check for the trade IDs might be redundant but there could be some edge cases around activation date, where we might have a duplicated trade. Not 100% if that can be the case but the check will not hurt. We remove it anyway some time after activation date.

wuhualiang: OK, got it.

HenrikJannsen: Actually after another thought, I think the check is wrong as we create the trade before that handler is called, thus it would always be found. I removed it in https://github.com/bisq-network/bisq2/pull/3409

wuhualiang: Yes, the two versions trade id of the check should be in the trade  creation place

PR #3411, opened 2025-04-23
fix(fiat-currency): fallback to USD

fernandolguevara:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when retrieving currency by country code by adding fallback handling. If the currency cannot be determined, the system now defaults to USD instead of failing.

### DIFF
--- a/common/src/main/java/bisq/common/currency/FiatCurrencyRepository.java
+++ b/common/src/main/java/bisq/common/currency/FiatCurrencyRepository.java
@@ -84,8 +84,12 @@ public class FiatCurrencyRepository {
 
         // The language and variant components of the locale at Currency.getInstance are ignored.
         Locale countryLocale = new Locale(locale.getLanguage(), countryCode);
-        Currency currency = Currency.getInstance(countryLocale);
-        return new FiatCurrency(currency);
+        try {
+            Currency currency = Currency.getInstance(countryLocale);
+            return new FiatCurrency(currency);
+        } catch (Exception e) {
+             return new FiatCurrency("USD");
+        }
     }
 
     public static Map<String, FiatCurrency> getCurrencyByCodeMap() {


common/src/main/java/bisq/common/currency/FiatCurrencyRepository.java
```diff
@@ -84,8 +84,12 @@ public static FiatCurrency getCurrencyByCountryCode(String countryCode, Locale l
 
         // The language and variant components of the locale at Currency.getInstance are ignored.
         Locale countryLocale = new Locale(locale.getLanguage(), countryCode);
-        Currency currency = Currency.getInstance(countryLocale);
-        return new FiatCurrency(currency);
+        try {
+            Currency currency = Currency.getInstance(countryLocale);
+            return new FiatCurrency(currency);
+        } catch (Exception e) {
+             return new FiatCurrency("EUR");
```
HenrikJannsen: We use USD as default usually.

HenrikJannsen (changes requested):

HenrikJannsen:
Has there been some bug reports regarding that? If so which country caused the exception?

fernandolguevara:
@HenrikJannsen On my local env app crashed with country-locale code **es_419**

HenrikJannsen (approved):
utACK

HenrikJannsen:
> @HenrikJannsen On my local env app crashed with country-locale code **es_419**

Ah strange. Which locale is that?

HenrikJannsen:
Ah its Generic Latin America...

You need to have pgp signed your commit. If its too complicate to do that, no worry I can resubmit the PR as well...

HenrikJannsen (approved):
utACK

HenrikJannsen:
Replaced by https://github.com/bisq-network/bisq2/pull/3426

For some unknown reason GH does not recognize the code signature as valid. It shows as green though... Maybe cross check your code signing setup.

PR #3395, opened 2025-04-18
Show only consistent trade notifications

mvp1983:
Trade notification will be shown only if has both trade object and trade channel.

Fixed #3189, #3359

## Summary by CodeRabbit

- **New Features**
  - Trade notifications now update automatically in response to changes in trades or open trade channels.
- **Improvements**
  - Trade notifications are more accurately filtered to show only relevant, valid notifications.
  - Enhanced management of notification observers for better performance and reliability.
  - Trades are now displayed regardless of missing contract information, improving visibility in trade lists.

PR #3418, opened 2025-04-28
Proper resource cleanup in all *Service classes

### LINKED ISSUE #3414: Add proper resource cleanups in all shutdown methods of all service classes
We have not always handled the resource cleanups (e.g. remove listeners, bindings, stop schedulers, etc.) in service classes as they are usually singleton like which are only initialized once and shutdown only at app shutdown.

With the re-use for mobile and the different life cycle management requirements there (pause, resume) we might get called some service classes shutdown calls and initialize calls repeatedly without a complete app shutdown. There might be cases where it will require larger changes, if that is observed, please add a comment, but do not apply any more complex changes with that issues.

mvp1983:
Fixed #3414

42 Services were changed  in shutdown() method. Some of them have other trivial changes like introducing a variable to have ability to proper cleanup.

BisqEasyNotificationsService was omitted, because proper shutdown was done in #3395.

## Summary by CodeRabbit

- **New Features**
  - Added observable property for MuSig activation in settings, with developer mode enforcement.
  - Introduced new method for retrieving banned user moderator data.

- **Bug Fixes**
  - Improved shutdown procedures across multiple services to ensure proper cleanup of resources, observers, listeners, schedulers, and internal state.
  - Enhanced idempotency and state checks for initialization and shutdown to prevent redundant operations.
  - Added explicit nullability annotations to various fields for better resource management and reliability.
  - Added null checks before unbinding observers to prevent potential null pointer exceptions.

- **Refactor**
  - Standardized observer and listener management, ensuring all are unbound and cleared during shutdown.
  - Stored observer subscriptions and scheduled tasks to enable explicit unbinding and stopping during shutdown.
  - Updated method signatures and internal logic for improved lifecycle handling and code clarity.
  - Moved listener registrations from constructors to initialization methods for better lifecycle control.
  - Enhanced shutdown methods to clear references and reset flags to avoid stale state retention.


bonded-roles/src/main/java/bisq/bonded_roles/bonded_role/AuthorizedBondedRolesService.java
```diff
@@ -136,6 +136,9 @@ public CompletableFuture<Boolean> shutdown() {
         if (initialDataScheduler != null) {
             initialDataScheduler.stop();
         }
+        if (reprocessScheduler != null) {
+            reprocessScheduler.stop();
+        }
```
HenrikJannsen, line 151: Setting reprocessScheduler and initialDataScheduler to null is missing.

HenrikJannsen (changes requested):
Missing to handle the pins from observer registrations in `NotificationService`.
`bisqEasyOpenTradeChannelService.getChannels().addObserver ...`

I stopped review here as its very cumbersome and time consuming. Be sure to make that PR without any additional failures, otherwise I need to reconsider my time allocation for reviews. This is pure technical work which only require careful attention.

mvp1983:
What I did:

1) introduced *Pin vars for observers where it was missed;
2) after .unbind/interrupt/stop - set null to var (except if var is final);
3) Marked *Pin, Executor vars  as @Nullable;
4) cleared private and observable collections;
5) also removed listeners in some places;
6) if I saw something like "isStarted" - added early exit from initialize/shutdown;
7) cleared collections of Pin/Listeners after remove/unbind

What I didn't do:

1) didn't change final vars;
2) didn't touch other non *Service classes (except of a few of them - it was unintentionally, during searching of suspicious places, and I didn't pay attention, that some classes were not *Service classes, but decided to leave changes, because these changes );

HenrikJannsen (approved):
utACK

PR #3423, opened 2025-04-29
Fix market badge issue in the long run

### LINKED ISSUE #2781: Notification icons in markets list not updated
At a long running app I observed new notifications at offerbook but the icons at the markets have not been visible.

mvp1983:
Fixed #2781

Actually, bug could be observed if a user left OfferBookTab and returned again.

Issue was that createMarketChannels() called in constructor of BisqEasyOfferbookController (which called only once), but onActivate-onDeactivate were called many times (when the user leaves/enters OfferBook). In onDeactivate, for each marketChannel was called dispose(), which unbind Pin of chatMessages and updateNumOffers method stops to be called.

Actually, the solution looks simple, but I spent a lot of time, researching different places (even checked Badge component). When I was ready to give up, my eyes touched the problem ) 

## Summary by CodeRabbit

- **Refactor**
  - Improved the activation and deactivation process for the offer book view, ensuring better resource management.
  - Removed outdated and unused code for enhanced maintainability.


HenrikJannsen (approved):
ACK

Good finding!

PR #3433, opened 2025-05-03
Fix NPE in BisqEasy offerbook when updating maket price

axpoems:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability when updating market prices by adding safeguards against missing or unavailable market data.


HenrikJannsen (approved):
utACK

PR #3446, opened 2025-05-04
Use effectiveDuration instead of Transitions.useAnimations

mvp1983:
Simplified code, according to @HenrikJannsen suggestion in #3428: instead of checking Transtions.useAnimation and creating if-else statement - just use animation as is except calling Transition.getDuration() for duration.

### Before review

What I did:

0. Removed "else" part of  if (Transitions.useAnimation)
1. Removed external using of Transitions.useAnimation (except ProgresBarWithLabel and WalletGuideRecieveView - fade transition will be replaces by Carousel in other PR);
2. Added Math.min(1, duration) for Transitions.getDuration() to handle case where duration is 0;
3. Called Trasitions.getDuration(duration) even if it is clear that duration is 0, but I was trying to keep readibility.
4. Removed internal "getDuration" in some classes and used Transiton.getDuration()

What I didn't:

1. Didn't touch classes which use Transitions.* method directly (this class has in-build checks for duration and animations);
2. Didn't change Duration.millis(0) to Duration.millis(Transition.getDuration(0)). I'm not sure, if the variant with Transitions will be more readable.

## Summary by CodeRabbit

- **New Features**
  - Introduced a centralized animation duration management system, ensuring consistent and dynamic animation timings throughout the application.

- **Refactor**
  - Standardized all animation durations to use managed duration utilities, replacing hardcoded values and conditional animation checks.
  - Simplified animation logic by always applying animations, regardless of user settings.
  - Updated method signatures to use more precise duration types for improved consistency.

- **Style**
  - Improved code readability through formatting and reordering of import statements.

- **Chores**
  - Refined import statements across multiple modules for clarity and specificity.


apps/desktop/desktop/src/main/java/bisq/desktop/common/Transitions.java
```diff
@@ -373,7 +373,8 @@ private static void removeEffect(Node node, int duration) {
     }
 
     public static int effectiveDuration(int duration) {
-        return useAnimations() ? duration : 1;
+        // sometimes duration can be zero, Math.min allows returning 0 in such a case
+        return useAnimations() ? duration : Math.min(1, duration);
```
HenrikJannsen: Can be always 0. The 1 was not needed, 0 behaves correctly.

HenrikJannsen: As we mostly need `Duration` better use a method which provides that.

 
```
public class ManagedDuration {
    public static Duration millis(long value) {
        return Transitions.useAnimations() ? Duration.millis(value) : Duration.ZERO;
    }
}
```

Further review comments (not related to that code here):
Duration is usually a long type not int.
Lets stick close to the Duration API. 

We only need to convert if it is used in the method applying it, not in methods calling others.

Add for the static values getters with Managed Duration, there are some times DEFAULT_DURATION/2 of DEFAULT_DURATION/4 used, best to add constants for those.

```
  public static final int DEFAULT_DURATION = 600;
    public static final int CROSS_FADE_IN_DURATION = 1500;
    public static final int CROSS_FADE_OUT_DURATION = 1000;
    ```

HenrikJannsen (changes requested):

HenrikJannsen (approved):
utACK

PR #3478, opened 2025-05-14
Fix: Chat message edit field handles Enter/Shift+Enter key presses multiple times

54corbin:
Issue: pressing the Enter key (to save) or Shift+Enter (to create a new line) within the chat message edit input field could result in the respective action being triggered multiple times.

The root cause was that the KeyEvent.KEY_PRESSED event filter, containing logic for both saving and new line creation, was added to the editInputField each time the edit mode was initiated. However, the corresponding removal logic in onCloseEditMessage (which previously used setOnKeyPressed(null)) did not correctly remove these event filters. This led to an accumulation of active filters, each processing the key press.

## Summary by CodeRabbit

- **Refactor**
  - Improved the internal handling of the Enter key in the chat message edit input for better maintainability. No changes to user-facing behavior.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/list/message_box/MyTextMessageBox.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/list/message_box/MyTextMessageBox.java
@@ -26,6 +26,7 @@ import bisq.desktop.components.controls.BisqTextArea;
 import bisq.desktop.main.content.chat.message_container.list.ChatMessageListItem;
 import bisq.desktop.main.content.chat.message_container.list.ChatMessagesListController;
 import bisq.i18n.Res;
+import javafx.event.EventHandler;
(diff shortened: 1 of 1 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/list/message_box/MyTextMessageBox.java
```diff
@@ -44,6 +45,8 @@ public final class MyTextMessageBox extends BubbleMessageBox {
     private Button saveEditButton, cancelEditButton;
     private HBox editButtonsHBox;
 
+    private final javafx.event.EventHandler<KeyEvent> editFieldEnterKeyFilter;
```
HenrikJannsen, line 48: NIT: can remove package name

HenrikJannsen (approved):
ACK

HenrikJannsen:
Good job!

Issue #3124, opened 2025-01-22
Add trade history view


HenrikJannsen:
Bisq Easy did not provide a trade history view to avoid that users keep the trade history forever and leak privacy of their trade peer in case their computer gets compromised.
But the current solution can lead to 2 not desired behavior:
- Some users might keep the trade open forever to have their trade history present. This leads to bad UX as the trade list gets long.
- Users might use the export to csv feature and keep that file on their computer forever and by that countering our motivation to protect the peers privacy.

I think its better to add a trade history view where a trade after completion gets moved automatically.
After a certain time (maybe 4 weeks) the peers data in the trade will get deleted automatically to protect their privacy. 
For this we would need to tag the log messages with the trade state so that we can find the relevant log messages containing the account data easier. In case users have exchanged in regular chat messages private information we cannot do much. Deleting all chat messages should be considered as well.

See Bisq 1 issue to the topic: https://github.com/bisq-network/bisq/issues/5396 and https://github.com/bisq-network/bisq/pull/6001

54corbin:
I took

54corbin:
@HenrikJannsen Here’s the approach I'm planning to take:

1.  **Introduction of New Trade States:**
    *   I will modify our `BisqEasyTradeState` enumeration to include `HISTORY` (for trades that are finished, with their data still complete) and `PRUNED` (for older historical trades where peer information has been removed).
    *   Additionally, I will add a field named `originalTerminalState` to the `BisqEasyTrade` class. This field will record how a trade concluded (for instance, `BTC_CONFIRMED` or `FAILED`) before it is moved to the `HISTORY` state.

2.  **Adaptation of Trade Protocols:**
    *   I will adjust the existing trade protocols (the `BisqEasy...Protocol.java` files). When a trade reaches its natural end point, rather than just stopping or being removed, the protocol will now perform these actions:
        *   Set the `tradeCompletedDate`, which marks when the trade officially becomes part of the history.
        *   Record the `originalTerminalState`.
        *   Change the trade's main state to `HISTORY`.
    *   A significant change is that trades will no longer be removed from the `BisqEasyTradeService` storage when they are completed or have failed.

3.  **Implementation of Automated Pruning in `BisqEasyTradeService`:**
    *   I intend to improve the existing periodic task (which is currently `maybeRedactDataOfCompletedTrades`). This task will be updated to:
[...]

PR #3485, opened 2025-05-16
Introduce TradeLifecycleState for enhanced trade management/future "History Views"

54corbin:
**Description:**

Introduces a foundational `TradeLifecycleState` to `Trade` objects (including `BisqEasyTrade` and `MuSigTrade`) as a first step towards implementing the trade [history view](https://github.com/bisq-network/bisq2/issues/3124#issuecomment-2868677389).

**Next Steps (Planned, not in this PR):**

*   Modify `BisqEasyTradeService` and `BisqEasyTradeStore` to correctly handle the single `trades` set and implement methods for lifecycle state transitions.
*   Implement FSM Event Handlers to set `TradeLifecycleState.FAILED` on error.
*   Develop the UI components for "History" and "Failed Trades" tabs.
*   Implement the data pruning mechanism and transition to `REDACTED`.
*   Add UI elements for user-driven lifecycle changes.

I appreciate any feedback or suggestions on this initial set of changes!

## Summary by CodeRabbit

- **New Features**
  - Introduced a new trade lifecycle state to categorize trades as Active, Historical, or Failed.
  - Trades now display and update their lifecycle state in the user interface.

- **Bug Fixes**
  - Improved handling of error messages related to trade peers.

- **Chores**
  - Enhanced trade data serialization to support the new lifecycle state.

### DIFF
--- a/trade/src/main/java/bisq/trade/Trade.java
+++ b/trade/src/main/java/bisq/trade/Trade.java
@@ -100,6 +100,8 @@ public abstract class Trade<T extends Offer<?, ?>, C extends Contract<T>, P exte
     // Set at protocol creation and not updated later, thus no need to be observable
     private final Observable<String> protocolVersion = new Observable<>();
 
+    private final Observable<TradeLifecycleState> lifecycleState = new Observable<>();
+
     public Trade(C contract,
                  State state,
                  boolean isBuyer,
@@ -107,14 +109,16 @@ public abstract class Trade<T extends Offer<?, ?>, C extends Contract<T>, P exte
                  Identity myIdentity,
                  T offer,
                  P taker,
-                 P maker) {
+                 P maker,
+                 TradeLifecycleState lifecycleState) {
         this(contract,
                 state,
                 createId(offer.getId(), taker.getNetworkId().getId(), contract.getTakeOfferDate()),
                 createRole(isBuyer, isTaker),
                 myIdentity,
                 taker,
-                maker);
+                maker,
+                lifecycleState);
     }
 
     protected Trade(C contract,
@@ -123,7 +127,8 @@ public abstract class Trade<T extends Offer<?, ?>, C extends Contract<T>, P exte
                     TradeRole tradeRole,
                     Identity myIdentity,
                     P taker,
-                    P maker) {
+                    P maker,
+                    TradeLifecycleState lifecycleState) {
         super(state);
 
         this.contract = contract;
@@ -132,6 +137,7 @@ public abstract class Trade<T extends Offer<?, ?>, C extends Contract<T>, P exte
         this.myIdentity = myIdentity;
         this.taker = taker;
         this.maker = maker;
+        this.setLifecycleState(lifecycleState);
     }
 
     protected bisq.trade.protobuf.Trade.Builder getTradeBuilder(boolean serializeForHash) {
@@ -147,6 +153,7 @@ public abstract class Trade<T extends Offer<?, ?>, C extends Contract<T>, P exte
         Optional.ofNullable(getErrorStackTrace()).ifPresent(builder::setErrorStackTrace);
         Optional.ofNullable(getPeersErrorMessage()).ifPresent(builder::setPeersErrorMessage);
         Optional.ofNullable(getPeersErrorStackTrace()).ifPresent(builder::setPeersErrorStackTrace);
+        Optional.ofNullable(getLifecycleState()).ifPresent(s -> builder.setLifecycleState(s.toProtoEnum()));
         return builder;
     }
 
@@ -209,6 +216,17 @@ public abstract class Trade<T extends Offer<?, ?>, C extends Contract<T>, P exte
         return this.protocolVersion.get();
     }
 
+    public void setLifecycleState(TradeLifecycleState lifecycleState) {
+        this.lifecycleState.set(lifecycleState);
+    }
+
+    public TradeLifecycleState getLifecycleState() {
+        return this.lifecycleState.get();
+    }
+
+    public ReadOnlyObservable<TradeLifecycleState> lifecycleState() {
+        return lifecycleState;
+    }
 
     /* --------------------------------------------------------------------- */
     // Delegates
--- a/trade/src/main/java/bisq/trade/Trade.java
+++ b/trade/src/main/java/bisq/trade/Trade.java
@@ -148,12 +148,12 @@ public abstract class Trade<T extends Offer<?, ?>, C extends Contract<T>, P exte
                 .setMyIdentity(myIdentity.toProto(serializeForHash))
                 .setTaker(taker.toProto(serializeForHash))
                 .setMaker(maker.toProto(serializeForHash))
-                .setState(getState().name());
+                .setState(getState().name())
+                .setLifecycleState(getLifecycleState().toProtoEnum());
         Optional.ofNullable(getErrorMessage()).ifPresent(builder::setErrorMessage);
         Optional.ofNullable(getErrorStackTrace()).ifPresent(builder::setErrorStackTrace);
         Optional.ofNullable(getPeersErrorMessage()).ifPresent(builder::setPeersErrorMessage);
         Optional.ofNullable(getPeersErrorStackTrace()).ifPresent(builder::setPeersErrorStackTrace);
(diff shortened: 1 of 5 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/BisqEasyOpenTradesController.java
```diff
@@ -359,7 +361,12 @@ private void handleClearTradesAndChannels() {
 
     // Misc
     private void updatePredicate() {
-        model.getFilteredList().setPredicate(e -> bisqEasyTradeService.findTrade(e.getTradeId()).isPresent());
+        model.getFilteredList().setPredicate(openTradeListItem -> {
+            Optional<BisqEasyTrade> tradeOpt = bisqEasyTradeService.findTrade(openTradeListItem.getTradeId());
+
+            boolean isActive = tradeOpt.isPresent()&&tradeOpt.get().getLifecycleState().equals(TradeLifecycleState.ACTIVE);
```
HenrikJannsen: Please apply formatting at commit in Intellij (space around &&). No need for declaring variable`boolean isActive` (I guess was left over from debugging...).

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/BisqEasyOpenTradesController.java
```diff
@@ -359,7 +361,12 @@ private void handleClearTradesAndChannels() {
 
     // Misc
     private void updatePredicate() {
-        model.getFilteredList().setPredicate(e -> bisqEasyTradeService.findTrade(e.getTradeId()).isPresent());
+        model.getFilteredList().setPredicate(openTradeListItem -> {
+            Optional<BisqEasyTrade> tradeOpt = bisqEasyTradeService.findTrade(openTradeListItem.getTradeId());
```
HenrikJannsen: You can use `openTradeListItem.getTrade()`. The item can only be created if trade was present.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/BisqEasyOpenTradesController.java
```diff
@@ -359,7 +361,12 @@ private void handleClearTradesAndChannels() {
 
     // Misc
     private void updatePredicate() {
-        model.getFilteredList().setPredicate(e -> bisqEasyTradeService.findTrade(e.getTradeId()).isPresent());
+        model.getFilteredList().setPredicate(openTradeListItem -> {
```
HenrikJannsen: Best to set the specific predicate in the model and combine various predicates in the controller.
With that patterns its more clear and less error prone. We could even use an Observable for the predicates and listen on any change and then trigger the combined predicate setting. 
Also such predicate can be used at initial filtering when creating the list item.

trade/src/main/java/bisq/trade/Trade.java
```diff
@@ -209,6 +217,13 @@ public String getProtocolVersion() {
         return this.protocolVersion.get();
     }
 
+    public void setLifecycleState(TradeLifecycleState lifecycleState){
+        this.lifecycleState.set(lifecycleState);
+    }
+
+    public TradeLifecycleState getLifecycleState(){
+        return this.lifecycleState.get();
+    }
```
HenrikJannsen, line 225: Add 

```
  public ReadOnlyObservable<TradeLifecycleState> lifecycleState() {
        return lifecycleState;
    }
    ```

File: trade/src/main/java/bisq/trade/TradeLifecycleState.java [hunk 5]
```diff
@@ -0,0 +1,24 @@
+package bisq.trade;
+
+import bisq.common.proto.ProtoEnum;
+import bisq.common.proto.ProtobufUtils;
+
+public enum TradeLifecycleState implements ProtoEnum {
+    ACTIVE,     // In progress, or completed but user hasn't moved it / not auto-moved yet
+    HISTORICAL, // Finished and moved to history view
+    FAILED,     // Concluded with an FSM error
+    REDACTED    // Historical or Failed, and sensitive peer data pruned
```
HenrikJannsen: Not only peer data also one's own data (if that would not be the case, its missing in the impl.). Maybe better stick with the redacted term, as that describes it more exactly.

trade/src/main/java/bisq/trade/TradeLifecycleState.java
```diff
@@ -0,0 +1,24 @@
...
+    ;
+
+    @Override
+    public bisq.trade.protobuf.TradeLifecycleState toProtoEnum() {
+        return bisq.trade.protobuf.TradeLifecycleState.valueOf(getProtobufEnumPrefix() + name());
+
```
HenrikJannsen: Remove empty lines

trade/src/main/proto/trade.proto
```diff
@@ -42,6 +42,14 @@ enum TradeRole{
   TRADEROLE_SELLER_AS_TAKER = 3;
   TRADEROLE_SELLER_AS_MAKER = 4;
 }
+
+enum TradeLifecycleState{
+  TRADELIFECYCLESTATE_ACTIVE = 0;
```
HenrikJannsen: We always add an UNSPECIFIED entry as first to handle certain failure cases. The proto enum utils have option to deal with that case (define fallback).

trade/src/main/proto/trade.proto
```diff
@@ -54,6 +62,7 @@ message Trade {
   optional string errorStackTrace = 9;
   optional string peersErrorMessage = 10;
   optional string peersErrorStackTrace = 11;
+  optional TradeLifecycleState lifecycleState = 12;
```
HenrikJannsen: I assume you use optional for backward compatibility, but I guess thats not needed, as we can add fields without breaking things. The value will be the first entry in the enum (the one with 0 index), we can map that in the ProtoEnumUtil to the TRADELIFECYCLESTATE_ACTIVE state.

HenrikJannsen: Removing optional must be reflected in the class as well. We try to keep the java class and the proto as closely mirrored as possible.

trade/src/main/proto/trade.proto
```diff
@@ -42,6 +42,14 @@ enum TradeRole{
   TRADEROLE_SELLER_AS_TAKER = 3;
   TRADEROLE_SELLER_AS_MAKER = 4;
 }
+
+enum TradeLifecycleState{
+  TRADELIFECYCLESTATE_ACTIVE = 0;
+  TRADELIFECYCLESTATE_HISTORICAL = 1;
+  TRADELIFECYCLESTATE_FAILED = 2;
+  TRADELIFECYCLESTATE_REDACTED = 3;
```
HenrikJannsen: Not sure the redacted should be a state, as it could be applied to any other state. It just applied after a certain time independent of its state. Better model that as an abstract method in Trade which evaluates the state of the redacted fields to provide the return value. We can cache it in a transient field if evaluation is costly.

HenrikJannsen (changes requested):

trade/src/main/java/bisq/trade/Trade.java
```diff
@@ -147,6 +153,7 @@ protected bisq.trade.protobuf.Trade.Builder getTradeBuilder(boolean serializeFor
         Optional.ofNullable(getErrorStackTrace()).ifPresent(builder::setErrorStackTrace);
         Optional.ofNullable(getPeersErrorMessage()).ifPresent(builder::setPeersErrorMessage);
         Optional.ofNullable(getPeersErrorStackTrace()).ifPresent(builder::setPeersErrorStackTrace);
+        Optional.ofNullable(getLifecycleState()).ifPresent(s -> builder.setLifecycleState(s.toProtoEnum()));
```
HenrikJannsen: It cannot be null as we set a default value of its not present in the proto. We use the `Optional.ofNullable` pattern only for optional fields.

HenrikJannsen (changes requested):
Only one small NIT

HenrikJannsen (approved):
utACK

Issue #3441, opened 2025-05-04
Add missing AccountPayload, Accounts and FiatPaymentRails


HenrikJannsen:
In the `bisq.account.accounts` package we have many payment methods from bisq 1 already ported but there are still   many missing.

Here is a list which might contain all (but need to cross check with Bisq 1 if there are more missing). Any which are marked as deprecated or not used, dont need to be ported.

```
SAME_BANK=Transfer with same bank
SPECIFIC_BANKS=Transfers with specific banks
MONEY_GRAM=MoneyGram
WESTERN_UNION=Western Union
JAPAN_BANK=Japan Bank Furikomi
UPHOLD=Uphold
MONEY_BEAM=MoneyBeam (N26)
POPMONEY=Popmoney
PERFECT_MONEY=Perfect Money
ALI_PAY=AliPay
WECHAT_PAY=WeChat Pay
SEPA_INSTANT=SEPA Instant Payments
SWISH=Swish
CHASE_QUICK_PAY=Chase QuickPay
HAL_CASH=HalCash
PROMPT_PAY=PromptPay
ADVANCED_CASH=Advanced Cash
WISE_USD=Wise-USD
PAYSERA=Paysera
PAXUM=Paxum
NEFT=India/NEFT
RTGS=India/RTGS
IMPS=India/IMPS
PAYTM=India/PayTM
NEQUI=Nequi
CAPITUAL=Capitual
CELPAY=CelPay
MONESE=Monese
SATISPAY=Satispay
TIKKIE=Tikkie
VERSE=Verse
SWIFT=SWIFT International Wire Transfer
DOMESTIC_WIRE_TRANSFER=Domestic Wire Transfer
CIPS=Cross-Border Interbank Payment System
```

HenrikJannsen:
PS: If you take that task be sure to have it free of errors, it is nasty to review and I don't want to review repeatedly...

namloan:
Linking here a previous conversation about the missing payment methods https://github.com/bisq-network/bisq2/pull/1124#issuecomment-1826439592

We may want to merge Wise-USD and into the already existing Wise. Also some of the Bisq1 payment methods may be deprecated or not used at all.

@pazza83 could you take a look and maybe give us a prioritized list of missing payment methods?

mvp1983:
looks like I took this task

PR #3489, opened 2025-05-16
Add SepaInstantAccount and SepaInstantAccountPayload messages into proto

mvp1983:
Corresponding java classes were created too and added unit-tests.

Part of #3441

## Summary by CodeRabbit

- **New Features**
  - Added support for SEPA Instant accounts, enabling users to manage SEPA Instant payment accounts with holder name, IBAN, BIC, and accepted country codes.
  - Enhanced protobuf schema to include SEPA Instant account types for improved data handling.

- **Bug Fixes**
  - Improved deserialization logic to correctly handle SEPA Instant account types.

- **Tests**
  - Added comprehensive tests covering SEPA Instant account creation, validation, serialization, and deserialization to ensure accuracy and robustness.

### DIFF
--- a/account/src/main/java/bisq/account/accounts/SepaInstantAccount.java
+++ b/account/src/main/java/bisq/account/accounts/SepaInstantAccount.java
@@ -1,0 +1,62 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.account.accounts;
+
+import bisq.account.payment_method.FiatPaymentMethod;
+import bisq.account.payment_method.FiatPaymentRail;
+import bisq.common.locale.Country;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
+
+@Getter
+@Slf4j
+@ToString
+@EqualsAndHashCode(callSuper = true)
+public final class SepaInstantAccount extends CountryBasedAccount<SepaInstantAccountPayload, FiatPaymentMethod> {
+
+    private static final FiatPaymentMethod PAYMENT_METHOD = FiatPaymentMethod.fromPaymentRail(FiatPaymentRail.SEPA_INSTANT);
+
+    public SepaInstantAccount(String accountName,
+                              SepaInstantAccountPayload sepaInstantAccountPayload,
+                              Country country) {
+        super(accountName, PAYMENT_METHOD, sepaInstantAccountPayload, country);
+    }
+
+    public static SepaInstantAccount fromProto(bisq.account.protobuf.Account proto) {
+        return new SepaInstantAccount(
+                proto.getAccountName(),
+                SepaInstantAccountPayload.fromProto(proto.getAccountPayload()),
+                Country.fromProto(proto.getCountryBasedAccount().getCountry()));
+    }
+
+    @Override
+    protected bisq.account.protobuf.CountryBasedAccount.Builder getCountryBasedAccountBuilder(boolean serializeForHash) {
+        return super.getCountryBasedAccountBuilder(serializeForHash)
+                .setSepaInstantAccount(toSepaInstantAccountProto(serializeForHash));
+    }
+
+    private bisq.account.protobuf.SepaInstantAccount toSepaInstantAccountProto(boolean serializeForHash) {
+        return resolveBuilder(getSepaInstantAccountBuilder(serializeForHash), serializeForHash).build();
+    }
+
+    private bisq.account.protobuf.SepaInstantAccount.Builder getSepaInstantAccountBuilder(boolean serializeForHash) {
+        return bisq.account.protobuf.SepaInstantAccount.newBuilder();
+    }
+}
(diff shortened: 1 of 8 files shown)


account/src/main/java/bisq/account/accounts/SepaInstantAccount.java
```diff
@@ -0,0 +1,62 @@
...
+                              Country country) {
+        super(accountName, PAYMENT_METHOD, sepaInstantAccountPayload, country);
+    }
+
+    public static SepaInstantAccount fromProto(bisq.account.protobuf.Account proto) {
+        return new SepaInstantAccount(
+                proto.getAccountName(),
+                SepaInstantAccountPayload.fromProto(proto.getAccountPayload()),
+                Country.fromProto(proto.getCountryBasedAccount().getCountry()));
+    }
```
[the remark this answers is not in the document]
mvp1983: other similar classes don't have these checks

HenrikJannsen:
@namloan Would you be available for review?

account/src/main/java/bisq/account/accounts/SepaInstantAccountPayload.java
```diff
@@ -0,0 +1,98 @@
...
+    private final String iban;
+    private final String bic;
+    private final List<String> acceptedCountryCodes;
+
+    public SepaInstantAccountPayload(String id,
+                                     String paymentMethodName,
+                                     String holderName,
+                                     String iban,
+                                     String bic,
+                                     String countryCode,
```
namloan, line 45: I think we wouldn't expect these fields to be null or empty at any point of time. Can we check for nullability?

account/src/test/java/bisq/account/accounts/SepaInstantAccountPayloadTest.java
```diff
@@ -0,0 +1,246 @@
...
+                                .setSepaInstantAccountPayload(SepaInstantAccountPayload.newBuilder()
+                                        .setHolderName("holderName")
+                                        .setIban("DE89370400440532013000")
+                                        .setBic("DEUTDEBBXXX")))
+                .build();
+
+        assertEquals(protoWithEmptyList, payloadWithEmptyList.completeProto());
+    }
+
+    @Test
```
namloan, line 103: You can remove this test  - it's redundant as you're already building valid payloads in other tests

common/src/main/java/bisq/common/validation/NetworkDataValidation.java
```diff
@@ -147,4 +148,19 @@ public static void validateBondUserName(String bondUserName) {
...
+                "Invalid IBAN format. Must start with country code (2 letters) followed by check digits (2 numbers) and BBAN. iban=" + iban);
+    }
+
+    // Format: institution code (4 letters) + country code (2 letters) + location code (2 alphanumeric) +
+    // optional branch code (3 alphanumeric).
+    public static void validateBicFormat(String bic) {
+        checkArgument(!StringUtils.isEmpty(bic), "BIC must not be null or empty");
+        checkArgument(bic.matches("[A-Z]{4}[A-Z]{2}[A-Z0-9]{2}([A-Z0-9]{3})?"),
+                "Invalid BIC/SWIFT format. Must follow pattern of institution code (4 letters) + country code (2 letters) + location code (2 alphanumeric) + optional branch code (3 alphanumeric). bic=" + bic);
+    }
```
namloan, line 175: I think these validations would be better placed as private methods in the payload itself. We can later refactor to use them in other payloads that need them (like the standard SEPA). Or if we later see that we need them elsewhere we can extract to a common thing, but for sure I think I'd avoid to keep extending this```NetworkDataValidation``` to avoid making it a generic bag of validations for everything.

account/src/main/java/bisq/account/accounts/SepaInstantAccount.java
```diff
@@ -0,0 +1,62 @@
...
+
+@Getter
+@Slf4j
+@ToString
+@EqualsAndHashCode(callSuper = true)
+public final class SepaInstantAccount extends CountryBasedAccount<SepaInstantAccountPayload, FiatPaymentMethod> {
+
+    private static final FiatPaymentMethod PAYMENT_METHOD = FiatPaymentMethod.fromPaymentRail(FiatPaymentRail.SEPA_INSTANT);
+
+    public SepaInstantAccount(String accountName,
```
namloan, line 36: check for nulls?

namloan (commented):
Good job! I like it. Some comments

account/src/main/java/bisq/account/accounts/SepaInstantAccount.java
mvp1983: I don't see null checking in other existed Accounts, so, didn't add here too. My thought was that if checks are absent everywhere, looks like this is not needed  here too.

account/src/main/java/bisq/account/accounts/SepaInstantAccountPayload.java
mvp1983: id, paymentName, countryCode are validated in parent(s) class(es)
iban and bic are validated and check for null is present.

however, old validation like validateText will fire NullPointerException if argument will be null. I will improve this.

common/src/main/java/bisq/common/validation/NetworkDataValidation.java
mvp1983: From Bisq1 paymentAccounts list, only Sepa and SepaInstant use bic, but IBAN also used by Tikkie. 

I think should be some common place for this validations, because make it private isn't working solution during other payment method adding

account/src/main/java/bisq/account/accounts/SepaInstantAccount.java
HenrikJannsen: In general we do not excessive null checks, only at user input (UI). If absent data is a valid case then we use Optional.

common/src/main/java/bisq/common/validation/NetworkDataValidation.java
namloan: ok fine for now, not a blocker

account/src/main/java/bisq/account/accounts/SepaInstantAccount.java
namloan: me too when I added the existing payment methods I think I was too influenced by Bisq1 and by not being sure which fields are required or not in the less known ones. Looking at this now, I should have added the checks; no reason for these objects to be loaded in memory at any point of time with these fields null / empty.

But yes fair point that the others don't have the checks, so I'm OK with leaving these as they are.

account/src/test/java/bisq/account/accounts/SepaInstantAccountPayloadTest.java
mvp1983: testVerifyWithValidFields was removed

namloan (approved):
utACK

PR #3506, opened 2025-05-20
Calling of verify() was absent in constructor of AchTransferAccountPayload

mvp1983:
Add call of verify() into constructor of AchTransferAccountPayload

Fixed AchTransferAccountPayloadTest/AchTransferAccountTest which become broken. Before this fix, verify() never wasn't called and tests were passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for ACH transfer account creation to ensure address information is verified during setup.

- **Tests**
  - Updated test data to use a specific country code ("US") instead of a placeholder for ACH transfer account tests.

### DIFF
--- a/account/src/main/java/bisq/account/accounts/AchTransferAccountPayload.java
+++ b/account/src/main/java/bisq/account/accounts/AchTransferAccountPayload.java
@@ -1,5 +1,6 @@
 package bisq.account.accounts;
 
+import bisq.common.validation.NetworkDataValidation;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.ToString;
@@ -7,7 +8,8 @@ import lombok.extern.slf4j.Slf4j;
 
 import java.util.Optional;
 
-import static bisq.common.util.OptionalUtils.*;
+import static bisq.common.util.OptionalUtils.normalize;
+import static bisq.common.util.OptionalUtils.toOptional;
 
 @Slf4j
 @ToString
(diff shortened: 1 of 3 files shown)


account/src/main/java/bisq/account/accounts/AchTransferAccountPayload.java
```diff
@@ -39,6 +41,14 @@ public AchTransferAccountPayload(String id,
                 null,
                 null);
         this.holderAddress = normalize(holderAddress);
+
+        verify();
```
HenrikJannsen, line 45: verify() should be called from the base class, thus avoiding the need to call it at each implementation and risk forgetting it.

HenrikJannsen: We can keep that here, and apply the refactoring in a follow up PR.

HenrikJannsen (approved):
utACK

account/src/main/java/bisq/account/accounts/AchTransferAccountPayload.java
mvp1983: I thought about calling verify() in parent constructor. But it works only if class doesn't override verify() method to check additional fields, which parent class doesn't check.

When class override verify():
in constructor first things should be super(...), which calls parent constructor, which calls verify() in child class (not parent). Child can call super.verify() to call parent verify(), however, additional data (class fields) are still not presented, because calling of parent constructor should be first line of code in child constructor, so child verify() will failed with NPE or IllegalStateException, because all new child fields will be null. 

So, if we have to override verify(), we should call it only in child, and don't call in parent at all (otherwise - exception).

If we don't need to override verify(), it can be called in parent constructor

mvp1983: Also, at least 14 *PaymentPayload.java don't call verify. And somewhere in future it can be fixed (task for newcomers) + some tests can be broken after this and maybe some tests should be added

HenrikJannsen: Ah yes, you are right. Lets stick then to the super.verify() call for any new classes. The existing ones missing can be tackled with a new issue.

Issue #3410, opened 2025-04-22
Make the notification icon show the current notification setting (e.g., Off, Mentions)


54corbin:
Right now, the notification bell icon always looks the same, no matter if you've set notifications to 'Off', 'If mentioned', or 'All messages' for a specific chat/channel/etc. You have to click it to remember or check the current setting.
It would be really helpful if the icon itself changed visually to reflect the active notification status.

![Image](https://github.com/user-attachments/assets/31aa0f9b-ece4-4389-bccd-0f23acda4495)

54corbin:
Is this relevant?
Should I work on this?

HenrikJannsen:
Yes, I think that would be a nice fix.

PR #3512, opened 2025-05-21
Make the notification icon show the current notification setting

54corbin:
issue: #3410

## Summary by CodeRabbit

- **New Features**
  - Notification settings menu icons in chat now update dynamically to reflect the selected notification preference, including support for global default settings.
  - New notification icons have been added for different notification types (all, mention, off), providing clearer visual feedback.
  - Dropdown menus can now update their icons dynamically at runtime, improving visual responsiveness to user interactions.

- **Style**
  - Introduced new notification icon images and corresponding styles for improved visual distinction in the user interface.


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/BaseChatView.java
```diff
@@ -215,4 +248,14 @@ void updateSelection(boolean isSelected) {
             getContent().pseudoClassStateChanged(SELECTED_PSEUDO_CLASS, isSelected);
         }
     }
+
+    private static class IconIdPair {
```
HenrikJannsen: As IconIdPair is only used inside the class we could use the Pair<> class as well. But no need to change...

HenrikJannsen (changes requested):
The icons need to have high resolution (@2x) version as otherwise they are blurry on retina / HD screens. Also we should stick to the outlined version as filled gives them too much weight. But if you prefer we can ask @axpoems to update the icons.

HenrikJannsen (approved):
utACK

PR #3516, opened 2025-05-22
Handle redacted PartialSignatures

HenrikJannsen:
## Summary by CodeRabbit

- **New Features**
  - Introduced support for redacted partial signatures in MuSig trade messages, enhancing privacy by sharing only necessary signature components between peers.
  - Added handling and exchange of swap transaction input partial signatures in relevant trade message flows.

- **Bug Fixes**
  - Improved serialization to prevent backend errors when handling empty signature fields.

- **Refactor**
  - Updated message formats and internal handling to use redacted partial signatures instead of full signatures where appropriate, streamlining the signature exchange process.

- **Documentation**
  - Removed outdated or redundant inline comments for improved code clarity.

### DIFF
--- a/trade/src/main/java/bisq/trade/mu_sig/messages/grpc/PartialSignaturesMessage.java
+++ b/trade/src/main/java/bisq/trade/mu_sig/messages/grpc/PartialSignaturesMessage.java
@@ -18,12 +18,14 @@
 package bisq.trade.mu_sig.messages.grpc;
 
 import bisq.common.proto.Proto;
+import bisq.common.util.OptionalUtils;
 import bisq.trade.mu_sig.messages.network.vo.PartialSignatures;
 import bisq.trade.mu_sig.messages.network.vo.RedactedPartialSignatures;
 import com.google.protobuf.ByteString;
 import lombok.Getter;
 
 import java.util.Arrays;
+import java.util.Optional;
 
 @Getter
 public final class PartialSignaturesMessage implements Proto {
@@ -32,7 +34,7 @@ public final class PartialSignaturesMessage implements Proto {
                 peersPartialSignatures.getPeersWarningTxBuyerInputPartialSignature(),
                 peersPartialSignatures.getPeersWarningTxSellerInputPartialSignature(),
                 peersPartialSignatures.getPeersRedirectTxInputPartialSignature(),
-                peersPartialSignatures.getSwapTxInputPartialSignature()
+                Optional.of(peersPartialSignatures.getSwapTxInputPartialSignature())
         );
     }
 
@@ -41,19 +43,19 @@ public final class PartialSignaturesMessage implements Proto {
                 peersPartialSignatures.getPeersWarningTxBuyerInputPartialSignature(),
                 peersPartialSignatures.getPeersWarningTxSellerInputPartialSignature(),
                 peersPartialSignatures.getPeersRedirectTxInputPartialSignature(),
-                new byte[]{}
+                Optional.empty()
         );
     }
 
     private final byte[] peersWarningTxBuyerInputPartialSignature;
     private final byte[] peersWarningTxSellerInputPartialSignature;
     private final byte[] peersRedirectTxInputPartialSignature;
-    private final byte[] swapTxInputPartialSignature;
+    private final Optional<byte[]> swapTxInputPartialSignature;
 
     public PartialSignaturesMessage(byte[] peersWarningTxBuyerInputPartialSignature,
                                     byte[] peersWarningTxSellerInputPartialSignature,
                                     byte[] peersRedirectTxInputPartialSignature,
-                                    byte[] swapTxInputPartialSignature) {
+                                    Optional<byte[]> swapTxInputPartialSignature) {
         this.peersWarningTxBuyerInputPartialSignature = peersWarningTxBuyerInputPartialSignature;
         this.peersWarningTxSellerInputPartialSignature = peersWarningTxSellerInputPartialSignature;
         this.peersRedirectTxInputPartialSignature = peersRedirectTxInputPartialSignature;
@@ -65,15 +67,8 @@ public final class PartialSignaturesMessage implements Proto {
         bisq.trade.protobuf.PartialSignaturesMessage.Builder builder = bisq.trade.protobuf.PartialSignaturesMessage.newBuilder()
                 .setPeersWarningTxBuyerInputPartialSignature(ByteString.copyFrom(peersWarningTxBuyerInputPartialSignature))
                 .setPeersWarningTxSellerInputPartialSignature(ByteString.copyFrom(peersWarningTxSellerInputPartialSignature))
-                .setPeersRedirectTxInputPartialSignature(ByteString.copyFrom(peersRedirectTxInputPartialSignature))
-                .setSwapTxInputPartialSignature(ByteString.copyFrom(swapTxInputPartialSignature));
-
-        // TODO  @stejbac: if swapTxInputPartialSignature is empty byte array,
-        //       passing that to setPeersPartialSignatures results in a backend error as it seems the backend expects
-        //       ByteString.EMPTY. With clearSwapTxInputPartialSignature we get ByteString.EMPTY.
-        if (swapTxInputPartialSignature.length == 0) {
-            builder.clearSwapTxInputPartialSignature();
-        }
+                .setPeersRedirectTxInputPartialSignature(ByteString.copyFrom(peersRedirectTxInputPartialSignature));
+        swapTxInputPartialSignature.ifPresent(e -> builder.setSwapTxInputPartialSignature(ByteString.copyFrom(e)));
         return builder;
     }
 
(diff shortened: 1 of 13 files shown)


HenrikJannsen:
@stejbac Could you have a look to that?

trade/src/main/java/bisq/trade/mu_sig/messages/network/handler/seller_as_maker/MuSigSetupTradeMessage_C_Handler.java
```diff
@@ -69,20 +69,12 @@ protected void process(MuSigSetupTradeMessage_C message) {
...
-        //       if swapTxInputPartialSignature is empty byte array (redacted by the peer),
-        //       passing that to setPeersPartialSignatures results in a backend error.
-        //       Currently redaction is done on the sellers side, with clearSwapTxInputPartialSignature which results in
-        //       ByteString.EMPTY.
-        var signaturesProto = peersPartialSignaturesMessage.toProto(true)
-                .toBuilder()
-                .clearSwapTxInputPartialSignature();
+        // swapTxInputPartialSignature is not set in peersRedactedPartialSignatures, thus the
+        // PartialSignaturesMessage has a cleared swapTxInputPartialSignature field.
+        PartialSignaturesMessage peersredactedPartialSignaturesMessage = PartialSignaturesMessage.from(peersRedactedPartialSignatures);
```
stejbac: Typo (case): Local variable should probably be named `peersRedactedPartialSignaturesMessage`.

trade/src/main/java/bisq/trade/mu_sig/messages/network/vo/PartialSignatures.java
```diff
@@ -35,16 +35,26 @@ public static PartialSignatures from(PartialSignaturesMessage partialSignaturesM
...
+                partialSignatures.getPeersWarningTxSellerInputPartialSignature().clone(),
+                partialSignatures.getPeersRedirectTxInputPartialSignature().clone(),
+                swapTxInputPartialSignature.clone()
+        );
+    }
+
     private final byte[] peersWarningTxBuyerInputPartialSignature;
     private final byte[] peersWarningTxSellerInputPartialSignature;
     private final byte[] peersRedirectTxInputPartialSignature;
     // todo swapTxInputPartialSignature can be empty byte array (redacted), consider to make optional or make 2 classes
```
stejbac: Is this TODO still applicable, now that the `RedactedPartialSignatures` class has been added? (I guess that `b.t.m.m.n.v.PartialSignatures.swapTxInputPartialSignature` should never be empty/missing now.)

trade/src/main/java/bisq/trade/mu_sig/messages/network/MuSigPaymentInitiatedMessage_E.java
```diff
@@ -28,12 +29,16 @@
 @Getter
 @EqualsAndHashCode(callSuper = true)
 public final class MuSigPaymentInitiatedMessage_E extends MuSigTradeMessage {
+    private final byte[] swapTxInputPartialSignature;
```
stejbac, line 32: By adding a byte array field, the Lombok `@EqualsAndHashCode` annotation should be replaced with a manual `equals`/`hashCode` implementation, as was done for the other such DTOs.

trade/src/main/java/bisq/trade/mu_sig/messages/grpc/PartialSignaturesMessage.java
```diff
@@ -52,11 +62,19 @@ public PartialSignaturesMessage(byte[] peersWarningTxBuyerInputPartialSignature,
...
     @Override
     public bisq.trade.protobuf.PartialSignaturesMessage.Builder getBuilder(boolean serializeForHash) {
-        return bisq.trade.protobuf.PartialSignaturesMessage.newBuilder()
+        bisq.trade.protobuf.PartialSignaturesMessage.Builder builder = bisq.trade.protobuf.PartialSignaturesMessage.newBuilder()
                 .setPeersWarningTxBuyerInputPartialSignature(ByteString.copyFrom(peersWarningTxBuyerInputPartialSignature))
                 .setPeersWarningTxSellerInputPartialSignature(ByteString.copyFrom(peersWarningTxSellerInputPartialSignature))
                 .setPeersRedirectTxInputPartialSignature(ByteString.copyFrom(peersRedirectTxInputPartialSignature))
                 .setSwapTxInputPartialSignature(ByteString.copyFrom(swapTxInputPartialSignature));
+
+        // TODO  @stejbac: if swapTxInputPartialSignature is empty byte array,
```
stejbac: In the corresponding `PartialSignaturesMessage` proto (defined in `rpc.proto`), I made the `swapTxInputPartialSignature` field an optional byte array, as it is intended to sometimes be missing (specifically, when the seller doesn't yet have a value to pass to the Rust server, at the trade start). I guess making the field non-optional in this class and using an empty byte array to represent 'missing' is OK, as obviously a valid partial signature will never be empty. But perhaps this class could also make the field optional, to exactly match the underlying proto, then maybe throw an `IllegalArgumentException` if converted into a `PartialSignatures` (as opposed to `RedactedPartialSignature`) instance (since the former is now intended specifically for the case where it isn't missing/empty). (I'm not sure offhand if the Java protobuf implementation is confusing 'empty' and 'missing' for `optional bytes` fields, or if the protobuf spec makes them indistinguishable -- I would presume not.)

stejbac (changes requested):
The changes look OK (and don't give any obvious errors testing against the Rust server), aside from the missed reference equality bug in `equals`/`hashCode` and the minor local variable casing typo, as indicated.

I probably should have added a buyer-as-maker/seller-as-taker trade test/example to `TradeProtocolClient` in the bisq-musig project. In that case, it would be the seller sending Message C and leaving the `swapTxInputPartialSignature` field in, the buyer sending Message D and leaving it out, and then the seller broadcasting the deposit tx (so the taker would always be the one who broadcasts it, as opposed to the seller in bisq1). So with the changes in this PR making `PartialSignatures` and `RedactedPartialSignatures` separate proto types, it may create the small complication of requiring two different versions of Message C and two different versions of Message D when the buyer-as-maker/seller-as-taker path is implemented. Of course, if an entirely distinct set of Messages A, B, C, D would be used anyway, then that won't be an issue.

trade/src/main/java/bisq/trade/mu_sig/messages/network/vo/PartialSignatures.java
HenrikJannsen: I will delete the TODO.

trade/src/main/java/bisq/trade/mu_sig/messages/network/MuSigPaymentInitiatedMessage_E.java
HenrikJannsen: Ah thanks!

trade/src/main/java/bisq/trade/mu_sig/messages/grpc/PartialSignaturesMessage.java
HenrikJannsen: Ah sorry missed that. Will make it optional as well.

HenrikJannsen: I removed now also the following as the clearSwapTxInputPartialSignature seems to be not needed anymore when using optional.
```
 // TODO  @stejbac: if swapTxInputPartialSignature is empty byte array,
        //       passing that to setPeersPartialSignatures results in a backend error as it seems the backend expects
        //       ByteString.EMPTY. With clearSwapTxInputPartialSignature we get ByteString.EMPTY.
        if (swapTxInputPartialSignature.isEmpty() || swapTxInputPartialSignature.get().length == 0) {
            builder.clearSwapTxInputPartialSignature();
        }
```

HenrikJannsen:
> The changes look OK (and don't give any obvious errors testing against the Rust server), aside from the missed reference equality bug in `equals`/`hashCode` and the minor local variable casing typo, as indicated.
> 
> I probably should have added a buyer-as-maker/seller-as-taker trade test/example to `TradeProtocolClient` in the bisq-musig project. In that case, it would be the seller sending Message C and leaving the `swapTxInputPartialSignature` field in, the buyer sending Message D and leaving it out, and then the seller broadcasting the deposit tx (so the taker would always be the one who broadcasts it, as opposed to the seller in bisq1). So with the changes in this PR making `PartialSignatures` and `RedactedPartialSignatures` separate proto types, it may create the small complication of requiring two different versions of Message C and two different versions of Message D when the buyer-as-maker/seller-as-taker path is implemented. Of course, if an entirely distinct set of Messages A, B, C, D would be used anyway, then that won't be an issue.

> I probably should have added a buyer-as-maker/seller-as-taker trade test/example to `TradeProtocolClient`

That would be great, then I can see better the differences. If the distinction of PartialSignatures/RedactedPartialSignatures cause too much extra complexity we might drop it and go back to the grpc message solution by using an optional field.

PR #3523, opened 2025-05-24
Add create offer wizard in new MuSig offerbook

axpoems:
* Add wizard access into the offerbook.
* Introduce step to select market and direction.
* Add create offer button in market header.

## Summary by CodeRabbit

- **New Features**
  - Introduced a dedicated step for selecting trade direction (buy/sell) and market when creating a MuSig offer, featuring an interactive UI with search, filtering, and reputation info.
  - Added a "Create Offer" button to the offer book screen for quicker access to offer creation.
  - Enhanced progress indicator to display the new direction and market selection step.

- **Improvements**
  - Streamlined the offer creation flow by separating direction and market selection into its own step and controller.
  - Improved validation and user feedback when attempting to create sell offers without sufficient reputation.
  - Unified navigation to the offer book regardless of trade direction.
  - Updated offer book layout for clearer organization and usability.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offerbook/MuSigOfferbookView.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offerbook/MuSigOfferbookView.java
@@ -236,6 +236,7 @@ public final class MuSigOfferbookView extends View<VBox, MuSigOfferbookModel, Mu
         marketListVBox = new VBox(header, Layout.hLine(), subheader, appliedFiltersSection, /*favouritesTableView,*/
                 marketListView);
         VBox.setVgrow(marketListView, Priority.ALWAYS);
+        VBox.setVgrow(marketListVBox, Priority.ALWAYS);
         marketListVBox.setMaxWidth(MARKET_LIST_WIDTH);
         marketListVBox.setPrefWidth(MARKET_LIST_WIDTH);
         marketListVBox.setMinWidth(MARKET_LIST_WIDTH);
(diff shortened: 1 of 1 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offerbook/MuSigOfferbookView.java
```diff
@@ -61,23 +61,20 @@ public final class MuSigOfferbookView extends View<VBox, MuSigOfferbookModel, Mu
...
     public MuSigOfferbookView(MuSigOfferbookModel model, MuSigOfferbookController controller) {
         super(new VBox(), model, controller);
 
-        muSigOfferListView = new RichTableView<>(model.getSortedMuSigOfferListItems());
-        muSigOfferListView.getFooterVBox().setVisible(false);
-        muSigOfferListView.getFooterVBox().setManaged(false);
-        configMuSigOfferListView();
-        createAndconfigMarketListView();
         configTitleHBox();
+        crateAndConfigMuSigOfferListView();
```
HenrikJannsen: Using the method makes muSigOfferListView non-final when it should be final as created in constructor. 
I think the little code in the method does not justify the extraction. Using comments as separators also help to structure the different parts. Or return the muSigOfferListView from the method, but that will become cumbersome when more then one object gets created which requires a class field.

axpoems: I did it to be symmetric with the markets table. But I agree, not so much code for now so let's have them both final.

PR #3548, opened 2025-05-31
Enhance amount slider thumbs with hover and pressed states

54corbin:
This PR improves the user experience by adding distinct hover and pressed visual feedback to the amount selection slider thumbs.
Previously, the slider thumbs lacked clear interaction cues. This update makes them more noticeable and interactive when hovered over or dragged.

How it looks like:
image

## Summary by CodeRabbit

- **Style**
  - Enhanced slider thumbs with new hover and pressed state styles, including color change, pointer cursor, and scaling effect for improved interactive feedback.

### DIFF
--- a/apps/desktop/desktop/src/main/resources/css/bisq_easy.css
+++ b/apps/desktop/desktop/src/main/resources/css/bisq_easy.css
@@ -1119,6 +1119,14 @@
     -fx-translate-y: -8;
 }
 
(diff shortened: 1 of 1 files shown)


axpoems:
Thanks for your efforts. However, the issue with the sliders wasn't addressed in this PR and the style you're proposing isn't aligned with the app style nor imo necessary. Since I'm currently reviewing this whole wizard screen, I'll follow-up with a solution for it.

PR #3549, opened 2025-06-01
Fix key handling in trade wizard

axpoems:
* Fix key handling in take offer.
* Add handler for ESC key.
* Call controller directly in setOnKeyPressed.

## Summary by CodeRabbit

- **New Features**
  - Improved keyboard navigation for overlays, allowing users to use Enter and Escape keys for overlay interactions across various offer and trade screens.

- **Refactor**
  - Simplified and centralized key event handling for overlays, reducing code duplication and streamlining event handler assignment.

- **Bug Fixes**
  - Ensured keyboard event handlers are properly registered and removed when overlays are shown or hidden, preventing unintended behavior.


HenrikJannsen (approved):
utACK

Issue #3492, opened 2025-05-17
Investigate ConcurrentModificationException


HenrikJannsen:
I got that exception at startup (localhost dev environment). 

Investigate what cause that exception.

```
Initializing applicationService failed java.util.concurrent.CompletionException: java.util.ConcurrentModificationException
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
	at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1194)
	at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2341)
	at bisq.user.reputation.ReputationService.initialize(ReputationService.java:106)
	at bisq.user.UserService.lambda$initialize$2(UserService.java:80)
	at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1187)
	at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2341)
	at bisq.user.UserService.initialize(UserService.java:80)
	at bisq.desktop_app.DesktopApplicationService.lambda$initialize$10(DesktopApplicationService.java:305)
	at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1150)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
	at java.base/java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:2179)
	at bisq.common.util.CompletableFutureUtils.lambda$anyOf$1(CompletableFutureUtils.java:84)
	at java.base/java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:863)
	at java.base/java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:841)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run$$$capture(CompletableFuture.java:1773)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
[...]

PR #3568, opened 2025-06-17
Introduction of access lock around THREAD_FACTORY_BY_NAME

tsjk:
Should fix https://github.com/bisq-network/bisq2/issues/3492.

The problem was that the `THREAD_FACTORY_BY_NAME` `HashMap`, which is not thread-safe, could be updated concurrently.


tsjk:
The rabbit of course suggests the path not taken. The advantage of the rabbit's suggestion is that reads may be concurrent. But as the dataset seems small the performance benefit is likely negligible. I guess that a third variant is to use a `HashTable`, which would be equivalent with the current implementation, but then the `HashTable` implementation would take care of the locking. The reviewer may give feedback. Changing how to fix the issue is easier than was finding its cause. :)

tsjk:
Hmm, that outside diff range comment by the rabbit is nice. I think it found a typo. The method argument says the unit should be seconds. I'll change this. Even if it's dead code.

HenrikJannsen (approved):
utACK

PR #3598, opened 2025-07-06
Wallet UI - Create Wallet flow

nostrbuddha:
Hi,

Pushing a basic PR for Create wallet dialog flow.

Todo:
 1. Font styles and spacing between elements to match with other similar screens.
 2. Icons in Setup Wallet screen, to be updated with relevant images.

Notes:
 1. Added `WalletException` to MockWalletService :: getSeedWords(). It throws error on first call and return right seeds words on further calls.
 2. Added this exception handling logic, loading logic CreateWalletBackup MVC. Allows the user the `retry` the same functionality again. Should also add `Report to GH` button here.
Like to specifically. get review on this Exception handling logic.

Some suggestions:
 1. We can have a additional OverlayWizardController over `OverlayController`, that manages the Progress steps, next/back navigation between different steps. This logic is now duplicated across various dialogs. I can do this generic component in Wallet Creation flow and later update other flows to make use of this.

Video preview:

https://github.com/user-attachments/assets/ce3c2e10-8183-40fb-aa99-5958628c5e16

## Summary by CodeRabbit

* **New Features**
  * Introduced a multi-step wallet creation wizard, including password protection, seed backup, and seed verification with a quiz interface.
  * Added new UI for users without an initialized wallet, featuring clear instructions and options to create or restore a wallet.
  * Enhanced wallet service with observable wallet initialization and backup states, and support for encryption and seed word management.
  * Expanded navigation to support wallet creation steps and overlays.
  * Added comprehensive localization for all wallet setup screens and actions.
  * Added a new vertical card UI component for instructional content in the wallet setup.

* **Bug Fixes**
  * Improved asynchronous handling and error feedback during wallet seed loading and initialization.

* **Style**
  * Updated password field visuals for improved clarity and usability.
  * Added new font weight and size styles to enhance UI consistency.

* **Documentation**
  * Added new localization keys for wallet setup and improved user guidance throughout the wallet creation process.


HenrikJannsen:
Looks good from my side, just dont get the retry after password.

Wallet setup is only done initially and then the seed verification. After that the wallet is expected to be available. If wrong password user cannot enter and cannot decrypt wallet.

Restore wallet from seed word is another workflow.

The text should follow the style in the other popup screens with a headline style and a smaller content style. I will review code once its not in draft state.

nostrbuddha:
> Looks good from my side, just dont get the retry after password.
'Retry' is to showcase, how if some Exception handling. In this context,
1. MockWalletService.getSeedWords() throws an Exception on first call
2. It's handled in the UI as: "Failed to load seed words" and a 'retry' button.
3. On clicking 'Retry', MockWalletService.getSeedWords() is called again.
4. But this time, that function is hardcoded to return the right set of words.
5. Since it's an async function, with a 1sec simulated delayed, a "Loading..." is shown for 1 second.
Basically, the controller/view now handled Error/Loading states and update UI accordingly.

This is the specific flow, that I like to get feedback on.

> Wallet setup is only done initially and then the seed verification. After that the wallet is expected to be available. If wrong password user cannot enter and cannot decrypt wallet.

Got it.

> Restore wallet from seed word is another workflow.

Yes. Will do it next.

> The text should follow the style in the other popup screens with a headline style and a smaller content style.

Yes. Will do it.

nostrbuddha:
@coderabbitai review

nostrbuddha:
@coderabbitai full review

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/wallet/create_wallet/CreateWalletController.java
```diff
@@ -0,0 +1,214 @@
...
+    private void handleFinalStep() {
+        NavigationTarget currentTarget = model.getNavigationTarget();
+        log.info("Handling final step for {}", currentTarget);
+        if (currentTarget == NavigationTarget.CREATE_WALLET_VERIFY) {
+            CreateWalletVerifyModel.ScreenState state = createWalletVerifyController.getModel().getCurrentScreenState().get();
+            if (state == CreateWalletVerifyModel.ScreenState.SUCCESS) {
+                log.info("Wallet verified successfully, initializing wallet");
+                walletService.purgeSeedWords();
+                walletService.initializeWallet(null, Optional.empty());
+                OverlayController.hide();
```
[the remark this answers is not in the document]
nostrbuddha: I guess, current implementation is fine with MockWalletService. During real wallet service integration, will update this logic accordingly.

HenrikJannsen:
Here feedback from UI and content POV:

## First screen
First screen should look more like https://xd.adobe.com/view/a83c2327-4730-4ec2-8938-e318b2749588-fd6f/screen/a306a26f-b13c-49e4-bb8d-d7db21eb6133/

Fonts are more light, and we use the mid grey tone for sub titles. Best to use the same styles as on other popups screens like create offer. The numbers are also larger and more light font weight.

## Password

Also here it should match the existing styles of similar screens. The distance between the 2 input fields is too large IMO.
There is no min. length required yet, but I think we should require at least 8 chars. We use input validators for such. See the new account screens I just merged for examples. If input is invalid we show the red text below the input, but thats all handled by the validation framework. You just need to set a validator and call validate on the fields at button click. But follow the patters like in the classes at `bisq.desktop.main.content.user.accounts.create.data.payment_form`.

## Seed words
I am not sure if it would be better to use a light grey instead of a dark grey for the words boxes. Maybe the same style as in payment screen at create offer? @axpoems What do you think?

## Seed words verify
Maybe better to use the outlined style for the selected word. Otherwise it competes with the next button.
The success screen feels a bit naked. I think we can use instead a overlay from top like used in create offer screen at completed.

apps/desktop/desktop/src/main/java/bisq/desktop/components/containers/VerticalCard.java
```diff
@@ -0,0 +1,102 @@
+package bisq.desktop.components.containers;
```
HenrikJannsen: As it is wallet specific and not intended to be used as generic component, I think its better to move it to the wallet UI package.

apps/desktop/desktop/src/main/java/bisq/desktop/components/containers/VerticalCard.java
```diff
@@ -0,0 +1,102 @@
...
+                        int imageHeight) {
+        super(SPACING);
+        setMinWidth(DEFAULT_WIDTH);
+        setWidth(DEFAULT_WIDTH);
+        setMaxWidth(DEFAULT_WIDTH);
+        buildUI(step, imageSrc, caption, imageWidth, imageHeight);
+    }
+
+    // Overloaded constructor with default image size
+    public VerticalCard(@Nullable String step,
```
HenrikJannsen: We use @Nullable  only if it is explicitely used for null values. Otherwise by default arguments are expected to be not null. Mostly its also better to use Optional to make it more explicit if values are not mandatory.

apps/desktop/desktop/src/main/java/bisq/desktop/components/containers/VerticalCard.java
```diff
@@ -0,0 +1,102 @@
...
+        this(step, imageSrc, caption, DEFAULT_IMAGE_WIDTH, DEFAULT_IMAGE_HEIGHT);
+    }
+
+    private void buildUI(@Nullable String step,
+                         @Nullable String imageSrc,
+                         @Nullable String caption,
+                         int imageWidth,
+                         int imageHeight) {
+
+        if (step != null && !step.isBlank()) {
```
HenrikJannsen: We have the `StringUtils.isNotEmpty(imageSrc)` util for checks against null and empty strings.

apps/desktop/desktop/src/main/java/bisq/desktop/components/containers/VerticalCard.java
```diff
@@ -0,0 +1,102 @@
...
+            this.getChildren().add(createStepLabel(step));
```
HenrikJannsen: We don't use `this.` if not required

apps/desktop/desktop/src/main/java/bisq/desktop/components/containers/VerticalCard.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    private Label createStepLabel(String step) {
+        Label label = new Label(step);
+        label.getStyleClass().add("very-large-text");
+        return label;
+    }
+
+    private VBox createImageContainer(String imageSrc, int imageWidth, int imageHeight) {
+        Image image = loadImage(imageSrc);
```
HenrikJannsen: We use `ImageUtil.getImageViewById(imageId)` and the `images.css` with the corresponding entry. This handles also the high resolution image automatically with the @2x postfix.

apps/desktop/desktop/src/main/java/bisq/desktop/components/containers/VerticalCard.java
```diff
@@ -0,0 +1,102 @@
...
+        imageView.setFitHeight(imageHeight);
+
+        VBox container = new VBox(SPACING, imageView);
+        container.setAlignment(Pos.CENTER);
+        VBox.setMargin(container, IMAGE_MARGIN);
+        VBox.setVgrow(container, Priority.ALWAYS);
+        return container;
+    }
+
+    private Label createCaptionLabel(String caption) {
```
HenrikJannsen: As the method does not do anything extra beside creating the label, it can be omitted.

HenrikJannsen (changes requested):

axpoems:
It looks good in general. I agree with @HenrikJannsen's UI review points, in particular the seed words in grey.
I haven't looked at the code yet, so I'll only give UI feedback.

1. On the first wallet screen, I would make the images' outlines thinner if it's not too much work. It looks good but would fit better with the app's thin style.

2. On the second screen, top to bottom:

- The header path is not centered.
- The subtitle is too long. I would cut/edit it to fit in two lines. Maybe spread it a little bit more. Also, apply the Create Offer Wizard letter style that HenrikJannsen mentioned.
- The set password input field feels very wide; maybe a 32-character visible length would feel better.
- The circles for the hidden characters are very big; they should be smaller.
- I would put the view/hide icon centered and maybe a little bit bigger.

3. On the third screen, use the same font style as the Create Offer Wizard, and HenrikJannsen's grey suggestion. Maybe add a go-back button here?

4. In the 'Confirm Seed words' section, I would remove the numbering of the subtitle as it sometimes causes confusion with the seed number being asked. The first letter of the words in the title should be in all caps.

5. On the 'seed words confirmation' page, I would remove the 'Confirm Seed words' title and replace it with 'Seed Words Confirmed!' instead. Or use an overlay as suggested by HenrikJannsen. The button prior to that should be 'Finish', 'Create wallet', or 'Complete wallet' instead of 'Next word' as there isn't a next word.

6. The 'Failed to load seedwords' and 'Wrong word!' page should be an overlay warning like you have currently when passwords don't match.

nostrbuddha:
In the WalletView (when wallet is not initialized), I added 3 dummy images. I am bad at creating images. Would be great, If I could get some assistance there.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/wallet/create_wallet/verify/CreateWalletVerifyController.java
```diff
@@ -0,0 +1,120 @@
...
+                });
+    }
+
+    void onCreateWallet() {
+        closeAndNavigateToHandler.accept(NavigationTarget.WALLET);
+    }
+
+    private Region getPopupOwner() {
+        return (Region) view.getRoot().getParent().getParent();
+    }
```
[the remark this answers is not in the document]
nostrbuddha: I guess, the Popup follows the same view hierarchy across the application, hence this is hardcoded like this everywhere.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/wallet/WalletController.java
```diff
@@ -26,22 +31,47 @@
...
+    public void onActivate() {
+        super.onActivate();
+        isWalletInitializedPin = FxBindings.bind(model.getIsWalletInitialized())
+                .to(walletService.getIsWalletInitialized());
+    }
+
+    @Override
+    public void onDeactivate() {
+        super.onDeactivate();
+        if (isWalletInitializedPin != null) {
```
HenrikJannsen, line 69: Not needed as it is guaranteed that onActivate is called before onDeactivate

HenrikJannsen (approved):
ACK

PR #3601, opened 2025-07-08
Integrate payment accounts into offers and trades

HenrikJannsen:


HenrikJannsen:
Here feedback from UI and content POV:

## First screen
First screen should look more like https://xd.adobe.com/view/a83c2327-4730-4ec2-8938-e318b2749588-fd6f/screen/a306a26f-b13c-49e4-bb8d-d7db21eb6133/

Fonts are more light, and we use the mid grey tone for sub titles. Best to use the same styles as on other popups screens like create offer. The numbers are also larger and more light font weight.

## Password

Also here it should match the existing styles of similar screens. The distance between the 2 input fields is too large IMO.
There is no min. length required yet, but I think we should require at least 8 chars. We use input validators for such. See the new account screens I just merged for examples. If input is invalid we show the red text below the input, but thats all handled by the validation framework. You just need to set a validator and call validate on the fields at button click. But follow the patters like in the classes at `bisq.desktop.main.content.user.accounts.create.data.payment_form`.

## Seed words
I am not sure if it would be better to use a light grey instead of a dark grey for the words boxes. Maybe the same style as in payment screen at create offer? @axpoems What do you think?

## Seed words verify
Maybe better to use the outlined style for the selected word. Otherwise it competes with the next button.
The success screen feels a bit naked. I think we can use instead a overlay from top like used in create offer screen at completed.

PR #3611, opened 2025-07-12
feat: Add version info to Support/Resources screen

### LINKED ISSUE #3603: Add version info to resources tab
The version info is a bit hidden in network/p2p network - local version info

We should move or duplicate that content to Support/Resources before Web resources.

mustardcaviar:
Resolves #3603 

Local version info was hidden and hard to find.  This makes it easier to find by adding it to the Support -> Resources tab.  

Changes:
- The Bisq version, commit hash, and tor version are now displayed in the Support -> Resources tab under the local data section.  

- The new headline is consistent with other headlines on the Resources screen

- The content is styled as static text so it stands out from the surrounding hyperlinks

- Added internationalization (i18n) keys to support.properties

## Summary by CodeRabbit

* **New Features**
  * Added a new section in the support resources view displaying local application version details, including Bisq version, commit hash, and Tor version.

* **Documentation**
  * Updated support resources with new labels and formatted descriptions for the local version information section.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/support/resources/ResourcesView.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/support/resources/ResourcesView.java
@@ -17,6 +17,7 @@
 
 package bisq.desktop.main.content.support.resources;
 
+import bisq.common.application.ApplicationVersion;
 import bisq.desktop.common.view.View;
 import bisq.desktop.components.controls.BisqHyperlink;
 import bisq.desktop.components.controls.MaterialTextField;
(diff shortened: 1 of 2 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/support/resources/ResourcesView.java
```diff
@@ -74,10 +74,15 @@ public ResourcesView(ResourcesModel model, ResourcesController controller) {
 
         Label localVersionHeadline = SettingsViewUtils.getHeadline(Res.get("support.resources.localVersion.headline"));
 
+        String versionString = ApplicationVersion.getVersion() != null ?
+            ApplicationVersion.getVersion().getVersionAsString() : "Unknown";
```
HenrikJannsen: For undefined data we usually use `Res.get("na)` which is hte i18n string for `N/A` (Not Available).

HenrikJannsen (changes requested):

HenrikJannsen:
I think it would be better to place it after backup, as backup is more related to local data. 

Also I think the font color might be better the same as in the links as otherwise it gets too much visual weight.  

Screenshot 2025-07-12 at 11 52 13

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/support/resources/ResourcesView.java
mustardcaviar: Thanks for pointing out the standard practice. I've made the change.

mustardcaviar:
> I think it would be better to place it after backup, as backup is more related to local data.

Thanks for feedback. In the latest commit:

- The "Local version info" section has been moved to after the "Backup" section
- The text style has been updated to match the other secondary informational text, 

Screenshot from 2025-07-12 11-41-24

PR #3620, opened 2025-07-15
Feature/for mobile based on 2.1.7 update user identity api

rodvar:
- Rest API implementation for updating trade terms and statement (support for bisq-mobile project)

### DIFF
--- a/http-api/src/main/java/bisq/http_api/rest_api/domain/user_identity/UserIdentityRestApi.java
+++ b/http-api/src/main/java/bisq/http_api/rest_api/domain/user_identity/UserIdentityRestApi.java
@@ -149,26 +149,48 @@ public class UserIdentityRestApi extends RestApiBase {
         }
     }
 
-// Create UserIdentityDto
-   /* @GET
-    @Path("/{id}")
+    @PATCH
     @Operation(
-            summary = "Get User Identity",
-            description = "Retrieves the user identity for the specified ID.",
+            summary = "Update User Identity and Publish User Profile",
+            description = "Updates user identity trade & terms and publishes the associated user profile changes.",
+            requestBody = @RequestBody(
+                    description = "Request payload containing user nickname, terms, statement, and prepared data.",
+                    content = @Content(schema = @Schema(implementation = UpdateUserIdentityRequest.class))
+            ),
             responses = {
-                    @ApiResponse(responseCode = "200", description = "User identity retrieved successfully",
-                            content = @Content(schema = @Schema(implementation = UserIdentity.class))),
-                    @ApiResponse(responseCode = "404", description = "User identity not found"),
+                    @ApiResponse(responseCode = "201", description = "User identity updated successfully",
+                            content = @Content(schema = @Schema(example = "{ \"userProfileId\": \"d22d7b62ef442b5df03378f134bc8f54a2171cba\" }"))),
+                    @ApiResponse(responseCode = "400", description = "Invalid input"),
                     @ApiResponse(responseCode = "500", description = "Internal server error")
             }
     )
-    public Response getUserIdentity(@PathParam("id") String id) {
-        Optional<UserIdentity> userIdentity = userIdentityService.findUserIdentity(id);
-        if (userIdentity.isEmpty()) {
-            return buildNotFoundResponse("Could not find user identity for ID: " + id);
+    public void updateAndPublishUserProfile(UpdateUserIdentityRequest request,
+                                                @Suspended AsyncResponse asyncResponse) {
+        asyncResponse.setTimeout(10, TimeUnit.SECONDS);
+        asyncResponse.setTimeoutHandler(response -> {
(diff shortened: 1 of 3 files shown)


rodvar:
we might want to update coderabbitai settings to allow when merging branches into branch pattern `for-mobile-based-on*`

http-api/src/main/java/bisq/http_api/rest_api/domain/user_identity/UserIdentityRestApi.java
```diff
@@ -176,15 +178,17 @@ public void updateAndPublishUserProfile(UpdateUserIdentityRequest request,
                 asyncResponse.resume(buildResponse(Response.Status.NOT_FOUND, "No selected user identity found"));
             }
             UserProfile userProfile = UserProfile.forEdit(
-                    selectedUserIdentity.getUserProfile(),
+                    Objects.requireNonNull(selectedUserIdentity).getUserProfile(),
                     request.getTerms(),
                     request.getStatement()
             );
+            UserIdentity newUserIdentity = new UserIdentity(selectedUserIdentity.getIdentity(), userProfile);
```
HenrikJannsen: This is not needed.
See editUserProfile code:
```
 public CompletableFuture<BroadcastResult> editUserProfile(UserIdentity oldUserIdentity,
                                                              String terms,
                                                              String statement) {
        Identity oldIdentity = oldUserIdentity.getIdentity();
        UserProfile oldUserProfile = oldUserIdentity.getUserProfile();
        UserProfile newUserProfile = UserProfile.forEdit(oldUserProfile, terms, statement);
        UserIdentity newUserIdentity = new UserIdentity(oldIdentity, newUserProfile);

        synchronized (lock) {
            getUserIdentities().remove(oldUserIdentity);
            getUserIdentities().add(newUserIdentity);
            persistableStore.setSelectedUserIdentity(newUserIdentity);
        }
        persist();

        KeyPair keyPair = oldIdentity.getNetworkIdWithKeyPair().getKeyPair();
        return networkService.removeAuthenticatedData(oldUserProfile, keyPair)
                .thenCompose(result -> publishUserProfile(newUserProfile, keyPair));
    }
    
    ```

Comment (rodvar) [reply http-api/src/main/java/bisq/http_api/rest_api/domain/user_identity/UserIdentityRestApi.java]: thanks for spotting this - will fix now!

File: http-api/src/main/java/bisq/http_api/rest_api/domain/user_identity/UserIdentityRestApi.java [hunk 2]
```diff
@@ -177,16 +177,10 @@ public void updateAndPublishUserProfile(UpdateUserIdentityRequest request,
...
-                    request.getStatement()
-            );
-            UserIdentity newUserIdentity = new UserIdentity(selectedUserIdentity.getIdentity(), userProfile);
-
-            userIdentityService.editUserProfile(newUserIdentity, request.getTerms(), request.getStatement())
+            userIdentityService.editUserProfile(Objects.requireNonNull(selectedUserIdentity), request.getTerms(), request.getStatement())
                     .thenAccept(result -> {
-                        UserProfileDto userProfileDto = DtoMappings.UserProfileMapping.fromBisq2Model(userProfile);
+                        UserIdentity userIdentity = userIdentityService.getSelectedUserIdentity();
+                        UserProfileDto userProfileDto = DtoMappings.UserProfileMapping.fromBisq2Model(userIdentity.getUserProfile());
```
HenrikJannsen, line 183: I guess you want to return the UserIdentity which contains the new UserProfile. Otherwise you have a old stale UserIdentity. But not sure how the mobile code deals with it... You can do similar as in editUserProfile, but I think its more clean to pass the UserIdentity back as that is what was actually changed.

rodvar: Isn't this code doing just that? 

Inside the `editUserProfile` it saves the new identity in the persistableStore , hence why I'm recovering it after it finishes and building the response with it..?

I just retested it, its working 👍

HenrikJannsen: I refer to the UserIdentity on mobile. I dont know if we keep a reference there, if so we would have a stale one as there is a new one created here, but we only pass back the UserProfile.

rodvar: we don't keep a reference per se in the mobile apps, its all handled by the bisq2 jars which I believe already would handle this properly?

HenrikJannsen: Yes on the backend side the UserIdentity is updated. I just looked up the mobile code and it seems we do not hold any reference to UserIdentity, only inside of BisqEasyOpenTradeChannel but that should be delivered from the backend anyway with the updated UserIdentity.

PR #3633, opened 2025-07-18
fix(tor): remove control port file if it exists

### LINKED ISSUE #3631: Tor control port poller fails with new data directory
When starting the app with a new data directory I get below error. After that it works. I assume the tor install process takes more time and the polling error handling is getting triggered too fast. Is likely caused by the recent changes in that area. 

```
15:54:59.794 ERROR [ControlPortFilePoller.pollCycle-10] b.n.t.p.c.ControlPortFilePoller: ControlPortFilePoller for <HOME_DIR>/Library/Application Support/bisq2_Bob_tor/tor/control/control failed after 10 consecutive read failures (during initial parse, attempt 10). Giving up. 
15:54:59.797 ERROR [Network.IO-0.ServiceNodesByTransport.getInitializedDefaultNode-TOR] b.d.DesktopApplicationService: Error at networkFuture.initialize java.util.concurrent.CompletionException: bisq.network.tor.process.control_port.ControlPortFileParseFailureException: Failed to parse/stabilize port file '/Users/dev/Library/Application Support/bisq2_Bob_tor/tor/control/control' after 10 consecutive read failures (phase: initial parse, attempt: 10).
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:347)
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire$$$capture(CompletableFuture.java:708)
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2194)
	at bisq.network.tor.process.control_port.ControlPortFilePoller.checkConsecutiveFailuresAndCompleteExceptionally(ControlPortFilePoller.java:189)
	at bisq.network.tor.process.control_port.ControlPortFilePoller.handleReadFailure(ControlPortFilePoller.java:180)
	at bisq.network.tor.process.control_port.ControlPortFilePoller.executePollCycle(ControlPortFilePoller.java:103)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
	at java.base/java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:317)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: bisq.network.tor.process.control_port.ControlPortFileParseFailureException: Failed to parse/stabilize port file '/Users/dev/Library/Application Support/bisq2_Bob_tor/tor/control/control' after 10 consecutive read failures (phase: initial parse, attempt: 10).
	... 10 common frames omitted
```

wodoro:
fixes #3631

from the errors mentioned in the issue it seems likely that a previously existing control port file is causing wrong port to be used. 
this fix aims to fix it by deleting the file before launching tor and after reading the file, hopefully mitigating the issue.

I updated `FileUtils#deleteFile` to use `deleteIfExists`, as it seemed cleaner and ignores the `NoSuchFileException` internally. previously there was a very small chance of throwing between check and delete in case the file gets deleted before `Files.delete` is called.

### DIFF
--- a/network/tor/tor/src/main/java/bisq/network/tor/TorService.java
+++ b/network/tor/tor/src/main/java/bisq/network/tor/TorService.java
@@ -141,15 +141,30 @@ public class TorService implements Service {
         PasswordDigest hashedControlPassword = PasswordDigest.generateDigest();
         Map<String, String> torrcConfigMap = createTorrcConfigFile(torDataDirPath, hashedControlPassword);
 
+        // remove the control port file before we launch tor, just in case it exists from
+        // a previous run
+        Path controlDirPath = torDataDirPath.resolve(BaseTorrcGenerator.CONTROL_DIR_NAME);
+        Path controlPortFilePath = controlDirPath.resolve("control");
+        try {
+            FileUtils.deleteFile(controlPortFilePath.toFile());
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to delete tor control port file: " + controlPortFilePath, e);
+        }
+
+
         var embeddedTorProcess = new EmbeddedTorProcess(torBinaryPath, torDataDirPath);
         torProcess = Optional.of(embeddedTorProcess);
         embeddedTorProcess.start();
 
-        Path controlDirPath = torDataDirPath.resolve(BaseTorrcGenerator.CONTROL_DIR_NAME);
-        Path controlPortFilePath = controlDirPath.resolve("control");
         return new ControlPortFilePoller(controlPortFilePath)
                 .parsePort()
                 .thenAccept(controlPort -> {
+                    // Remove control port file after we have read the port.
+                    try {
+                        FileUtils.deleteFile(controlPortFilePath.toFile());
+                    } catch (IOException e) {
+                        log.warn("Failed to delete tor control port file after read: {}", controlPortFilePath, e);
+                    }
                     torController.initialize(controlPort);
                     torController.authenticate(hashedControlPassword);
                     torController.bootstrap();
--- a/common/src/main/java/bisq/common/file/FileUtils.java
+++ b/common/src/main/java/bisq/common/file/FileUtils.java
@@ -131,10 +131,12 @@ public class FileUtils {
         }
     }
 
+    /**
+     * Deletes a file if it exists. On some operating systems it may not be possible
+     * to remove a file when it is open and in use by this Java virtual machine or other programs
+     */
     public static void deleteFile(File file) throws IOException {
-        if (file.exists()) {
-            Files.delete(file.toPath());
-        }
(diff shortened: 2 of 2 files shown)


network/tor/tor/src/main/java/bisq/network/tor/TorService.java
```diff
@@ -141,15 +141,30 @@ public CompletableFuture<Boolean> initialize() {
         PasswordDigest hashedControlPassword = PasswordDigest.generateDigest();
         Map<String, String> torrcConfigMap = createTorrcConfigFile(torDataDirPath, hashedControlPassword);
 
+        // remove the control port file before we launch tor, just in case it exists from
+        // a previous run
+        Path controlDirPath = torDataDirPath.resolve(BaseTorrcGenerator.CONTROL_DIR_NAME);
+        Path controlPortFilePath = controlDirPath.resolve("control");
+        try {
+            FileUtils.deleteFile(controlPortFilePath.toFile());
```
HenrikJannsen: For a user upgrading the file exists, but the delete might be delayed a bit by the OS and the poller might then read the old file. To not depend on the pollers resiliance handling, I think its better the wait in a loop for the state that the file does not exist anymore in case the file existed before. E.g. a while loop with a 100 ms sleep and check that the previously existing file does now not exist anymore. Then we can be sure that the poller cannot read a stale file.

HenrikJannsen (changes requested):
The delete after read seems to not delete the file or Tor recreates it. When I set a break point it is deleted but without the file remains. the result oft the `FileUtils.deleteFile` is true (I added a return value - would be good to do in the FileUtils), so the delete worked. I suspect that Tor is creating it again. Either we find out when Tor is recreating it, or we delay the delete to a later point...

I tested to move the code to the end of the method and then the file got deleted.

For extra redundancy we could add `File.deleteOnExit();`

HenrikJannsen (approved):
utACK

PR #3643, opened 2025-07-20
Improve broadcast

### LINKED ISSUE #3637: Make republishing of oracle data less stressful for network
When a oracle nodes restarts it republished all its data. This causes some stress level for the gossip network. We should look into options it lower the impact on the network.

wodoro:
fixes #3637

sudden increase in broadcasts should no longer cause deadlocks
main cause of it was using join() inside the suppliers while the function that was being called was submitting work to another thread.

I tried to contain and separate `reBroadcasts` by using a more limited thread pool and a lower priority


wodoro:
@coderabbitai review

network/network/src/main/java/bisq/network/NetworkService.java
```diff
@@ -92,7 +92,7 @@
 @Slf4j
 public class NetworkService implements PersistenceClient<NetworkServiceStore>, Service {
     public static final ExecutorService NETWORK_IO_POOL = ExecutorFactory.newCachedThreadPool("Network.IO", 5, 3000, 5);
-    public static final ExecutorService NETWORK_IO_LOW_PRIORITY_POOL = ExecutorFactory.newCachedThreadPool("Network.IO.Low-Priority", 2, 4, 5, ExecutorFactory.LOW_PRIORITY);
+    public static final ExecutorService NETWORK_IO_LOW_PRIORITY_POOL = ExecutorFactory.newCachedThreadPool("Network.IO.Low-Priority", 5, 2000, 5, ExecutorFactory.LOW_PRIORITY);
```
HenrikJannsen: I think we can lower the upper limits for the NETWORK_IO_POOL and NETWORK_IO_LOW_PRIORITY_POOL.

It was increased from 300 to 3000 at commit hash: `a5109c828ac20dccdb80a718a548220219c73821`
(Commit msg: `Increase poolsize. There is some threading issue for long running nodes like seeds. This is a temporary mitigation until the cause of the issue is resolved.`)

If you need a higher limit for stress testing, you could add a config parameter to define it.

apps/settings.gradle.kts
```diff
@@ -20,5 +20,6 @@ include("http-api-app")
 include("seed-node-app")
 include("oracle-node-app")
 include("node-monitor-web-app")
+include("chatterbox-app")
```
HenrikJannsen: Maybe better name it more generic for stress testing the network.  There might be more tools in future, like simulation tools. So maybe the module should be like `network-lab`? There could be also future integration tests/test harness or other stuff... In that sense its better to make the app also more generic and use config entries to define which particular stresstest/simutation you want to run. 
The dev can then pass the jvm arg to run a particular scenario.

The changes in config and app can be also done in a separate PR, but the module rename would be good to get changed here.

HenrikJannsen (changes requested):

network/network/src/main/java/bisq/network/NetworkService.java
wodoro: throughput suffers at 300 (I just ran a test)
something around 800 is the amount where it does near it's maximum throughput
of course the test is run with 1 node + 1 desktop, so it does not map to real world case but still it would help I think

PR #3647, opened 2025-07-21
Sync RPC client code with Musig security enhancements

stejbac:
Bring the MuSig2 protocol code into sync with pending (breaking) API changes on the Rust side, which attempt to enhance the trade protocol security, made in the PR https://github.com/bisq-network/bisq-musig/pull/75.

This requires an extra `swapTxInputSighash` (optional) byte array to be passed from the seller to the buyer at the trade start, in Messages C + D. (Message C for the seller-as-taker and Message D for the seller-as-maker.)

The changes also make the Bisq2 client call the gRPC methods `GetPartialSignatures` and `SignSwapTx` twice, once with a redacted response and once with the relevant secret to reveal to the peer (further on in the trade when it's safe to reveal it), to avoid any cryptographic secrets from having to be held on the Java side (even momentarily, as none of the server responses should be sensitive if the API is used correctly).

_These changes should probably only be merged at the same time as https://github.com/bisq-network/bisq-musig/pull/75, as they won't work otherwise._

## Summary by CodeRabbit

* **New Features**
  * Added support for buyer and seller readiness flags during the signature exchange process, improving trade state signaling.
  * Introduced handling and transmission of an additional optional signature hash (swapTxInputSighash) in trade messages.

* **Enhancements**
  * Improved message structures to include new fields for readiness and signature data, enabling more precise communication between trading parties.
  * Updated trade workflows to utilize the new readiness flags and signature hash fields for smoother transaction progression.

### DIFF
--- a/trade/src/main/java/bisq/trade/mu_sig/events/buyer/PaymentInitiatedEventHandler.java
+++ b/trade/src/main/java/bisq/trade/mu_sig/events/buyer/PaymentInitiatedEventHandler.java
@@ -20,23 +20,37 @@ package bisq.trade.mu_sig.events.buyer;
 import bisq.common.util.StringUtils;
 import bisq.trade.ServiceProvider;
 import bisq.trade.mu_sig.MuSigTrade;
+import bisq.trade.mu_sig.MuSigTradeParty;
 import bisq.trade.mu_sig.handler.MuSigTradeEventHandlerAsMessageSender;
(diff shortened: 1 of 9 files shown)


HenrikJannsen:
Note: It requires to run https://github.com/stejbac/bisq2/tree/sync-rpc-client-with-musig-security-enhancements on the backend side.

PR #3682, opened 2025-07-31
Fix transifex workflow dependency

hiciefte:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the German translation for the wallet protection step's skip button to use a shorter, more concise label.


bisqadmin:
utACK

PR #3684, opened 2025-07-31
Change tx call

hiciefte:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the German localization to use formal language in the subheadline message when a trade partner leaves the chat.

* **Chores**
  * Adjusted workflow to use a relative path for the Transifex CLI, improving reliability in translation file syncing.


bisqadmin:
utACK

PR #3701, opened 2025-08-02
Use custom pool for sending messages

HenrikJannsen:
## Summary by CodeRabbit

* **New Features**
  * Introduced asynchronous message sending across the network layer, improving responsiveness and scalability.
  * Added enhanced thread pool management for network read, send, node, and worker operations, including custom queue and rejection policies for better performance and reliability.

* **Refactor**
  * Replaced synchronous and thread pool-based message sending with direct asynchronous methods throughout the network services.
  * Simplified error handling and logging for asynchronous operations, providing clearer feedback on failures.
  * Updated connection closing methods to use asynchronous implementations for improved resource management.
  * Refactored outbound connection creation for better modularity and error handling.
  * Streamlined application initialization and shutdown flows to run off the main thread, improving startup and shutdown sequences.
  * Updated UI thread handling for desktop application initialization.
  * Restructured confidential message sending logic to use CompletableFuture chaining, improving clarity and error handling.

* **Bug Fixes**
  * Improved handling of rejected tasks and network backpressure to prevent dropped messages and improve stability.

* **Chores**
  * Removed unused lifecycle methods returning immediate completions from multiple services for cleaner codebase.
  * Removed deprecated and unused send methods across network node classes.
  * Streamlined imports and removed obsolete executor references.
  * Added logging to default service lifecycle methods for better observability.


common/src/main/java/bisq/common/threading/CallerRunsPolicyWithLogging.java
```diff
@@ -0,0 +1,38 @@
...
+    public CallerRunsPolicyWithLogging() {
+    }
+
+    public void rejectedExecution(Runnable runnable, ThreadPoolExecutor executor) {
+        if (executor.isShutdown()) {
+            log.warn("Executor was already shut down");
+            return;
+        }
+        log.warn("Task rejected. We run the task on the calling thread instead.");
+        runnable.run();
```
wodoro, line 36: I'm not sure if this is a good idea, I think it's better to just reject rather than block a thread unexpectedly

network/network/src/main/java/bisq/network/p2p/services/peer_group/network_load/NetworkLoadExchangeService.java
```diff
@@ -99,7 +99,12 @@ public void onMessage(EnvelopePayloadMessage envelopePayloadMessage, Connection
...
             NetworkLoadExchangeResponse response = new NetworkLoadExchangeResponse(request.getNonce(),
                     myNetworkLoad);
-            NetworkService.NETWORK_IO_POOL.submit(() -> node.send(response, connection));
+            node.sendAsync(response, connection)
+                    .whenComplete((result, throwable) -> {
+                        if (throwable != null) {
+                            log.warn("Sending {} to {} failed. {}", response.getClass().getSimpleName(), connection.getPeerAddress(), ExceptionUtil.getRootCauseMessage(throwable));
+                        }
+                    });
             log.debug("Sent NetworkLoadResponse with nonce {} and my networkLoad {} to {}. Connection={}",
```
wodoro: shouldn't this log.debug be inside the whenComplete as well ?

wodoro (approved):
utACK

I really liked the new  `Async` method signature and impl, it also helps with future changes if we decide to do things differently

common/src/main/java/bisq/common/threading/CallerRunsPolicyWithLogging.java
HenrikJannsen: I use it in the NetworkSendExecutor. We dont want to get too high number of threads from sending messages, thus the pool size and queue are rather limited. This is the only way to further add scaling without failing and without increasing threads by adding backpressure. But once the thread changes are completed we can see if that is needed or if there are better ways.

network/network/src/main/java/bisq/network/p2p/services/peer_group/network_load/NetworkLoadExchangeService.java
HenrikJannsen: Yes, better.

apps/desktop/desktop-app/src/main/java/bisq/desktop_app/DesktopExecutable.java
```diff
@@ -80,12 +81,19 @@ protected void launchApplication(String[] args) {
                 });
     }
 
+    @Override
+    protected void onApplicationLaunched() {
+        supplyAsync(() -> applicationService.initialize().join())
```
wodoro: I think it's better to actually make the initialize() itself Async, so that we can just call it without the need to supplyAsync and then join, that way it requires 1 less thread

common/src/main/java/bisq/common/threading/CallerRunsPolicyWithLogging.java
wodoro: I understand the purpose, It's just that I think it's not a good idea because of the unexpected side effects we may get from doing this
if you are sure this is fine and does not cause any issues all is good

apps/desktop/desktop-app/src/main/java/bisq/desktop_app/DesktopExecutable.java
wodoro: my comment is not correct. since initialize does sync work right now and just returns a future, so it makes no difference other than looking nicer.

wodoro (approved):

apps/desktop/desktop-app/src/main/java/bisq/desktop_app/DesktopExecutable.java
HenrikJannsen: Currently the UI thread starts the initialisation, then at the Network init, a network thread completed the init and continues with the rest of the app init. I wanted to avoid that the UI thread gets potentially slowed down by init code, and that network thread is used for the rest of the init. 
For non UI apps thats not a problem as its then the main thread. The network thread issue was resoled by packing the network init into its own forkjoin pool thread (as we dont provide an executor). 

I did initially do the supplyAsync in initialize but then it would be an unneeded delegation for headless apps.

apps/desktop/desktop-app/src/main/java/bisq/desktop_app/DesktopExecutable.java
```diff
@@ -80,12 +81,19 @@ protected void launchApplication(String[] args) {
...
+                .whenComplete(this::onApplicationServiceInitialized);
+    }
```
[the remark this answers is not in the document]
HenrikJannsen: The reason for that is to avoid that the UI thread is used for the initialisation which could potentially make UI less responsive.

apps/desktop/desktop-app/src/main/java/bisq/desktop_app/DesktopApplicationService.java
```diff
@@ -277,33 +276,13 @@ public DesktopApplicationService(String[] args, ShutDownHandler shutDownHandler)
...
-                        setState(State.INITIALIZE_SERVICES);
-                    }
-                })
+                .thenCompose(result -> securityService.initialize()
+                        .whenComplete((r, t) -> setState(State.INITIALIZE_NETWORK)))
+                .thenCompose(result -> supplyAsync(() -> networkService.initialize().join())
+                        .whenComplete((r, t) -> setState(State.INITIALIZE_WALLET)))
+                .thenCompose(result -> walletService.map(Service::initialize)
+                        .orElse(CompletableFuture.completedFuture(true))
+                        .whenComplete((r, t) -> setState(State.INITIALIZE_SERVICES)))
```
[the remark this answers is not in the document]
HenrikJannsen: The reason for that supplyAsync call at networkService.initialize is because the network thread present at end of initialisation would stick for further service initialisation, which could potentially slow down that thread.

common/src/main/java/bisq/common/threading/CallerRunsPolicyWithLogging.java
HenrikJannsen: I am not sure about potential effects. Worst case is that the caller gets slowed down or even blocked. The alternative to drop the task or throw an exceptions have also hard to evaluate risks. To me it feels the former risk is less severe, specially when we consider that now we had only 1 thread handling a large chunk of network layer tasks and it did not cause issues beside the deadlock we are investigating, so that at least suggests that the callers can deal with the backpressure (the deadlock is another issue, but i think that will be solved be more precise thread management).

I would prefer to keep it for now but observe and test it with stress scenarios where that handler gets activated.

HenrikJannsen:
@coderabbitai full review

HenrikJannsen:
Closed in favor of smaller PRs.
Split up PRs to: #3702, #3703 , #3704

PR #3728, opened 2025-08-08
feat: Implement WalletService with gRPC client

54corbin:
This PR introduces the Java-side implementation for the new wallet architecture.
It creates a GrpcWalletService class which implements the existing WalletService interface. Each method in this class makes a gRPC call to the Rust BDK backend, which will act as the gRPC server.
This change allows the rest of the Java application to interact with the wallet without being aware that the core logic is inside the Rust. The service contract is defined in the new wallet.proto file.

## Summary by CodeRabbit

- New Features
  - Adds a gRPC-backed Wallet service: balance, addresses, transactions/UTXOs, send, encrypt/decrypt, seed words, and readiness checks; desktop can enable/configure it.

- Refactor
  - Wallet API modernized to use protobuf-backed types, read-only observables, and integer coin amounts.

- Chores
  - Wallet configuration added to multiple apps (desktop, HTTP API, node monitor, oracle, seed).

- Tests
  - In-process unit tests for the wallet gRPC client.

- Removal
  - Development mock wallet implementation removed.


gradle/libs.versions.toml
```diff
@@ -40,6 +40,7 @@ google-guava-lib = { strictly = '33.2.1-jre' }
 google-libphonenumber-lib = { strictly = '9.0.8' }
 
 grpc = { strictly = '1.61.0' }
+grpc-testing = { strictly = '1.63.0' }
```
HenrikJannsen: Is it required to use a diff. version here?  If not better use the grpc. you can also update that version.

wallet/src/main/proto/wallet.proto
```diff
@@ -0,0 +1,124 @@
+syntax = "proto3";
+
+package bisq.wallets.grpc;
+
+option java_multiple_files = true;
+option java_package = "bisq.wallets.grpc.pb";
```
HenrikJannsen: Use `protobuf` instead of `pb`. and not plural in wallets. -> `bisq.wallet.protobuf`

wallet/src/main/proto/wallet.proto
```diff
@@ -0,0 +1,124 @@
...
+option java_outer_classname = "WalletProto";
```
HenrikJannsen: Better not use that so the mapping is WalletService->WalletServiceGrpc

wallet/src/main/proto/wallet.proto
```diff
@@ -0,0 +1,124 @@
...
+
+// The WalletService definition.
+service WalletService {
```
HenrikJannsen: In musig we did not use the Service post fix here. -> `Wallet` will be `WalletGrpc` in the java code

wallet/src/main/proto/wallet.proto
```diff
@@ -0,0 +1,124 @@
...
+  repeated string addresses = 1;
+}
+
+message ListTransactionsRequest {}
+
+message ListTransactionsResponse {
+  repeated Transaction transactions = 1;
+}
+
+message ListUnspentRequest {}
```
HenrikJannsen: Maybe better use Utxo -> ListUtxosRequest

wallet/src/main/proto/wallet.proto
```diff
@@ -0,0 +1,124 @@
...
+
+message ListUnspentResponse {
+  repeated Utxo utxos = 1;
+}
+
+message SendToAddressRequest {
+  optional string passphrase = 1;
+  string address = 2;
+  double amount = 3;
```
HenrikJannsen: Amount should be long (sats)

wallet/src/main/proto/wallet.proto
```diff
@@ -0,0 +1,124 @@
...
+
+message IsWalletEncryptedResponse {
+  bool encrypted = 1;
+}
+
+message GetBalanceRequest {}
+
+message GetBalanceResponse {
+  // Balance in satoshis
+  int64 balance = 1;
```
HenrikJannsen: We use sint64 as java always use signed integers.

wallet/src/main/proto/wallet.proto
```diff
@@ -0,0 +1,124 @@
...
+}
+
+message GetSeedWordsRequest {}
```
HenrikJannsen, line 95: I expect that this will require password. but can be changed once we impl it.

wallet/src/main/proto/wallet.proto
```diff
@@ -0,0 +1,124 @@
...
+
+message GetSeedWordsResponse {
+  repeated string seed_words = 1;
+}
+
+message EncryptWalletRequest {
```
HenrikJannsen, line 101: Decrypt method and messages missing

wallet/src/main/proto/wallet.proto
```diff
@@ -0,0 +1,124 @@
...
+  string password = 1;
+}
+
+message EncryptWalletResponse {
+  bool success = 1;
+}
+
+message TransactionInput {
+  string prev_out_tx_id = 1;
```
HenrikJannsen: We use camelCase in protobuf. I know its not protobuf standard but its such a pain and cause for errors to always convert between snake case and camel case. We use exact same field name on java and proto, thus with copy paste its faster and less error prone.

wallet/src/main/proto/wallet.proto
```diff
@@ -0,0 +1,124 @@
...
+  int32 prev_out_index = 2;
+  int64 sequence = 3;
```
HenrikJannsen: sequence -> sequenceNumber

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
```diff
@@ -0,0 +1,248 @@
...
+        return walletAddresses;
+    }
+
+    @Override
+    public CompletableFuture<ObservableSet<String>> requestWalletAddresses() {
+        GetWalletAddressesRequest request = GetWalletAddressesRequest.newBuilder().build();
+        return toCompletableFuture(futureStub.getWalletAddresses(request))
+                .thenApply(response -> {
+                    walletAddresses.clear();
+                    walletAddresses.addAll(response.getAddressesList());
```
HenrikJannsen: You can omit clear and use setAll instead

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
```diff
@@ -0,0 +1,248 @@
...
+
+    @Override
+    public CompletableFuture<String> getUnusedAddress() {
+        GetUnusedAddressRequest request = GetUnusedAddressRequest.newBuilder().build();
+        return toCompletableFuture(futureStub.getUnusedAddress(request))
+                .thenApply(GetUnusedAddressResponse::getAddress);
+    }
+
+    @Override
+    public ObservableSet<String> getWalletAddresses() {
```
HenrikJannsen: We should change interface to rpturn ReadOnlyObservableSet

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
```diff
@@ -0,0 +1,248 @@
...
+        return toCompletableFuture(futureStub.getWalletAddresses(request))
+                .thenApply(response -> {
+                    walletAddresses.clear();
+                    walletAddresses.addAll(response.getAddressesList());
+                    return walletAddresses;
+                });
+    }
+
+    @Override
+    public CompletableFuture<List<? extends TransactionInfo>> listTransactions() {
```
HenrikJannsen: I dont think we will have multiple impl for TransactionInfo -> make a class out of TransactionInfo. Was leftover from multi wallets support.

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
```diff
@@ -0,0 +1,248 @@
...
+        ListTransactionsRequest request = ListTransactionsRequest.newBuilder().build();
+        return toCompletableFuture(futureStub.listTransactions(request))
+                .thenApply(response -> response.getTransactionsList().stream()
+                        .map(GrpcTransactionInfo::new)
+                        .collect(Collectors.toList()));
+    }
+
+    @Override
+    public CompletableFuture<List<? extends Utxo>> listUnspent() {
```
HenrikJannsen: Same here. No need for interface for Utxo

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
```diff
@@ -0,0 +1,248 @@
...
+        return toCompletableFuture(futureStub.getBalance(request))
+                .thenApply(response -> {
+                    Coin newBalance = Coin.asBtcFromValue(response.getBalance());
+                    balance.set(newBalance);
+                    return newBalance;
+                });
+    }
+
+    @Override
+    public Observable<Coin> getBalance() {
```
HenrikJannsen: Again use ReadOnly version -> for all observables

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
```diff
@@ -0,0 +1,248 @@
...
+                });
+    }
+
+    @Override
+    public CompletableFuture<Boolean> shutdown() {
+        channel.shutdown();
+        return CompletableFuture.completedFuture(true);
+    }
+
+    private Transaction toTransaction(bisq.wallets.grpc.pb.Transaction tx) {
```
HenrikJannsen: We use a common pattern in the VO itself to convert from/to proto. Follow that patter please.

I see that most messages here are so simple and have only primitive types so that it feels like an overhead, but at least for those with reference fields I prefer to use the established pattern.

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
```diff
@@ -0,0 +1,248 @@
...
+        return new Transaction(
+                tx.getTxId(),
+                tx.getInputsList().stream().map(this::toTransactionInput).toList(),
+                tx.getOutputsList().stream().map(this::toTransactionOutput).toList(),
+                tx.getLockTime(),
+                tx.getHeight(),
+                Optional.of(new Date(tx.getDate() * 1000)),
```
HenrikJannsen: I guess its optional as it was used for non confirmed txs and use the block date once mined. I guess its better to remove that and leave it to the client to get the block data. But we can leave it for later when implementation requirements are more clear.

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
```diff
@@ -0,0 +1,248 @@
...
+            @Override
+            public void onSuccess(T result) {
+                completableFuture.complete(result);
+            }
+
+            @Override
+            public void onFailure(Throwable t) {
+                completableFuture.completeExceptionally(t);
+            }
+        }, Executors.newSingleThreadExecutor());
```
HenrikJannsen: Use MoreExecutors.directExecutor(). We dont want to map to a new thread here.

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
```diff
@@ -0,0 +1,248 @@
...
+
+    private bisq.wallet.vo.TransactionOutput toTransactionOutput(bisq.wallets.grpc.pb.TransactionOutput output) {
+        return new bisq.wallet.vo.TransactionOutput(
+                output.getValue(),
+                output.getAddress(),
+                output.getScriptPubKey()
+        );
+    }
+
+    private static <T> CompletableFuture<T> toCompletableFuture(ListenableFuture<T> future) {
```
HenrikJannsen: Move method to CompletableFutureUtils

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
```diff
@@ -0,0 +1,248 @@
...
+
+            @Override
+            public void onFailure(Throwable t) {
+                completableFuture.completeExceptionally(t);
+            }
+        }, Executors.newSingleThreadExecutor());
+        return completableFuture;
+    }
+
+    private record GrpcTransactionInfo(bisq.wallets.grpc.pb.Transaction tx) implements TransactionInfo {
```
HenrikJannsen: Move out into vo package. Dont prefix with Grpc remove interface and use that impl directly.

HenrikJannsen: Actually I dont see the reason for that class. Its just a subset of tx. I guess we can remove it and use Tx instead

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
```diff
@@ -0,0 +1,248 @@
...
+import io.grpc.ManagedChannelBuilder;
+
+import java.util.Date;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executors;
+import java.util.stream.Collectors;
+
+public class GrpcWalletService implements WalletService {
```
HenrikJannsen: The wallet service might handle more then just grpc calls. Better extract grpc stuff to WalletGrpcClient and use that here for delegations.

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
```diff
@@ -0,0 +1,248 @@
...
+
+    public record ServerConfig(String host, int port) {
```
HenrikJannsen: We have the pattern to name that Config (as static inner class of the service it belongs to). And use the typesafe config for the values. Add the config node (I removed the wallets config in the PR where wallets got removed) to all configs also those where no wallet is expected like seed node. Add an enable flag to signal usage. We want to keep all configs in sync otherwise its more cumbersome to deal with missing fields.

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
```diff
@@ -0,0 +1,248 @@
...
+    private final ManagedChannel channel;
+    private final WalletServiceGrpc.WalletServiceBlockingStub blockingStub;
+    private final WalletServiceGrpc.WalletServiceFutureStub futureStub;
+
+    private final Observable<Coin> balance = new Observable<>();
+    private final ObservableSet<Transaction> transactions = new ObservableSet<>();
+    private final ObservableSet<String> walletAddresses = new ObservableSet<>();
+
+    public GrpcWalletService(ServerConfig config) {
+        this.channel = ManagedChannelBuilder.forAddress(config.host(), config.port())
```
HenrikJannsen: Dont do init code in constructor. Service impl have the initialize and shutdown methods.

For mobile there can be the use case to actually shut down service and then init them again.

HenrikJannsen (changes requested):

wallet/src/main/proto/wallet.proto
54corbin: I get what you mean about Java longs, but BTC balances are never negative, so I would say uint64 makes more sense. sint64 is only better if we expect lots of negatives, which we don’t. In Java it’s still a long, and Bitcoin’s max supply is way below Long.MAX_VALUE

HenrikJannsen: Yes, its just a bit confusing to have unsigned on proto and signed when deserialized to java, as java has no unsigned int/long. But of course it can be seen also that its confusing in proto to see it signed... 
But as we applied that to all other proto (using sint) I prefer to stick to that.

wallet/src/main/java/bisq/wallet/GrpcWalletService.java
54corbin: Renamed the Config as static inner class of the `BDKWalletService`.
As for the "Add the config node", my understanding is that  should be handled at the place where the `BDKWalletService` was instanced. Not sure if I should address that part in this PR or not?

54corbin:
@HenrikJannsen Thank you very much for the comments. I've addressed all of the comments. PTAL
Thanks

wallet/src/main/proto/wallet.proto
```diff
@@ -0,0 +1,133 @@
+syntax = "proto3";
+
+package bisq.wallet.grpc;
```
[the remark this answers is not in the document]
HenrikJannsen: Yes, should be `package wallet;`

common/src/main/java/bisq/common/util/CompletableFutureUtils.java
```diff
@@ -109,7 +113,21 @@ public static <T> CompletableFuture<T> logOnFailure(CompletableFuture<T> future,
...
+            @Override
+            public void onSuccess(T result) {
+                completableFuture.complete(result);
+            }
+
+            @Override
+            public void onFailure(Throwable t) {
+                completableFuture.completeExceptionally(t);
+            }
+        }, Executors.newSingleThreadExecutor());
```
HenrikJannsen: As said we usually dont want a new thread here, best to add the param to the method signature for passing the executor and use MoreExecutors.directExecutor() which is same as the caller.

wallet/src/main/java/bisq/wallet/BDKWalletService.java
```diff
@@ -0,0 +1,150 @@
...
+import bisq.wallet.protobuf.SendToAddressRequest;
+import bisq.wallet.protobuf.SendToAddressResponse;
+import bisq.wallet.vo.Transaction;
+import bisq.wallet.vo.Utxo;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+public class BDKWalletService implements WalletService {
```
HenrikJannsen: We use PascalCase for acronyms -> BdkWalletService.

wallet/src/main/java/bisq/wallet/BDKWalletService.java
```diff
@@ -0,0 +1,150 @@
...
+                });
+    }
+
+    @Override
+    public CompletableFuture<Boolean> shutdown() {
+        client.shutdown();
+        return CompletableFuture.completedFuture(true);
+    }
+
+    public record Config(String host, int port) {
```
HenrikJannsen: Use same style as elsewhere where we apply typesafe config

wallet/src/main/java/bisq/wallet/BDKWalletService.java
```diff
@@ -0,0 +1,150 @@
...
+
+public class BDKWalletService implements WalletService {
+
+    private final BDKWalletGrpcClient client;
+    private final Observable<Coin> balance = new Observable<>();
+    private final ObservableSet<Transaction> transactions = new ObservableSet<>();
+    private final ObservableSet<String> walletAddresses = new ObservableSet<>();
+
+    public BDKWalletService(Config config) {
+        this(new BDKWalletGrpcClient(config.host(), config.port()));
```
HenrikJannsen: We use initialize for any resource or time consuming operation.  Same for BDKWalletGrpcClient. Follow pattern in Musig...

apps/desktop/desktop/src/main/java/bisq/desktop/common/observable/FxBindings.java
```diff
@@ -129,6 +131,11 @@ public Pin to(ObservableSet<S> observable) {
         public Pin to(ObservableArray<S> observable) {
             return observable.addCollectionChangeMapper(observableList, filterFunction, mapFunction, UIThread::run);
         }
+
+        public Pin to(ReadOnlyObservableSet<S> observable) {
+            var collectionChangeMapper = new CollectionChangeMapper<>(observableList, filterFunction, mapFunction, UIThread::run);
+            return observable.addObserver(collectionChangeMapper);
```
HenrikJannsen: This should work but has a slightly difference to the addCollectionChangeMapper version as it does call addAll instead of setAll. This is usually not an issue as the collection is normally empty, but for sake of being compliant with the other method I think its better to use below solution. We could also log a warning in the else case, as its not expected to happen. We could add the same to the other ReadOnly versions but not need to do in that PR, we can do it on demand... 

```
public Pin to(ReadOnlyObservableSet<S> observable) {
            if (observable instanceof ObservableSet<S> observableSet) {
                return to(observableSet);
            } else {
                // Not expected case
                CollectionChangeMapper<S, T> collectionChangeMapper = new CollectionChangeMapper<>(observableList, filterFunction, mapFunction, UIThread::run);
                return observable.addObserver(collectionChangeMapper);
            }
        }
```

gradle/libs.versions.toml
```diff
@@ -40,6 +40,7 @@ google-guava-lib = { strictly = '33.2.1-jre' }
 google-libphonenumber-lib = { strictly = '9.0.8' }
 
 grpc = { strictly = '1.61.0' }
+grpc-testing = { strictly = '1.61.0' }
```
HenrikJannsen: My suggestion was to remove the grpc-testing entry as we normally want to use same versions. Not that important.... just to avoid overload the version handling

wallet/src/main/java/bisq/wallet/WalletService.java
```diff
@@ -20,42 +20,171 @@
...
+                .thenApply(response -> {
+                    transactions.setAll(response.getTransactionsList().stream()
+                            .map(Transaction::fromProto)
+                            .toList());
+                    return transactions;
+                });
+    }
+
+    @Override
+    public CompletableFuture<Boolean> initialize() {
```
HenrikJannsen, line 167: Please follow the code style we use elsewhere to put the init and shutdown after constructor and Config or other static classes, enums at start of class.

wallet/src/main/java/bisq/wallet/WalletService.java
```diff
@@ -20,42 +20,171 @@
...
+        if (!config.enabled) {
+            log.info("WalletService was disabled.");
+            return CompletableFuture.completedFuture(true);
+        }
+        try {
+            this.client = new WalletGrpcClient(config.host, config.port);
```
HenrikJannsen: If you would use the initialize in WalletGrpcClient you could make the field final and create it in the constructor. We prefer final fields when it makes sense. As the client is actually doing the resource allocation and take time, to do it in initialize would be better IMO. In general we try to keep constructors as light as possible and only assign values but dont do any heavy work

wallet/src/main/proto/wallet.proto
```diff
@@ -0,0 +1,132 @@
...
+    string password = 1;
+}
+
+message DecryptWalletResponse {
+    bool success = 1;
+}
+
+message TransactionInput {
+    string prevOutTxId = 1;
+    int32 prevOutIndex = 2;
```
HenrikJannsen, line 119: As mentioned before we should use `sint32`

wallet/src/main/java/bisq/wallet/WalletGrpcClient.java
```diff
@@ -0,0 +1,158 @@
...
+    private WalletGrpc.WalletFutureStub futureStub;
+    private final String host;
+    private final int port;
+
+    public WalletGrpcClient(String host, int port) {
+        this.host = host;
+        this.port = port;
+    }
+
+    public WalletGrpcClient(ManagedChannel managedChannel) {
```
54corbin, line 64: allow unitests to init the` WalletGrpcClient`

Issue #3605, opened 2025-07-09
Text not wrapped at create user profile


HenrikJannsen:
<img width="2062" height="1302" alt="Image" src="https://github.com/user-attachments/assets/24fbd00d-c92e-4ccb-a55b-ea34bb8c5c37" />

sidhantmourya:
hi @HenrikJannsen it looks like the wrapping is not set to true for the above text in the image in the following file.
https://github.com/bisq-network/bisq2/blob/c5cb047c2ed1c91623f87907b26c5b4a25569af6/apps/desktop/desktop/src/main/java/bisq/desktop/overlay/onboarding/create_profile/CreateProfileView.java#L60
can you set wrap to true and check if it fixes your issue?
subtitleLabel.setWrapText(true);

let me know if it resolves your issue.
Thanks

martinovjulian:
<img width="1158" height="678" alt="Image" src="https://github.com/user-attachments/assets/f2098be9-c436-420e-b304-119eb2143841" />
It appears to be wrapping just fine for me. Line 64 in the aforementioned file is as follows - marking wrap-text.

`        subtitleLabel.getStyleClass().addAll("bisq-text-3", "wrap-text");
`
@sidhantmourya @HenrikJannsen

HenrikJannsen:
It happens at the create user profile screen from the User Options menu.

mathcoder23:
I can reproduce this issue on macOS, and simply calling `subtitleLabel.setWrapText(true)` still doesn’t work. I’ll work on this issue as my first PR

PR #3759, opened 2025-08-16
Fix text not wrapped at create user profile

mathcoder23:
**Background:** 
- When clicking "Create new profile" in User Options, the title section does not wrap automatically, causing the UI to display incomplete information.
- Related issue: https://github.com/bisq-network/bisq2/issues/3605
- Reproduce this UI issue locally
image

**Changes:**
- We have already set wrap-text in the style, but it didn’t take effect because minHeight was fixed at 40. This likely caused the wrap to fail, as the label height wasn’t correctly calculated during UI rendering. Therefore, we set minHeight to Region.USE_PREF_SIZE to calculate the label height dynamically.

**How to verify:**
- I have reproduced the issue locally and verified that it is resolved after the code changes.
- 
image

## Summary by CodeRabbit

- Style
  - Improved the onboarding “Create Profile” screen subtitle for better readability and responsiveness. The subtitle now wraps across multiple lines and adjusts its height dynamically, preventing truncation or clipping on smaller windows and with longer or localized text. Enhances layout consistency and user experience across different display sizes and languages.


HenrikJannsen (approved):
utACK

PR #3786, opened 2025-08-23
fix: wallet.proto sint -> uint

54corbin:
## Summary by CodeRabbit

- Bug Fixes
  - Prevented negative/overflow issues for large balances, amounts, confirmations, timestamps, and block heights.
  - Improved reliability and correctness when reading and displaying transaction scripts.

- Chores
  - Migrated wallet data to more precise numeric and binary types for amounts, counts, timestamps, and scripts to improve robustness.

- Refactor
  - Updated internal conversions and data handling to align with the new schema for consistent transaction and balance presentation.

### DIFF
--- a/wallet/src/main/proto/wallet.proto
+++ b/wallet/src/main/proto/wallet.proto
@@ -72,7 +72,7 @@ message ListUtxosResponse {
 message SendToAddressRequest {
     optional string passphrase = 1;
     string address = 2;
-    sint64 amount = 3;
+    uint64 amount = 3;
 }
 
 message SendToAddressResponse {
@@ -89,7 +89,7 @@ message GetBalanceRequest {}
 
 message GetBalanceResponse {
     // Balance in satoshis
-    sint64 balance = 1;
+    uint64 balance = 1;
 }
 
 message GetSeedWordsRequest {}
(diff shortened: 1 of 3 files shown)


wallet/src/main/proto/wallet.proto
```diff
@@ -89,7 +89,7 @@ message GetBalanceRequest {}
 
 message GetBalanceResponse {
     // Balance in satoshis
-    sint64 balance = 1;
+    uint64 balance = 1;
```
HenrikJannsen, line 92: Shouldn't we use then always int64/int32 instead of mixing uint with int?

54corbin:
The corresponding PR:https://github.com/bisq-network/bisq-musig/pull/82 on rust side has been merged.

HenrikJannsen (changes requested):
Open question

wallet/src/main/proto/wallet.proto
54corbin: I just feel like a nagative balance value doesn't make sense.

stejbac: I'm pretty sure that for `int64` versus `uint64`, the protobuf Java bindings have identical (de)serialisation behaviour. The 1-10 byte VarInt encoding on the wire represents a 64-bit signed or unsigned integer, encoded as a 2's complement binary representation in the latter case, and coercion to/from a Java `long` behaves identically in either case. (For `uint64`, a Java `long` is reinterpreted as an unsigned 64-bit integer in the obvious way, so -1 <-> 18446744073709551615 == 0xFFFFFFFFFFFFFFFF.)

For `int32` versus `uint32`, the protobuf deserialisation behaves identically, with wrapping/truncating behaviour to fit the lower 32 bits into a Java `int`. The serialisation of a Java `int` appears to behave slightly differently in each case. For `int32`, the `int` is sign-extended to a `long` which is then encoded on the wire as a VarInt in the same way as above. For `uint32`, the `int` is _zero_-extended to a `long`. But in either case, the Rust side will _also_ silently wrap/truncate to fit in the lower 32 bits when deserialising, so there's no practical difference between zero-extension and sign-extension.

So it makes very little difference whether int32/64 or uint32/64 is specified in the proto, or even if one side is using int32/64 in its proto and the other side is using uint32/64 in its proto, because from the perspective of the Java code it will behave identically. (But using signed integers to represent clearly unsigned quantities is a bit less nice on the Rust side.)

HenrikJannsen: What would you recommend to use in the proto, uint64 or int64?

stejbac: I would lean towards uint64, since that would make the bindings a bit nicer on the Rust side, and it shouldn't make any functional difference to the bindings on the Java side.

I guess one could add validation to the Java DTO that the serialised/deserialised balance doesn't overflow and wrap to a negative `long`, to help avoid bugs when formatting the balance or doing any integer comparisons/conversions. Of course, any real BTC balance will fall well short of `Long.MAX_VALUE` anyway, so such checks shouldn't do any obvious harm.

HenrikJannsen (approved):
utACK

Issue #3771, opened 2025-08-19
Use RequestResponseHandler for ReportRequestService


HenrikJannsen:
Similar like other request/response patterns we should use RequestResponseHandler. See KeepAliveService as simple example.

neminatha:
I would like to do that.

HenrikJannsen:
Oh, I just asked @mathcoder23  who works on https://github.com/bisq-network/bisq2/issues/3727 to take that as well.

Sorry for not updating the issue here

HenrikJannsen:
@mathcoder23 Can you take that issue as well? I only can assign you after a comment...

mathcoder23:
> [@mathcoder23](https://github.com/mathcoder23) Can you take that issue as well? I only can assign you after a comment...

@HenrikJannsen Thanks, I can do this. It looks like a good start for the RequestResponseHandler.

mathcoder23:
Hi @HenrikJannsen I’d like to expand the refactoring scope of this issue to include the refactoring of [ReportRequestService](https://github.com/mathcoder23/bisq2/blob/main/network/network/src/main/java/bisq/network/p2p/services/reporting/ReportResponseService.java#L66) within ReportResponseService, because [ReportRequestService](https://github.com/mathcoder23/bisq2/blob/main/network/network/src/main/java/bisq/network/p2p/services/reporting/ReportRequestService.java#L84) doesn’t fully follow the request–response pattern at the moment. It’s essentially an external API call that forwards into the P2P network, while ReportResponseService is a standard request–response implementation. Changing them together would make testing easier. What do you think?

HenrikJannsen:
Sounds good to me!

PR #3789, opened 2025-08-23
Refactor RequestResponseHandler to a delegation pattern and apply it in Report

mathcoder23:
# Context:
- Related issue: https://github.com/bisq-network/bisq2/issues/3771 
- This PR refactors `ReportRequestService` and `ReportResponseService` to adopt the `RequestResponseHandler` pattern. The goal is to improve code reusability and centralize the management of P2P connection listening, timeouts, and concurrent request handling, ensuring adherence to the request-response pattern.
- In particular, some services are 'request-only'(leech) and do not adhere to a strict Request/Response model. To accommodate this, this PR creates the LeechRequestResponseHandler for use cases that only require sending requests.

# Changes Made:
- Refactor: ReportResponseService now adopts the RequestResponseHandler pattern (in practice, it only functions in 'response' mode).
- New Abstract Class: Introduced LeechRequestResponseHandler, an abstract class designed for services operating in 'Leech' mode within the P2P network (i.e., they only send requests and do not process responses). ReportRequestService, which forwards requests from a RESTful API to the P2P network, is a typical use case. LeechRequestResponseHandler inherits from RequestResponseHandler but overrides response-handling logic to function solely as a request-side client.
- Refactor: ReportRequestService now adopts the LeechRequestResponseHandler pattern, which eliminates all its manual lifecycle management logic.
- Logging Enhancement: The RequestResponseHandler base class has been enhanced. All its key log outputs now automatically include the specific service class name as a prefix (e.g., [KeepAliveService]), significantly improving debugging efficiency in complex environments.

# Verification & Testing
- I have successfully verified and tested these changes locally.
- Testing was primarily conducted on a CLEAR network using the `seed-node-app` and `node-monitor-web-app`. Since ReportRequestService and ReportResponseService are split into two distinct P2PServiceNode types, "REPORT_RESPONSE" and "REPORT_REQUEST", I have also tested various combinations to verify that both services can operate concurrently on the same Node without interference.

## Test Steps(brief):
1. Configure a CLEAR network for both the seed-node-app and the node-monitor-web-app.
2. Configure different combinations of the "REPORT_RESPONSE" and "REPORT_REQUEST" services in the P2PServiceNode for each app.
3. Access http://localhost:8083/api/v1/report/127.0.0.1:8000. It should return the correct report information.

## Summary by CodeRabbit

- Refactor
  - Reworked P2P request/response handling into reusable handler/delegate components, simplifying request routing and response generation.
  - Migrated reporting flows to the new handler pattern and removed legacy per-request helper behavior.
  - Inventory now reads pending-request counts via the unified handler interface.

- Bug Fixes
  - Centralized timeout and disconnect cleanup to reduce stuck or orphaned requests.

- Chores
  - Added a lightweight lifecycle interface to standardize initialize/shutdown across network handlers.

### DIFF
--- a/network/network/src/main/java/bisq/network/p2p/common/HandlerLifecycle.java
+++ b/network/network/src/main/java/bisq/network/p2p/common/HandlerLifecycle.java
@@ -16,7 +16,7 @@
  */
 
 package bisq.network.p2p.common;
-
+// TODO Define a superclass Service and a Lifecycle interface in the common package, and consider better naming.
 public interface HandlerLifecycle {
     void initialize();
 
--- a/network/network/src/main/java/bisq/network/p2p/common/RequestHandler.java
+++ b/network/network/src/main/java/bisq/network/p2p/common/RequestHandler.java
@@ -24,8 +24,9 @@ import bisq.network.p2p.message.Response;
 import bisq.network.p2p.node.CloseReason;
 import bisq.network.p2p.node.Connection;
 import bisq.network.p2p.node.Node;
+import lombok.extern.slf4j.Slf4j;
 
-
+@Slf4j
 public abstract class RequestHandler<T extends Request, R extends Response> extends BaseHandler implements Node.Listener, HandlerLifecycle {
     protected final Node node;
     private final RequestHandlerDelegate<T, R> requestHandlerDelegate;
@@ -46,6 +47,20 @@ public abstract class RequestHandler<T extends Request, R extends Response> exte
     }
 
     /* --------------------------------------------------------------------- */
+    // HandlerLifecycle implementation
+    /* --------------------------------------------------------------------- */
+
+    public void initialize() {
+        node.addListener(this);
+        requestHandlerDelegate.initialize();
+    }
+
+    public void shutdown() {
+        node.removeListener(this);
+        requestHandlerDelegate.shutdown();
+    }
+
+    /* --------------------------------------------------------------------- */
     // RequestHandler implementation
     /* --------------------------------------------------------------------- */
 
@@ -61,26 +76,17 @@ public abstract class RequestHandler<T extends Request, R extends Response> exte
     }
 
     /* --------------------------------------------------------------------- */
-    // HandlerLifecycle implementation
-    /* --------------------------------------------------------------------- */
-
-    public void initialize() {
-        node.addListener(this);
-        requestHandlerDelegate.initialize();
-    }
-
-    public void shutdown() {
-        node.removeListener(this);
-        requestHandlerDelegate.shutdown();
-    }
-
-    /* --------------------------------------------------------------------- */
     // Node.Listener implementation
     /* --------------------------------------------------------------------- */
 
     @Override
     public void onMessage(EnvelopePayloadMessage envelopePayloadMessage, Connection connection, NetworkId networkId) {
-        requestHandlerDelegate.resolveRequest(envelopePayloadMessage).ifPresent(request -> processRequest(connection, request));
+        try {
+            requestHandlerDelegate.resolveRequest(envelopePayloadMessage)
+                    .ifPresent(request -> processRequest(connection, request));
+        } catch (Throwable t) {
+            log.error("Error handling message: {}", connection, t);
+        }
     }
 
     @Override
--- a/network/network/src/main/java/bisq/network/p2p/common/RequestHandlerDelegate.java
+++ b/network/network/src/main/java/bisq/network/p2p/common/RequestHandlerDelegate.java
@@ -48,6 +48,18 @@ class RequestHandlerDelegate<T extends Request, R extends Response> implements H
         this.callback = callback;
     }
 
+    /* --------------------------------------------------------------------- */
+    // HandlerLifecycle implementation
+    /* --------------------------------------------------------------------- */
+
+    @Override
+    public void initialize() {
+    }
+
+    @Override
+    public void shutdown() {
+    }
+
     protected Optional<T> resolveRequest(EnvelopePayloadMessage message) {
         return requestClass.isInstance(message)
                 ? Optional.of(requestClass.cast(message))
@@ -67,18 +79,4 @@ class RequestHandlerDelegate<T extends Request, R extends Response> implements H
                     }
                 });
     }
-
-    /* --------------------------------------------------------------------- */
-    // HandlerLifecycle implementation
-    /* --------------------------------------------------------------------- */
-
-    @Override
-    public void initialize() {
-    }
-
-    @Override
-    public void shutdown() {
-    }
-
-
 }
--- a/network/network/src/main/java/bisq/network/p2p/common/RequestResponseHandler.java
+++ b/network/network/src/main/java/bisq/network/p2p/common/RequestResponseHandler.java
@@ -50,7 +50,22 @@ public abstract class RequestResponseHandler<T extends Request, R extends Respon
             }
         });
         this.responseHandlerDelegate = new ResponseHandlerDelegate<>(node, timeout, getResponseClass());
+    }
 
+    /* --------------------------------------------------------------------- */
+    // HandlerLifecycle implementation
+    /* --------------------------------------------------------------------- */
+
+    public void initialize() {
+        node.addListener(this);
+        requestHandlerDelegate.initialize();
+        responseHandlerDelegate.initialize();
+    }
+
+    public void shutdown() {
+        node.removeListener(this);
+        requestHandlerDelegate.shutdown();
+        responseHandlerDelegate.shutdown();
     }
 
     /* --------------------------------------------------------------------- */
@@ -87,30 +102,19 @@ public abstract class RequestResponseHandler<T extends Request, R extends Respon
     }
 
     /* --------------------------------------------------------------------- */
-    // HandlerLifecycle implementation
-    /* --------------------------------------------------------------------- */
-
-    public void initialize() {
-        node.addListener(this);
-        requestHandlerDelegate.initialize();
-        responseHandlerDelegate.initialize();
-    }
-
-    public void shutdown() {
-        node.removeListener(this);
-
-        requestHandlerDelegate.shutdown();
-        responseHandlerDelegate.shutdown();
-    }
-
-    /* --------------------------------------------------------------------- */
     // Node.Listener implementation
     /* --------------------------------------------------------------------- */
 
     @Override
     public void onMessage(EnvelopePayloadMessage envelopePayloadMessage, Connection connection, NetworkId networkId) {
-        responseHandlerDelegate.resolveResponse(envelopePayloadMessage).ifPresent(response -> processResponse(connection, response));
-        requestHandlerDelegate.resolveRequest(envelopePayloadMessage).ifPresent(request -> processRequest(connection, request));
+        try {
+            responseHandlerDelegate.resolveResponse(envelopePayloadMessage)
+                    .ifPresent(response -> processResponse(connection, response));
+            requestHandlerDelegate.resolveRequest(envelopePayloadMessage)
+                    .ifPresent(request -> processRequest(connection, request));
+        } catch (Throwable t) {
+            log.error("Error handling message: {}", connection, t);
+        }
     }
 
     @Override
--- a/network/network/src/main/java/bisq/network/p2p/common/ResponseHandlerDelegate.java
+++ b/network/network/src/main/java/bisq/network/p2p/common/ResponseHandlerDelegate.java
@@ -24,6 +24,7 @@ import bisq.network.p2p.message.Response;
 import bisq.network.p2p.node.CloseReason;
 import bisq.network.p2p.node.Connection;
 import bisq.network.p2p.node.Node;
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
 import java.util.Map;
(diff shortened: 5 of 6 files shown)


network/network/src/main/java/bisq/network/p2p/services/reporting/ReportHandler.java
```diff
@@ -1,101 +0,0 @@
...
-        return future;
-    }
-
-
-    /* --------------------------------------------------------------------- */
-    // Connection.Listener implementation
-    /* --------------------------------------------------------------------- */
-
-    @Override
-    public void onNetworkMessage(EnvelopePayloadMessage envelopePayloadMessage) {
```
mathcoder23, line 70: The listening of a Connection can be captured within the Node listener, so it is no longer necessary to register a separate Connection listener.

network/network/src/main/java/bisq/network/p2p/services/reporting/ReportRequestService.java
```diff
@@ -18,94 +18,40 @@
...
-//TODO Extend RequestResponseHandler
-@Slf4j
-public class ReportRequestService implements Node.Listener {
-    private static final long TIMEOUT_SEC = 60;
+import static java.util.concurrent.TimeUnit.SECONDS;
 
-    private final Node node;
-    private final Map<String, ReportHandler> requestHandlerMap = new ConcurrentHashMap<>();
+@Slf4j
+public class ReportRequestService extends LeechRequestResponseHandler<ReportRequest, ReportResponse> {
```
mathcoder23: `LeechRequestResponseHandler` is critical for `ReportRequestService`. Since `ReportRequestService` and `ReportResponseService` share the same `ReportRequest`, once `ReportRequestService` inherits `RequestResponseHandler`, it will start handling `ReportRequest` messages, which leads to incorrect responses when both "REPORT_RESPONSE" and "REPORT_REQUEST" are enabled through the Node's p2pServiceNode at the same time.

network/network/src/main/java/bisq/network/p2p/common/RequestResponseHandler.java
```diff
@@ -39,12 +39,12 @@
 @Slf4j
 public abstract class RequestResponseHandler<T extends Request, R extends Response> implements Node.Listener {
     protected final Node node;
-    protected final long timeout;
+    protected final long requestTimeoutMs;
```
mathcoder23: nit: I think renaming it to `requestTimeoutMs` would better fit the context.

network/network/src/main/java/bisq/network/p2p/common/RequestResponseHandler.java
```diff
@@ -149,6 +149,6 @@ protected String createRequestId() {
     }
 
     protected int createNonce() {
-        return new Random().nextInt();
+        return ThreadLocalRandom.current().nextInt();
```
mathcoder23: Suggestion from coderabbitai: replace new Random() on each call to improve performance.

network/network/src/main/java/bisq/network/p2p/services/reporting/ReportRequestService.java
```diff
@@ -18,94 +18,39 @@
...
-                                if (throwable != null) {
-                                    log.error("storageReporting failed for address {} with exception: {}",
-                                            address, throwable.getMessage());
-                                } else {
-                                    log.info("storageReporting successful for address {} with report: {}",
-                                            address, report);
-                                }
-                                requestHandlerMap.remove(connectionId);
-                            });
+                    ReportRequest reportRequest = new ReportRequest(createRequestId());
```
mathcoder23, line 51: Correct the PR description: replace UUID with Nonce (random int value).

network/network/src/main/java/bisq/network/p2p/common/RequestResponseHandler.java
```diff
@@ -82,19 +82,19 @@ protected Optional<R> resolveResponse(EnvelopePayloadMessage message) {
     protected CompletableFuture<R> request(Connection connection, T request) {
         return requestFuturesByConnectionId.compute(connection.getId(), (k, existing) -> {
             if (existing != null && !existing.isDone()) {
-                log.warn("Reusing pending request future for {}", connection.getPeerAddress());
+                log.warn("[{}]Reusing pending request future for {}", this.getClass().getSimpleName(), connection.getPeerAddress());
```
HenrikJannsen: We could remove the `@Slf4j` annotation and instantiate the logger by:

`log = LoggerFactory.getLogger(getClass().getName());`

The the class name of the implementing class is displayed in the logs.

HenrikJannsen: Though not sure if we want that really. In the IDE you can click the class and then jump to the log, but that would not work then... So maybe better leave it as you have it. the `this.` we omit if not needed for brevity.

network/network/src/main/java/bisq/network/p2p/common/LeechRequestResponseHandler.java
```diff
@@ -0,0 +1,35 @@
+package bisq.network.p2p.common;
+
+import bisq.network.identity.NetworkId;
+import bisq.network.p2p.message.EnvelopePayloadMessage;
+import bisq.network.p2p.message.Request;
+import bisq.network.p2p.message.Response;
+import bisq.network.p2p.node.Connection;
+import bisq.network.p2p.node.Node;
+
+public abstract class LeechRequestResponseHandler<T extends Request, R extends Response> extends RequestResponseHandler<T, R> {
```
HenrikJannsen: Maybe a better approach is to split the RequestResponseHandler into the client and server functionality.
Client: send request, listen on response
Server: listen on requests, reply with response

The RequestResponseHandler would then use both (prob. requires delegation as we cannot inherit both, or maybe default functions in the interface are sufficient)

The this class would be just the client and we dont need UnsupportedOperationExceptions.

HenrikJannsen (changes requested):

network/network/src/main/java/bisq/network/p2p/common/RequestResponseHandler.java
mathcoder23: Agreed. Not using `@Slf4j` makes it difficult to locate logs during local debugging. Moreover, when I debug locally, if I don’t log the concrete implementation class, it’s hard to trace the logs. Therefore, I added some subclass name markers for tracking (after refactoring request/response into the delegation pattern, I used request.getClass().getSimpleName() as a log marker).

Thanks for your suggestions regarding code style. I have removed all uses of this.. in the PR Since I’m not yet fully familiar with the project’s code style details, I’ll follow the existing style as a reference. If you have any further style recommendations, I’d be happy to hear them.

network/network/src/main/java/bisq/network/p2p/common/LeechRequestResponseHandler.java
mathcoder23: Hi @HenrikJannsen Sounds good. Since interfaces cannot have member variables (such as `requestFuturesByConnectionId`), I chose to refactor the `Request` and `Response` handling using the `delegation pattern.` I reworked the entire Request and Response processing, adding a `BaseHandler` superclass and a `LifecycleHandler` interface to standardize initialization and shutdown management.

`RequestResponseHandler` remains fully compatible with the previous implementation. I also introduced abstract classes `ResponseHandler` and `RequestHandler` for external extension, as well as two basic delegates — `RequestHandlerDelegate` and `ResponseHandlerDelegate` — to handle the core request/response processing.

If you notice any issues, please let me know — I really appreciate your review.

network/network/src/main/java/bisq/network/p2p/common/ResponseHandlerDelegate.java
```diff
@@ -0,0 +1,111 @@
...
+                        if (throwable instanceof TimeoutException) {
+                            log.warn("[{}] Request to {} timed out after {} ms", request.getClass().getSimpleName(), connection.getPeerAddress(), timeout);
+                        } else if (throwable != null) {
+                            log.warn("[{}] Request to {} failed: {}", request.getClass().getSimpleName(), connection.getPeerAddress(), ExceptionUtil.getRootCauseMessage(throwable));
+                        } else {
+                            log.debug("[{}] Request to {} completed", request.getClass().getSimpleName(), connection.getPeerAddress());
+                        }
+                    });
+            return requestFuture;
+        });
```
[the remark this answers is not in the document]
mathcoder23: I agree with this race condition issue. The requestFuturesByConnectionId.remove(k) comes from an asynchronous callback triggered by the Node listener. Once the compute method returns the requestFuture, the key-level lock is released. If a second thread happens to execute between the moment existing.isDone() == true and the requestFuturesByConnectionId.remove(k) call, the requestFuture created by the second thread can be removed. As a result, requestFuturesByConnectionId.get(connection.getId()) will return null, causing the second thread’s requestFuture to timeout.

network/network/src/main/java/bisq/network/p2p/common/RequestHandlerDelegate.java
```diff
@@ -0,0 +1,84 @@
...
+        R response = callback.createResponse(connection, request);
+        node.sendAsync(response, connection)
+                .whenComplete((result, throwable) -> {
+                    if (throwable != null) {
+                        log.warn("Sending {} to {} failed. {}", StringUtils.truncate(response), connection.getPeerAddress(), ExceptionUtil.getRootCauseMessage(throwable));
+                    } else {
+                        log.info("Sent {} to {}", StringUtils.truncate(response), connection.getPeerAddress());
+                    }
+                });
+    }
```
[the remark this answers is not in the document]
mathcoder23: You are correct. We are not handling exceptions in the Node’s onMessage method. I tested the exception locally, and if we don’t use try/catch, the exception will be swallowed!!!

network/network/src/main/java/bisq/network/p2p/common/ResponseHandlerDelegate.java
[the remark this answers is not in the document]
HenrikJannsen: > I agree with this race condition issue. The requestFuturesByConnectionId.remove(k) comes from an asynchronous callback triggered by the Node listener. Once the compute method returns the requestFuture, the key-level lock is released. If a second thread happens to execute between the moment existing.isDone() == true and the requestFuturesByConnectionId.remove(k) call, the requestFuture created by the second thread can be removed. As a result, requestFuturesByConnectionId.get(connection.getId()) will return null, causing the second thread’s requestFuture to timeout.

I am not that sure about the existing policy to return a pending request in case we would get an additional call on the same connection.
If we would manage the futures not in a map using connectionId but the requestId as key, we could avoid that policy and be more generic. For the use cases in network it mostly does not make sense to send a repeated request before the old one has been received, but it would need some support for order by adding a sequence number to avoid that a response from an older requests overwrites data we got from a more response from more recent state.
With that policy we avoid that so, that simplifies things.... 

Another alternative would be to use computeIfAbsent and return any existing future, also completed (if that tiny race opportunity window).  

Just wanted to throw in those 2 thoughts, not suggesting that we should change it.

network/network/src/main/java/bisq/network/p2p/common/ResponseHandlerDelegate.java
```diff
@@ -0,0 +1,111 @@
...
+            return null; // removes entry
+        });
+    }
+
+    public Map<String, RequestFuture<T, R>> getRequestFuturesByConnectionId(){
+        return requestFuturesByConnectionId;
+    }
+
+    /* --------------------------------------------------------------------- */
+    // LifecycleHandler implementation
```
HenrikJannsen: Small style nit: We usually place the initialize and shutdown after constructor.

network/network/src/main/java/bisq/network/p2p/common/RequestHandler.java
```diff
@@ -0,0 +1,94 @@
...
+        requestHandlerDelegate.shutdown();
+    }
+
+    /* --------------------------------------------------------------------- */
+    // Node.Listener implementation
+    /* --------------------------------------------------------------------- */
+
+    @Override
+    public void onMessage(EnvelopePayloadMessage envelopePayloadMessage, Connection connection, NetworkId networkId) {
+        requestHandlerDelegate.resolveRequest(envelopePayloadMessage).ifPresent(request -> processRequest(connection, request));
```
HenrikJannsen: Small style NIT: We usually place the continuation on streams/optionals in the next line of its not very short.
```
requestHandlerDelegate.resolveRequest(envelopePayloadMessage)
    .ifPresent(request -> processRequest(connection, request));
```

network/network/src/main/java/bisq/network/p2p/common/ResponseHandlerDelegate.java
```diff
@@ -0,0 +1,111 @@
...
+    public void processOnDisconnect(Connection connection, CloseReason closeReason) {
+        requestFuturesByConnectionId.computeIfPresent(connection.getId(), (key, handler) -> {
+            if (!handler.isDone()) {
+                handler.cancel(true);
+            }
+            return null; // removes entry
+        });
+    }
+
+    public Map<String, RequestFuture<T, R>> getRequestFuturesByConnectionId(){
```
HenrikJannsen: We usually use Lombok Getters. But if you want to make it more explicit that its part of the API (what it is), its find to leave it. Motivation for Lombok Getters is primarily to avoid lot of boiler plate by getters/setters but that is here not the case as its only one getter.

network/network/src/main/java/bisq/network/p2p/common/HandlerLifecycle.java
```diff
@@ -17,7 +17,7 @@
 
 package bisq.network.p2p.common;
 
-public interface LifecycleHandler {
+public interface HandlerLifecycle {
```
HenrikJannsen, line 20: Maybe we should move that to common module next to the Service interface. 
I don't have a good name suggestion. In mobile we use `LifecycleAware` but not super happy with that but did not find a better one. If you have some suggestion, please share. 

We could also rename Service to AsyncService and use Service for that one... If we do that lets do it in a dedicated PR with only that refactoring.

HenrikJannsen (approved):
utACK

Some NITs and discussion points...

mathcoder23:
> utACK
> 
> Some NITs and discussion points...

Hi @HenrikJannsen I’ve addressed all your comments, and renamed the PR title to (Refactor RequestResponseHandler to a delegation pattern and apply it in Report), and performed a rebase.

Compared to your last review, I’ve made the following changes:
- Fix code style use Lombok, line breaks for streams/Optionals, place initialize and shutdown after the constructor
- Add TODOs to refactor Service and Lifecycle naming
- Fix map requestFuturesByConnectionId's race condition
- Add try-catch in Node's onMessage method.

HenrikJannsen:
Great thanks @mathcoder23! Excellent work!

Issue #3563, opened 2025-06-07
Show custom popup if trade amount validation fails


blackdelph:
Error message:
The trade failed at your peer: 'Takers (buyers) Bitcoin amount is too high. This can be caused by differences in the 2 traders market price or by an attempt by the taker to manipulate the price.
'

Stack trace: java.lang.IllegalArgumentException
    at bisq.trade.bisq_easy.protocol.messages.BisqEasyTakeOfferRequestHandler.validateAmount(BisqEasyTakeOfferRequestHandler.java:246)
    at bisq.trade.bisq_easy.protocol.messages.BisqEasyTakeOfferRequestHandler.verifyMessage(BisqEasyTakeOfferRequestHandler.java:150)
    at bisq.trade.bisq_easy.protocol.messages.BisqEasyTakeOfferRequestHandler.handle(BisqEasyTakeOfferRequestHandler.java:60)
    at bisq.common.fsm.Fsm.handle(Fsm.java:83)
    at bisq.trade.bisq_easy.protocol.BisqEasyProtocol.handle(BisqEasyProtocol.java:49)
    at bisq.trade.bisq_easy.BisqEasyTradeService.handleBisqEasyTakeOfferMessage(BisqEasyTradeService.java:211)
    at bisq.trade.bisq_easy.BisqEasyTradeService.onMessage(BisqEasyTradeService.java:196)
    at bisq.network.p2p.services.confidential.ConfidentialMessageService.lambda$processConfidentialMessage$10(ConfidentialMessageService.java:430)
    at java.base/java.util.concurrent.CopyOnWriteArrayList.forEach(CopyOnWriteArrayList.java:891)
    at java.base/java.util.concurrent.CopyOnWriteArraySet.forEach(CopyOnWriteArraySet.java:425)
    at bisq.network.p2p.services.confidential.ConfidentialMessageService.lambda$processConfidentialMessage$11(ConfidentialMessageService.java:428)
    at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1804)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
    at java.base/java.lang.Thread.run(Thread.java:1570)

[bisq2-logs.zip](https://github.com/user-attachments/files/20638570/bisq2-logs.zip)
[...]

PR #3808, opened 2025-08-29
Show custom warning popup if trade amount validation fails

54corbin:
Fixes: #3563

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/TradeStateController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/TradeStateController.java
@@ -18,7 +18,6 @@
 package bisq.desktop.main.content.bisq_easy.open_trades.trade_state;
 
 import bisq.account.payment_method.BitcoinPaymentRail;
-import bisq.desktop.navigation.NavigationTarget;
 import bisq.chat.ChatService;
 import bisq.chat.bisq_easy.open_trades.BisqEasyOpenTradeChannel;
 import bisq.chat.bisq_easy.open_trades.BisqEasyOpenTradeChannelService;
@@ -36,6 +35,7 @@ import bisq.desktop.common.view.Navigation;
 import bisq.desktop.components.overlay.Popup;
 import bisq.desktop.main.content.bisq_easy.open_trades.trade_details.TradeDetailsController;
 import bisq.desktop.main.content.bisq_easy.open_trades.trade_state.states.*;
+import bisq.desktop.navigation.NavigationTarget;
 import bisq.i18n.Res;
 import bisq.network.NetworkService;
 import bisq.network.p2p.services.confidential.ack.MessageDeliveryStatus;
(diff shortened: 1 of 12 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/TradeStateController.java
```diff
@@ -137,31 +137,34 @@ public void onActivate() {
...
-                            }
-                        }
+                if (errorMessage == null) {
+                    return;
+                }
+                String key = "errorMessage_" + model.getBisqEasyTrade().get().getId();
+                if (dontShowAgainService.showAgain(key)) {
+                    switch (bisqEasyTrade.getErrorCode()) {
+                        case TRADE_AMOUNT_VALIDATION_FAILED ->
+                                UIThread.run(() -> new Popup().warning(errorMessage).dontShowAgainId(key).show());
```
axpoems, line 147: I would still show this as an error (not warning) as it's done by default, since the trade fails after this.
Also, perhaps in this case, since we know the error, it would be better to use a custom string (which can be translated) explaining the why instead of the error message.

common/src/main/java/bisq/common/fsm/Fsm.java
```diff
@@ -116,23 +116,36 @@ public <E extends Event> void handle(E event) {
...
+        // We apply that only if the event which triggered the exception was not the FsmErrorEvent itself
+        // to avoid potential recursive calls if the error handling code causes a follow-up exception.
+        if (!(event instanceof FsmErrorEvent)) {
+            persist();
+            handle(new FsmErrorEvent(fsmException));
+        }
+        throw fsmException;
+    }
+
+    private ErrorCode determineErrorCode(Exception exception) {
```
HenrikJannsen, line 142: The Fsm must stay domain neutral. It must not contain specific stuff like TradeAmountValidationException or ErrorCode.TRADE_AMOUNT_VALIDATION_FAILED

trade/src/main/java/bisq/trade/Trade.java
```diff
@@ -92,6 +94,10 @@ private static TradeRole createRole(boolean isBuyer, boolean isTaker) {
     // transient fields are excluded by default for EqualsAndHashCode
     private transient final TradeRole tradeRole;
     private final C contract;
+    @Setter
+    private ErrorCode errorCode = ErrorCode.UNSPECIFIED;
```
HenrikJannsen, line 98: If we add it here we should also add it to the protobuf to have it persisted. But I will look into alternative implementations.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/TradeStateController.java
```diff
@@ -137,31 +137,34 @@ public void onActivate() {
                     }));
 
             errorMessagePin = bisqEasyTrade.errorMessageObservable().addObserver(errorMessage -> {
```
HenrikJannsen, line 139: If error code is set on error message change we have dependency on order how those 2 fields are set. I will look into alternative impl. by modelling the error data into an observable object.

HenrikJannsen (changes requested):
NACK

I will provide a suggestion PR for an alternative implementation.

HenrikJannsen:
You just display the error message in a warning popup instead of the generic error popup.
We can simply drop that generic error popup and display the error message.
The `errorMessage = ExceptionUtil.getRootCauseMessage(fsmException);` in BisqEasyFsmErrorEventHandler is already taking the message of the cause of the exception.

So to achieve the same as that PR tried to do is by changing the popup from error type to warn and pass the error message.

HenrikJannsen:
Actually after a second thought, I think its best to improve the generic error popup to be more user friendly by hiding the stacktrace and show it only on demand (button click).

PR #3856, opened 2025-09-09
Sync `rpc.proto` with bisq-musig & add receiver amounts calculation

stejbac:
Bring `rpc.proto` into sync with the latest Rust version, which makes the following changes to it:

1. Change sint32/64 field types back to uint32/64;
2. Rename `PartialSignaturesRequest.receivers` to `redirectionReceivers` and `PublishDepositTxRequest.depositPsbt` to `peersDepositPsbt`;
3. Remove unneeded `SubscribeTxConfirmationStatusRequest.depositPsbt`;
4. Add `tradeFeeReceiver` field to `NonceSharesRequest`;
5. Add `redirectionAmountMsat` field to `NonceSharesMessage`;
6. Add fields to `(NonceShares|PartialSignatures)Message` to create a prepared Claim Tx in addition to the other prepared txs. Previously, it was assumed that Claim spends would always be script spends with a tx generated on the fly. Now there's an additional keyspend path.

Also make corresponding changes to the Java DTOs, in particular exchanging extra Claim Tx-related fields with the peer in the `PartialSignatures` and `NonceShares` messages defined in `trade.proto`. Additionally, provide a dummy trade fee recipient to add to the Deposit Tx with the new `NonceSharesRequest.tradeFeeRecipent` field, and replace all the test/dummy Signet addresses with Regtest addresses that the Rust side now expects.

Finally, add code to `DelayedPayoutTxReceiverService` to compute a filtered list of receiver outputs (addresses + output values) from the provided BM shares, amount of available funds and target fee rate. Small outputs are filtered out under the same criteria currently used in Bisq1. No further capping or sanity checking post-filtering is being done, however, as there is no designated Legacy BM in the oracle data to act as a bin for any surplus funds in the event that caps or sanity checks prevent the adjusted shares from summing to 100%.

--

The changes fix all the soft (missing field, wrong receiver total, wrong network) warnings/errors currently showing up on the Rust side when carrying out a trade, due to changes to the API, which I plan to make into hard errors.

I'm not certain whether its better to keep the receiver amounts calculation on the Java side or move it to the Rust side. There isn't any complicated Bisq-specific logic (involving caps or LBM) in the calculation at present, but moving it to the Rust side may lead to issues with strict cross-platform floating-point determinism (that Java provides automatically) -- I'm not sure.

### DIFF
--- a/trade/src/main/java/bisq/trade/mu_sig/messages/network/handler/NonceSharesRequestUtil.java
+++ b/trade/src/main/java/bisq/trade/mu_sig/messages/network/handler/NonceSharesRequestUtil.java
@@ -18,8 +18,8 @@
 package bisq.trade.mu_sig.messages.network.handler;
 
 import bisq.common.market.Market;
+import bisq.trade.mu_sig.MuSigFeeRateProvider;
 import bisq.trade.mu_sig.MuSigTrade;
-import bisq.trade.mu_sig.MusSigFeeRateProvider;
 import bisq.trade.protobuf.ReceiverAddressAndAmount;
 
 public class NonceSharesRequestUtil {
@@ -35,11 +35,11 @@ public class NonceSharesRequestUtil {
     }
 
     public static long getDepositTxFeeRate() {
-        return MusSigFeeRateProvider.getDepositTxFeeRate();
+        return MuSigFeeRateProvider.getDepositTxFeeRate();
     }
 
     public static long getPreparedTxFeeRate() {
-        return MusSigFeeRateProvider.getPreparedTxFeeRate();
+        return MuSigFeeRateProvider.getPreparedTxFeeRate();
     }
 
     public static long getBuyerSecurityDeposit() {
--- a/trade/src/main/java/bisq/trade/mu_sig/messages/network/handler/NonceSharesRequestUtil.java
+++ b/trade/src/main/java/bisq/trade/mu_sig/messages/network/handler/NonceSharesRequestUtil.java
@@ -20,6 +20,7 @@ package bisq.trade.mu_sig.messages.network.handler;
 import bisq.common.market.Market;
 import bisq.trade.mu_sig.MuSigTrade;
 import bisq.trade.mu_sig.MusSigFeeRateProvider;
+import bisq.trade.protobuf.ReceiverAddressAndAmount;
 
 public class NonceSharesRequestUtil {
     public static long getTradeAmount(MuSigTrade trade) {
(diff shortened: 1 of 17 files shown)


HenrikJannsen:
> as there is no designated Legacy BM in the oracle data to act as a bin

We can just use it hardcoded in Bisq 2 code.

> I'm not certain whether its better to keep the receiver amounts calculation on the Java side or move it to the Rust side. 

No strong opinion on that, but as you stated it might be easier on the java side. Do you see any security risk when its provided from java to rust without rust double checking it? As its based on oracle data, this is likely anyways the weak spot but cannot be resolved without running a bisq 1 node itself.

trade/src/main/proto/rpc.proto
```diff
@@ -59,49 +59,55 @@ message PubKeySharesRequest {
...
+  uint64 sellersSecurityDeposit = 8; // sats
+  ReceiverAddressAndAmount tradeFeeReceiver = 9;
 }
 
 message NonceSharesMessage {
   string warningTxFeeBumpAddress = 1;
   string redirectTxFeeBumpAddress = 2;
+  string claimTxPayoutAddress = 11;
   bytes halfDepositPsbt = 3;
+  uint64 redirectionAmountMsat = 14; // (millisatoshis)
```
HenrikJannsen, line 82: Why is msat and not sat used here?

trade/src/main/java/bisq/trade/mu_sig/messages/network/handler/NonceSharesRequestUtil.java
```diff
@@ -50,4 +51,21 @@ public static long getSellersSecurityDeposit() {
         // TODO get from CollateralOptions
         return 30_000;
     }
+
+    public static ReceiverAddressAndAmount getTradeFeeReceiver() {
+        // NOTE: As implemented right now on the Rust side, the seller pays the entire trade fee (including the small
```
HenrikJannsen, line 56: What is the reason for that? From a user perspective I think we stick with same fee model as in Bisq 1, so both traders pay their fee. As stated below that is achieved by increasing buyers deposit (becomes deposit + fee) and reducing sellers trade amount, so (sellers fee + buyers fee) + deposit+ (trade amount - buyers fee) = sellers fee + deposit + trade amount.

trade/src/main/proto/rpc.proto
stejbac: To reach the target fee rate as closely as possible, millisatoshis need to be used internally in the calculation. If the Redirect Tx input has value `X` sats, the base Redirect Tx (metadata + signed input + 330 sat anchor output) has weight `W` wu and the feerate is `R` sats per kwu, then the funds in sats available to the receivers (including their output weight costs) are

> X - 330 - W * R / 1000

which is fractional in general. That is the quantity `redirectionAmountMsat` passed from the Rust server (divided by 1000). Each receiver output + output-weight-cost added will also be fractional, at least for a general feerate which is a whole number of sats per kwu but not necessarily a whole number of sats per vB.

trade/src/main/java/bisq/trade/mu_sig/messages/network/handler/NonceSharesRequestUtil.java
stejbac: I was mainly just trying to keep things as simple as possible on the Rust side, as there are only four degrees of freedom available to the deposits and payouts (excluding mining fees), namely:
1. Buyer deposit;
2. Seller deposit;
3. Trade fee;
4. How to split the escrow (= buyer-deposit + seller-deposit - trade-fee) into a buyer/seller payout.

The four degrees of freedom are already provided by the buyer/seller security deposit, the trade amount and the (seller) trade fee. A fifth (buyer-trade-fee) DoF could be added to the Musig API, but I think it would be technically redundant, so I was just leaving it the Java side to deal with.

stejbac:
I don't see any obvious security risk to leaving the receiver outputs calculation on the Java side, as the Rust side already checks that they sum to the correct total (which depends on their individual address types and the feerate). I guess an extremely large number of outputs could be added to make the Redirect Tx nonstandard, which I haven't included a check on the Rust side yet for. But the authenticity of the receivers and their shares needs to be trusted by the Rust side anyway.

(I don't think a bug in the amounts calculation is much less likely if implemented in Rust.)

PR #3866, opened 2025-09-15
Make gradle scrips work cross-platform

axpoems:
## Summary by CodeRabbit

- New Features
  - None.

- Bug Fixes
  - Fixed cross-platform build issues by using an OS-aware Gradle wrapper command.
  - Ensured proto sources are packaged only after proto generation when present.

- Chores
  - Aligned proto packaging to use the generated build output location.
  - Standardized OS detection with locale-aware checks and unified wrapper invocation across build tasks.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,12 +5,15 @@ plugins {
     id("bisq.gradle.maven_publisher.LocalMavenPublishPlugin")
 }
 
+fun getGradleCommand(): String {
+    return if (System.getProperty("os.name").lowercase(getDefault()).contains("win")) "gradlew.bat" else "./gradlew"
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -36,6 +39,7 @@ tasks.register("cleanAll") {
     description = "Cleans the entire project."
 
     doLast {
+        val gradleCmd = if (System.getProperty("os.name").lowercase(getDefault()).contains("win")) "gradlew.bat" else "./gradlew"
```
HenrikJannsen: I think there is a OS util available for checking for Windows. Also the repeated code should be extracted in a method.

PR #3896, opened 2025-10-01
Show a placeholder in the offerbook if there are no activities.

### LINKED ISSUE #3861: No placeholder shown if no offers are in a market
Image

KimStrand:
Fixes #3861


KimStrand:
@axpoems Thanks for the feedback! I’ve made the suggested changes and updated the PR.

i18n/src/main/resources/bisq_easy.properties
```diff
@@ -531,6 +531,10 @@ bisqEasy.offerbook.offerList.popup.offersWithInsufficientReputationWarning.messa
 bisqEasy.offerbook.offerList.popup.offersWithInsufficientReputationWarning.buildReputation=Learn how to build up reputation
 bisqEasy.offerbook.offerList.popup.offersWithInsufficientReputationWarning.removeOffers=Remove my invalid offers
 
+bisqEasy.offerbook.messagebox.noMessages.placeholder.title=No offers in {0} market
```
axpoems: I don't think this will work well with translations, thus my suggestion of putting the market name at the end.

KimStrand: understood! I updated the PR.

PR #3897, opened 2025-10-01
Provide & apply commonForkJoinPool in ExecutorFactory.

### LINKED ISSUE #3881: Make usage of ForkJoinPool explicit
When we use async CompletableFuture methods without explicitly setting an executor the ForkJoinPool is used, which can be dangerous if the tasks are long running or blocking and the ForkJoinPool gets exhausted (bounded by number of CPU cores).
We should add a method to the ExecutorFactory to provide the ForkJoinPool and apply that explicitly in the methods, to make it easier to find where we use that and to see if the task might be risky for that executor.

KimStrand:
Resolves #3881 

## Summary by CodeRabbit

- Performance
  - More consistent startup/shutdown and smoother responsiveness across desktop, services, networking, security, trade, and user modules; background and delayed tasks (retries, proofs-of-work, webcam, REST/WebSocket, Tor, I2P, tx confirmations) now run more predictably.
- Refactor
  - Unified asynchronous execution across components for steadier concurrency and timing app‑wide.
- Tests
  - Async tests updated to use the standardized executor for consistent test behavior.

### DIFF
--- a/security/src/main/java/bisq/security/keys/KeyBundleService.java
+++ b/security/src/main/java/bisq/security/keys/KeyBundleService.java
@@ -40,6 +40,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 
+import static bisq.common.threading.ExecutorFactory.commonForkJoinPool;
 import static com.google.common.base.Preconditions.checkArgument;
 
 @Slf4j
@@ -158,7 +159,7 @@ public class KeyBundleService implements PersistenceClient<KeyBundleStore>, Serv
                 && defaultI2pIdentityBase64.isEmpty()
                 && defaultBundle.isPresent()) {
             // Nothing to update, so we return
-            return CompletableFuture.supplyAsync(() -> true);
+            return CompletableFuture.supplyAsync(() -> true, commonForkJoinPool());
         } else {
             return createOrUpdateDefaultBundle(defaultBundle, defaultKeyId)
                     .thenApply(Objects::nonNull);
@@ -245,7 +246,7 @@ public class KeyBundleService implements PersistenceClient<KeyBundleStore>, Serv
             KeyBundle keyBundle = new KeyBundle(defaultKeyId, defaultKeyPair, defaultTorKeyPair, defaultI2pKeyPair);
             persistKeyBundle(defaultKeyId, keyBundle);
             return keyBundle;
-        });
+        }, commonForkJoinPool());
     }
 
     private KeyBundle createKeyBundle(String keyId) {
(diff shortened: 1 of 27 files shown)


security/src/main/java/bisq/security/keys/KeyBundleService.java
```diff
@@ -158,7 +159,7 @@ public CompletableFuture<Boolean> initialize() {
                 && defaultI2pIdentityBase64.isEmpty()
                 && defaultBundle.isPresent()) {
             // Nothing to update, so we return
-            return CompletableFuture.supplyAsync(() -> true);
+            return CompletableFuture.supplyAsync(() -> true, commonForkJoinPool());
```
HenrikJannsen, line 162: Maybe we should pass a `MoreExecutors.directExecutor()` in those cases. We could add the directExecutor as well to the Executor Factory to have all centralized there.

HenrikJannsen: But can be done in a new PR

security/src/main/java/bisq/security/pow/ProofOfWorkService.java
```diff
@@ -35,7 +37,7 @@ public ProofOfWorkService() {
     public CompletableFuture<ProofOfWork> mintAsync(byte[] payload,
                                                     byte[] challenge,
                                                     double difficulty) {
-        return CompletableFuture.supplyAsync(() -> mint(payload, challenge, difficulty));
+        return CompletableFuture.supplyAsync(() -> mint(payload, challenge, difficulty), commonForkJoinPool());
```
HenrikJannsen, line 40: I guess here we should use a dedicated thread, but can be discussed in a new issue/PR.

user/src/main/java/bisq/user/identity/UserIdentityStore.java
```diff
@@ -225,7 +230,7 @@ CompletableFuture<EncryptedData> encrypt() {
             EncryptedData encryptedData = encryptPlainTextProto(getPlainTextBuilder().build());
             log.info("encrypt took {} ms", System.currentTimeMillis() - ts);
             return encryptedData;
-        }).whenComplete((encrypted, throwable) -> {
+        }, commonForkJoinPool()).whenComplete((encrypted, throwable) -> {
```
HenrikJannsen, line 233: NIT: better break line at whenComplete

HenrikJannsen:
Searched for all usages of async methods of CompletableFuture and could not find any case where it was missing. Well done!

I will create a new issue for potential changes to a dedicated executor where it makes sense.

PR #3914, opened 2025-10-07
Use dedicated executors instead of common pool where needed.

### LINKED ISSUE #3898: Use dedicated executors instead of ForkJoinPool.commonPool() where appropriate
Based on https://github.com/bisq-network/bisq2/pull/3897 we can now check all cases where `ForkJoinPool.commonPool()` is used and consider to use a dedicated executor, pool or `directExecutor` where appropriate.

`directExecutor` should be used in all cases where we just follow the API like in initialize and shutdown methods where no async work is done, to avoid unnecessary overhead.

For other cases the dev working on that should evaluate which strategy is best. We can then further discuss in the the PR the various cases. There might be often not a clear decisions. In doubt leaving it with `ForkJoinPool.commonPool()` would carry the least risk as that is the status quo.

KimStrand:
Resolves #3898 

## Summary by CodeRabbit

- New Features
  - None.

- Bug Fixes
  - Improved reliability of request retries and subscription recovery.
  - Safer shutdown sequencing to avoid lingering background tasks.

- Refactor
  - Unified delayed-execution behavior across services for consistent scheduling.
  - Replaced some shared thread pools with dedicated executors in trade flows.
  - Simplified asynchronous paths for lower overhead and quicker responsiveness.
  - Certain shutdown paths now complete immediately for faster app exit.

- Chores
  - Minor cleanups and formatting tweaks.

### DIFF
--- a/apps/oracle-node-app/src/main/java/bisq/oracle_node/bisq1_bridge/grpc/services/BridgeSubscriptionGrpcService.java
+++ b/apps/oracle-node-app/src/main/java/bisq/oracle_node/bisq1_bridge/grpc/services/BridgeSubscriptionGrpcService.java
@@ -18,6 +18,7 @@
 package bisq.oracle_node.bisq1_bridge.grpc.services;
 
 import bisq.common.application.Service;
+import bisq.common.timer.Delay;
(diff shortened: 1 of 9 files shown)


HenrikJannsen (approved):
utACK

PR #4008, opened 2025-10-27
feat: implement auth for http api

wodoro:
I refactored common config into an abstract class to pull the complexity down to BaseRestApiResourceConfig

Adds a "password" config to websocket and rest api configs

## Summary by CodeRabbit

* **New Features**
  * Optional password-based HMAC authentication for HTTP and WebSocket endpoints; unauthenticated requests are rejected.
  * WebSocket proxy forwards auth headers to the REST backend.
  * New configurable password field for REST and WebSocket servers.

* **Refactor**
  * Unified shared API configuration for REST and WebSocket including password.
  * Simplified wiring for request validation/authentication and optional server-side auth addon registration.

* **Bug Fixes**
  * Expanded allowed characters in request paths/queries (improved I2P compatibility).

* **Tests**
  * Added normalization tests for path/query handling.

### DIFF
--- a/http-api/src/main/java/bisq/http_api/auth/HttpApiAuthFilter.java
+++ b/http-api/src/main/java/bisq/http_api/auth/HttpApiAuthFilter.java
@@ -38,9 +38,14 @@ public class HttpApiAuthFilter implements ContainerRequestFilter {
         String timestamp = ctx.getHeaderString(AuthUtils.AUTH_TIMESTAMP_HEADER);
         String receivedHmac = ctx.getHeaderString(AuthUtils.AUTH_HEADER);
         String nonce = ctx.getHeaderString(AuthUtils.AUTH_NONCE_HEADER);
-        String bodySha256Hex = getBodySha256Hex(ctx);
-        if (!AuthUtils.isValidAuthentication(secretKey, ctx.getMethod(), normalizedPathAndQuery, nonce, timestamp, receivedHmac, bodySha256Hex)) {
-            log.warn("HttpRequest rejected: Invalid or missing authorization token");
+        try {
+            String bodySha256Hex = getBodySha256Hex(ctx);
+            if (!AuthUtils.isValidAuthentication(secretKey, ctx.getMethod(), normalizedPathAndQuery, nonce, timestamp, receivedHmac, bodySha256Hex)) {
+                log.warn("HttpRequest rejected: Invalid or missing authorization token");
+                ctx.abortWith(Response.status(Response.Status.UNAUTHORIZED).build());
+            }
+        } catch (Exception e) {
+            log.error("Error while reading body of request for authentication", e);
             ctx.abortWith(Response.status(Response.Status.UNAUTHORIZED).build());
         }
     }
@@ -56,8 +61,7 @@ public class HttpApiAuthFilter implements ContainerRequestFilter {
 
         // If length is known and too large, return early
         if (declaredLength > MAX_BODY_SIZE_BYTES) {
-            log.warn("Request body too large: {} bytes, max: {}. This will result in auth failure", declaredLength, MAX_BODY_SIZE_BYTES);
-            return null;
+            throw new RuntimeException("Request body exceeds maximum allowed size of " + MAX_BODY_SIZE_BYTES + " bytes");
         }
 
         try {
@@ -72,9 +76,7 @@ public class HttpApiAuthFilter implements ContainerRequestFilter {
                 while ((read = entityStream.read(chunk)) != -1) {
                     total += read;
                     if (total > MAX_BODY_SIZE_BYTES) {
-                        log.warn("Request body too large: {} bytes, max: {}. This will result in auth failure", total, MAX_BODY_SIZE_BYTES);
-                        ctx.setEntityStream(new ByteArrayInputStream(new byte[0])); // replace stream to not leave it in half read state
-                        return null;
+                        throw new RuntimeException("Request body exceeds maximum allowed size of " + MAX_BODY_SIZE_BYTES + " bytes");
                     }
                     buffer.write(chunk, 0, read);
                 }
@@ -89,8 +91,7 @@ public class HttpApiAuthFilter implements ContainerRequestFilter {
                 return Hex.encode(DigestUtil.sha256(bytes));
             }
         } catch (Exception e) {
-            log.error("Failed to read request body for authentication. This will result in auth failure", e);
-            return null;
+            throw new RuntimeException("Failed to read request body for authentication. This will result in auth failure", e);
         }
     }
 }


http-api/src/main/java/bisq/http_api/auth/HttpApiAuthFilter.java
```diff
@@ -0,0 +1,47 @@
...
+import jakarta.ws.rs.ext.Provider;
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+import java.security.MessageDigest;
+
+@Provider
+@Priority(Priorities.AUTHENTICATION)
+@Slf4j
+public class HttpApiAuthFilter implements ContainerRequestFilter {
```
HenrikJannsen, line 24: Nit: Maybe use same name pattern for HttpApiAuthFilter and WebSocketAuthenticationFilter

apps/http-api-app/src/main/resources/http_api_app.conf
```diff
@@ -44,6 +45,7 @@ application {
         whiteListEndPoints = []
         blackListEndPoints = []
         supportedAuth = []   // supported auth schemes. If empty no authentication is required
+        password = ""
```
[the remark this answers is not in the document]
wodoro: This is intended

http-api/src/main/java/bisq/http_api/config/CommonApiConfig.java
```diff
@@ -0,0 +1,45 @@
...
+        this.host = host;
+        this.port = port;
+        this.localhostOnly = localhostOnly;
+        this.whiteListEndPoints = whiteListEndPoints;
+        this.blackListEndPoints = blackListEndPoints;
+        this.supportedAuth = supportedAuth;
+        this.password = password;
+
+        restApiBaseAddress = protocol + host + ":" + port;
+        restApiBaseUrl = restApiBaseAddress + REST_API_BASE_PATH;
```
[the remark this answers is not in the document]
wodoro: protocol contains :// already in config file

http-api/src/main/java/bisq/http_api/auth/AuthUtils.java
```diff
@@ -1,9 +1,64 @@
...
+
+    public static final String AUTH_HEADER = "AUTH-TOKEN";
+    public static final String AUTH_TIMESTAMP_HEADER = "AUTH-TS";
+    public static final String AUTH_HMAC_ALGORITHM = "HmacSHA256";
+    // timestamp validity is long to allow Tor requests to succeed
+    // but also increases the window of vulnerability for replay attacks
+    public static final long AUTH_TIMESTAMP_VALIDITY_MS = 45_000;
+
+    public static boolean isValidAuthentication(String password,
+                                                @Nullable String timestamp,
```
HenrikJannsen, line 70: Is it a valid case that the timestamp and receivedHmac are null? If not I would remove the annotation, as we use that only when its an expected/supported case.

http-api/src/main/java/bisq/http_api/auth/AuthConstants.java
```diff
@@ -1,9 +1,64 @@
...
+import javax.annotation.Nullable;
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+import java.nio.charset.StandardCharsets;
+import java.security.InvalidKeyException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+@Slf4j
 public final class AuthConstants {
```
HenrikJannsen: Maybe rename as the class has grown from scope.

http-api/src/main/java/bisq/http_api/rest_api/BaseRestApiResourceConfig.java
```diff
@@ -20,8 +20,10 @@ public BaseRestApiResourceConfig(CommonApiConfig config) {
                 .register(RestApiException.Mapper.class)
                 .register(mapper);
 
-        if (!config.getPassword().isEmpty()) {
-            HttpApiAuthFilter httpApiAuthFilter = HttpApiAuthFilter.from(config);
+
+        String password = config.getPassword();
+        if (password != null && !password.isEmpty()) {
```
HenrikJannsen: We have a StringUtils.isNotEmpty() method for that

http-api/src/main/java/bisq/http_api/auth/AuthUtils.java
wodoro: yes, they both can be null when header is missing

http-api/src/main/java/bisq/http_api/rest_api/BaseRestApiResourceConfig.java
wodoro: StringUtils.isNotEmpty should be renamed to StringUtils.isNotBlank, as it's not the same

http-api/src/main/java/bisq/http_api/web_socket/rest_api_proxy/WebSocketRestApiService.java
```diff
@@ -93,6 +97,10 @@ private WebSocketRestApiResponse sendToRestApiServer(WebSocketRestApiRequest req
                 .header("Content-Type", "application/json")
                 .header("Accept", "application/json")
                 .method(method, HttpRequest.BodyPublishers.ofString(body));
+        if (authToken != null && authTs != null) {
+            requestBuilder.header(AuthConstants.AUTH_HEADER, authToken);
+            requestBuilder.header(AuthConstants.AUTH_TIMESTAMP_HEADER, authTs);
+        }
         try {
             HttpRequest httpRequest = requestBuilder.build();
             log.info("Send {} httpRequest to {}. httpRequest={} ", method, url, httpRequest);
```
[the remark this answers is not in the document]
wodoro: I'll remove the httpRequest object from it

http-api/src/main/java/bisq/http_api/web_socket/WebSocketService.java
```diff
@@ -147,7 +126,8 @@ public WebSocketService(Config config,
                 userService,
                 bisqEasyService,
                 openTradeItemsService);
-        webSocketRestApiService = new WebSocketRestApiService(objectMapper, restApiBaseAddress, requestValidator);
+        WebSocketRequestValidator requestValidator = WebSocketRequestValidator.from(config);
+        webSocketRestApiService = new WebSocketRestApiService(objectMapper, config.getRestApiBaseAddress(), requestValidator);
         webSocketConnectionHandler = new WebSocketConnectionHandler(subscriptionService, webSocketRestApiService);
```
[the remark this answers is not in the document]
wodoro: done

http-api/src/main/java/bisq/http_api/web_socket/WebSocketService.java
```diff
@@ -101,26 +85,23 @@ public static Config from(com.typesafe.config.Config config) {
                     config.getBoolean("localhostOnly"),
                     config.getStringList("whiteListEndPoints"),
                     config.getStringList("blackListEndPoints"),
-                    config.getStringList("supportedAuth")
+                    config.getStringList("supportedAuth"),
+                    config.getString("password")
             );
```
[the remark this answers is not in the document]
wodoro: password always exists in default config as empty string,
the same can be said for the rest of config as well
this is fine

http-api/src/main/java/bisq/http_api/auth/AuthUtils.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public static boolean isValidAuthentication(SecretKey secretKey,
+                                                @Nullable String timestamp,
+                                                @Nullable String receivedHmac,
+                                                String method,
+                                                String encodedPathAndQuery) {
+        if (timestamp == null || receivedHmac == null) {
+            return false;
+        }
```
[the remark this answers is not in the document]
wodoro: method and encodedPathAndQuery are never null, and not expected to be null

http-api/src/main/java/bisq/http_api/auth/HttpApiAuthFilter.java
```diff
@@ -0,0 +1,74 @@
...
+            ctx.setEntityStream(new ByteArrayInputStream(bytes));
+        } catch (Exception e) {
+            try {
+                entityStream.close();
+            } catch (Exception ex) {
+                // no need
+            }
+        }
+        return bytes;
+    }
```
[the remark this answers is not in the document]
wodoro: authentication cannot succeed when there's a read error, because if the client has already used body for auth token and body cannot be read, it will result in a failed auth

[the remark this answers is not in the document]
wodoro: isn't it recommended to close entityStream on IoException ?

wodoro:
@HenrikJannsen I think this is good to merge now

wodoro:
will create a PR for mobile branch after,
please squash when merging, so that it can be easier to apply changes to the other branch

http-api/src/main/java/bisq/http_api/auth/WebSocketAuthFilter.java
```diff
@@ -29,7 +30,10 @@ public NextAction handleRead(FilterChainContext ctx) {
             if ("websocket".equalsIgnoreCase(upgradeHeader)) {
                 String timestamp = request.getHeader(AuthUtils.AUTH_TIMESTAMP_HEADER);
                 String receivedHmac = request.getHeader(AuthUtils.AUTH_HEADER);
-                if (!AuthUtils.isValidAuthentication(secretKey, timestamp, receivedHmac)) {
+                String method = "GET";
```
HenrikJannsen: Why not use `request.getMethod().getMethodString()`

http-api/src/main/java/bisq/http_api/auth/WebSocketAuthFilter.java
```diff
@@ -29,7 +30,10 @@ public NextAction handleRead(FilterChainContext ctx) {
...
+                String normalizedPathAndQuery = AuthUtils.normalizePathAndQuery(URI.create(request.getRequestURI()));
+
+                if (!AuthUtils.isValidAuthentication(secretKey, timestamp, receivedHmac, method, normalizedPathAndQuery)) {
```
HenrikJannsen: param name in `isValidAuthentication` is `encodedPathAndQuery`, maybe harmonize the 2 diff. names (`normalizedPathAndQuery`, `encodedPathAndQuery`)

http-api/src/main/java/bisq/http_api/auth/HttpApiAuthFilter.java
```diff
@@ -24,13 +31,44 @@ public HttpApiAuthFilter(String password) {
...
         URI requestUri = ctx.getUriInfo().getRequestUri();
         String normalizedPathAndQuery = AuthUtils.normalizePathAndQuery(requestUri);
-        if (!AuthUtils.isValidAuthentication(secretKey, timestamp, receivedHmac, ctx.getMethod(), normalizedPathAndQuery)) {
+        String timestamp = ctx.getHeaderString(AuthUtils.AUTH_TIMESTAMP_HEADER);
+        String receivedHmac = ctx.getHeaderString(AuthUtils.AUTH_HEADER);
+        String bodySha256Hex = null;
+        byte[] bodyBytes = getBody(ctx);
+        if (bodyBytes != null && bodyBytes.length > 0) {
+            try {
+                bodySha256Hex = Hex.encode(MessageDigest.getInstance("SHA-256").digest(bodyBytes));
```
HenrikJannsen: You can use `DigestUtil.sha256`

http-api/src/main/java/bisq/http_api/auth/HttpApiAuthFilter.java
```diff
@@ -24,13 +31,44 @@ public HttpApiAuthFilter(String password) {
...
+
+    @Nullable
+    private byte[] getBody(ContainerRequestContext ctx) {
+        byte[] bytes = null;
+        if (ctx.getLength() <= 0) {
+            return bytes;
+        }
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        InputStream entityStream = ctx.getEntityStream();
+        try {
```
HenrikJannsen, line 67: Can use try with resources

http-api/src/main/java/bisq/http_api/auth/HttpApiAuthFilter.java
```diff
@@ -24,13 +31,44 @@ public HttpApiAuthFilter(String password) {
...
+            entityStream.transferTo(baos);
+            bytes = baos.toByteArray();
+            ctx.setEntityStream(new ByteArrayInputStream(bytes));
+        } catch (Exception e) {
+            try {
+                entityStream.close();
+            } catch (Exception ex) {
+                // no need
```
HenrikJannsen: We usually call the param (`ex`) then `ignore`, then IDE does not complain about unused param

http-api/src/main/java/bisq/http_api/auth/HttpApiAuthFilter.java
```diff
@@ -52,23 +52,19 @@ public void filter(ContainerRequestContext ctx) throws IOException {
...
             ctx.setEntityStream(new ByteArrayInputStream(bytes));
+            return bytes;
         } catch (Exception e) {
-            try {
-                entityStream.close();
-            } catch (Exception ex) {
-                // no need
-            }
+            log.error("Failed to read request body for authentication, will result in auth failure", e);
+            return null;
```
HenrikJannsen: Using null will exclude the body, not causing an auth failure. I guess we should throw in that method and use that in the caller for failing auth.

HenrikJannsen (changes requested):

http-api/src/main/java/bisq/http_api/auth/HttpApiAuthFilter.java
wodoro: it will cause an auth failure, if there's a body and we fail to read it, as client has already included it in its token, the token will differ, causing the auth to fail

http-api/src/main/java/bisq/http_api/auth/AuthUtils.java
```diff
@@ -22,10 +23,24 @@ private AuthUtils() {
 
     public static final String AUTH_HEADER = "AUTH-TOKEN";
     public static final String AUTH_TIMESTAMP_HEADER = "AUTH-TS";
+    public static final String AUTH_NONCE_HEADER = "AUTH-NONCE";
     private static final String AUTH_HMAC_ALGORITHM = "HmacSHA256";
     // timestamp validity is long to allow Tor requests to succeed
     // but also increases the window of vulnerability for replay attacks
-    private static final long AUTH_TIMESTAMP_VALIDITY_MS = 45_000;
+    private static final long AUTH_TIMESTAMP_VALIDITY_MS = 3_600_000; // 1 hour, to account for out of sync clocks
```
HenrikJannsen: TimeUnit.HOURS.toMillis(1) makes it better readable

http-api/src/main/java/bisq/http_api/auth/HttpApiAuthFilter.java
```diff
@@ -34,28 +33,23 @@ public void filter(ContainerRequestContext ctx) throws IOException {
...
+        String bodySha256Hex = getBodySha256Hex(ctx);
         if (!AuthUtils.isValidAuthentication(secretKey, ctx.getMethod(), normalizedPathAndQuery, timestamp, receivedHmac, bodySha256Hex)) {
             log.warn("HttpRequest rejected: Invalid or missing authorization token");
             ctx.abortWith(Response.status(Response.Status.UNAUTHORIZED).build());
         }
     }
 
     @Nullable
-    private byte[] getBody(ContainerRequestContext ctx) {
+    private String getBodySha256Hex(ContainerRequestContext ctx) {
```
HenrikJannsen, line 54: Instead of returning null and relying that the caller interpret that correctly, I think it would be more clear if we throw exceptions for all invalid cases.

This would also justify better that the do not apply the 
`ctx.setEntityStream(new ByteArrayInputStream(bytes));` on early returns. Currently when hitting the max size the original stream would stay half consumed.

wodoro: But not having a body is a valid case, so it still needs to return null on some cases

I would like to not throw here, as I actually want the server to return with 401, not with 500 or anything else

wodoro: actually I just imagined a case where this is exploited
will update accordingly

HenrikJannsen (approved):
utACK

PR #4042, opened 2025-11-03
fix(auth): prevent indefinite read on normal conditions

wodoro:
Sorry, I should have tested the latest change

So the problem is that when the body is empty, the read never returns, blocking the function

Client can spoof the content-length but thats an important indicator for us to know if body is present, so I'm not sure if we can work around that

## Summary by CodeRabbit

* **Refactor**
  * Optimized HTTP request handling for improved performance in authentication processing.


HenrikJannsen (approved):
utACK

PR #4059, opened 2025-11-05
Implement triggerUserActivityDetection endpoint

salottodev:
## Summary by CodeRabbit

* **New Features**
  * Added a new endpoint to trigger user activity detection. This enhancement enables applications to signal when user activity is detected, allowing the system to respond appropriately by updating user profile information and maintaining accurate user data records.

### DIFF
--- a/http-api/src/main/java/bisq/http_api/HttpApiService.java
+++ b/http-api/src/main/java/bisq/http_api/HttpApiService.java
@@ -90,8 +90,10 @@ public class HttpApiService implements Service {
             MarketPriceRestApi marketPriceRestApi = new MarketPriceRestApi(bondedRolesService.getMarketPriceService());
             SettingsRestApi settingsRestApi = new SettingsRestApi(settingsService);
             PaymentAccountsRestApi paymentAccountsRestApi = new PaymentAccountsRestApi(accountService);
-            UserProfileRestApi userProfileRestApi = new UserProfileRestApi(userService.getUserProfileService(),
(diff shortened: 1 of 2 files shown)


HenrikJannsen:
@salottodev In Bisq2 we trigger userActivityDetected by each mouse movement and keyboard event, but for mobile client that would too much of course. We have a 5 min. pause between user profile refresh calls, thus that can be used as a throttle when sending the API call.

salottodev:
@HenrikJannsen we are triggering userActivityDetected by each new screen navigation. I have just added the throttling logic. https://github.com/bisq-network/bisq-mobile/pull/901

rodvar:
a 5 min user activity window doesn't make any sense in mobile, user usage spans are low even when trading, and they can jump in and out to other apps anytime.

Let's accept it for now but we should consider changing that window in http-api when its used for mobile? what do you think @HenrikJannsen

HenrikJannsen:
I can be less then 5 min or at each getting into foreground. Its' just if the mobile app sends more requests in 5 min, it will not have effect on refreshing the user profile and by that updating the activity state in the network.

rodvar:
> I can be less then 5 min or at each getting into foreground. Its' just if the mobile app sends more requests in 5 min, it will not have effect on refreshing the user profile and by that updating the activity state in the network.

we are on the same page on that - its merged. What I'm asking is if we can configure the core to update more frequently when its running in mobile..?

PR #4068, opened 2025-11-08
Add deleted chats indicator

### LINKED ISSUE #1995: Add indicator to chat list view about expired messages

iMac7:
resolves #1995 

indicator added to public channels chats/discussions, chats/support

image

## Summary by CodeRabbit

* **New Features**
  * Added a Deleted Chats indicator in public chat channels with system-style text and a close icon.
  * Indicator can be dismissed just once or permanently (persistence supported); dismissal shows a confirmation.

* **Bug Fixes / Behavior**
  * Message rendering and validation now correctly handle Deleted Chats indicators so they display reliably and don’t trigger date validation errors.


HenrikJannsen:
@axpoems Can you review?

axpoems:
Thanks. I have made a new PR based on your changes. See #4084.

PR #4079, opened 2025-11-19
bugfix/for-mobile-based-on-2.1.7_guards_npe_fixes

rodvar:
- guards NPE fixes making the http-api server crash randomly
 - found whilst testing for first Bisq Connect release on a real prod environment

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened peer management services with comprehensive null-safety checks to prevent runtime errors when processing peer address information across ban lists and peer selection flows.

### DIFF
--- a/network/network/src/main/java/bisq/network/p2p/services/peer_group/BanList.java
+++ b/network/network/src/main/java/bisq/network/p2p/services/peer_group/BanList.java
@@ -55,6 +55,9 @@ public class BanList {
     }
 
     public boolean isBanned(Address address) {
+        if (address == null) {
+            return false;
+        }
         return entryMap.containsKey(address);
     }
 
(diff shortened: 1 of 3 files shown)


rodvar:
Note for myself after discussion with reviewer:

`with transport type changes issues are expected as that is not a supported use case in desktop. but i think i know where we need to guard against that. its when accessing the addressByTransportType map. in main thats already improved as i run into that with i2p where we need to support now switching transports` -> I'll try to repro the issue again without the guards to make sure / confirm if we need this or any other change to make transport changes not cause this problems

HenrikJannsen:
As discussed on Matrix, I prefer to wait for info what exactly caused the NP. Will open a new PR once we know more.

Issue #4142, opened 2025-12-08
Add support for Bisq Connect mobile app


HenrikJannsen:
As discussed at https://github.com/bisq-network/bisq-mobile/discussions/976 we should add in Bisq 2 support for acting as backend for a Bisq Connect mobile client. 

- Add a toggle in Settings to enable/disable Websockets
- Add password protection
- Show QR code with onion address:port or LAN address:port
- If Lan address is used, add util to detect the LAN address
- Allow to enter text input for address and port for WAN IP addresses
- Add warning when using WAN address
- Add info about connection status of a remote client and client info (version)
- Add info that multiple profiles are not supported on mobile, thus should not be used on the desktop either
- If user has enabled WS, show warning when user creates a new profile.
- When enabling WS when activating and user has multiple profiles

We need to test out which risks/issues arise when backend has multiple profiles. If there are critical issues we should not allow it. Otherwise we can use a warning.

wodoro:
I think we should avoid letting user enable websocket server on WAN altogether, as we also only allow connecting to WAN using HTTPS on client, which is not possible without using a reverse proxy
so LAN or TOR option should suffice

PR #4189, opened 2025-12-19
feat: Allow desktop to start WebSocket API through settings

wodoro:
toward #4142 and [#977](https://github.com/bisq-network/bisq-mobile/issues/977)
Implements settings tab for starting websocket server on desktop client
Haven't implemented warning or checks for profiles yet, wanted to get this out first to get a feedback before continuing

image

## Summary by CodeRabbit

* **New Features**
  * Bisq Connect settings section in desktop Settings: LAN/TOR exposure modes, enable toggle, password, save/restart flow
  * Live API URL preview, QR code generation, and connected clients table with real-time WebSocket client monitoring

* **Configuration**
  * New publishOnionService boolean to control Tor onion publishing for REST and WebSocket services (added across apps)

* **Localization**
  * Added Bisq Connect UI strings and expanded tooltips in i18n resources

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
@@ -139,6 +139,7 @@ public class BisqConnectController implements Controller {
                     .useShutDownButton()
                     .show();
         }
+        updatePreview();
     }
 
     private void loadFromConfig() {
@@ -184,7 +185,9 @@ public class BisqConnectController implements Controller {
                 model.getApiUrl().set("");
                 model.getQrCodeImage().set(null);
 
-                if (!model.getWebsocketRunning().get() && !initialEnabled) {
+                if (model.getEnabled().get() && !initialEnabled) {
+                    model.getQrPlaceholder().set(Res.get("settings.bisqConnect.connectionDetails.placeholderRestartRequired"));
+                } else if (!initialEnabled) {
                     model.getQrPlaceholder().set(Res.get("settings.bisqConnect.connectionDetails.placeholderDisabled"));
                 } else {
                     if (initialExposureMode == BisqConnectExposureMode.LAN) {
@@ -250,7 +253,6 @@ public class BisqConnectController implements Controller {
         boolean changeDetected = model.getEnabled().get() != savedEnabled
                 || model.getSelectedMode().get() != savedMode
                 || !Objects.equals(model.getPassword().get(), savedPassword);
-        boolean enabled = changeDetected && model.getPasswordIsValid().get();
-        model.getIsChangeDetected().set(enabled);
+        model.getIsChangeDetected().set(changeDetected);
     }
 }
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectModel.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectModel.java
@@ -20,6 +20,7 @@ package bisq.desktop.main.content.settings.bisq_connect;
 import bisq.common.network.Address;
 import bisq.desktop.common.view.Model;
 import bisq.desktop.components.controls.validator.RequiredFieldValidator;
+import bisq.desktop.components.controls.validator.TextMinLengthValidator;
 import bisq.desktop.components.controls.validator.ValidatorBase;
 import bisq.http_api.web_socket.BisqConnectClientInfo;
 import bisq.i18n.Res;
@@ -51,6 +52,7 @@ public class BisqConnectModel implements Model {
     private final StringProperty password = new SimpleStringProperty("");
     private final BooleanProperty passwordIsValid = new SimpleBooleanProperty();
     private final ValidatorBase passwordRequiredValidator = new RequiredFieldValidator(Res.get("validation.empty"));
+    private final ValidatorBase passwordMinimumLengthValidator = new TextMinLengthValidator(Res.get("validation.password.tooShort"), 8);
 
     private final ObjectProperty<Image> qrCodeImage = new SimpleObjectProperty<>();
     private final StringProperty qrPlaceholder = new SimpleStringProperty("");
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectView.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectView.java
@@ -69,12 +69,10 @@ public class BisqConnectView extends View<VBox, BisqConnectModel, BisqConnectCon
         VBox.setMargin(description, new Insets(5, 0, 10, 0));
 
         enableSwitch = new Switch(Res.get("settings.bisqConnect.enable"));
-        passwordField = new MaterialPasswordField(
-                Res.get("settings.bisqConnect.password.label"),
-                ""
-        );
+        passwordField = new MaterialPasswordField(Res.get("settings.bisqConnect.password.label"));
         passwordField.setHelpText(null);
         passwordField.setPrefWidth(SettingsViewUtils.TEXT_FIELD_WIDTH);
+        passwordField.setValidators(model.getPasswordRequiredValidator(), model.getPasswordMinimumLengthValidator());
         saveChangesButton = new Button(Res.get("settings.bisqConnect.saveChanges"));
         saveChangesButton.setDefaultButton(true);
 
@@ -138,13 +136,13 @@ public class BisqConnectView extends View<VBox, BisqConnectModel, BisqConnectCon
         BisqTableColumn<BisqConnectClientInfo> indexCol = IndexColumnUtil.getIndexColumn(sortedClients);
         BisqTableColumn<BisqConnectClientInfo> ipCol = new BisqTableColumn.Builder<BisqConnectClientInfo>()
                 .title(Res.get("settings.bisqConnect.clients.ip"))
-                .valueSupplier(info -> info.getAddress().orElse("-"))
+                .valueSupplier(info -> info.getAddress().orElse(Res.get("data.na")))
                 .left()
                 .minWidth(200)
                 .build();
         BisqTableColumn<BisqConnectClientInfo> uaCol = new BisqTableColumn.Builder<BisqConnectClientInfo>()
                 .title(Res.get("settings.bisqConnect.clients.userAgent"))
-                .valueSupplier(info -> info.getUserAgent().orElse("-"))
+                .valueSupplier(info -> info.getUserAgent().orElse(Res.get("data.na")))
                 .left()
                 .minWidth(300)
                 .build();
@@ -170,12 +168,12 @@ public class BisqConnectView extends View<VBox, BisqConnectModel, BisqConnectCon
         });
 
         passwordField.setText(model.getPassword().get());
-        passwordField.setValidator(model.getPasswordRequiredValidator());
         passwordField.isValidProperty().bindBidirectional(model.getPasswordIsValid());
         passwordSubscription = EasyBind.subscribe(passwordField.textProperty(), newVal -> {
             passwordField.validate();
             controller.onPasswordChanged(newVal);
         });
+        passwordField.resetValidation(); // to reset error message set by current value of the password
 
         qrPlaceholder.textProperty().bind(model.getQrPlaceholder());
 
@@ -244,5 +242,4 @@ public class BisqConnectView extends View<VBox, BisqConnectModel, BisqConnectCon
         saveChangesButton.disableProperty().unbind();
         saveChangesButton.setOnAction(null);
     }
-
 }
--- a/i18n/src/main/resources/settings.properties
+++ b/i18n/src/main/resources/settings.properties
@@ -113,6 +113,7 @@ settings.bisqConnect.saveChanges=Save Changes
 settings.bisqConnect.url.label=API address
 settings.bisqConnect.connectionDetails.headline=Connection details
 settings.bisqConnect.connectionDetails.unexpectedErrorHint=Check logs for more details.
+settings.bisqConnect.connectionDetails.placeholderRestartRequired=Restart required to start WebSocket API
 settings.bisqConnect.connectionDetails.placeholderDisabled=WebSocket API is disabled
 settings.bisqConnect.connectionDetails.placeholderLan.error=Cannot determine Lan address. Make sure you are connected to a network.
 settings.bisqConnect.connectionDetails.placeholderLan=Waiting for the Lan address
--- a/apps/desktop/desktop/build.gradle.kts
+++ b/apps/desktop/desktop/build.gradle.kts
@@ -46,7 +46,7 @@ dependencies {
     implementation("bisq:application")
     implementation("bisq:evolution")
     implementation("bisq:wallet")
-    implementation("bisq:http_api")
+    implementation("bisq:http-api")
 
     implementation("network:network")
     implementation("network:network-identity")


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
```diff
@@ -0,0 +1,252 @@
...
+                clientsPin.unbind();
+                websocketInitPin = null;
+                websocketErrorPin = null;
+                websocketAddressPin = null;
+                clientsPin = null;
+            }
+        } catch (Exception t) {
+            log.debug("Error removing websocket listeners", t);
+        }
+    }
```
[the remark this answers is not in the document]
HenrikJannsen: We do not do excessive null checks. All pins are expected to be not null at onDeactivate. Only if a pin is optionally set we check for null.

http-api/src/main/java/bisq/http_api/web_socket/RemoteControlClient.java
```diff
@@ -0,0 +1,36 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.http_api.web_socket;
+
+import lombok.Getter;
+
+@Getter
+public class RemoteControlClient {
```
HenrikJannsen: Adopt to new naming style: BisqConnectClientInfo (BisqConnectClient might sound confusing -> connect client)

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectModel.java
```diff
@@ -0,0 +1,57 @@
...
+    private final ObjectProperty<BisqConnectHostNetwork> selectedMode = new SimpleObjectProperty<>(BisqConnectHostNetwork.TOR);
+    private final IntegerProperty port = new SimpleIntegerProperty();
+
+    private final StringProperty apiUrl = new SimpleStringProperty("");
+    private final BooleanProperty isChangeDetected = new SimpleBooleanProperty();
+    private final StringProperty password = new SimpleStringProperty("");
+
+    private final ObjectProperty<Image> qrCodeImage = new SimpleObjectProperty<>();
+    private final StringProperty qrPlaceholder = new SimpleStringProperty("");
+    private final SimpleIntegerProperty qrCodeSize = new SimpleIntegerProperty(220);
```
HenrikJannsen: As this is potentially configurable value, I think its better to pass it to the model constructor from the controller, so that the controller defines the value. We keep the model dumb and use only simple default values there but not using specific values defined in the model.
The controller is the central element in our MVC pattern.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectView.java
```diff
@@ -0,0 +1,248 @@
...
+                Res.get("settings.bisqConnect.password.label"),
+                Res.get("settings.bisqConnect.password.prompt")
+        );
+        passwordField.setHelpText(null);
+        passwordField.setPrefWidth(SettingsViewUtils.TEXT_FIELD_WIDTH);
+        saveChangesButton = new Button(Res.get("settings.bisqConnect.saveChanges"));
+        saveChangesButton.getStyleClass().addAll("grey-transparent-outlined-button", "font-size-12");
+
+        ObservableList<BisqConnectHostNetwork> modes = FXCollections.observableArrayList(BisqConnectHostNetwork.values());
+        modeComboBox = new AutoCompleteComboBox<>(modes, Res.get("settings.bisqConnect.exposure.label"));
```
HenrikJannsen: exposure/modeComboBox - Better keep naming of UI element and i18n string synced so its easier to associate.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
```diff
@@ -0,0 +1,252 @@
...
+public class BisqConnectController implements Controller {
+
+
+    @Getter
+    private final BisqConnectView view;
+    private final BisqConnectModel model;
+    private final ServiceProvider serviceProvider;
+
+    private boolean initialEnabled;
+    private BisqConnectHostNetwork initialMode = BisqConnectHostNetwork.LAN;
```
HenrikJannsen: Controller does not store state, should be in model. Applies to the other fields as well.

HenrikJannsen: Also better to keep field name and type in sync (-> initialHostNetwork)

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
```diff
@@ -0,0 +1,252 @@
...
+    public BisqConnectController(ServiceProvider serviceProvider) {
+        this.serviceProvider = serviceProvider;
+
+        model = new BisqConnectModel();
+        view = new BisqConnectView(model, this);
+    }
+
+    @Override
+    public void onActivate() {
+        loadFromConfig();
```
HenrikJannsen, line 74: The config is available at constructor. The resulting values can be passed to model constructor.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
```diff
@@ -0,0 +1,252 @@
...
+            log.error("Could not write config file {}", customConfigFilePath.toAbsolutePath(), e);
+            throw new RuntimeException(e);
+        }
+    }
+
+    private void updateChangeDetectedFlag() {
+        boolean changeDetected = model.getEnabled().get() != savedEnabled
+                || model.getSelectedMode().get() != savedMode
+                || !Objects.equals(model.getPassword().get(), savedPassword);
+        UIThread.run(() -> model.getIsChangeDetected().set(changeDetected));
```
HenrikJannsen: UIThread.run is not needed here as that method is only called from UI handlers. We only need UIThread.run when we get called from domain layer (non Java FX).

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
```diff
@@ -0,0 +1,252 @@
...
+        Path appDataDirPath = appConfig.getAppDataDirPath();
+
+        boolean torMode = model.getSelectedMode().get().isTor();
+        String password = Optional.ofNullable(model.getPassword().get()).orElse("");
+
+        Config newConfig = ConfigFactory.parseMap(Map.of(
+                "application.websocket.enabled", model.getEnabled().get(),
+                "application.websocket.localhostOnly", torMode,
+                "application.websocket.publishOnionService", torMode,
+                "application.websocket.server.host", torMode ? "localhost" : "0.0.0.0",
```
HenrikJannsen, line 231: For Tor I would expect the onion address as host not localhost.

Using `0.0.0.0` is potentially risky to accidentally get exposed to public or untrusted network (internet cafe or co-working LAN, Router NAT + port forwarding, UPnP, VPN interface added).

HenrikJannsen: If the actual values (LAN address, onion address) are not present at that moment, I think its better to leave those fields empty to signal "not available yet"

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    private void applyUrlToModel(String url) {
+        model.getApiUrl().set(url);
+        model.getQrPlaceholder().set("");
+        try {
+            model.getQrCodeImage().set(QrCodeDisplay.toImage(url, model.getQrCodeSize().get()));
+        } catch (RuntimeException e) {
+            log.warn("QR code generation failed", e);
+            model.getQrCodeImage().set(null);
```
HenrikJannsen, line 215: Shouldn't we reset model.getApiUrl as well in that case?

apps/desktop/desktop/src/main/java/bisq/desktop/navigation/NavigationTarget.java
```diff
@@ -244,6 +244,7 @@ public enum NavigationTarget {
     NOTIFICATION_SETTINGS(SETTINGS),
     TRADE_SETTINGS(SETTINGS),
     NETWORK_SETTINGS(SETTINGS),
+    REMOTE_CONTROL_SETTINGS(SETTINGS),
```
HenrikJannsen: Update with new naming pattern

i18n/src/main/resources/settings.properties
```diff
@@ -101,6 +102,28 @@ settings.network.i2p.bi2pGrpcAddress.help=Default: 127.0.0.1:6159
 settings.network.resetToDefaults=Reset to default values
 settings.network.shutdown=Apply changes and restart
 
+settings.bisqConnect.headline=Bisq Connect
+settings.bisqConnect.description=Enabling the WebSocket API server allows companion clients (like Bisq Connect mobile application) to connect and control your Bisq node. Choose how the API is exposed and share the URL or QR code only with trusted devices.
+settings.bisqConnect.enable=Enable WebSocket API server
+settings.bisqConnect.exposure.label=Exposure mode
```
HenrikJannsen: Maybe better: `Accessible via`

i18n/src/main/resources/settings.properties
```diff
@@ -101,6 +102,28 @@ settings.network.i2p.bi2pGrpcAddress.help=Default: 127.0.0.1:6159
...
+settings.bisqConnect.exposure.lan=Local network (LAN)
+settings.bisqConnect.exposure.tor=Tor onion service
```
HenrikJannsen: Maybe better: `Tor network`

i18n/src/main/resources/settings.properties
```diff
@@ -101,6 +102,28 @@ settings.network.i2p.bi2pGrpcAddress.help=Default: 127.0.0.1:6159
 settings.network.resetToDefaults=Reset to default values
 settings.network.shutdown=Apply changes and restart
 
+settings.bisqConnect.headline=Bisq Connect
+settings.bisqConnect.description=Enabling the WebSocket API server allows companion clients (like Bisq Connect mobile application) to connect and control your Bisq node. Choose how the API is exposed and share the URL or QR code only with trusted devices.
```
HenrikJannsen: Maybe better:

To allow the Bisq Connect mobile client to remotely control your Bisq Desktop application, enable the WebSocket API server and choose how the API should be accessible. Then connect the mobile application by scanning the QR code.

i18n/src/main/resources/settings.properties
```diff
@@ -101,6 +102,28 @@ settings.network.i2p.bi2pGrpcAddress.help=Default: 127.0.0.1:6159
...
+settings.bisqConnect.exposure.label=Exposure mode
+settings.bisqConnect.exposure.lan=Local network (LAN)
+settings.bisqConnect.exposure.tor=Tor onion service
+settings.bisqConnect.password.label=API password
+settings.bisqConnect.password.prompt=Leave blank to disable authentication (not recommended).
+settings.bisqConnect.saveChanges=Save Changes
+settings.bisqConnect.url.label=Connection URL
+settings.bisqConnect.connectionDetails.headline=Connection details
+settings.bisqConnect.connectionDetails.unexpectedErrorHint=Check logs for more details.
+settings.bisqConnect.connectionDetails.placeholderDisabled=Enable the WebSocket API to generate a QR code.
```
HenrikJannsen: The QR code is not generated when first toggle the switch

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectView.java
```diff
@@ -0,0 +1,248 @@
...
+
+        enableSwitch = new Switch(Res.get("settings.bisqConnect.enable"));
+        passwordField = new MaterialPasswordField(
+                Res.get("settings.bisqConnect.password.label"),
+                Res.get("settings.bisqConnect.password.prompt")
+        );
+        passwordField.setHelpText(null);
+        passwordField.setPrefWidth(SettingsViewUtils.TEXT_FIELD_WIDTH);
+        saveChangesButton = new Button(Res.get("settings.bisqConnect.saveChanges"));
+        saveChangesButton.getStyleClass().addAll("grey-transparent-outlined-button", "font-size-12");
```
HenrikJannsen: As you set `saveChangesButton.setDefaultButton(true);` the style should be removed. Just use the standard button with setDefaultButton as that is the default action button.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectView.java
```diff
@@ -0,0 +1,248 @@
...
+
+        ObservableList<BisqConnectHostNetwork> modes = FXCollections.observableArrayList(BisqConnectHostNetwork.values());
```
HenrikJannsen: Data should be in model. Sure here its trivial but better stick with the pattern.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectView.java
```diff
@@ -0,0 +1,248 @@
...
+        modeComboBox = new AutoCompleteComboBox<>(modes, Res.get("settings.bisqConnect.exposure.label"));
+        modeComboBox.setPrefWidth(SettingsViewUtils.TEXT_FIELD_WIDTH);
+        modeComboBox.setConverter(new StringConverter<>() {
+            @Override
+            public String toString(BisqConnectHostNetwork mode) {
+                if (mode == null) return "";
+                return mode == BisqConnectHostNetwork.LAN
```
HenrikJannsen: You could add a `displayString` method to the enum instead.

i18n/src/main/resources/settings.properties
```diff
@@ -101,6 +102,28 @@ settings.network.i2p.bi2pGrpcAddress.help=Default: 127.0.0.1:6159
...
+settings.bisqConnect.description=Enabling the WebSocket API server allows companion clients (like Bisq Connect mobile application) to connect and control your Bisq node. Choose how the API is exposed and share the URL or QR code only with trusted devices.
+settings.bisqConnect.enable=Enable WebSocket API server
+settings.bisqConnect.exposure.label=Exposure mode
+settings.bisqConnect.exposure.lan=Local network (LAN)
+settings.bisqConnect.exposure.tor=Tor onion service
+settings.bisqConnect.password.label=API password
+settings.bisqConnect.password.prompt=Leave blank to disable authentication (not recommended).
+settings.bisqConnect.saveChanges=Save Changes
+settings.bisqConnect.url.label=Connection URL
+settings.bisqConnect.connectionDetails.headline=Connection details
```
HenrikJannsen, line 114: Maybe better: API address

common/src/main/java/bisq/common/util/NetworkUtils.java
```diff
@@ -137,6 +140,11 @@ public static Optional<String> findLANHostAddress(Optional<NetworkInterface> pre
         }
         return networkInterfaceHostAddressPairs.stream()
                 .map(Pair::getSecond)
-                .findFirst();
+                .distinct()
```
HenrikJannsen, line 143: log message 
`log.warn("No networkInterface found which matches the preferredNetworkInterface. We return all LAN addresses.");` was wrong already before the change. Can you update it with?

`log.warn("Preferred network interface not found. Falling back to auto-selected LAN address.");`

http-api/src/main/java/bisq/http_api/auth/WebSocketMetadataAddOn.java
```diff
@@ -0,0 +1,35 @@
...
+
+package bisq.http_api.auth;
+
+import org.glassfish.grizzly.filterchain.FilterChainBuilder;
+import org.glassfish.grizzly.http.server.AddOn;
+import org.glassfish.grizzly.http.server.HttpServerFilter;
+import org.glassfish.grizzly.http.server.NetworkListener;
+
+public class WebSocketAddressAddOn implements AddOn {
+
```
HenrikJannsen, line 26: Add `@Override`

http-api/src/main/java/bisq/http_api/auth/WebSocketMetadataAddOn.java
```diff
@@ -0,0 +1,35 @@
...
+    public void setup(NetworkListener networkListener, FilterChainBuilder builder) {
+        int httpServerFilterIdx = builder.indexOfType(HttpServerFilter.class);
+        if (httpServerFilterIdx >= 0) {
```
HenrikJannsen, line 32: Maybe add comment to give context that we want out filter to be added before HttpServerFilter and require that HttpServerFilter is present

http-api/src/main/java/bisq/http_api/auth/WebSocketRequestMetadataFilter.java
```diff
@@ -0,0 +1,56 @@
...
+import org.glassfish.grizzly.http.HttpContent;
+import org.glassfish.grizzly.http.HttpRequestPacket;
+
+import java.net.SocketAddress;
+import java.util.Objects;
+
+@Slf4j
+public class WebSocketAddressFilter extends BaseFilter {
+
+    public NextAction handleRead(FilterChainContext ctx) {
```
HenrikJannsen, line 36: Add @Override

http-api/src/main/java/bisq/http_api/auth/WebSocketAddressFilter.java
```diff
@@ -0,0 +1,56 @@
...
+@Slf4j
+public class WebSocketAddressFilter extends BaseFilter {
+
+    public NextAction handleRead(FilterChainContext ctx) {
+        Object message = ctx.getMessage();
+
+        if (message instanceof HttpContent httpContent && httpContent.getHttpHeader() instanceof HttpRequestPacket request) {
+            String upgradeHeader = request.getHeader("Upgrade");
+            if ("websocket".equalsIgnoreCase(upgradeHeader)) {
+                if (request.getAttribute("ws_user_agent") != null) {
```
HenrikJannsen: Add the magic strings as constants

public static final String ATTR_WS_USER_AGENT = "ws_user_agent";
pub static final String ATTR_WS_REMOTE_ADDRESS = "ws_remote_address";

http-api/src/main/java/bisq/http_api/auth/WebSocketAddressFilter.java
```diff
@@ -0,0 +1,56 @@
...
+import org.glassfish.grizzly.filterchain.FilterChainContext;
+import org.glassfish.grizzly.filterchain.NextAction;
+import org.glassfish.grizzly.http.HttpContent;
+import org.glassfish.grizzly.http.HttpRequestPacket;
+
+import java.net.SocketAddress;
+import java.util.Objects;
+
+@Slf4j
+public class WebSocketAddressFilter extends BaseFilter {
```
HenrikJannsen: Maybe Interceptor is a better name as Filter as we do not filter, just sniff and set attributes.
Or WebSocketRequestMetadataFilter to stick with grizzly Filter base class (they dont have Interceptors).

http-api/src/main/java/bisq/http_api/web_socket/WebSocketConnectionHandler.java
```diff
@@ -24,17 +24,22 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutorService;
+import java.util.function.Consumer;
 
 @Slf4j
 public class WebSocketConnectionHandler extends WebSocketApplication implements Service {
     public final ExecutorService executor = ExecutorFactory.newCachedThreadPool("WebSocketConnectionHandler", 1, 50, 30);
     private final SubscriptionService subscriptionService;
     private final WebSocketRestApiService webSocketRestApiService;
+    private final CopyOnWriteArrayList<Consumer<List<RemoteControlClient>>> clientsListeners = new CopyOnWriteArrayList<>();
```
HenrikJannsen: For type declaration we use List in such cases as long the specific type is not required by client code.

http-api/src/main/java/bisq/http_api/web_socket/WebSocketConnectionHandler.java
```diff
@@ -81,4 +88,46 @@ public void onMessage(WebSocket webSocket, String message) {
...
+    }
+
+    public List<RemoteControlClient> getClients() {
+        return getWebSockets().stream().map(ws -> {
+            try {
+                if (ws instanceof DefaultWebSocket dws) {
+                    Object addr = null;
+                    Object ua = null;
+                    try {
+                        addr = dws.getUpgradeRequest().getAttribute("ws_remote_address");
```
HenrikJannsen: Use static constant instead of magic strings.

http-api/src/main/java/bisq/http_api/web_socket/WebSocketConnectionHandler.java
```diff
@@ -81,4 +88,46 @@ public void onMessage(WebSocket webSocket, String message) {
...
+
+    public void removeClientsListener(Consumer<List<RemoteControlClient>> listener) {
+        clientsListeners.remove(listener);
+    }
+
+    public List<RemoteControlClient> getClients() {
+        return getWebSockets().stream().map(ws -> {
+            try {
+                if (ws instanceof DefaultWebSocket dws) {
+                    Object addr = null;
```
HenrikJannsen: If null values are expected Optional would be more clear. Also inside RemoteControlClient and other areas where the "-" is used.

http-api/src/main/java/bisq/http_api/ApiTorOnionService.java
```diff
@@ -37,23 +39,28 @@ public class ApiTorOnionService implements Service {
...
         this.networkService = networkService;
         this.port = port;
+        this.publishOnionService = publishOnionService;
     }
 
     @Override
     public CompletableFuture<Boolean> initialize() {
-        if (networkService.findServiceNode(TransportType.TOR).isEmpty()) {
-            // If tor is not used we don't do anything
+        if (!publishOnionService) {
```
HenrikJannsen: The `if (networkService.findServiceNode(TransportType.TOR).isEmpty()) {` is still useful IMO.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
```diff
@@ -0,0 +1,252 @@
...
+        model.getPassword().set(password);
+        model.getIsChangeDetected().set(false);
+    }
+
+    private void updatePreview() {
+        Optional<Address> address = model.getWebsocketAddress().get();
+
+        try {
+            if (address.isPresent()) {
+                if (address.get().getHost().equals("0.0.0.0")) {
```
HenrikJannsen, line 179: We should do that in the WebSocketService instead of setting to 0.0.0.0. The LAN address is immediately available.

HenrikJannsen (changes requested):

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
wodoro: This is configuration for the httpapi server, and will take effect only on restarts

for tor we want loopback-only, thats why we use localhost, then the onion service will later forward to this loopback-only server
it doesn't make any sense to have onion address here

using 0.0.0.0 is the only way that allows us to reliably expose the server to LAN network, otherwise the user has to restart the app every time they change the network, and detecting the correct LAN interface (correct from user's perspective) is challenging. 
not exposing server on LAN on unsafe network is really the user's responsibility, they can always choose to set a password for safety. 

the reason for choosing to directly set 0.0.0.0 instead of empty (in WebSocketService) is to indicate that server is listening on all interfaces, then later detect the LAN address on every activation of the tab

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
wodoro: The reason was to reflect the latest changes from the file, as user changes the file inside this settings tab as well
otherwise the change detection logic would fail (assuming the configuration is read once and passed everywhere)

i18n/src/main/resources/settings.properties
wodoro: I intentionally made it a bit vague to convey that this is a risky feature and it allows anyone with the right tools and information connect and control their app
I will change to your version, let me know if you want to change it again.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
wodoro: The lan address may change over time, and getting it in WebSocketService will not reflect the changes

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
HenrikJannsen: OK.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
HenrikJannsen: Now its a bit distributed. Maybe worth to extract the config as setting to a class which is used by WebSocketService and controller and has the single source of state and gets updated at any change.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
HenrikJannsen: If we have to keep 0.0.0.0 we should enforce usage of password. There are some subtle possibilities to leak to the public network.

wodoro: Okay
I think we should enforce the password then
We can detect and set a random password if enabled and no password is set to prevent accidents as well

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectView.java
```diff
@@ -71,7 +71,7 @@ public class BisqConnectView extends View<VBox, BisqConnectModel, BisqConnectCon
         enableSwitch = new Switch(Res.get("settings.bisqConnect.enable"));
         passwordField = new MaterialPasswordField(
                 Res.get("settings.bisqConnect.password.label"),
-                Res.get("settings.bisqConnect.password.prompt")
+                ""
```
HenrikJannsen: prompt can be omitted as there is a `MaterialPasswordField(String description)` constructor

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectView.java
```diff
@@ -140,21 +133,21 @@ public BisqConnectHostNetwork fromString(String string) {
...
-                .left()
-                .minWidth(200)
-                .build();
-        BisqTableColumn<RemoteControlClient> uaCol = new BisqTableColumn.Builder<RemoteControlClient>()
-                .title(Res.get("settings.bisqConnect.clients.userAgent"))
-                .valueSupplier(RemoteControlClient::getUserAgent)
-                .left()
-                .minWidth(300)
-                .build();
+        SortedList<BisqConnectClientInfo> sortedClients = new SortedList<>(model.getConnectedClients());
```
HenrikJannsen, line 134: sortedClients is not needed if not used elsewhere as we have an RichTableView constructor for it:
 ```
 public RichTableView(ObservableList<T> observableList) {
        this(new SortedList<>(observableList));
    }
```

Comment (HenrikJannsen) [reply apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectView.java:134]: Ah you use it for IndexColumn. Not sure if IndexColumn is needed as its expected to have only one client or at least very few. We use IndexColumn in tables where its useful to see number of entries and ranking like in reputation.

File: apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectView.java [hunk 32]
```diff
@@ -140,21 +133,21 @@ public BisqConnectHostNetwork fromString(String string) {
...
+        RichTableView<BisqConnectClientInfo> clientsTable = new RichTableView<>(sortedClients, Res.get("settings.bisqConnect.clients.headline"));
+        BisqTableColumn<BisqConnectClientInfo> indexCol = IndexColumnUtil.getIndexColumn(sortedClients);
+        BisqTableColumn<BisqConnectClientInfo> ipCol = new BisqTableColumn.Builder<BisqConnectClientInfo>()
+            .title(Res.get("settings.bisqConnect.clients.ip"))
+            .valueSupplier(info -> info.getAddress().orElse(""))
```
HenrikJannsen: We use usually `Res.get("data.na")` (`N/A`) for data not present

HenrikJannsen (changes requested):

HenrikJannsen:
The password field should not be in error state before the user tries to click "Save". I think we should also require a min. length of 8 chars for the password. 
 
Screenshot 2025-12-30 at 22 51 47

The placeholder text is not correct:

Screenshot 2025-12-30 at 22 56 02

I think best would be to show a "Establishing onion address..." text when the user enable the API and no onion is known yet. 
If the user has already an onion created we can show that immediately. The state when the onion is available can be detected by observing the state in the network layer.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectView.java
wodoro: I think it's useful to see at a quick glance
should I remove it ?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
wodoro: the thing is the service only uses config once, so its not really needed
it would be needed if it was reacting to changes imo

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
wodoro: after thinking about it, we should allow passwordless configuration, as it needs significant work to comply with the authentication scheme required by the http api
so we should only enforce password in the UI, and not generate it if it's empty in the config and server is enabled

wodoro:
No, when the api is just enabled, its actually still disabled until next restart, so we should show a "Restart to start WebSocket API" placeholder instead

the onion service usually gets published quite fast, and we already observe the state of onion service (done by ApiTorOnionService, observable by WebSocketService.addressObservable)

wodoro:
I still believe localhostOnly needs to be removed, because it is confusing, and host alone will determine this

PR #4244, opened 2026-01-16
feature/new_bisq_mobile_langs_support

rodvar:
- enable new languages codes in locale repo

## Summary by CodeRabbit

* **New Features**
  * Added support for French, Hindi, and Indonesian languages.

### DIFF
--- a/common/src/main/java/bisq/common/locale/LanguageRepository.java
+++ b/common/src/main/java/bisq/common/locale/LanguageRepository.java
@@ -86,30 +86,30 @@ public class LanguageRepository {
             "en", // English
             "de", // German
             "es", // Spanish
+            "fr", // French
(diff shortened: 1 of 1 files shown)


rodvar:
> The new language bundles are missing. I am not sure if those for core are resolved by the mobile resources, so I still merge it but please check if that is the case and otherwise add the missing bundles.

thanks Henrik. The final impl we did with @hiciefte back in the day is independent from Bisq2 core bundles so yes - we are good 👍 

That said, bear in mind this will affect Bisq2 desktop next release as the langs dropdown is populated from LanguageRepository.LANGUAGE_TAGS

So it should be either disabled for Desktop or ideally bundles should be added ( but this is contained in the mobile branch 👍 )

PR #4379, opened 2026-02-10
refactor: extract super class for request service

### LINKED ISSUE #4305: Extract superclass from request service classes
We use duplicated code in `ExplorerService`, `MarketPriceRequestService`, `ReferenceTimeService` and `MobileNotificationRelayClient`.

We should extract a common base class to avoid code duplication.

Config for ReferenceTimeService

sdmg15:
This PR essentially refactor the request service classes by extracting a super class `BaseService` which will be inherited by other service providing common functionalities.

Closes #4305

## Summary by CodeRabbit

* **Refactor**
  * Centralized provider and configuration management into a shared BaseService, unifying lifecycle, provider selection/rotation and removing duplicated provider plumbing across related services.
  * Simplified public APIs for several services to rely on the shared base behavior.
* **Breaking Change**
  * Legacy public config/provider types and some public accessors were removed or simplified.
* **New Features**
  * Added a public helper that exposes the currently selected provider’s base URL for improved observability.


bonded-roles/src/main/java/bisq/bonded_roles/explorer/ExplorerService.java
```diff
@@ -52,7 +52,7 @@ public ExplorerService(Config conf, NetworkService networkService) {
     @Override
     public CompletableFuture<Boolean> shutdown() {
         shutdownStarted = true;
-        return httpClient.map(BaseHttpClient::shutdown)
+        return httpClient.get().map(BaseHttpClient::shutdown)
```
HenrikJannsen: This could be moved to the base class.

network/network/src/main/java/bisq/network/BaseService.java
```diff
@@ -0,0 +1,187 @@
...
+import bisq.common.util.CollectionUtil;
+import bisq.network.http.BaseHttpClient;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public abstract class BaseService implements Service {
```
HenrikJannsen: I think name should be more descriptive to what it does. Maybe `HttpRequestService`.

HenrikJannsen (changes requested):

HenrikJannsen:
I think there should be moved more of the structural logic code to the base class to avoid code duplication. 
The concrete classes should ideally just call the request and receive the result. You can add generic request and response objects to handle the different use cases.

network/network/src/main/java/bisq/network/BaseService.java
```diff
@@ -0,0 +1,187 @@
...
+                return TransportType.CLEAR;
+            }
+        }
+
+        private final Set<Provider> providers;
+        private final Set<Provider> fallbackProviders;
+        private final long timeoutInSeconds;
+        private final long interval;
+        @Setter
+        private long timeSinceLastResponse;
```
HenrikJannsen, line 121: Config should not have mutable fields. It should be just a java representation of the type safe config.
You can move that field to the service class.

network/network/src/main/java/bisq/network/BaseService.java
```diff
@@ -0,0 +1,187 @@
...
+                        String operator = config.getString("operator");
+                        TransportType transportType = getTransportTypeFromUrl(url);
+                        return new Provider(url, apiPath, operator, transportType);
+                    })
+                    .collect(Collectors.toUnmodifiableSet());
+            long interval = typesafeConfig.getLong("interval");
+            return new Config(apiPath, timeoutInSeconds, interval, providers, fallbackProviders);
+        }
+
+        private static TransportType getTransportTypeFromUrl(String url) {
```
HenrikJannsen, line 106: We used an improved version in MobileNotificationRelayClient. Better use that.

```
public static TransportType getTransportTypeFromUrl(String url) {
        try {
            URI uri = URI.create(url);
            String host = uri.getHost();
            if (host != null) {
                if (host.endsWith(".i2p")) {
                    return TransportType.I2P;
                } else if (host.endsWith(".onion")) {
                    return TransportType.TOR;
                }
            }
        } catch (IllegalArgumentException e) {
            log.warn("Failed to parse URL for transport type detection: {}", url);
        }
        return TransportType.CLEAR;
    }
```

network/network/src/main/java/bisq/network/BaseService.java
```diff
@@ -0,0 +1,187 @@
...
+
+            Set<Provider> fallbackProviders = typesafeConfig.getConfigList("fallbackProviders").stream()
+                    .map(config -> {
+                        String url = config.getString("url");
+                        String operator = config.getString("operator");
+                        TransportType transportType = getTransportTypeFromUrl(url);
+                        return new Provider(url, apiPath, operator, transportType);
+                    })
+                    .collect(Collectors.toUnmodifiableSet());
+            long interval = typesafeConfig.getLong("interval");
```
HenrikJannsen, line 102: interval is only used in MarketPriceRequestService and its a special feature, thus I think its better to extend the Config for MarketPriceRequestService and add it only there.

network/network/src/main/java/bisq/network/BaseService.java
```diff
@@ -0,0 +1,187 @@
...
+                        String operator,
+                        String apiPath,
+                        TransportType transportType) {
+            this.baseUrl = baseUrl;
+            this.operator = operator;
+            this.apiPath = apiPath;
+            this.transportType = transportType;
+        }
+
+        public boolean isClearnetProvider() {
```
HenrikJannsen, line 152: Can be removed as not used anymore

network/network/src/main/java/bisq/network/BaseService.java
```diff
@@ -0,0 +1,187 @@
...
+            this.interval = interval;
+            this.apiPath = apiPath;
+        }
+    }
+
+    @Getter
+    @ToString
+    @EqualsAndHashCode
+    public static final class Provider {
+        protected final String baseUrl;
```
HenrikJannsen, line 137: protected can be changed to private as the class is final. Though for ExplorerService it would be good it extend that and add the extra paths for tx and addess, so removing final. Still the fields can stay protected as they are exposed by getters anyway.

HenrikJannsen: Ah I see you handle with with 

```
@Setter
protected String apiPath;
```

HenrikJannsen: As providers is part of config and therefore meant to be immutable, I think using inheritence for the ExplorerService provider might be more clean.

network/network/src/main/java/bisq/network/HttpRequestBaseService.java
```diff
@@ -0,0 +1,313 @@
...
+        numTotalCandidates = providersFromConfig.size() + fallbackProviders.size();
+        if (noProviderAvailable) {
+            log.warn("We do not have any matching provider setup for supportedTransportTypes {}", supportedTransportTypes);
+        } else {
+            selectedProvider.set(selectNextProvider());
+        }
+    }
+
+    @Override
+    public CompletableFuture<Boolean> shutdown() {
```
HenrikJannsen, line 99: Executor shutdown is missing (was in some classes, in others it was missing before):

```
public CompletableFuture<Boolean> shutdown() {
        shutdownStarted = true;

        return httpClient.map(BaseHttpClient::shutdown)
                .orElse(CompletableFuture.completedFuture(true))
                .thenApply(result -> {
                    ExecutorFactory.shutdownAndAwaitTermination(executorService, 100);
                    return result;
                });
    }
```

network/network/src/main/java/bisq/network/HttpRequestBaseService.java
```diff
@@ -0,0 +1,313 @@
...
+            this.transportType = transportType;
+        }
+
+        public boolean isClearnetProvider() {
+            String fullAddress = baseUrl.replaceFirst("^https?://", "");
+            return Address.fromFullAddress(fullAddress).isClearNetAddress();
+        }
+    }
+
+    protected Provider selectNextProvider() {
```
HenrikJannsen, line 284: In MobileNotificationRelayClient we had a better version:

```
private HttpRequestUrlProvider selectNextProvider() {
        if (candidates.isEmpty()) {
            fillCandidates(0);
        }
        // Guard against null return from getRandomElement (can happen if candidates is empty)
        HttpRequestUrlProvider selected = CollectionUtil.getRandomElement(candidates);
        if (selected == null) {
            log.error("No provider available - candidates list is empty after fillCandidates");
            // Return first available provider from config as fallback
            return providersFromConfig.stream().findFirst()
                    .orElseGet(() -> fallbackProviders.stream().findFirst()
                            .orElse(null));
        }
        candidates.remove(selected);
        return selected;
    }
```

network/network/src/main/java/bisq/network/HttpRequestBaseService.java
```diff
@@ -0,0 +1,313 @@
...
+                            log.warn("Request to provider {} timed out after {} seconds",
+                                    currentProvider.getBaseUrl(), conf.getTimeoutInSeconds());
+                        }
+                        return CompletableFuture.failedFuture(new RetryException("Timeout", new AtomicInteger(1)));
+                    }
+                    return CompletableFuture.completedFuture(result);
+                });
+    }
+
+    protected void handleRequestException(Exception e, Provider provider, BaseHttpClient client) throws RuntimeException {
```
HenrikJannsen, line 185: It throws other exceptions as well. As they are all RuntimeExceptions either don't add the throws or list all if we want to document it explicitly. As its a private method, I think we can omit the throws.

HenrikJannsen: The numRecursions handling could be added here as well instead of the entry method.

bonded-roles/src/main/java/bisq/bonded_roles/market_price/MarketPriceRequestService.java
```diff
@@ -136,98 +134,74 @@ private void periodicRequest() {
...
-                        // We only use those market prices for which we have a market in the repository
-                        Map<Market, MarketPrice> filtered = map.entrySet().stream()
-                                .filter(e -> e.getValue().isValidDate())
-                                .filter(e -> MarketRepository.findAnyMarketByMarketCodes(e.getKey().getMarketCodes()).isPresent())
-                                .collect(Collectors.toMap(e -> MarketRepository.findAnyMarketByMarketCodes(e.getKey().getMarketCodes()).orElseThrow(),
-                                        Map.Entry::getValue));
-                        marketPriceByCurrencyMap.clear();
-                        marketPriceByCurrencyMap.putAll(filtered);
-                        mostRecentProvider = Optional.of(selectedProvider.get());
+                        T result = requestFunction.apply(provider).join();
```
HenrikJannsen, line 195: mostRecentProvider could be set here instead of the MarketPriceRequestService as it would be a useful info.

network/network/src/main/java/bisq/network/HttpRequestBaseService.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    protected void shutdownHttpClient(BaseHttpClient client) {
+        try {
+            client.shutdown();
+        } catch (Exception ignore) {
+        }
+    }
+
+    protected <T> CompletableFuture<T> executeWithRetry(Supplier<CompletableFuture<T>> requestSupplier) {
```
HenrikJannsen, line 116: The solution with Supplier<CompletableFuture<T>> requestSupplier comes still with lot of duplicated code in the clients and make the code more complex.

network/network/src/main/java/bisq/network/HttpRequestBaseService.java
```diff
@@ -0,0 +1,313 @@
...
+                    httpClient.set(Optional.of(client));
+
+                    try {
+                        T result = requestFunction.apply(provider).join();
+                        selectedProvider.set(selectNextProvider());
+                        shutdownHttpClient(client);
+                        return result;
+                    } catch (Exception e) {
+                        handleRequestException(e, provider, client);
+                        throw new RuntimeException("Unexpected error", e);
```
HenrikJannsen, line 167: We wrap the exception from handleRequestException into another RuntimeException.
I think if the exception handling is kept inside the catch, it is more explicit which exception is thrown under which condition.

HenrikJannsen:
To make the review easier and be sure to spot differences from the refactoring I did it myself, and then compared the results.  
See comments in https://github.com/bisq-network/bisq2/pull/4426

HenrikJannsen:
As your version came with considerable effort and parts of it was reused in https://github.com/bisq-network/bisq2/pull/4426 I consider your closed PR as subject for a compensation request despite it is not merged.

PR #4403, opened 2026-02-17
Apply unified calculation of security deposit as BTC to all areas.

KimStrand:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security deposit calculation to use both takers' base and quote side amounts with market context, ensuring accurate BTC-based deposit display across offer review flows.
  * Removed an extra setter that could trigger unintended payment-method resets, preventing unexpected changes to the selected payment method during offer creation review.


HenrikJannsen (approved):
utACK

PR #4424, opened 2026-02-19
Fix MuSig BTC and non-BTC side mapping in open trades.

KimStrand:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected payment method shown across trade details, open-trade list, phase views and closed trades so the Non‑BTC side payment method is used when Bitcoin is the base currency.
  * Aligned trade duration and related displays with the corrected payment method source.

* **Refactor**
  * Added contract-level accessors to explicitly expose BTC-side and Non‑BTC-side payment method specs for consistent UI sourcing.


HenrikJannsen (approved):
utACK

PR #4457, opened 2026-02-25
Fix cancel trade flow in rest api

salottodev:
Fixes: https://github.com/bisq-network/bisq-mobile/issues/1143

Fixed an issue where closing a trade after cancellation failed with the error: Open trade channel not found. The TradeID was being cleared during the cancellation process; I've updated the logic to match the desktop app, ensuring the ID is retained until the trade is fully closed

## Release Notes

* **Bug Fixes**
  * Fixed an issue where canceling a trade would unexpectedly remove you from the private chat channel. You can now remain in the chat after canceling a trade.

### DIFF
--- a/http-api/src/main/java/bisq/http_api/rest_api/domain/trades/TradeRestApi.java
+++ b/http-api/src/main/java/bisq/http_api/rest_api/domain/trades/TradeRestApi.java
@@ -284,7 +284,6 @@ public class TradeRestApi extends RestApiBase {
         String encoded = Res.encode("bisqEasy.openTrades.tradeLogMessage.rejected", userName);
         bisqEasyOpenTradeChannelService.sendTradeLogMessage(encoded, channel).get();
         bisqEasyTradeService.cancelTrade(trade);
-        leavePrivateChatManager.leaveChannel(channel);
     }
 
     private void handleCloseTrade(BisqEasyOpenTradeChannel channel, BisqEasyTrade trade) throws Exception {


HenrikJannsen:
@rodvar In desktop the trade stays in the open trades and only at "archive to history" (in 2.19 i guess its close trade), the leave channel is called. If that flow is the same on mobile, then its ok to remove the leave here. Can you confirm? 

Screenshot 2026-02-25 at 17 14 29

rodvar:
> @rodvar In desktop the trade stays in the open trades and only at "archive to history" (in 2.19 i guess its close trade), the leave channel is called. If that flow is the same on mobile, then its ok to remove the leave here. Can you confirm?
> 
> Screenshot 2026-02-25 at 17 14 29

@HenrikJannsen  there is no `trade history` in mobile yet. Effectively, the trade is staying in open trades for the bug described in the associated issue. The issue I think is that both `handleCancel` and `handleClose` close the channel, so for the described issue first a cancel happens (closes channel), then a close happens (tries to cancel an already closed channel -> boom)

`TradeRestApi.java`
```
private void handleCancelTrade(BisqEasyOpenTradeChannel channel,
                                   BisqEasyTrade trade,
                                   String userName) throws Exception {
        String encoded = Res.encode("bisqEasy.openTrades.tradeLogMessage.rejected", userName);
        bisqEasyOpenTradeChannelService.sendTradeLogMessage(encoded, channel).get();
        bisqEasyTradeService.cancelTrade(trade);
        leavePrivateChatManager.leaveChannel(channel);
    }

    private void handleCloseTrade(BisqEasyOpenTradeChannel channel, BisqEasyTrade trade) throws Exception {
        bisqEasyTradeService.removeTrade(trade);
        leavePrivateChatManager.leaveChannel(channel);
    }
```

hence why salotto is proposing deleting that line.

So yes, to your question on mobile we end up executing the "handleCloseTrade" when the user taps on close trade.

PR #4475, opened 2026-03-03
Add mediation result accept/reject flow and wire banner actions

KimStrand:
## Release Notes

* **New Features**
  * Users can now accept or reject mediation results in MuSig trades via new action buttons
  * Mediation result decisions are communicated to the trading peer
  * Trade history logs mediation result acceptance and rejection actions


trade/src/main/proto/trade.proto
```diff
@@ -221,6 +221,7 @@ message MuSigTradeParty {
   optional musigrpc.CloseTradeResponse myCloseTradeResponse = 11;
   optional common.ByteArray peersOutputPrvKeyShare = 12;
   optional account.AccountPayload accountPayload = 13;
+  optional bool mediationResultAcceptance = 14;
```
HenrikJannsen, line 224: Does it mean mediationResultAccepted?

PR #4497, opened 2026-03-08
Update MuSig RPC client to pick up missing txids & exchange `multisigScriptKey` fields

stejbac:
Add a `ContractualTxIds` proto, holding the five Deposit, Warning and Redirect txids of the MuSig trade, together with a `contractualTxIds` field to `PartialSignaturesMessage`, to bring `rpc.proto` up to date with the latest [bisq-musig version](https://github.com/bisq-network/bisq-musig/blob/main/rpc/src/main/proto/rpc.proto). This provides the Bisq2 client with the missing Deposit txid needed from the start of the trade.

(I'm not sure if `ContractualTxIds` is quite the best name -- the idea was that the Deposit, Warning & Redirect txids could be considered an implicit/supplemental part of the trade contract. A trader would ideally require a signed message from the peer holding the five txids, and cross-check it before signing his Deposit Tx inputs, as the txids are relevant to arbitration/mediation cases. In particular, this would prevent any dispute over whose Warning Tx was actually used, which might otherwise be hard to prove.)

NOTE: These changes are dependent on the followup PR https://github.com/bisq-network/bisq-musig/pull/131 in order to work properly, which updates the Rust service to supply the txids as hex strings instead of raw bytes (for convenience to the Java client).

--

Additionally, update `rpc.proto` & `trade.proto` to exchange provisional multisig pubkeys at the trade start, for use in custom script-path payouts for mediation, by adding an extra field to the exchanged `PubKeyShares` value object (and the `PubKeySharesResponse` & `NonceSharesRequest` gRPC protos).

The corresponding Rust changes are still work-in-progress and not yet merged, but I'm fairly certain the extra multisig keys need to exchanged in the first round with the MuSig2 keys (and not later with the fee bump or claim payout addresses, say), and I don't expect any further alterations to the protos will be needed for that.

(Modifying the client first is a little easier, as it's just forwarding the data, and any missing proto fields default to empty.)

## Summary by CodeRabbit

* **New Features**
  * Peers' multisig script key is now exchanged and preserved across flows.
  * Explicit contractual transaction identifiers added to track deposit/warning/redirect transactions.
  * Optional trade fee recipient support for flexible fee routing.

* **Chores**
  * Protocol and message formats updated to carry the new fields and optional values throughout setup and signing flows.

### DIFF
--- a/trade/src/main/proto/rpc.proto
+++ b/trade/src/main/proto/rpc.proto
@@ -72,38 +72,38 @@ message PubKeySharesRequest {
 message PubKeySharesResponse {
   bytes buyerOutputPubKeyShare = 1;
   bytes sellerOutputPubKeyShare = 2;
-  bytes multisigScriptKey = 4;
-  uint32 currentBlockHeight = 3;
+  bytes multisigScriptKey = 3;
+  uint32 currentBlockHeight = 4;
 }
 
 message NonceSharesRequest {
   string tradeId = 1;
   bytes buyerOutputPeersPubKeyShare = 2;
   bytes sellerOutputPeersPubKeyShare = 3;
-  bytes peersMultisigScriptKey = 10;
-  uint64 depositTxFeeRate = 4;       // sats per kwu
-  uint64 preparedTxFeeRate = 5;      // sats per kwu
-  uint64 tradeAmount = 6;            // sats
-  uint64 buyersSecurityDeposit = 7;  // sats
-  uint64 sellersSecurityDeposit = 8; // sats
-  optional ReceiverAddressAndAmount tradeFeeReceiver = 9;
+  bytes peersMultisigScriptKey = 4;
+  uint64 depositTxFeeRate = 5;       // sats per kwu
+  uint64 preparedTxFeeRate = 6;      // sats per kwu
+  uint64 tradeAmount = 7;            // sats
+  uint64 buyersSecurityDeposit = 8;  // sats
+  uint64 sellersSecurityDeposit = 9; // sats
+  optional ReceiverAddressAndAmount tradeFeeReceiver = 10;
 }
 
 message NonceSharesMessage {
   string warningTxFeeBumpAddress = 1;
   string redirectTxFeeBumpAddress = 2;
-  string claimTxPayoutAddress = 11;
-  bytes halfDepositPsbt = 3;
-  uint64 redirectionAmountMsat = 14; // (millisatoshis)
-  bytes swapTxInputNonceShare = 4;
-  bytes buyersWarningTxBuyerInputNonceShare = 5;
-  bytes buyersWarningTxSellerInputNonceShare = 6;
-  bytes sellersWarningTxBuyerInputNonceShare = 7;
-  bytes sellersWarningTxSellerInputNonceShare = 8;
-  bytes buyersRedirectTxInputNonceShare = 9;
-  bytes sellersRedirectTxInputNonceShare = 10;
-  bytes buyersClaimTxInputNonceShare = 12;
-  bytes sellersClaimTxInputNonceShare = 13;
+  string claimTxPayoutAddress = 3;
+  bytes halfDepositPsbt = 4;
+  uint64 redirectionAmountMsat = 5; // (millisatoshis)
+  bytes swapTxInputNonceShare = 6;
+  bytes buyersWarningTxBuyerInputNonceShare = 7;
+  bytes buyersWarningTxSellerInputNonceShare = 8;
+  bytes sellersWarningTxBuyerInputNonceShare = 9;
+  bytes sellersWarningTxSellerInputNonceShare = 10;
+  bytes buyersRedirectTxInputNonceShare = 11;
+  bytes sellersRedirectTxInputNonceShare = 12;
+  bytes buyersClaimTxInputNonceShare = 13;
+  bytes sellersClaimTxInputNonceShare = 14;
 }
 
 message PartialSignaturesRequest {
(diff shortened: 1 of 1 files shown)


trade/src/main/java/bisq/trade/mu_sig/messages/grpc/ContractualTxIds.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+package bisq.trade.mu_sig.messages.grpc;
+
+import bisq.common.proto.Proto;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+@Getter
+@EqualsAndHashCode
```
HenrikJannsen, line 27: Maybe we should add `@toString` as well

trade/src/main/proto/rpc.proto
```diff
@@ -110,6 +118,7 @@ message PartialSignaturesMessage {
   bytes peersClaimTxInputPartialSignature = 6;
   optional bytes swapTxInputPartialSignature = 4;
   optional bytes swapTxInputSighash = 5;
+  ContractualTxIds contractualTxIds = 7;
```
HenrikJannsen: Should we declare it as optional as it is in the Java class?

trade/src/main/proto/rpc.proto
```diff
@@ -72,13 +72,15 @@ message PubKeySharesRequest {
...
   bytes sellerOutputPubKeyShare = 2;
+  bytes multisigScriptKey = 4;
   uint32 currentBlockHeight = 3;
 }
 
 message NonceSharesRequest {
   string tradeId = 1;
   bytes buyerOutputPeersPubKeyShare = 2;
   bytes sellerOutputPeersPubKeyShare = 3;
+  bytes peersMultisigScriptKey = 10;
```
HenrikJannsen: I think we can break compatibility as its only used in dev yet, and use index 4 here

trade/src/main/proto/rpc.proto
```diff
@@ -72,13 +72,15 @@ message PubKeySharesRequest {
 message PubKeySharesResponse {
   bytes buyerOutputPubKeyShare = 1;
   bytes sellerOutputPubKeyShare = 2;
+  bytes multisigScriptKey = 4;
```
HenrikJannsen: I think we can break compatibility as its only used in dev yet, and use index 3 here

trade/src/main/java/bisq/trade/mu_sig/messages/grpc/ContractualTxIds.java
stejbac: OK, I'll add that. (I wasn't sure whether to, as the other messages in the package haven't been given `toString()` implementations.)

trade/src/main/proto/rpc.proto
stejbac: Yes, I think it may be better to make the message fields in `rpc.proto` explicitly optional where they can sometimes be missing during normal API usage, and to agree with the Java classes. (I don't think it affects the proto serialisation at all, at least in this case.)

trade/src/main/proto/rpc.proto
stejbac: OK, I'll make the corresponding changes on the Rust side as an addition to PR https://github.com/bisq-network/bisq-musig/pull/131 (which isn't merged yet), to keep the two sides in sync.

trade/src/main/proto/rpc.proto
stejbac: (Same as above.)

trade/src/main/proto/trade.proto
```diff
@@ -174,20 +174,21 @@ message BisqEasyReportErrorMessage {
...
+  bytes swapTxInputNonceShare = 5;
+  bytes buyersWarningTxBuyerInputNonceShare = 6;
+  bytes buyersWarningTxSellerInputNonceShare = 7;
+  bytes sellersWarningTxBuyerInputNonceShare = 8;
+  bytes sellersWarningTxSellerInputNonceShare = 9;
+  bytes buyersRedirectTxInputNonceShare = 10;
+  bytes sellersRedirectTxInputNonceShare = 11;
   bytes buyersClaimTxInputNonceShare = 12;
   bytes sellersClaimTxInputNonceShare = 13;
 }
```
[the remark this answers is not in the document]
stejbac: Yes, it does indeed appear to persist the trade parties to disk, which I missed, though I oddly didn't spot persistence errors upon restart from archived trades, during testing.

As there don't appear to be past changes to `trade.proto` which (seriously) break wire compatibility since persistence for MuSig trades was added, I'll rebase the last two commits to revert the field ordering changes to `trade.proto` (but not `rpc.proto` where persistence isn't an issue), to be on the safe side.

trade/src/main/java/bisq/trade/mu_sig/messages/grpc/PubKeySharesResponse.java
```diff
@@ -27,13 +27,16 @@
...
 
     public PubKeySharesResponse(byte[] buyerOutputPubKeyShare,
                                 byte[] sellerOutputPubKeyShare,
+                                byte[] multisigScriptKey,
                                 int currentBlockHeight) {
         this.buyerOutputPubKeyShare = buyerOutputPubKeyShare;
         this.sellerOutputPubKeyShare = sellerOutputPubKeyShare;
+        this.multisigScriptKey = multisigScriptKey;
         this.currentBlockHeight = currentBlockHeight;
     }
```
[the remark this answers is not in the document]
stejbac: I think that `MockMusigServer` is essentially dead code that hasn't been kept up to date with the API changes for a long time. Removing it or bringing it up to date should probably be done in a separate PR.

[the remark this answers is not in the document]
stejbac: No, I think that task is too low priority compared to the currently open issues, and will probably be addressed later with other cleanups.

trade/src/main/java/bisq/trade/mu_sig/messages/grpc/ContractualTxIds.java
HenrikJannsen: Yes, most messages might contain sensitive data, thus we do not add it to avoid that it would leak in logs.

HenrikJannsen:
Maybe better we leave it open until the backend is merged, thus devs dont need to switch backend to the working branch... @stejbac Can you ping me once merged?

HenrikJannsen:
@stejbac  told me its already merged. `multisigScriptKey` not yes, but that is not needed atm

PR #4570, opened 2026-03-22
Fix email and phone validation

salottodev:
## Bug Fixes
* Email validation now correctly rejects malformed addresses with leading, trailing, or consecutive dots in the local part
* Phone number validation improved to count actual digits instead of total length, properly supporting formatted numbers with spaces and separators while enforcing digit limits


HenrikJannsen:
Can you make a PR to main as well?

PR #4578, opened 2026-03-23
UI: Refactor Language & Region settings and fix real-time ticker sync

puchis0323:
Description
This PR refactors the "Language" settings into "Language & Region" to provide a cleaner and more integrated user experience. It also resolves a synchronization issue with the top bar market ticker and implements smart defaults for regional settings.

Key Changes:
- UI Redesign: Combined Language, Country, and Currency selectors into a single horizontal row (`HBox`) and removed redundant sub-headlines for a minimalist, professional look.
- Smart Currency Auto-Selection: Implemented logic in the controller to automatically select the most likely default currency when a country is picked (e.g., selecting "Colombia" automatically suggests "COP"), streamlining the onboarding process.
- Locale & es_US Fix: Improved locale handling in `SettingsService` to ensure valid regional defaults are applied correctly, specifically addressing edge cases with certain locale strings like `es_US`.
- Real-time Ticker Sync: Integrated `MarketPriceService` to trigger an immediate refresh of the Top Bar Market Ticker upon currency selection, eliminating the need for an app restart.
- Persistence: Added `countryCode` and `currencyCode` fields to `SettingsStore` and updated the `settings.proto` definitions.

Testing
- Navigated to `Settings -> Language & Region`.
- Verified the new single-row layout.
- Selected different countries and confirmed that the "Default Currency" box updates automatically to the corresponding fiat.
- Verified that the Top Bar Ticker updates instantly when the currency changes.
- Confirmed that all regional settings persist correctly after restarting the application.

## Summary by CodeRabbit

* **New Features**
  * Country and currency selectors added to Language & Region; selections auto-save and restore on launch. Currency list comes from available fiat markets; selecting a currency updates market preference.

* **UI Changes**
  * Section renamed to "Language & Region". Language, country and currency selectors shown together in one row. New labels for country and currency added.

* **Bug Fixes**
  * Better persistence with system locale fallbacks and safer restore on startup.

* **Chores**
  * Settings storage extended to include country and currency.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsController.java
@@ -17,17 +17,27 @@
 
 package bisq.desktop.main.content.settings.language;
 
+import bisq.bonded_roles.market_price.MarketPriceService;
+import bisq.common.asset.FiatCurrencyRepository;
+import bisq.common.locale.CountryRepository;
 import bisq.common.locale.LanguageRepository;
+import bisq.common.market.MarketRepository;
 import bisq.common.observable.Pin;
 import bisq.common.util.StringUtils;
 import bisq.desktop.ServiceProvider;
 import bisq.desktop.common.observable.FxBindings;
 import bisq.desktop.common.view.Controller;
+import bisq.desktop.components.overlay.Popup;
+import bisq.i18n.Res;
 import bisq.settings.SettingsService;
+
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
+import javax.annotation.Nullable;
+import java.util.Currency;
 import java.util.List;
+import java.util.Locale;
 
 @Slf4j
 public class LanguageSettingsController implements Controller {
@@ -35,31 +45,36 @@ public class LanguageSettingsController implements Controller {
     private final LanguageSettingsView view;
     private final LanguageSettingsModel model;
     private final SettingsService settingsService;
-
+    private final MarketPriceService marketPriceService;
     private Pin supportedLanguageTagsPin;
 
     public LanguageSettingsController(ServiceProvider serviceProvider) {
-        settingsService = serviceProvider.getSettingsService();
-        model = new LanguageSettingsModel();
+        this.settingsService = serviceProvider.getSettingsService();
+        this.marketPriceService = serviceProvider.getBondedRolesService().getMarketPriceService();
+
+        this.model = new LanguageSettingsModel(settingsService);
+
+        model.getCountryCodes().sort((c1, c2) ->
+                convertCountryCode(c1).compareToIgnoreCase(convertCountryCode(c2)));
+
+        List<String> supportedCurrencies = MarketRepository.getAllFiatMarkets().stream()
+                .map(market -> market.getQuoteCurrencyCode())
+                .distinct()
+                .sorted()
+                .toList();
+        model.getCurrencyCodes().setAll(supportedCurrencies);
+
         view = new LanguageSettingsView(model, this);
     }
 
     @Override
     public void onActivate() {
         model.getLanguageTags().setAll(getSortedLanguageTags());
-        model.setSelectedLanguageTag(settingsService.getLanguageTag().get());
         model.getSupportedLanguageTags().setAll(getSortedLanguageTags());
 
         supportedLanguageTagsPin = FxBindings.<String, String>bind(model.getSelectedSupportedLanguageTags())
                 .to(settingsService.getSupportedLanguageTags());
-
         model.getSupportedLanguageTagsFilteredList().setPredicate(e -> !model.getSelectedSupportedLanguageTags().contains(e));
-    }
-
-    private List<String> getSortedLanguageTags() {
-        return LanguageRepository.LANGUAGE_TAGS.stream()
-                .sorted((tag1, tag2) -> getDisplayLanguage(tag1).compareTo(getDisplayLanguage(tag2)))
-                .toList();
     }
 
     @Override
@@ -67,19 +82,76 @@ public class LanguageSettingsController implements Controller {
         supportedLanguageTagsPin.unbind();
     }
 
-    void onSelectLanguageTag(String languageTag) {
-        model.setSelectedLanguageTag(languageTag);
-        settingsService.setLanguageTag(languageTag);
+    public String getDisplayCountry(String countryCode) {
+        return convertCountryCode(countryCode);
     }
 
-    String getDisplayLanguage(String languageTag) {
+    public String getDisplayLanguage(String languageTag) {
+        return convertLanguageTag(languageTag);
+    }
+
+    public String getDisplayCurrency(@Nullable String currencyCode) {
+        return convertCurrencyCode(currencyCode);
+    }
+
+    String convertCountryCode(String countryCode) {
+        return CountryRepository.getLocalizedCountryDisplayString(countryCode);
+    }
+
+    String convertLanguageTag(String languageTag) {
         return LanguageRepository.getDisplayString(languageTag);
     }
 
-    void onSelectSupportedLanguage(String languageTag) {
-        if (languageTag != null) {
-            model.getSelectedLSupportedLanguageTag().set(languageTag);
+    String convertCurrencyCode(@Nullable String currencyCode) {
+        if (currencyCode == null) {
+            return "";
         }
+        return FiatCurrencyRepository.findDisplayName(currencyCode)
+                .map(displayName -> displayName + " (" + currencyCode + ")")
+                .orElse(currencyCode);
+    }
+
+    void onSelectCountryCode(@Nullable String countryCode) {
+        if (countryCode == null) return;
+        model.setSelectedCountryCode(countryCode);
+        settingsService.setCountryCode(countryCode);
+        try {
+            Locale locale = new Locale.Builder().setRegion(countryCode).build();
+            Currency defaultCurrency = Currency.getInstance(locale);
+            if (defaultCurrency != null) {
+                onSelectCurrencyCode(defaultCurrency.getCurrencyCode());
+            }
+        } catch (Exception e) {
+            log.warn("Could not derive default currency for country: {}", countryCode);
+        }
+    }
+
+    void onSelectCurrencyCode(@Nullable String currencyCode) {
+        if (currencyCode == null) return;
+        log.debug("User selected default currency: {}", currencyCode);
+        MarketRepository.getAllFiatMarkets().stream()
+                .filter(m -> m.getQuoteCurrencyCode().equalsIgnoreCase(currencyCode))
+                .findFirst()
+                .ifPresentOrElse(market -> {
+                    model.setSelectedCurrencyCode(currencyCode);
+                    settingsService.setCountryCode(model.getSelectedCountryCode());
+                    settingsService.setCurrencyCode(currencyCode);
+                    // R118: Sin Platform.runLater (Contexto de UI garantizado)
+                    if (marketPriceService != null) {
+                        marketPriceService.setSelectedMarket(market);
+                    }
+                }, () -> log.debug("Market not found for: {}", currencyCode));
+    }
+
+    void onSelectLanguageTag(@Nullable String languageTag) {
+        if (languageTag == null) return;
+        model.setSelectedLanguageTag(languageTag);
+        settingsService.setLanguageTag(languageTag);
+        new Popup().feedback(Res.get("settings.language.restart")).useShutDownButton().show();
+    }
+
+    void onSelectSupportedLanguage(@Nullable String languageTag) {
+        if (languageTag != null) model.getSelectedLSupportedLanguageTag().set(languageTag);
     }
 
     void onAddSupportedLanguage() {
@@ -87,14 +159,17 @@ public class LanguageSettingsController implements Controller {
         if (StringUtils.isNotEmpty(selected)) {
             settingsService.getSupportedLanguageTags().add(selected);
             model.getSelectedLSupportedLanguageTag().set(null);
-            model.getSupportedLanguageTagsFilteredList().setPredicate(e -> !model.getSelectedSupportedLanguageTags().contains(e));
         }
     }
 
-    void onRemoveSupportedLanguage(String languageTag) {
-        if (StringUtils.isNotEmpty(languageTag)) {
-            settingsService.getSupportedLanguageTags().remove(languageTag);
-            model.getSupportedLanguageTagsFilteredList().setPredicate(e -> !model.getSelectedSupportedLanguageTags().contains(e));
-        }
+    void onRemoveSupportedLanguage(@Nullable String languageTag) {
+        if (StringUtils.isNotEmpty(languageTag)) settingsService.getSupportedLanguageTags().remove(languageTag);
+    }
+
+
+    private List<String> getSortedLanguageTags() {
+        return LanguageRepository.LANGUAGE_TAGS.stream()
+                .sorted((tag1, tag2) -> convertLanguageTag(tag1).compareTo(convertLanguageTag(tag2)))
+                .toList();
     }
 }
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsModel.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsModel.java
@@ -17,12 +17,16 @@
 
 package bisq.desktop.main.content.settings.language;
 
+import bisq.common.locale.CountryRepository;
 import bisq.desktop.common.view.Model;
+import bisq.settings.SettingsService;
+
 import javafx.beans.property.SimpleStringProperty;
 import javafx.beans.property.StringProperty;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
 import javafx.collections.transformation.FilteredList;
+
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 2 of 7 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsController.java
```diff
@@ -26,23 +28,40 @@
...
     private Pin supportedLanguageTagsPin;
 
     public LanguageSettingsController(ServiceProvider serviceProvider) {
-        settingsService = serviceProvider.getSettingsService();
-        model = new LanguageSettingsModel();
+        this.settingsService = serviceProvider.getSettingsService();
+        this.marketPriceService = serviceProvider.getBondedRolesService().getMarketPriceService();
+        this.model = new LanguageSettingsModel();
+
+        model.getCountryCodes().setAll(List.of(Locale.getISOCountries()));
```
HenrikJannsen: Better to use `CountryRepository.getAllCountyCodes()` so we have more control. 
Can you also put it in the model constructor as its set initially.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsController.java
```diff
@@ -26,23 +28,40 @@
...
+        model.getCountryCodes().sort((c1, c2) -> getDisplayCountry(c1).compareToIgnoreCase(getDisplayCountry(c2)));
+
+        // Filtramos las monedas para mostrar solo las que tienen un mercado fiat en Bisq
```
HenrikJannsen: Maybe better stick to English so other devs can read without translation

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsController.java
```diff
@@ -26,23 +28,40 @@
...
+        List<String> supportedCurrencies = MarketRepository.getAllFiatMarkets().stream()
+                .map(market -> market.getQuoteCurrencyCode())
+                .distinct()
+                .sorted()
+                .toList();
+        model.getCurrencyCodes().setAll(supportedCurrencies);
```
HenrikJannsen, line 65: Can be also set in model constructor

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsController.java
```diff
@@ -51,35 +70,105 @@ public void onActivate() {
...
                 .to(settingsService.getSupportedLanguageTags());
-
         model.getSupportedLanguageTagsFilteredList().setPredicate(e -> !model.getSelectedSupportedLanguageTags().contains(e));
     }
 
-    private List<String> getSortedLanguageTags() {
-        return LanguageRepository.LANGUAGE_TAGS.stream()
-                .sorted((tag1, tag2) -> getDisplayLanguage(tag1).compareTo(getDisplayLanguage(tag2)))
-                .toList();
+    void onSelectCountryCode(@Nullable String countryCode) {
```
HenrikJannsen: Better move the package scope methods down after the public methods.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsController.java
```diff
@@ -51,35 +70,105 @@ public void onActivate() {
...
+                .filter(m -> m.getQuoteCurrencyCode().equalsIgnoreCase(currencyCode))
+                .findFirst()
+                .ifPresentOrElse(market -> {
+                    // Si existe, ENTONCES guardamos la moneda en los ajustes y el modelo
+                    model.setSelectedCurrencyCode(currencyCode);
+                    settingsService.setCurrencyCode(currencyCode);
+
+                    log.debug("Syncing Top Bar Ticker with market: {}", market.getMarketCodes());
+
+                    Platform.runLater(() -> {
```
HenrikJannsen: We use `UIThread.run` instead of `Platform.runLater`. Also why is it used here. We are called from UI context, thus delegating it to the UI thread is not required.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsController.java
```diff
@@ -51,35 +70,105 @@ public void onActivate() {
...
+                            marketPriceService.setSelectedMarket(market);
+                        }
+                    });
+                }, () -> log.debug("No active exchange market found in Bisq for currency: {}", currencyCode));
     }
 
     @Override
     public void onDeactivate() {
-        supportedLanguageTagsPin.unbind();
+        if (supportedLanguageTagsPin != null) {
```
HenrikJannsen: The null check is not required as it is guaranteed by the MVC framework that onDeactivate only called only after onActivate. We only need the null check if the pin is set conditionally.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsView.java
```diff
@@ -44,33 +44,61 @@
...
 
+        // 2. Country Selector
+        countrySelection = new AutoCompleteComboBox<>(model.getCountryCodes(), Res.get("settings.language.region.country"));
+        countrySelection.setPrefWidth(300); // Increased width
+        countrySelection.setConverter(new StringConverter<>() {
             @Override
-            public String fromString(String string) {
-                return null;
+            public String toString(String countryCode) {
+                return countryCode != null ? controller.getDisplayCountry(countryCode) : "";
```
HenrikJannsen: You can use ` CountryRepository.getLocalizedCountryDisplayString(countryCode)`. 
We usually never call getters on the controller. The view binds on the model and calls on controller only UI handlers. If we would not have a util method and the conversion is non-trivial a custom string converter class would be the better option here.

HenrikJannsen: Actually it was not correct before with `getDisplayLanguage`. This also should be avoided to match better the used MVC pattern. (Was my own old code...)

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsController.java
```diff
@@ -51,35 +70,105 @@ public void onActivate() {
...
     }
 
-    void onSelectSupportedLanguage(String languageTag) {
+    // --- NUEVO MÉTODO PARA MOSTRAR EL NOMBRE LARGO DE LA MONEDA ---
+    public String getDisplayCurrency(@Nullable String currencyCode) {
+        if (currencyCode == null) {
+            return "";
+        }
+        try {
+            java.util.Currency currency = java.util.Currency.getInstance(currencyCode);
```
HenrikJannsen: You can use `FiatCurrencyRepository.findDisplayName(currencyCode)`

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsController.java
```diff
@@ -51,35 +70,105 @@ public void onActivate() {
...
     }
 
-    String getDisplayLanguage(String languageTag) {
+    public String getDisplayLanguage(String languageTag) {
         return LanguageRepository.getDisplayString(languageTag);
     }
 
-    void onSelectSupportedLanguage(String languageTag) {
+    // --- NUEVO MÉTODO PARA MOSTRAR EL NOMBRE LARGO DE LA MONEDA ---
+    public String getDisplayCurrency(@Nullable String currencyCode) {
```
HenrikJannsen: Can be package scope (as the other getters). 

Those getter methods would be better renamed to convertCurrencyCode (as used in StringConverter context) or alternatively add custom StringConverters to the model.

I am aware it was used that getter style before but it is a bit confusing as the view should never call the controller for getting data, only the model. A custom StringConverter inside the model would be most clean approach to fit our MVC architecture.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsController.java
```diff
@@ -51,35 +70,105 @@ public void onActivate() {
...
+            supportedLanguageTagsPin.unbind();
+        }
     }
 
-    void onSelectLanguageTag(String languageTag) {
-        model.setSelectedLanguageTag(languageTag);
-        settingsService.setLanguageTag(languageTag);
-        new Popup().feedback(Res.get("settings.language.restart")).useShutDownButton().show();
+    public String getDisplayCountry(String countryCode) {
+        return new Locale.Builder().setRegion(countryCode).build().getDisplayCountry(Locale.getDefault());
```
HenrikJannsen: You can use  CountryRepository.getLocalizedCountryDisplayString(countryCode).

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsController.java
```diff
@@ -51,35 +70,105 @@ public void onActivate() {
...
+        try {
+            java.util.Currency currency = java.util.Currency.getInstance(currencyCode);
+            // Returns: "US Dollar (USD)" or "Euro (EUR)"
+            return currency.getDisplayName() + " (" + currencyCode + ")";
+        } catch (Exception e) {
+            return currencyCode; // Fallback por si falla
+        }
+    }
+
+    private List<String> getSortedLanguageTags() {
```
HenrikJannsen: Private methods are usually at the end of the class

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsView.java
```diff
@@ -44,33 +44,61 @@
...
             }
+            @Override
+            public String fromString(String string) { return null; }
         });
-        languageSelection.validateOnNoItemSelectedWithMessage(Res.get("settings.language.select.invalid"));
 
-        // Supported languages
+        // 3. Currency Selector
+        currencySelection = new AutoCompleteComboBox<>(model.getCurrencyCodes(), Res.get("settings.language.region.currency"));
+        currencySelection.setPrefWidth(300); // Increased width as requested by maintainers
```
HenrikJannsen, line 85: This comment can be removed.

i18n/src/main/resources/settings.properties
```diff
@@ -129,12 +132,9 @@ settings.bisqConnect.apiAccessTransportType.CLEAR=Clearnet
 # suppress inspection "UnusedProperty"
 settings.bisqConnect.apiAccessTransportType.I2P=I2P
 
-
-
-
 settings.bisqConnect.apiConfig.headline=API config
 settings.bisqConnect.apiConfig.info=It is recommended to use Tor for connecting to the API. \
-  If you choose clearnet, using LAN is the next best option — apply the detected LAN host. \
+  If you choose clearnet, using LAN is the next best option apply the detected LAN host. \
```
HenrikJannsen: The dash should not get removed.

i18n/src/main/resources/settings.properties
```diff
@@ -199,4 +199,4 @@ settings.backup.totalMaxBackupSizeInMB.info.tooltip=Important data is automatica
   - Previous Years: Keep one backup per year.\n\n\
   Please note, this backup mechanism does not protect against hard disk failures.\n\
   For better protection, we strongly recommend creating manual backups on an alternative hard drive.
-settings.backup.totalMaxBackupSizeInMB.invalid=Must be a value between {0} MB and {1} MB
+settings.backup.totalMaxBackupSizeInMB.invalid=Must be a value between {0} MB and {1} MB
```
HenrikJannsen: A trailing blank line is part of our formatting rules.

settings/src/main/java/bisq/settings/SettingsService.java
```diff
@@ -171,302 +197,85 @@ public void onPersistedApplied(SettingsStore persisted) {
...
-
-    public Map<String, Market> getMuSigLastSelectedMarketByBaseCurrencyMap() {
-        return Collections.unmodifiableMap(persistableStore.muSigLastSelectedMarketByBaseCurrencyMap);
-    }
-
-    public ReadOnlyObservable<Market> getSelectedWalletMarket() {
-        return persistableStore.selectedWalletMarket;
-    }
-
+    public ReadOnlyObservable<Market> getSelectedMuSigMarket() { return persistableStore.selectedMuSigMarket; }
```
HenrikJannsen: Our formatting rules are not using one liners in such case. Do you use IntelliJ IDE? There is the formatting rules checked in and it should be applied at safe or at commit to avoid changes of formatting.

settings/src/main/java/bisq/settings/SettingsService.java
```diff
@@ -171,302 +197,85 @@ public void onPersistedApplied(SettingsStore persisted) {
...
+    public void setDifficultyAdjustmentFactor(double value) { if (value >= NetworkLoad.MIN_DIFFICULTY_ADJUSTMENT && value <= NetworkLoad.MAX_DIFFICULTY_ADJUSTMENT) persistableStore.difficultyAdjustmentFactor.set(value); }
+    public void setMaxTradePriceDeviation(double value) { if (value >= MIN_TRADE_PRICE_DEVIATION && value <= MAX_TRADE_PRICE_DEVIATION) persistableStore.maxTradePriceDeviation.set(value); }
+    public void setChatNotificationType(ChatNotificationType chatNotificationType) { persistableStore.chatNotificationType.set(chatNotificationType); }
+    public void setIsTacAccepted(boolean isTacAccepted) { persistableStore.isTacAccepted.set(isTacAccepted); }
+    public void setCloseMyOfferWhenTaken(boolean closeMyOfferWhenTaken) { persistableStore.closeMyOfferWhenTaken.set(closeMyOfferWhenTaken); }
+    public void setLanguageTag(String languageTag) { if (languageTag != null && LanguageRepository.LANGUAGE_TAGS.contains(languageTag)) persistableStore.languageTag.set(languageTag); }
+    public void setCountryCode(String countryCode) { if (countryCode != null && !countryCode.isBlank()) persistableStore.countryCode.set(countryCode); }
```
HenrikJannsen: 'StringUtils.isNotEmpty' can be used instead

settings/src/main/java/bisq/settings/SettingsStore.java
```diff
@@ -267,6 +280,8 @@ public static SettingsStore fromProto(bisq.settings.protobuf.SettingsStore proto
                 new HashSet<>(proto.getConsumedAlertIdsList()),
                 proto.getCloseMyOfferWhenTaken(),
                 proto.getLanguageTag(),
+                proto.getCountryCode(),
```
HenrikJannsen: For users updating from an older version that value will be an empty String as it was not present in the protobuf data. We should apply the default values in such cases

HenrikJannsen:
Can you resolve the merge conflicts and rebase on main?

puchis0323:
> Can you resolve the merge conflicts and rebase on main?

Working

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsModel.java
```diff
@@ -17,27 +17,54 @@
...
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 @Getter
 public class LanguageSettingsModel implements Model {
+
+    public LanguageSettingsModel(SettingsService settingsService) {
+        this.settingsService = settingsService;
+
+        // Inicialización de datos
```
HenrikJannsen: No need for such comments which are not really required for making the code more understandable. We try to be not too verbose as every line cause mental costs for reading...

HenrikJannsen: And use English only

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/language/LanguageSettingsView.java
```diff
@@ -44,33 +44,61 @@
...
 
-        // Language
-        Label languageSelectionHeadline = SettingsViewUtils.getHeadline(Res.get("settings.language.headline"));
+        // --- MAIN HEADLINE ---
+        Label mainHeadline = SettingsViewUtils.getHeadline(Res.get("settings.language.headline"));
 
+        // 1. Language Selector
         languageSelection = new AutoCompleteComboBox<>(model.getLanguageTags(), Res.get("settings.language.select"));
-        languageSelection.setPrefWidth(300);
+        languageSelection.setPrefWidth(300); // Increased width to avoid truncation
```
HenrikJannsen: Comment not needed

settings/src/main/java/bisq/settings/SettingsService.java
```diff
@@ -128,6 +130,33 @@ public CompletableFuture<Boolean> initialize() {
...
+        if (getCountryCode().get() == null || getCountryCode().get().isBlank()) {
+            String osDefaultCountry = Locale.getDefault().getCountry();
+            setCountryCode(osDefaultCountry);
+            needsPersist = true;
+        }
+
+        // Fallback for currency code using OS locale
+        if (getCurrencyCode().get() == null || getCurrencyCode().get().isBlank()) {
+            try {
+                java.util.Currency defaultCurrency = java.util.Currency.getInstance(Locale.getDefault());
```
HenrikJannsen: Use import instead of package name if possible.

settings/src/main/java/bisq/settings/SettingsService.java
```diff
@@ -146,15 +175,13 @@ public CompletableFuture<Boolean> shutdown() {
 
     @Override
     public CompletableFuture<Boolean> persist() {
-        // We don't want to call persist from the addObserver calls at initialize
```
HenrikJannsen, line 149: Do not remove the old comments as they are useful for context

settings/src/main/java/bisq/settings/SettingsService.java
```diff
@@ -180,123 +207,40 @@ private void applyLanguageTag(String languageTag) {
...
-    public Map<String, Market> getMuSigLastSelectedMarketByBaseCurrencyMap() {
-        return Collections.unmodifiableMap(persistableStore.muSigLastSelectedMarketByBaseCurrencyMap);
-    }
-
-    public ReadOnlyObservable<Market> getSelectedWalletMarket() {
-        return persistableStore.selectedWalletMarket;
-    }
-
+    public ReadOnlyObservable<Market> getSelectedMuSigMarket() { return persistableStore.selectedMuSigMarket; }
+    public ReadOnlyObservable<Boolean> getUseAnimations() { return persistableStore.useAnimations; }
```
HenrikJannsen: Do not use one liners.

puchis0323:
Hi @HenrikJannsen , I've updated the PR. I squashed the commits into a single clean one, rebased on the latest main, and removed all non-English comments and internal notes. It's ready for review now. Thanks!

HenrikJannsen:
Replaced by https://github.com/bisq-network/bisq2/pull/4619

The PR is still valid for compensation request.

PR #4579, opened 2026-03-24
Fix/linux tray icon transparency

puchis0323:
Fixes the background transparency issue on Linux system tray icons. 

`Toolkit.getImage()` flattens the alpha channel into a solid background when scaling via `setImageAutoSize(true)` on Linux desktop environments (like KDE Plasma). Switched to `ImageIO.read()`, which generates a proper `BufferedImage` preserving the transparency natively. 

Tested locally: the seed-node-app compiles and runs cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved system tray icon handling: the app now detects missing or unreadable tray icons, logs the issue, and aborts tray setup when none are available.
  * Added a safer image-loading flow with a reliable fallback to prevent startup failures from corrupted or missing icon files.
* **Chores**
  * Minor code cleanup and import updates.

### DIFF
--- a/apps/desktop/bi2p/src/main/java/bisq/bi2p/Bi2pApp.java
+++ b/apps/desktop/bi2p/src/main/java/bisq/bi2p/Bi2pApp.java
@@ -17,7 +17,6 @@
 
 package bisq.bi2p;
 
-
 import bisq.bi2p.common.standby.PreventStandbyModeService;
 import bisq.bi2p.common.utils.ImageUtil;
 import bisq.bi2p.common.utils.KeyHandlerUtil;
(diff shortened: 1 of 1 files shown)


apps/desktop/bi2p/src/main/java/bisq/bi2p/Bi2pApp.java
```diff
@@ -58,15 +59,12 @@ public class Bi2pApp extends Application {
     private Bi2pController controller;
     private String i2pRouterDir;
     private PreventStandbyModeService preventStandbyModeService;
-    // For now, we have it turned on always, but maybe we allow to turn off in UI or by options, thus we leave it
```
HenrikJannsen, line 61: Better not remove those comments. Same for others below

HenrikJannsen:
Replaced by https://github.com/bisq-network/bisq2/pull/4612

PR #4634, opened 2026-04-09
Enable native macOS arm64 Tor and ad-hoc sign extracted binaries

### LINKED ISSUE #2679: Tor binary for MacOS aarch64 not working
When using the Tor binary v13.5.2 for MacOs aarch64 Tor does not start up and does not get past `TorInstaller: Tor files installed to <HOME_DIR>/Library/Application Support/bisq2_dev1/tor`.

As that has been also observed with previous version on mostly Windows, I assume it might have a similar reason.

@alva Any idea?

dbast:
closes #2679

Switch macOS ARM Tor download to the native aarch64 expert bundle and ad-hoc sign extracted Tor binaries on macOS at install time. This avoids startup failure on Apple Silicon where unsigned arm64 Tor binaries are killed by macOS before control socket creation.

Root cause:
- Tor expert bundles are unsigned Mach-O binaries.
- On Apple Silicon, unsigned arm64 `tor` exits with SIGKILL (exit code 137), so Bisq times out waiting for `tor/control/control`.
- x86_64 Tor still runs via Rosetta, which masked this issue in previous fallback behavior.

Why ad-hoc signing here:
- `/usr/bin/codesign` is a standard macOS system tool.
- Ad-hoc signing (`-s -`) does not require Apple Developer certificates, notarization, or user keychain setup.
- It works for both Intel and Apple Silicon binaries, so one macOS code path covers x86_64 and arm64 consistently.
- Applying it to extracted runtime binaries is appropriate because Tor is unpacked into the user data directory, not executed directly from a pre-signed app bundle location.

Manual verification steps:
1) Download and unpack Tor bundles:
   `curl -fsSL https://archive.torproject.org/tor-package-archive/torbrowser/13.5.2/tor-expert-bundle-macos-aarch64-13.5.2.tar.gz -o arm.tar.gz && tar -xzf arm.tar.gz`
2) Confirm binary arch:
   `file tor/tor tor/libevent-2.1.7.dylib`
3) Confirm unsigned:
   `codesign --verify --verbose=4 tor/tor` (fails: not signed)
4) Reproduce failure:
   `tor/tor --version` (arm64 fails, typically RC 137)
5) Confirm workaround:
   `/usr/bin/codesign -s - -f --timestamp=none tor/libevent-2.1.7.dylib tor/tor && tor/tor --version` (succeeds)

Also bump Tor installer version marker to force reinstall so existing macOS users receive signed binaries.

This should unlock building full Apple Silicon aka darwin arm64 dmg files. The workflow [build.yml](https://github.com/bisq-network/bisq2/blob/main/.github/workflows/build.yml) runs on macos-latest which is already arm64 according to https://docs.github.com/en/actions/reference/runners/github-hosted-runners

## Summary by CodeRabbit

* **Bug Fixes**
  * macOS ARM64 now downloads the correct Apple Silicon Tor binary.

* **Updates**
  * Bundled Tor upgraded from 0.1.0 to 0.2.0.
  * Added a macOS-only post-install signing step for installed Tor binaries; installation will fail and report errors if signing does not complete successfully or times out, ensuring usable signed binaries.

### DIFF
--- a/build-logic/tor-binary/src/main/kotlin/bisq/gradle/tor_binary/TorBinaryUrlProvider.kt
+++ b/build-logic/tor-binary/src/main/kotlin/bisq/gradle/tor_binary/TorBinaryUrlProvider.kt
@@ -13,11 +13,7 @@ class TorBinaryUrlProvider(private val version: String) : PerPlatformUrlProvider
         get() = "tor-expert-bundle-macos-x86_64-$version.tar.gz"
 
     override val MACOS_ARM_64_URL: String
-        // Currently the Tor version for aarch64 does not work, thus we use the x64 version.
-        // See: https://github.com/bisq-network/bisq2/issues/2679
-        // Once resolved we can use the line below.
-        // get() = "tor-expert-bundle-macos-aarch64-$version.tar.gz"
-        get() = "tor-expert-bundle-macos-x86_64-$version.tar.gz"
+        get() = "tor-expert-bundle-macos-aarch64-$version.tar.gz"
 
     override val WIN_X86_64_URL: String
         get() = "tor-expert-bundle-windows-x86_64-$version.tar.gz"
--- a/network/tor/tor/src/main/java/bisq/network/tor/installer/TorInstaller.java
+++ b/network/tor/tor/src/main/java/bisq/network/tor/installer/TorInstaller.java
@@ -19,15 +19,24 @@ package bisq.network.tor.installer;
 
 import bisq.common.file.FileMutatorUtils;
 import bisq.common.file.FileReaderUtils;
+import bisq.common.platform.OS;
 import lombok.extern.slf4j.Slf4j;
 
+import java.io.BufferedReader;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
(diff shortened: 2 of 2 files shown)


HenrikJannsen:
@alvasw Could you review that PR?

HenrikJannsen (approved):
utACK

dbast:
Anything else I can do here? 


(The only other option I see is fully signing and notarising all the binaries in the .dmg file with real Apple certificates, which bisq doesn't do as explained here https://github.com/bisq-network/bisq/discussions/6341 ... so the solution of this PR seems to be the remaining option to get things working.)

HenrikJannsen:
> Commits must have verified signatures.

Can you sign your commits?

dbast:
@HenrikJannsen done :)

HenrikJannsen:
We implemented aarch64 support and signing in commit `785c14aa181e87707698835731e4fd222c161398`.
Instead of doing the signing at user level we do at at the build and ship the signed tor binaries. Do you see any drawbacks with that approach?

dbast:
Inspected 785c14aa181e87707698835731e4fd222c161398 ... Build-time ad-hoc signing is probably enough to solve the Apple Silicon Tor execution issue, because the signature is embedded in the Mach-O binary and is not limited to the machine where it was created.

PR #4643, opened 2026-04-10
feat: improve external tor UX, add file socket control support, add tails support

wodoro:
- Solves an issue I had before with the override config, which didn't allow to have multiple `Bridge` directives or similar things
- Fixes the `java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.String` issue mentioned in https://github.com/bisq-network/bisq2/issues/4606#issuecomment-4206218568
- Adds support for using a custom torrc file using the new `torrcOverrideFilePath` config
- Same `torrcOverrideFilePath` can now be used for configuring external tor connection details
- Adds support for using a Unix socket for connecting to tor
- Improves `useExternalTor` flag UX by automatically detecting it's tail os and setting the ControlPort to 951
- Improves UX by hard failing rather than failing back to using embedded tor when `useExternalTor` flag is true, which can be very confusing otherwise
- Includes tails and whonix onion-grater configs in the package now
- Adds a new `prepare_tails.sh` for easier setup on tails 

## Summary by CodeRabbit

* **New Features**
  * Tor overrides now support ordered multi-value directives, can be loaded from an external torrc-style file, and external Tor may use either UNIX-domain sockets or host:port endpoints. Improved control connection detection, cookie-auth handling, and clearer error messages. Tails packaging and a helper script added.

* **Documentation**
  * Config docs updated to add external torrc file option, describe value shapes (string or list) and precedence.

* **Tests**
  * Added tests for parsing, merging, endpoints, socket control, heuristics, and helpers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq2/pull/4643?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/desktop/desktop-app-launcher/package/linux/prepare_tails.sh
+++ b/apps/desktop/desktop-app-launcher/package/linux/prepare_tails.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+# Bisq2 - Tails OS preparation script
+# Copies the onion-grater profile required for Bisq2 to work on Tails.
+# Must be run with root privileges.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+BISQ_LIB_DIR="$SCRIPT_DIR"
+ONION_GRATER_SRC="$BISQ_LIB_DIR/onion-grater/40_bisq_tails.yml"
+ONION_GRATER_DEST="/etc/onion-grater.d/bisq.yml"
+DataDirectory="/home/amnesia/Persistent/bisq/Bisq2"
+DEFAULT_EXTERNAL_TOR_CONFIG="/home/amnesia/.local/share/Bisq2/tor/external_tor.config"
+
+if [ "$(id -u)" -ne 0 ]; then
+    echo "Error: This script must be run as root (use sudo)." >&2
+    exit 1
+fi
+
+if [ ! -f "$ONION_GRATER_SRC" ]; then
+    echo "Error: Onion-grater profile not found at $ONION_GRATER_SRC" >&2
+    exit 1
+fi
+
+if [ ! -d "/etc/onion-grater.d" ]; then
+    echo "Error: /etc/onion-grater.d directory not found. Is this a Tails system?" >&2
+    exit 1
+fi
+
+# Gracefully shut down Bisq2 if it is running
+if pgrep -x "Bisq2" > /dev/null 2>&1; then
+    echo "Bisq2 is running. Sending SIGTERM and waiting for exit ..."
+    pkill -SIGTERM -x "Bisq2"
+    for i in $(seq 1 30); do
+        if ! pgrep -x "Bisq2" > /dev/null 2>&1; then
+            echo "Bisq2 has exited."
+            break
+        fi
+        sleep 1
+    done
+    if pgrep -x "Bisq2" > /dev/null 2>&1; then
+        echo "Warning: Bisq2 did not exit within 30 seconds. Exiting..." >&2
+        exit 1
+    fi
+fi
+
+# Remove stale external Tor config if present
+if [ -f "$DEFAULT_EXTERNAL_TOR_CONFIG" ]; then
+    rm -f "$DEFAULT_EXTERNAL_TOR_CONFIG"
+    echo "Removed existing external Tor config: $DEFAULT_EXTERNAL_TOR_CONFIG"
+fi
+
+cp "$ONION_GRATER_SRC" "$ONION_GRATER_DEST"
+chmod 644 "$ONION_GRATER_DEST"
+echo "Onion-grater profile installed to $ONION_GRATER_DEST"
+echo "Restart onion-grater service ..."
+systemctl restart onion-grater.service
+
+echo "Redirect user data to Tails Persistent Storage at $DataDirectory ..."
+if [ ! -d "$DataDirectory" ]; then
+    mkdir -p "$DataDirectory"
+    chown -R amnesia:amnesia "$DataDirectory"
+fi
+
+TARGET_LINK="/home/amnesia/.local/share/Bisq2"
+if [ -e "$TARGET_LINK" ]; then
+    if [ -L "$TARGET_LINK" ]; then
+        echo "Symlink already exists: $TARGET_LINK"
+    else
+        echo "Warning: $TARGET_LINK exists and is not a symlink; skipping creation" >&2
+    fi
+else
+    ln -s "$DataDirectory" "$TARGET_LINK"
+    echo "Created symlink: $TARGET_LINK -> $DataDirectory"
+fi
+
+echo "Preparations complete."
--- a/apps/desktop/desktop-app-launcher/package/linux/prepare_tails.sh
+++ b/apps/desktop/desktop-app-launcher/package/linux/prepare_tails.sh
@@ -74,11 +74,10 @@ if [ -L "$TARGET_LINK" ]; then
 elif [ -d "$TARGET_LINK" ]; then
     BACKUP_DIR="${TARGET_LINK}.pre-tails-migration.$(date +%Y%m%d%H%M%S)"
     echo "Found existing non-persistent data at $TARGET_LINK; migrating into $DataDirectory ..."
-    if ! command -v rsync >/dev/null 2>&1; then
-        echo "Error: rsync not found; cannot migrate existing $TARGET_LINK safely." >&2
-        exit 1
-    fi
-    rsync -aHAX --info=stats2 -- "$TARGET_LINK"/ "$DataDirectory"/
+    # Use cp (always present) instead of rsync (not guaranteed). cp -a == -dR --preserve=all,
+    # preserving mode, ownership, timestamps, symlinks, hard links, ACLs and xattrs. The trailing
+    # "/." copies the directory contents (including dotfiles) into the existing $DataDirectory.
+    cp -a -- "$TARGET_LINK"/. "$DataDirectory"/
     mv -- "$TARGET_LINK" "$BACKUP_DIR"
     echo "Backed up original directory to $BACKUP_DIR"
     chown -R amnesia:amnesia "$DataDirectory"
--- a/apps/desktop/desktop-app-launcher/package/linux/prepare_tails.sh
+++ b/apps/desktop/desktop-app-launcher/package/linux/prepare_tails.sh
@@ -1,0 +1,95 @@
+#!/bin/bash
+# Bisq2 - Tails OS preparation script
+# Copies the onion-grater profile required for Bisq2 to work on Tails.
+# Must be run with root privileges.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+BISQ_LIB_DIR="$SCRIPT_DIR"
+ONION_GRATER_SRC="$BISQ_LIB_DIR/onion-grater/40_bisq_tails.yml"
+ONION_GRATER_DEST="/etc/onion-grater.d/bisq.yml"
+DataDirectory="/home/amnesia/Persistent/bisq/Bisq2"
+DEFAULT_EXTERNAL_TOR_CONFIG="/home/amnesia/.local/share/Bisq2/tor/external_tor.config"
+
+if [ "$(id -u)" -ne 0 ]; then
+    echo "Error: This script must be run as root (use sudo)." >&2
+    exit 1
+fi
+
+if [ ! -f "$ONION_GRATER_SRC" ]; then
+    echo "Error: Onion-grater profile not found at $ONION_GRATER_SRC" >&2
+    exit 1
+fi
+
+if [ ! -d "/etc/onion-grater.d" ]; then
+    echo "Error: /etc/onion-grater.d directory not found. Is this a Tails system?" >&2
+    exit 1
+fi
+
+# Gracefully shut down Bisq2 if it is running
+if pgrep -x "Bisq2" > /dev/null 2>&1; then
+    echo "Bisq2 is running. Sending SIGTERM and waiting for exit ..."
+    pkill -SIGTERM -x "Bisq2"
+    for i in $(seq 1 30); do
+        if ! pgrep -x "Bisq2" > /dev/null 2>&1; then
+            echo "Bisq2 has exited."
+            break
+        fi
+        sleep 1
+    done
+    if pgrep -x "Bisq2" > /dev/null 2>&1; then
+        echo "Warning: Bisq2 did not exit within 30 seconds. Exiting..." >&2
+        exit 1
+    fi
+fi
+
+# Remove stale external Tor config if present
+if [ -f "$DEFAULT_EXTERNAL_TOR_CONFIG" ]; then
+    rm -f "$DEFAULT_EXTERNAL_TOR_CONFIG"
+    echo "Removed existing external Tor config: $DEFAULT_EXTERNAL_TOR_CONFIG"
+fi
+
+cp "$ONION_GRATER_SRC" "$ONION_GRATER_DEST"
+chmod 644 "$ONION_GRATER_DEST"
+echo "Onion-grater profile installed to $ONION_GRATER_DEST"
+echo "Restart onion-grater service ..."
+systemctl restart onion-grater.service
+
+echo "Redirect user data to Tails Persistent Storage at $DataDirectory ..."
+if [ ! -d "$DataDirectory" ]; then
+    mkdir -p "$DataDirectory"
+fi
+chown -R amnesia:amnesia "$DataDirectory"
+
+TARGET_LINK="/home/amnesia/.local/share/Bisq2"
+if [ -L "$TARGET_LINK" ]; then
+    CURRENT_TARGET="$(readlink -f "$TARGET_LINK")"
+    EXPECTED_TARGET="$(readlink -f "$DataDirectory")"
+    if [ "$CURRENT_TARGET" != "$EXPECTED_TARGET" ]; then
+        echo "Error: $TARGET_LINK is a symlink to $CURRENT_TARGET, expected $EXPECTED_TARGET. Remove it manually and re-run." >&2
+        exit 1
+    fi
+    echo "Symlink already exists: $TARGET_LINK -> $CURRENT_TARGET"
+elif [ -d "$TARGET_LINK" ]; then
+    BACKUP_DIR="${TARGET_LINK}.pre-tails-migration.$(date +%Y%m%d%H%M%S)"
+    echo "Found existing non-persistent data at $TARGET_LINK; migrating into $DataDirectory ..."
+    if ! command -v rsync >/dev/null 2>&1; then
+        echo "Error: rsync not found; cannot migrate existing $TARGET_LINK safely." >&2
+        exit 1
+    fi
+    rsync -aHAX --info=stats2 -- "$TARGET_LINK"/ "$DataDirectory"/
+    mv -- "$TARGET_LINK" "$BACKUP_DIR"
+    echo "Backed up original directory to $BACKUP_DIR"
+    chown -R amnesia:amnesia "$DataDirectory"
+    ln -s "$DataDirectory" "$TARGET_LINK"
+    echo "Created symlink: $TARGET_LINK -> $DataDirectory"
+elif [ -e "$TARGET_LINK" ]; then
+    echo "Error: $TARGET_LINK exists and is neither a symlink nor a directory; refusing to overwrite. Remove it manually and re-run." >&2
+    exit 1
+else
+    ln -s "$DataDirectory" "$TARGET_LINK"
+    echo "Created symlink: $TARGET_LINK -> $DataDirectory"
+fi
+
+echo "Preparations complete."
--- a/apps/desktop/desktop-app-launcher/package/linux/prepare_tails.sh
+++ b/apps/desktop/desktop-app-launcher/package/linux/prepare_tails.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 # Bisq2 - Tails OS preparation script
-# Copies the onion-grater profile required for Bisq2 to work on Tails.
+# Copies the onion-grater profile required for Bisq2 to work on Tails and migrates data
+# written to the volatile home directory by earlier versions into Persistent Storage.
 # Must be run with root privileges.
 
 set -euo pipefail
@@ -9,8 +10,9 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 BISQ_LIB_DIR="$SCRIPT_DIR"
 ONION_GRATER_SRC="$BISQ_LIB_DIR/onion-grater/40_bisq_tails.yml"
 ONION_GRATER_DEST="/etc/onion-grater.d/bisq.yml"
-DataDirectory="/home/amnesia/Persistent/bisq/Bisq2"
-DEFAULT_EXTERNAL_TOR_CONFIG="/home/amnesia/.local/share/Bisq2/tor/external_tor.config"
+PERSISTENT_STORAGE_DIR="/home/amnesia/Persistent"
+PERSISTENT_DATA_DIR="$PERSISTENT_STORAGE_DIR/Bisq2"
+VOLATILE_DATA_DIR="/home/amnesia/.local/share/Bisq2"
 
 if [ "$(id -u)" -ne 0 ]; then
     echo "Error: This script must be run as root (use sudo)." >&2
@@ -44,51 +46,51 @@ if pgrep -x "Bisq2" > /dev/null 2>&1; then
     fi
 fi
 
-# Remove stale external Tor config if present
-if [ -f "$DEFAULT_EXTERNAL_TOR_CONFIG" ]; then
-    rm -f "$DEFAULT_EXTERNAL_TOR_CONFIG"
-    echo "Removed existing external Tor config: $DEFAULT_EXTERNAL_TOR_CONFIG"
-fi
-
 cp "$ONION_GRATER_SRC" "$ONION_GRATER_DEST"
 chmod 644 "$ONION_GRATER_DEST"
 echo "Onion-grater profile installed to $ONION_GRATER_DEST"
 echo "Restart onion-grater service ..."
 systemctl restart onion-grater.service
 
-echo "Redirect user data to Tails Persistent Storage at $DataDirectory ..."
-if [ ! -d "$DataDirectory" ]; then
-    mkdir -p "$DataDirectory"
-fi
-chown -R amnesia:amnesia "$DataDirectory"
+# Bisq2 uses Persistent Storage as its data directory whenever it is unlocked, so no symlink is
+# needed. Running amnesically is a supported choice, so a locked volume is a warning, not an error.
+if [ -d "$PERSISTENT_STORAGE_DIR" ]; then
+    DATA_DIR="$PERSISTENT_DATA_DIR"
+    mkdir -p "$DATA_DIR"
 
-TARGET_LINK="/home/amnesia/.local/share/Bisq2"
-if [ -L "$TARGET_LINK" ]; then
-    CURRENT_TARGET="$(readlink -f "$TARGET_LINK")"
-    EXPECTED_TARGET="$(readlink -f "$DataDirectory")"
-    if [ "$CURRENT_TARGET" != "$EXPECTED_TARGET" ]; then
-        echo "Error: $TARGET_LINK is a symlink to $CURRENT_TARGET, expected $EXPECTED_TARGET. Remove it manually and re-run." >&2
-        exit 1
+    # Earlier versions wrote to the volatile home directory; migrate that data once.
+    if [ -L "$VOLATILE_DATA_DIR" ]; then
+        # Symlink created by an earlier version of this script; the app no longer needs it.
+        rm -f "$VOLATILE_DATA_DIR"
+        echo "Removed obsolete symlink: $VOLATILE_DATA_DIR"
+    elif [ -d "$VOLATILE_DATA_DIR" ]; then
+        if [ -n "$(find "$DATA_DIR" -mindepth 1 -maxdepth 1 -print -quit)" ]; then
+            echo "Error: Both $VOLATILE_DATA_DIR and $DATA_DIR contain data." >&2
+            echo "Refusing to merge them because that would overwrite the persistent profile." >&2
+            echo "Back up both directories and keep the one you want at $DATA_DIR, then re-run." >&2
+            exit 1
+        fi
+        echo "Migrating $VOLATILE_DATA_DIR into $DATA_DIR ..."
+        # Use cp (always present) instead of rsync (not guaranteed). cp -a == -dR --preserve=all,
+        # preserving mode, ownership, timestamps, symlinks, hard links, ACLs and xattrs. The trailing
+        # "/." copies the directory contents (including dotfiles) into the existing $DATA_DIR.
+        cp -a -- "$VOLATILE_DATA_DIR"/. "$DATA_DIR"/
+        echo "Migration complete. The original is left at $VOLATILE_DATA_DIR until the next reboot."
     fi
-    echo "Symlink already exists: $TARGET_LINK -> $CURRENT_TARGET"
-elif [ -d "$TARGET_LINK" ]; then
-    BACKUP_DIR="${TARGET_LINK}.pre-tails-migration.$(date +%Y%m%d%H%M%S)"
-    echo "Found existing non-persistent data at $TARGET_LINK; migrating into $DataDirectory ..."
-    # Use cp (always present) instead of rsync (not guaranteed). cp -a == -dR --preserve=all,
-    # preserving mode, ownership, timestamps, symlinks, hard links, ACLs and xattrs. The trailing
-    # "/." copies the directory contents (including dotfiles) into the existing $DataDirectory.
-    cp -a -- "$TARGET_LINK"/. "$DataDirectory"/
-    mv -- "$TARGET_LINK" "$BACKUP_DIR"
-    echo "Backed up original directory to $BACKUP_DIR"
-    chown -R amnesia:amnesia "$DataDirectory"
-    ln -s "$DataDirectory" "$TARGET_LINK"
-    echo "Created symlink: $TARGET_LINK -> $DataDirectory"
-elif [ -e "$TARGET_LINK" ]; then
-    echo "Error: $TARGET_LINK exists and is neither a symlink nor a directory; refusing to overwrite. Remove it manually and re-run." >&2
-    exit 1
+
+    chown -R amnesia:amnesia "$DATA_DIR"
 else
-    ln -s "$DataDirectory" "$TARGET_LINK"
-    echo "Created symlink: $TARGET_LINK -> $DataDirectory"
+    DATA_DIR="$VOLATILE_DATA_DIR"
+    echo "Warning: Persistent Storage is not unlocked, so Bisq2 uses $DATA_DIR." >&2
+    echo "All data, including identity keys and open offers, will be lost on shutdown." >&2
 fi
 
-echo "Preparations complete."
+# The external Tor config is only auto-detected when absent, so drop a stale file written
+# before Tails was detected. Bisq2 recreates it with the onion-grater control port.
+STALE_EXTERNAL_TOR_CONFIG="$DATA_DIR/tor/external_tor.config"
+if [ -f "$STALE_EXTERNAL_TOR_CONFIG" ]; then
+    rm -f "$STALE_EXTERNAL_TOR_CONFIG"
+    echo "Removed existing external Tor config: $STALE_EXTERNAL_TOR_CONFIG"
+fi
+
+echo "Preparations complete. Bisq2 uses $DATA_DIR."
--- a/network/network/src/main/resources/tor/external_tor.config
+++ b/network/network/src/main/resources/tor/external_tor.config
@@ -2,6 +2,7 @@
 
 ## This configuration follows the torrc format but only supports the options listed below.
 ## Ensure the torrc file of the external Tor process is configured accordingly.
+## Bisq will prefer ControlSocket when both ControlSocket and ControlPort are configured.
 
 ## If UseExternalTor is set to '1' Bisq use the external Tor process.
 #UseExternalTor 1
@@ -13,8 +14,13 @@ CookieAuthentication 0
 ## Example: /opt/homebrew/etc/tor/control_auth_cookie
 CookieAuthFile /path/to/control_auth_cookie
 
+## ControlSocket specifies the path to the Tor control socket.
+## Use this on systems where Tor exposes a Unix-domain control socket.
+## ControlSocket /path/to/tor/control.socket
+
 ## ControlPort specifies the address for the Tor control port.
 ## This must match the configuration in the external Tor process's torrc file.
+## When both ControlSocket and ControlPort are configured, Bisq prefers ControlSocket.
 ## By default, this is:
 ControlPort 127.0.0.1:9051
 
--- a/network/tor/tor/src/main/java/bisq/network/tor/TorService.java
+++ b/network/tor/tor/src/main/java/bisq/network/tor/TorService.java
@@ -18,7 +18,6 @@
 package bisq.network.tor;
 
 import bisq.common.application.Service;
-import bisq.common.data.Pair;
 import bisq.common.facades.FacadeProvider;
 import bisq.common.file.FileMutatorUtils;
 import bisq.common.file.FileReaderUtils;
@@ -28,8 +27,9 @@ import bisq.common.platform.OS;
 import bisq.common.util.StringUtils;
 import bisq.network.tor.common.torrc.BaseTorrcGenerator;
 import bisq.network.tor.common.torrc.TorrcFileGenerator;
-import bisq.network.tor.controller.TorControlAuthenticationFailed;
+import bisq.network.tor.controller.TorControlConnectionClosedException;
 import bisq.network.tor.controller.TorController;
+import bisq.network.tor.controller.exceptions.CannotConnectWithTorException;
 import bisq.network.tor.controller.events.events.TorBootstrapEvent;
 import bisq.network.tor.installer.TorInstaller;
 import bisq.network.tor.process.EmbeddedTorProcess;
@@ -37,19 +37,20 @@ import bisq.network.tor.process.control_port.ControlPortFileParser;
 import bisq.security.keys.TorKeyPair;
 import com.runjva.sourceforge.jsocks.protocol.Socks5Proxy;
 import lombok.Getter;
+import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import net.freehaven.tor.control.PasswordDigest;
 
 import java.io.IOException;
 import java.net.Inet4Address;
 import java.net.InetAddress;
-import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -57,10 +58,10 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.stream.Collectors;
 
 import static bisq.common.threading.ExecutorFactory.commonForkJoinPool;
 import static bisq.network.tor.common.torrc.Torrc.Keys.CONTROL_PORT;
+import static bisq.network.tor.common.torrc.Torrc.Keys.CONTROL_SOCKET;
 import static bisq.network.tor.common.torrc.Torrc.Keys.COOKIE_AUTHENTICATION;
 import static bisq.network.tor.common.torrc.Torrc.Keys.COOKIE_AUTH_FILE;
 import static com.google.common.base.Preconditions.checkArgument;
@@ -87,6 +88,7 @@ public class TorService implements Service {
     private Optional<EmbeddedTorProcess> torProcess = Optional.empty();
     private Optional<TorSocksProxyFactory> torSocksProxyFactory = Optional.empty();
     private final Map<String, String> externalTorConfigMap = new HashMap<>();
+    private Optional<Path> activeExternalTorConfigPath = Optional.empty();
 
     public TorService(TorTransportConfig transportConfig) {
         this.transportConfig = transportConfig;
@@ -111,13 +113,10 @@ public class TorService implements Service {
             if (connectedToExternalTor(externalTorConfigMap)) {
                 log.info("External Tor will be used");
                 return CompletableFuture.completedFuture(true);
-            } else if (LinuxDistribution.isWhonix()) {
-                // In case we failed at connectedToExternalTor, we do not start the embedded Tor as fallback if we are on Whonix.
-                throw new RuntimeException("Bisq runs on a Whonix system but it could not connect to the Whonix system Tor. " +
-                        "Please check the Bisq external_tor.config and the system Tor torrc config.");
             } else {
-                // In case connectedToExternalTor failed we try embedded Tor (if not running on Whonix)
-                useExternalTor.set(false);
+                isRunning.set(false);
+                throw new RuntimeException("`useExternalTor` is enabled, but Bisq could not connect to the external Tor control " +
+                        "server. Please check " + getExternalTorConfigDescription() + " and your system Tor configuration.");
             }
         }
 
@@ -141,7 +140,7 @@ public class TorService implements Service {
         torController = new TorController(transportConfig.getBootstrapTimeout(), transportConfig.getHsUploadTimeout(), bootstrapEvent);
 
         PasswordDigest hashedControlPassword = PasswordDigest.generateDigest();
-        Map<String, String> torrcConfigMap = createTorrcConfigFile(torDataDirPath, hashedControlPassword);
+        createTorrcConfigFile(torDataDirPath, hashedControlPassword);
 
         // remove the control port file before we launch tor, just in case it exists from
         // a previous run
@@ -187,29 +186,24 @@ public class TorService implements Service {
     }
 
     private boolean evaluateUseExternalTor() {
-        boolean useExternalTor = false;
-        if (LinuxDistribution.isWhonix()) {
-            log.info("Bisq runs on a Whonix system and use the Whonix system Tor");
-            return true;
-        } else {
-            // If environment variable is set we take that.
-            String torSkipLaunch = System.getenv(TOR_SKIP_LAUNCH);
-            if (StringUtils.isNotEmpty(torSkipLaunch)) {
-                log.info("Environment variable 'TOR_SKIP_LAUNCH' is set to '{}'", torSkipLaunch);
-                return torSkipLaunch.equals("1") ||
-                        torSkipLaunch.equalsIgnoreCase("true") ||
-                        torSkipLaunch.equalsIgnoreCase("yes");
-            } else {
-                // We check if external_tor.config value is set. If so we use that value
-                Optional<String> externalTorFromConfigMap = Optional.ofNullable(externalTorConfigMap.get("UseExternalTor"));
-                if (externalTorFromConfigMap.isPresent()) {
-                    return externalTorFromConfigMap.get().equals("1");
-                } else {
-                    // Last we take the value form the JVM config
-                    return transportConfig.isUseExternalTor();
-                }
-            }
+        boolean isWhonixOrTails = LinuxDistribution.isWhonix() || LinuxDistribution.isTails();
+        String torSkipLaunch = System.getenv(TOR_SKIP_LAUNCH);
+        String useExternalTorFromConfig = externalTorConfigMap.get("UseExternalTor");
+        boolean useExternalTorFromJvmConfig = transportConfig.isUseExternalTor();
+
+        boolean useExternalTor = UseExternalTorResolver.evaluateUseExternalTorValue(
+                isWhonixOrTails,
+                torSkipLaunch,
+                useExternalTorFromJvmConfig,
+                useExternalTorFromConfig);
+
+        if (isWhonixOrTails) {
+            log.info("Bisq runs on a Whonix/Tails system and use the system Tor");
+        } else if (StringUtils.isNotEmpty(torSkipLaunch)) {
+            log.info("Environment variable 'TOR_SKIP_LAUNCH' is set to '{}'", torSkipLaunch);
         }
+
+        return useExternalTor;
     }
 
     @Override
@@ -225,6 +219,7 @@ public class TorService implements Service {
             torSocksProxyFactory = Optional.empty();
             publishedOnionServices.clear();
             externalTorConfigMap.clear();
+            activeExternalTorConfigPath = Optional.empty();
             return true;
         }, commonForkJoinPool());
     }
@@ -294,20 +289,24 @@ public class TorService implements Service {
     }
 
     private boolean connectedToExternalTor(Map<String, String> torConfigMap) {
-        Pair<String, Integer> controlHostAndPort = getControlHostAndPort(torConfigMap);
-        String controlHost = controlHostAndPort.getFirst();
-        int controlPort = controlHostAndPort.getSecond();
-        if (!isTorControlAvailable(controlHost, controlPort)) {
-            log.warn("Couldn't connect to external Tor control server. " +
-                    "This could happen if the control port is not correctly set up in 'external_tor.config' file.");
-            return false;
-        }
+        ControlEndpoint controlEndpoint = getControlEndpoint(torConfigMap);
 
         if (torController != null) {
             torController.shutdown();
         }
         torController = new TorController(transportConfig.getBootstrapTimeout(), transportConfig.getHsUploadTimeout(), bootstrapEvent);
-        torController.initialize(controlPort);
+        try {
+            if (controlEndpoint instanceof UnixSocketControlEndpoint socketControlEndpoint) {
+                torController.initialize(socketControlEndpoint.controlSocketPath());
+            } else if (controlEndpoint instanceof TcpControlEndpoint tcpControlEndpoint) {
+                torController.initialize(tcpControlEndpoint.controlHost(), tcpControlEndpoint.controlPort());
+            }
+        } catch (CannotConnectWithTorException e) {
+            log.warn("Could not connect to external Tor control endpoint.", e);
+            torController.shutdown();
+            torController = null;
+            return false;
+        }
 
         boolean isAuthenticated = false;
         boolean noCookieAuthenticationRequired = Optional.ofNullable(torConfigMap.get(COOKIE_AUTHENTICATION)).orElse("0").equals("0");
@@ -317,23 +316,26 @@ public class TorService implements Service {
             try {
                 torController.authenticate();
                 isAuthenticated = true;
-            } catch (TorControlAuthenticationFailed e) {
-                log.warn("Authentication to Tor Control failed.");
+            } catch (Exception e) {
+                log.warn("Authentication to Tor Control failed.", e);
                 return false;
             }
         }
 
         if (!isAuthenticated) {
             if (torConfigMap.containsKey(COOKIE_AUTH_FILE)) {
-                String authCookiePath = torConfigMap.get(COOKIE_AUTH_FILE);
+                String authCookiePath = StringUtils.unquote(torConfigMap.get(COOKIE_AUTH_FILE));
                 try {
                     byte[] authCookie = Files.readAllBytes(Paths.get(authCookiePath));
                     torController.authenticate(authCookie);
                     isAuthenticated = true;
-                } catch (TorControlAuthenticationFailed e) {
-                    log.warn("Authentication with authCookie failed.");
                 } catch (IOException e) {
                     log.warn("Couldn't read the COOKIE_AUTH_FILE '{}' from the config.", authCookiePath);
+                } catch (TorControlConnectionClosedException e) {
+                    log.warn("Tor control connection was closed unexpectedly when trying to authenticate.", e);
+                    return false; // since pipe is broken
+                } catch (Exception e) {
+                    log.warn("Authentication to Tor Control failed.", e);
                 }
             } else {
                 log.warn("The user did not provide the path to the COOKIE_AUTH_FILE.");
@@ -344,8 +346,8 @@ public class TorService implements Service {
             log.info("We try to call authenticate without parameters for the case that authentication is not required.");
             try {
                 torController.authenticate();
-            } catch (TorControlAuthenticationFailed ex) {
-                log.warn("Authentication to Tor Control failed.");
+            } catch (Exception e) {
+                log.warn("Authentication to Tor Control failed.", e);
                 return false;
             }
         }
@@ -360,8 +362,16 @@ public class TorService implements Service {
         return true;
     }
 
-    private Pair<String, Integer> getControlHostAndPort(Map<String, String> externalTorConfigMap) {
+    private ControlEndpoint getControlEndpoint(Map<String, String> externalTorConfigMap) {
+        String controlSocketString = externalTorConfigMap.get(CONTROL_SOCKET);
+        if (controlSocketString != null && !controlSocketString.isBlank()) {
+            Path controlSocketPath = Path.of(StringUtils.unquote(controlSocketString));
+            return ControlEndpoint.forSocket(controlSocketPath);
+        }
+
         String controlPortString = externalTorConfigMap.get(CONTROL_PORT);
+        checkArgument(controlPortString != null && !controlPortString.isBlank(),
+                "Either ControlSocket or ControlPort must be set in " + getExternalTorConfigDescription());
         String[] tokens = controlPortString.split(":");
         String controlHost;
         int controlPort;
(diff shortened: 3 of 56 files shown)


wodoro:
@coderabbitai full review

wodoro:
I'll mark this as ready as soon as I can confirm the issue #4606 is resolved by this

wodoro:
@coderabbitai full review

wodoro:
@coderabbitai full review

network/network/src/main/resources/tor/external_tor.config
```diff
@@ -13,8 +14,13 @@ CookieAuthentication 0
...
 
+## ControlSocket specifies the path to the Tor control socket.
+## Use this on systems where Tor exposes a Unix-domain control socket.
+## ControlSocket /path/to/tor/control.socket
+
 ## ControlPort specifies the address for the Tor control port.
 ## This must match the configuration in the external Tor process's torrc file.
+## When both ControlSocket and ControlPort are configured, Bisq prefers ControlSocket.
 ## By default, this is:
 ControlPort 127.0.0.1:9051
```
HenrikJannsen, line 25: Should be `ControlPort 9051`

network/tor/tor/src/main/java/bisq/network/tor/TorService.java
```diff
@@ -430,21 +447,77 @@ private void installTorIfNotUpToDate() {
...
+
+    private record TcpControlEndpoint(String controlHost, int controlPort) implements ControlEndpoint {
+        @Override
+        @NonNull
+        public String toString() {
+            return String.format("TcpControlEndpoint(host: %s, port: %d)", controlHost, controlPort);
+        }
+    }
+
+    private record SocketControlEndpoint(Path controlSocketPath) implements ControlEndpoint {
```
HenrikJannsen: Maybe `UnixSocketControlEndpoint` would make it more distinct?

HenrikJannsen (changes requested):

HenrikJannsen:
@alvasw Could you review as well?

network/network/src/main/resources/tor/external_tor.config
wodoro: [Tor manual](https://man.archlinux.org/man/tor.1.en) states that this format is correct, and it's been that way for long, so why change now?
\+ I think the explicit host makes the behavior more clear

network/tor/tor/src/main/java/bisq/network/tor/TorService.java
wodoro: Should I also rename the other one to `TcpSocketControlEndpoint` (it lacks `Socket` keyword at the moment) ?
edit: No, I think that's unnecessary. no need to answer this.

HenrikJannsen:
@wodoro Can you look into the Ci errors?

wodoro:
Sounds like GitHub flakiness. trigger rerun ?

HenrikJannsen (approved):
utACK

HenrikJannsen:
@alvasw Could you review as well?

alvasw (dismissed):
## Conceptual Review
- Bisq won't receive incoming connections even though the onion verification will pass. Bisq won't work for makers (clobbering the orderbook and increasing mediation). Tails blocks global binds and this change is missing. 
Tails only allows loopback binds (see [Firewall rules](https://gitlab.tails.boum.org/tails/tails/-/blob/stable/config/chroot_local-includes/etc/ferm/ferm.conf?ref_type=heads)).

- It'd be better to parse `/etc/os-release` because all Linux distributions use it and it simplifies the other cases too.
	- `NAME="Tails"`
	- `ID="tails"`

- Firewall hardening possible. Tor Browser, OnionCirc, Onion-Share, ... all run in isolated network namespaces and have fail-safe rules:
	- https://gitlab.tails.boum.org/tails/tails/-/blob/stable/config/chroot_local-includes/usr/local/lib/tails-create-netns?ref_type=heads
	- https://gitlab.tails.boum.org/tails/tails/-/blob/stable/config/chroot_local-includes/etc/ferm/ferm.conf?ref_type=heads

- `JPackageTask` copies the `onion-grater` profiles for Tails and Whonix into package. `onion-grater` officially supports Bisq[^1].
	- Better only copy Tails' `onion-grater` profile

- `prepare-tails.sh`
	- Better handle directly at setup (deb install)
		- Use `postinst` and `prerm`/`postrm` scripts in deb to automatically install and enable Bisq's onion-grater profile on Tails.
		- User doesn't need to do anything except to install the deb.

- Check at startup if persistence is activated or data directory is on external drive.
	- Otherwise user loses access to Bisq, dead offers, more mediation, ...
	- Tell user and abort early - otherwise lost data (ask explicit permission)

- `40_bisq_tails.yml`
	- AppArmor profiles missing or argument is wrong.
		- Check reference in Tails repo, yes onionshare does have it
			- https://gitlab.tails.boum.org/tails/tails/-/blob/stable/config/chroot_local-includes/etc/apparmor.d/usr.bin.onionshare?ref_type=heads
			- https://gitlab.tails.boum.org/tails/tails/-/blob/stable/config/chroot_local-includes/etc/apparmor.d/usr.bin.onionshare-cli?ref_type=heads

- We should mirror Tor's original override behaviour, we had compatibility issues in the past with Whonix and Tails.

Tor's man page:

```
           Note

           Configuration options on the command line override those in configuration files. See THE
           CONFIGURATION FILE FORMAT for more information.

       The following options in this section are only recognized on the tor command line, not in a
       configuration file.

```

- Move integration tests from unit test module to intergration test module to speed-up build.

## Code Review

- The control port and control socket code paths can be merged by using `SocketAddress`

`TorController`:

```diff
@@ -32,16 +32,8 @@
         this.bootstrapEvent = bootstrapEvent;
     }
 
-    public void initialize(int controlPort) {
-        initialize("127.0.0.1", controlPort);
-    }
-
-    public void initialize(String controlHost, int controlPort) {
-        torControlProtocol.initialize(controlHost, controlPort);
-    }
-
-    public void initialize(Path controlSocketPath) {
-        torControlProtocol.initialize(controlSocketPath);
+    public void initialize(SocketAddress socketAddress) {
+        torControlProtocol.initialize(socketAddress);
     }
 
     public void authenticate(byte[] authCookie) {
```

`TorControlProtocol`:

```diff
@@ -45,64 +38,27 @@
         torControlReader = new TorControlReader();
     }
 
-    public void initialize(int port) {
-        initialize("127.0.0.1", port);
-    }
-
-    public void initialize(String host, int port) {
+    public void initialize(SocketAddress socketAddress) {
         try {
-            SocketChannel socketChannel = createAndConnectTcpControlSocket(host, port);
-            initializeControlConnection(socketChannel,
-                    Channels.newInputStream(socketChannel),
-                    Channels.newOutputStream(socketChannel),
-                    host + ":" + port);
-        } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();  // Restore interrupted state
-            log.error("TorControlProtocol initialization for control endpoint {}:{} was interrupted.", host, port, e);
-            close();
-            throw new CannotConnectWithTorException(e);
-        } catch (CannotConnectWithTorException e) {
-            log.error("TorControlProtocol failed to connect to Tor control endpoint {}:{}.", host, port, e);
-            close();
-            throw e;
-        } catch (Exception e) {
-            log.error("Unexpected exception during TorControlProtocol initialization for control endpoint {}:{}.", host, port, e);
+            SocketChannel socketChannel = new TorControlSocketChannelFactory(socketAddress).createAndConnect();
+            initializeControlConnection(socketChannel);
+
+        } catch (Exception e) {
+            if (e instanceof InterruptedException) {
+                Thread.currentThread().interrupt();  // Restore interrupted state
+            }
+
+            log.error("TorControlProtocol initialization for control endpoint {}.", socketAddress, e);
             close();
             throw new CannotConnectWithTorException(e);
         }
     }
 
-    public void initialize(Path controlSocketPath) {
-        try {
-            SocketChannel socketChannel = createAndConnectUnixControlSocket(controlSocketPath);
-            initializeControlConnection(socketChannel,
-                    Channels.newInputStream(socketChannel),
-                    Channels.newOutputStream(socketChannel),
-                    controlSocketPath.toString());
-        } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();
-            log.error("TorControlProtocol initialization for control socket {} was interrupted.", controlSocketPath, e);
-            close();
-            throw new CannotConnectWithTorException(e);
-        } catch (CannotConnectWithTorException e) {
-            log.error("TorControlProtocol failed to connect to Tor control socket {}.", controlSocketPath, e);
-            close();
-            throw e;
-        } catch (Exception e) {
-            log.error("Unexpected exception during TorControlProtocol initialization for control socket {}.", controlSocketPath, e);
-            close();
-            throw new CannotConnectWithTorException(e);
-        }
-    }
-
-    private void initializeControlConnection(AutoCloseable controlConnection,
-                                             InputStream inputStream,
-                                             OutputStream outputStream,
-                                             String endpointDescription) {
-        this.controlConnection = controlConnection;
-        this.outputStream = Optional.of(outputStream);
-        this.torControlReader.start(inputStream);
-        log.info("TorControlProtocol initialized successfully for control endpoint {}", endpointDescription);
+    private void initializeControlConnection(SocketChannel socketChannel) throws IOException {
+        this.controlConnection = socketChannel;
+        this.outputStream = Optional.of(Channels.newOutputStream(socketChannel));
+        this.torControlReader.start(Channels.newInputStream(socketChannel));
+        log.info("TorControlProtocol initialized successfully for control endpoint {}", socketChannel.getRemoteAddress());
     }
 
     @Override
@@ -303,117 +259,6 @@
         }
     }
 
-    private SocketChannel createAndConnectTcpControlSocket(String host,
-                                                           int port) throws InterruptedException, CannotConnectWithTorException {
-        int connectionAttempt = 0;
-        Exception lastException = null;
-
-        while (connectionAttempt < MAX_CONNECTION_ATTEMPTS) {
-            SocketChannel socketChannel = null;
-            try {
-                // The Tor Control Port communication is typically unencrypted.
-                // This is considered safe because the connection is made to 127.0.0.1 (localhost),
-                // ensuring that the communication does not leave the local machine and is not
-                // exposed to external networks. Authentication (via cookie or password)
-                // is handled by the Tor control protocol itself after connection.
-                //
-                // Explicitly use INET (IPv4) so the socket's local address appears as
-                // "127.0.0.1" in /proc/net/tcp.  Java's default NioSocketImpl creates
-                // AF_INET6 dual-stack sockets, causing the local address to appear as
-                // "::ffff:127.0.0.1" in /proc/net/tcp6.  On Tails, onion-grater resolves
-                // the client PID via psutil.net_connections() matching on local address;
-                // the IPv6-mapped form does not match, so PID lookup fails and
-                // onion-grater silently closes the connection.
-                socketChannel = SocketChannel.open(StandardProtocolFamily.INET);
-                var socketAddress = new InetSocketAddress(host, port);
-                log.debug("Attempting to connect to Tor control endpoint {} ({}/{})", socketAddress, connectionAttempt + 1, MAX_CONNECTION_ATTEMPTS);
-                socketChannel.connect(socketAddress);
-                log.info("Successfully connected control socket to Tor endpoint {} after {} attempts", socketAddress, connectionAttempt + 1);
-                return socketChannel;
-            } catch (ConnectException e) {
-                lastException = e;
-                log.warn("ConnectException on attempt {} to Tor control endpoint {}:{}: {}. Closing channel.",
-                        connectionAttempt + 1, host, port, e.getMessage());
-                if (socketChannel != null) {
-                    try {
-                        socketChannel.close();
-                    } catch (IOException closeEx) {
-                        log.warn("Failed to close channel after ConnectException on endpoint {}:{}: {}",
-                                host, port, closeEx.getMessage(), closeEx);
-                    }
-                }
-            } catch (IOException e) {
-                lastException = e;
-                log.warn("IOException on attempt {} to Tor control endpoint {}:{}: {}. Closing channel.",
-                        connectionAttempt + 1, host, port, e.getMessage());
-                if (socketChannel != null) {
-                    try {
-                        socketChannel.close();
-                    } catch (IOException closeEx) {
-                        log.warn("Failed to close channel after IOException on endpoint {}:{}: {}",
-                                host, port, closeEx.getMessage(), closeEx);
-                    }
-                }
-            }
-
-            connectionAttempt++;
-            if (connectionAttempt < MAX_CONNECTION_ATTEMPTS) {
-                log.debug("Connection attempt to Tor control endpoint {}:{} failed. Retrying in 200ms...", host, port);
-                Thread.sleep(200);
-            }
-        }
-
-        String errorMessage = "Failed to connect to Tor control endpoint " + host + ":" + port +
-                " after " + MAX_CONNECTION_ATTEMPTS + " attempts.";
-        IOException wrapperException = new IOException(errorMessage, lastException);
-        log.error(errorMessage, wrapperException);
-        throw new CannotConnectWithTorException(wrapperException);
-    }
-
-    private SocketChannel createAndConnectUnixControlSocket(Path controlSocketPath) throws InterruptedException, CannotConnectWithTorException {
-        int connectionAttempt = 0;
-        Exception lastException = null;
-
-        while (connectionAttempt < MAX_CONNECTION_ATTEMPTS) {
-            SocketChannel socketChannel = null;
-            try {
-                ProtocolFamily protocolFamily = StandardProtocolFamily.UNIX;
-                socketChannel = SocketChannel.open(protocolFamily);
-                UnixDomainSocketAddress socketAddress = UnixDomainSocketAddress.of(controlSocketPath);
-                log.debug("Attempting to connect to Tor control socket {} ({}/{})",
-                        controlSocketPath, connectionAttempt + 1, MAX_CONNECTION_ATTEMPTS);
-                socketChannel.connect(socketAddress);
-                log.info("Successfully connected control socket to Tor socket {} after {} attempts",
-                        controlSocketPath, connectionAttempt + 1);
-                return socketChannel;
-            } catch (IOException | UnsupportedOperationException e) {
-                lastException = e;
-                log.warn("Connection attempt {} to Tor control socket {} failed: {}. Closing channel.",
-                        connectionAttempt + 1, controlSocketPath, e.getMessage());
-                if (socketChannel != null) {
-                    try {
-                        socketChannel.close();
-                    } catch (IOException closeEx) {
-                        log.warn("Failed to close socket channel for control socket {}: {}",
-                                controlSocketPath, closeEx.getMessage(), closeEx);
-                    }
-                }
-            }
-
-            connectionAttempt++;
-            if (connectionAttempt < MAX_CONNECTION_ATTEMPTS) {
-                log.debug("Connection attempt to Tor control socket {} failed. Retrying in 200ms...", controlSocketPath);
-                Thread.sleep(200);
-            }
-        }
-
-        String errorMessage = "Failed to connect to Tor control socket " + controlSocketPath +
-                " after " + MAX_CONNECTION_ATTEMPTS + " attempts.";
-        IOException wrapperException = new IOException(errorMessage, lastException);
-        log.error(errorMessage, wrapperException);
-        throw new CannotConnectWithTorException(wrapperException);
-    }
-
     private void sendCommand(String command) {
         if (closeInProgress) {
             return;
```
```java
@Slf4j
public class TorControlSocketChannelFactory {
    private static final int MAX_CONNECTION_ATTEMPTS = 10;

    private final SocketAddress socketAddress;
    private int connectionAttempt;

    TorControlSocketChannelFactory(SocketAddress socketAddress) {
        this.socketAddress = socketAddress;
    }

    SocketChannel createAndConnect() throws InterruptedException {
        Exception lastException = null;

        while (connectionAttempt < MAX_CONNECTION_ATTEMPTS) {
            SocketChannel socketChannel = null;
            try {
                ++connectionAttempt;
                socketChannel = connectToControlSocketRaw();
                return socketChannel;

            } catch (Exception e) {
                lastException = e;
                log.warn("Connection attempt {} to Tor control endpoint {} failed. Closing channel.",
                        connectionAttempt, socketAddress, e);
                closeSocketChannel(socketChannel, socketAddress);
            }

            if (connectionAttempt < MAX_CONNECTION_ATTEMPTS) {
                log.debug("Connection attempt to Tor control endpoint {} failed. Retrying in 200ms...", socketAddress);
                //noinspection BusyWait
                Thread.sleep(200);
            }
        }

        String errorMessage = "Failed to connect to Tor control endpoint " + socketAddress +
                " after " + MAX_CONNECTION_ATTEMPTS + " attempts.";
        IOException wrapperException = new IOException(errorMessage, lastException);
        log.error(errorMessage, wrapperException);
        throw new CannotConnectWithTorException(wrapperException);
    }

    private SocketChannel connectToControlSocketRaw() throws IOException {
        log.debug("Attempting to connect to Tor control endpoint {} ({}/{})", socketAddress, connectionAttempt, MAX_CONNECTION_ATTEMPTS);
        SocketChannel socketChannel = SocketChannel.open(socketAddress);
        log.info("Successfully connected control socket to Tor endpoint {} after {} attempts", socketAddress, connectionAttempt);
        return socketChannel;
    }

    private void closeSocketChannel(SocketChannel socketChannel, SocketAddress socketAddress) {
        if (socketChannel != null) {
            try {
                socketChannel.close();
            } catch (IOException e) {
                log.warn("Failed to close channel on endpoint {}", socketAddress, e);
            }
        }
    }
}
```

`TorService`:

```diff
@@ -175,7 +177,8 @@
 
         FileMutatorUtils.deleteOnExit(controlPortFilePath);
 
-        torController.initialize(controlPort);
+        var socketAddress = new InetSocketAddress("127.0.0.1", controlPort);
+        torController.initialize(socketAddress);
         torController.authenticate(hashedControlPassword);
         torController.bootstrap();
 
@@ -288,69 +291,55 @@
         return socksProxyFactory.getSocks5Proxy(streamId);
     }
 
+    private void doCookieAuthentication(Map<String, String> torConfigMap) {
+        String authCookiePath = StringUtils.unquote(torConfigMap.get(COOKIE_AUTH_FILE));
+        try {
+            byte[] authCookie = Files.readAllBytes(Paths.get(authCookiePath));
+            torController.authenticate(authCookie);
+
+        } catch (IOException e) {
+            throw new TorControlAuthenticationFailed("Couldn't read the COOKIE_AUTH_FILE '" +
+                    authCookiePath + "' from the config.", e);
+
+        } catch (TorControlConnectionClosedException e) {
+            throw new TorControlAuthenticationFailed("Tor control connection was closed unexpectedly" +
+                    " when trying to authenticate.", e);
+
+        } catch (Exception e) {
+            throw new TorControlAuthenticationFailed("Authentication to Tor Control failed.", e);
+        }
+    }
+
     private boolean connectedToExternalTor(Map<String, String> torConfigMap) {
-        ControlEndpoint controlEndpoint = getControlEndpoint(torConfigMap);
-
         if (torController != null) {
             torController.shutdown();
         }
-        torController = new TorController(transportConfig.getBootstrapTimeout(), transportConfig.getHsUploadTimeout(), bootstrapEvent);
+
+        torController = new TorController(transportConfig.getBootstrapTimeout(),
+                transportConfig.getHsUploadTimeout(), bootstrapEvent);
+
         try {
-            if (controlEndpoint instanceof UnixSocketControlEndpoint socketControlEndpoint) {
-                torController.initialize(socketControlEndpoint.controlSocketPath());
-            } else if (controlEndpoint instanceof TcpControlEndpoint tcpControlEndpoint) {
-                torController.initialize(tcpControlEndpoint.controlHost(), tcpControlEndpoint.controlPort());
+            SocketAddress socketAddress = getControlSocketAddress(torConfigMap);
+            torController.initialize(socketAddress);
+
+            // No authentication required, but we still need to send an empty
+            // AUTHENTICATE call to be able to send control commands
+            torController.authenticate();
+
+            if (torConfigMap.containsKey(COOKIE_AUTH_FILE)) {
+                doCookieAuthentication(torConfigMap);
             }
+
         } catch (CannotConnectWithTorException e) {
             log.warn("Could not connect to external Tor control endpoint.", e);
             torController.shutdown();
             torController = null;
             return false;
-        }
 
-        boolean isAuthenticated = false;
-        boolean noCookieAuthenticationRequired = Optional.ofNullable(torConfigMap.get(COOKIE_AUTHENTICATION)).orElse("0").equals("0");
-        if (noCookieAuthenticationRequired) {
-            // No authentication required, but we still need to send an empty
-            // AUTHENTICATE call to be able to send control commands
-            try {
-                torController.authenticate();
-                isAuthenticated = true;
-            } catch (Exception e) {
-                log.warn("Authentication to Tor Control failed.", e);
-                return false;
-            }
+        } catch (TorControlAuthenticationFailed e) {
+            log.warn("Authentication to Tor Control failed.", e);
+            return false;
         }
-
-        if (!isAuthenticated) {
-            if (torConfigMap.containsKey(COOKIE_AUTH_FILE)) {
-                String authCookiePath = StringUtils.unquote(torConfigMap.get(COOKIE_AUTH_FILE));
-                try {
-                    byte[] authCookie = Files.readAllBytes(Paths.get(authCookiePath));
-                    torController.authenticate(authCookie);
-                    isAuthenticated = true;
-                } catch (IOException e) {
-                    log.warn("Couldn't read the COOKIE_AUTH_FILE '{}' from the config.", authCookiePath);
-                } catch (TorControlConnectionClosedException e) {
-                    log.warn("Tor control connection was closed unexpectedly when trying to authenticate.", e);
-                    return false; // since pipe is broken
-                } catch (Exception e) {
-                    log.warn("Authentication to Tor Control failed.", e);
-                }
-            } else {
-                log.warn("The user did not provide the path to the COOKIE_AUTH_FILE.");
-            }
-        }
-
-        if (!isAuthenticated) {
-            log.info("We try to call authenticate without parameters for the case that authentication is not required.");
-            try {
-                torController.authenticate();
-            } catch (Exception e) {
-                log.warn("Authentication to Tor Control failed.", e);
-                return false;
-            }
-        }
 
         // We do not call torController.bootstrap as we do not need to bootstrap Tor. Instead, we set the event to
         // trigger the application state update.
@@ -362,11 +351,11 @@
         return true;
     }
 
-    private ControlEndpoint getControlEndpoint(Map<String, String> externalTorConfigMap) {
+    private SocketAddress getControlSocketAddress(Map<String, String> externalTorConfigMap) {
         String controlSocketString = externalTorConfigMap.get(CONTROL_SOCKET);
         if (controlSocketString != null && !controlSocketString.isBlank()) {
             Path controlSocketPath = Path.of(StringUtils.unquote(controlSocketString));
-            return ControlEndpoint.forSocket(controlSocketPath);
+            return UnixDomainSocketAddress.of(controlSocketPath);
         }
 
         String controlPortString = externalTorConfigMap.get(CONTROL_PORT);
@@ -384,7 +373,7 @@
             controlHost = "127.0.0.1";
             controlPort = Integer.parseInt(tokens[0]);
         }
-        return ControlEndpoint.forPort(controlHost, controlPort);
+        return new InetSocketAddress(controlHost, controlPort);
     }
 
     private void readExternalTorConfigMap() {
```
```diff
-    private sealed interface ControlEndpoint permits TcpControlEndpoint, UnixSocketControlEndpoint {
-
-        static ControlEndpoint forPort(String controlHost, int controlPort) {
-            return new TcpControlEndpoint(controlHost, controlPort);
-        }
-
-        static ControlEndpoint forSocket(Path controlSocketPath) {
-            return new UnixSocketControlEndpoint(controlSocketPath);
-        }
-    }
-
-    private record TcpControlEndpoint(String controlHost, int controlPort) implements ControlEndpoint {
-        @Override
-        @NonNull
-        public String toString() {
-            return String.format("TcpControlEndpoint(host: %s, port: %d)", controlHost, controlPort);
-        }
-    }
-
-    private record UnixSocketControlEndpoint(Path controlSocketPath) implements ControlEndpoint {
-        @Override
-        @NonNull
-        public String toString() {
-            return String.format("UnixSocketControlEndpoint(controlSocketPath: %s)", controlSocketPath);
-        }
-    }
 }
```

- `torConfig.lines().toList()` converts a `Stream<String>` to a `List<String>` without using the list (only iterating).

```diff
@@ -392,20 +381,23 @@
             Path externalTorConfigPath = getExternalTorConfigPath();
             activeExternalTorConfigPath = Optional.of(externalTorConfigPath);
             externalTorConfigMap.clear();
-            String torConfig = readExternalTorConfig(externalTorConfigPath);
-            for (String line : torConfig.lines().toList()) {
-                line = line.trim();
-                if (!line.isEmpty() && !line.startsWith("#")) {
-                    int firstSpaceIndex = line.indexOf(" ");
-                    if (firstSpaceIndex != -1) {
-                        String key = line.substring(0, firstSpaceIndex);
-                        String value = line.substring(firstSpaceIndex + 1);
-                        externalTorConfigMap.put(key, value);
-                    } else {
-                        log.warn("Ignoring malformed line (no key/value separator) in external Tor config: '{}'", line);
-                    }
-                }
-            }
+
+            readExternalTorConfig(externalTorConfigPath)
+                    .lines()
+                    .map(String::trim)
+                    .filter(line -> !line.isEmpty() && !line.startsWith("#"))
+                    .forEach(line -> {
+                        int firstSpaceIndex = line.indexOf(" ");
+                        if (firstSpaceIndex != -1) {
+                            String key = line.substring(0, firstSpaceIndex);
+                            String value = line.substring(firstSpaceIndex + 1);
+                            externalTorConfigMap.put(key, value);
+                        } else {
+                            log.warn("Ignoring malformed line (no key/value separator) in external " +
+                                    "Tor config: '{}'", line);
+                        }
+                    });
+
         } catch (IOException e) {
             log.warn("Could not read external tor config file.", e);
         }
```

- The torrc config map to line conversion should happen in `TorrcFileGenerator` not `torrcClientConfigFactory`

`TorService`:

```diff
@@ -485,22 +477,21 @@
         torInstaller.installIfNotUpToDate();
     }
 
-    private List<String> createTorrcConfigFile(Path dataDirPath, PasswordDigest hashedControlPassword) {
+    private void createTorrcConfigFile(Path dataDirPath, PasswordDigest hashedControlPassword) {
         TorrcClientConfigFactory torrcClientConfigFactory = TorrcClientConfigFactory.builder()
                 .isTestNetwork(transportConfig.isTestNetwork())
                 .dataDirPath(dataDirPath)
                 .hashedControlPassword(hashedControlPassword)
                 .build();
 
-        Map<String, List<String>> overrides = resolveOverrides(dataDirPath);
-        List<String> torrcConfigLines = torrcClientConfigFactory.torrcClientConfigLines(overrides);
+        Map<String, List<String>> torrcOverrideConfigs = resolveOverrides(dataDirPath);
+        Map<String, List<String>> torrcConfigMap = torrcClientConfigFactory.torrcClientConfigMap(torrcOverrideConfigs);
 
         Path torrcPath = dataDirPath.resolve("torrc");
         var torrcFileGenerator = new TorrcFileGenerator(torrcPath,
-                torrcConfigLines,
+                torrcConfigMap,
                 transportConfig.getDirectoryAuthorities());
         torrcFileGenerator.generate();
-        return torrcConfigLines;
     }
 
     /**
```

`TorrcClientConfigFactory`:

```diff
@@ -46,20 +46,11 @@
         this.hashedControlPassword = hashedControlPassword;
     }
 
-    public List<String> torrcClientConfigLines(Map<String, List<String>> torrcOverrides) {
-        Map<String, String> baseConfig = clientTorrcGenerator().generate();
-        baseConfig.put(DISABLE_NETWORK, "1");
-
-        List<String> lines = new ArrayList<>();
-        baseConfig.forEach((key, value) -> {
-            if (!torrcOverrides.containsKey(key)) {
-                lines.add(key + " " + value);
-            }
-        });
-        torrcOverrides.forEach((key, values) ->
-                values.forEach(value -> lines.add(key + " " + value))
-        );
-        return lines;
+    public Map<String, List<String>> torrcClientConfigMap(Map<String, List<String>> torrcOverrides) {
+        Map<String, List<String>> torrcClientConfig = clientTorrcGenerator().generate();
+        torrcClientConfig.put(DISABLE_NETWORK, List.of("1"));
+        torrcClientConfig.putAll(torrcOverrides);
+        return torrcClientConfig;
     }
 
     TorrcConfigGenerator clientTorrcGenerator() {
```

`TorrcFileGenerator`:

```diff
@@ -21,42 +21,35 @@
 
 import java.io.IOException;
 import java.nio.file.Path;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 public class TorrcFileGenerator {
     private final Path torrcPath;
-    private final List<String> torrcConfigLines;
+    private final Map<String, List<String>> torrcConfigMap;
     private final Set<DirectoryAuthority> customDirectoryAuthorities;
 
     public TorrcFileGenerator(Path torrcPath,
-                              List<String> torrcConfigLines,
+                              Map<String, List<String>> torrcConfigMap,
                               Set<DirectoryAuthority> customDirectoryAuthorities) {
         this.torrcPath = torrcPath;
-        this.torrcConfigLines = torrcConfigLines;
+        this.torrcConfigMap = torrcConfigMap;
         this.customDirectoryAuthorities = customDirectoryAuthorities;
     }
 
-    // Constructor to keep the change to List<String> smaller in scope
-    public TorrcFileGenerator(Path torrcPath,
-                              Map<String, String> torrcConfigMap,
-                              Set<DirectoryAuthority> customDirectoryAuthorities) {
-        this(torrcPath, mapToLines(torrcConfigMap), customDirectoryAuthorities);
-    }
-
-    private static List<String> mapToLines(Map<String, String> map) {
-        List<String> lines = new ArrayList<>();
-        map.forEach((key, value) -> lines.add(key + " " + value));
-        return lines;
-    }
-
     public void generate() {
         StringBuilder torrcStringBuilder = new StringBuilder();
-        torrcConfigLines.forEach(line ->
-                torrcStringBuilder.append(line).append("\n")
-        );
+
+        for (Map.Entry<String, List<String>> entry : torrcConfigMap.entrySet()) {
+            String key = entry.getKey();
+            for (String value : entry.getValue()) {
+                torrcStringBuilder.append(key)
+                        .append(" ")
+                        .append(value)
+                        .append("\n");
+            }
+        }
 
         customDirectoryAuthorities.forEach(dirAuthority ->
                 torrcStringBuilder.append("DirAuthority ").append(dirAuthority.getNickname())
```

- It's better to add multi-value support at the TorrcConfigGenerator interface level.

`TorrcConfigGenerator`:

```diff
@@ -17,8 +17,9 @@
 
 package bisq.network.tor.common.torrc;
 
+import java.util.List;
 import java.util.Map;
 
 public interface TorrcConfigGenerator {
-    Map<String, String> generate();
+    Map<String, List<String>> generate();
 }
```

`BaseTorrcGenerator`:

```diff
@@ -48,20 +49,20 @@
     }
 
     @Override
-    public Map<String, String> generate() {
-        Map<String, String> torConfigMap = new LinkedHashMap<>();
-        torConfigMap.put(DATA_DIRECTORY, dataDirPath.toAbsolutePath().toString());
+    public Map<String, List<String>> generate() {
+        Map<String, List<String>> torConfigMap = new LinkedHashMap<>();
+        torConfigMap.put(DATA_DIRECTORY, List.of(dataDirPath.toAbsolutePath().toString()));
 
-        torConfigMap.put(CONTROL_PORT, CONTROL_PORT_AUTO);
-        torConfigMap.put(CONTROL_PORT_WRITE_TO_FILE, controlPortWriteFilePath.toAbsolutePath().toString());
-        torConfigMap.put(HASHED_CONTROL_PASSWORD, hashedControlPassword);
+        torConfigMap.put(CONTROL_PORT, List.of(CONTROL_PORT_AUTO));
+        torConfigMap.put(CONTROL_PORT_WRITE_TO_FILE, List.of(controlPortWriteFilePath.toAbsolutePath().toString()));
+        torConfigMap.put(HASHED_CONTROL_PASSWORD, List.of(hashedControlPassword));
 
         String logLevel = isTestNetwork ? "debug" : "notice";
         torConfigMap.put("Log",
-                logLevel + " file " + dataDirPath.resolve("debug.log").toAbsolutePath()
+                List.of(logLevel + " file " + dataDirPath.resolve("debug.log").toAbsolutePath())
         );
 
-        torConfigMap.put(SOCKS_PORT, SOCKS_PORT_DISABLED);
+        torConfigMap.put(SOCKS_PORT, List.of(SOCKS_PORT_DISABLED));
         return torConfigMap;
     }
 }
```

`ClientTorrcGenerator`:

```diff
@@ -33,9 +34,9 @@
     }
 
     @Override
-    public Map<String, String> generate() {
-        Map<String, String> torConfigMap = baseTorrcConfigGenerator.generate();
-        torConfigMap.put(SOCKS_PORT,  SOCKS_PORT_AUTO);
+    public Map<String, List<String>> generate() {
+        Map<String, List<String>> torConfigMap = baseTorrcConfigGenerator.generate();
+        torConfigMap.put(SOCKS_PORT,  List.of(SOCKS_PORT_AUTO));
         return torConfigMap;
     }
 }
```

`DirectoryAuthorityTorrcGenerator`:

```diff
@@ -29,24 +30,24 @@
     }
 
     @Override
-    public Map<String, String> generate() {
-        Map<String, String> torConfigMap = baseTorrcConfigGenerator.generate();
+    public Map<String, List<String>> generate() {
+        Map<String, List<String>> torConfigMap = baseTorrcConfigGenerator.generate();
 
-        torConfigMap.put("AuthoritativeDirectory", "1");
-        torConfigMap.put("V3AuthoritativeDirectory", "1");
-        torConfigMap.put("ContactInfo", "auth-" + nickname + "@test.test\n");
+        torConfigMap.put("AuthoritativeDirectory", List.of("1"));
+        torConfigMap.put("V3AuthoritativeDirectory", List.of("1"));
+        torConfigMap.put("ContactInfo", List.of("auth-" + nickname + "@test.test\n"));
 
-        torConfigMap.put("AssumeReachable", "1");
-        torConfigMap.put("TestingV3AuthInitialVotingInterval", "20");
+        torConfigMap.put("AssumeReachable", List.of("1"));
+        torConfigMap.put("TestingV3AuthInitialVotingInterval", List.of("20"));
 
-        torConfigMap.put("TestingV3AuthInitialVoteDelay", "4");
-        torConfigMap.put("TestingV3AuthInitialDistDelay", "4");
+        torConfigMap.put("TestingV3AuthInitialVoteDelay", List.of("4"));
+        torConfigMap.put("TestingV3AuthInitialDistDelay", List.of("4"));
 
-        torConfigMap.put("V3AuthVotingInterval", "20");
-        torConfigMap.put("V3AuthVoteDelay", "4");
-        torConfigMap.put("V3AuthDistDelay", "4");
+        torConfigMap.put("V3AuthVotingInterval", List.of("20"));
+        torConfigMap.put("V3AuthVoteDelay", List.of("4"));
+        torConfigMap.put("V3AuthDistDelay", List.of("4"));
 
-        torConfigMap.put("ExitPolicy", "accept *:*");
+        torConfigMap.put("ExitPolicy", List.of("accept *:*"));
 
         return torConfigMap;
```

`RelayTorrcGenerator`:

```diff
@@ -27,12 +28,12 @@
     }
 
     @Override
-    public Map<String, String> generate() {
-        Map<String, String> torConfigMap = baseTorrcConfigGenerator.generate();
+    public Map<String, List<String>> generate() {
+        Map<String, List<String>> torConfigMap = baseTorrcConfigGenerator.generate();
 
-        torConfigMap.put("ExitRelay", "1");
-        torConfigMap.put("ExitPolicy", "accept 127.0.0.0/8:*,accept private:*,accept *:*,reject *:*");
-        torConfigMap.put("ExitPolicyRejectPrivate", "0");
+        torConfigMap.put("ExitRelay", List.of("1"));
+        torConfigMap.put("ExitPolicy", List.of("accept 127.0.0.0/8:*,accept private:*,accept *:*,reject *:*"));
+        torConfigMap.put("ExitPolicyRejectPrivate", List.of("0"));
 
         return torConfigMap;
     }
```

`TestNetworkTorrcGenerator`:

```diff
@@ -40,33 +41,32 @@
     private final Optional<Integer> dirPort = Optional.empty();
 
     @Override
-    public Map<String, String> generate() {
-        Map<String, String> torConfigMap = baseTorrcConfigGenerator.generate();
+    public Map<String, List<String>> generate() {
+        Map<String, List<String>> torConfigMap = baseTorrcConfigGenerator.generate();
 
-        torConfigMap.put("TestingTorNetwork", "1");
-        torConfigMap.put("TestingDirAuthVoteExit", "*");
-        torConfigMap.put("TestingDirAuthVoteHSDir", "*");
+        torConfigMap.put("TestingTorNetwork", List.of("1"));
+        torConfigMap.put("TestingDirAuthVoteExit", List.of("*"));
+        torConfigMap.put("TestingDirAuthVoteHSDir", List.of("*"));
 
-        torConfigMap.put("V3AuthNIntervalsValid", "2");
-        torConfigMap.put("TestingDirAuthVoteGuard", "*");
-        torConfigMap.put("TestingMinExitFlagThreshold", "0");
+        torConfigMap.put("V3AuthNIntervalsValid", List.of("2"));
+        torConfigMap.put("TestingDirAuthVoteGuard", List.of("*"));
+        torConfigMap.put("TestingMinExitFlagThreshold", List.of("0"));
 
+        nickname.ifPresent(nickname -> torConfigMap.put("Nickname", List.of(nickname)));
+        torConfigMap.put("ShutdownWaitLength", List.of("2"));
+        torConfigMap.put("DisableDebuggerAttachment", List.of("0"));
 
-        nickname.ifPresent(nickname -> torConfigMap.put("Nickname", nickname));
-        torConfigMap.put("ShutdownWaitLength", "2");
-        torConfigMap.put("DisableDebuggerAttachment", "0");
+        torConfigMap.put("ProtocolWarnings", List.of("1"));
+        torConfigMap.put("SafeLogging", List.of("0"));
+        torConfigMap.put("LogTimeGranularity", List.of("1"));
 
-        torConfigMap.put("ProtocolWarnings", "1");
-        torConfigMap.put("SafeLogging", "0");
-        torConfigMap.put("LogTimeGranularity", "1");
+        orPort.ifPresent(orPort -> torConfigMap.put("OrPort", List.of(String.valueOf(orPort))));
+        torConfigMap.put("Address", List.of("127.0.0.1"));
+        torConfigMap.put("ServerDNSDetectHijacking", List.of("0"));
 
-        orPort.ifPresent(orPort -> torConfigMap.put("OrPort", String.valueOf(orPort)));
-        torConfigMap.put("Address", "127.0.0.1");
-        torConfigMap.put("ServerDNSDetectHijacking", "0");
+        torConfigMap.put("ServerDNSTestAddresses", List.of(""));
 
-        torConfigMap.put("ServerDNSTestAddresses", "");
-
-        dirPort.ifPresent(dirPort -> torConfigMap.put("DirPort", String.valueOf(dirPort)));
+        dirPort.ifPresent(dirPort -> torConfigMap.put("DirPort", List.of(String.valueOf(dirPort))));
 
         return torConfigMap;
     }
```
- The `torrcOverrides` config options doesn't restrict nested data types.

`TorTransportConfig`:

```diff
@@ -51,8 +49,8 @@
                 (int) TimeUnit.SECONDS.toMillis(config.getInt("socketTimeout")),
                 config.getBoolean("testNetwork"),
                 parseDirectoryAuthorities(config.getList("directoryAuthorities")),
-                parseTorrcOverrideConfig(config.getConfig("torrcOverrides")),
+                parseTorrcOverrideConfig(config.getObject("torrcOverrides")),
                 config.getInt("sendMessageThrottleTime"),
                 config.getInt("receiveMessageThrottleTime"),
                 config.getBoolean("useExternalTor")
@@ -75,49 +73,24 @@
         return allDirectoryAuthorities;
     }
 
-    private static Map<String, List<String>> parseTorrcOverrideConfig(com.typesafe.config.Config torrcOverrides) {
-        Map<String, List<String>> result = new LinkedHashMap<>();
-        torrcOverrides.entrySet().forEach(entry -> {
-            Object unwrapped = entry.getValue().unwrapped();
-            if (unwrapped instanceof List<?> values) {
-                List<String> stringValues = new ArrayList<>();
-                values.forEach(v -> stringValues.add(String.valueOf(v)));
-                result.put(entry.getKey(), stringValues);
-            } else {
-                result.put(entry.getKey(), List.of(String.valueOf(unwrapped)));
-            }
-        });
-        return result;
+    private static Map<String, List<String>> parseTorrcOverrideConfig(ConfigObject torrcOverrides) {
+        return TorrcOverrideConfigParser.parse(torrcOverrides);
     }
```

`TorrcOverrideConfigParser`:

```java
public class TorrcOverrideConfigParser {

    static Map<String, List<String>> parse(ConfigObject torrcOverrides) {
        return torrcOverrides.entrySet().stream()
                .map(TorrcOverrideConfigParser::tryToParseTorrcOverrideEntry)
                .filter(TorrcOverrideConfigParser::isValidTorrcOverrideEntry)
                .map(TorrcOverrideConfigParser::flattenTorrcOverrideOptionalValue)
                .collect(Collectors.toMap(Pair::getFirst, Pair::getSecond));
    }

    private static Pair<String, Optional<List<String>>> tryToParseTorrcOverrideEntry(Map.Entry<String, ConfigValue> entry) {
        String key = entry.getKey();
        ConfigValue configValue = entry.getValue();

        Optional<List<String>> value = tryToParseBasicDataType(configValue)
                .map(List::of)
                .or(tryToParseObject(configValue));

        return new Pair<>(key, value);
    }

    private static boolean isValidTorrcOverrideEntry(Pair<String, Optional<List<String>>> keyValuePair) {
        Optional<List<String>> optionalValue = keyValuePair.getSecond();
        return optionalValue.isPresent();
    }

    private static Pair<String, List<String>> flattenTorrcOverrideOptionalValue(Pair<String, Optional<List<String>>> keyValuePair) {
        //noinspection OptionalGetWithoutIsPresent
        return new Pair<>(keyValuePair.getFirst(), keyValuePair.getSecond().get());
    }

    private static Optional<String> tryToParseBasicDataType(ConfigValue configValue) {
        return tryToParseNumber(configValue)
                .or(tryToParseString(configValue))
                .or(tryToParseBoolean(configValue));
    }

    private static Supplier<Optional<? extends List<String>>> tryToParseObject(ConfigValue configValue) {
        return () -> {
            if (configValue.valueType() == ConfigValueType.OBJECT) {
                ConfigObject valueObject = (ConfigObject) configValue;

                List<String> values = valueObject.values()
                        .stream()
                        .map(TorrcOverrideConfigParser::tryToParseBasicDataType)
                        .filter(Optional::isPresent)
                        .map(Optional::get)
                        .toList();

                return Optional.of(values);
            }
            return Optional.empty();
        };
    }

    private static Optional<String> tryToParseNumber(ConfigValue configValue) {
        if (configValue.valueType() == ConfigValueType.NUMBER) {
            int value = (int) configValue.unwrapped();
            return Optional.of(String.valueOf(value));
        }
        return Optional.empty();
    }

    private static Supplier<Optional<? extends String>> tryToParseString(ConfigValue configValue) {
        return () -> {
            if (configValue.valueType() == ConfigValueType.STRING) {
                String value = (String) configValue.unwrapped();
                return Optional.of(value);
            }
            return Optional.empty();
        };
    }

    private static Supplier<Optional<? extends String>> tryToParseBoolean(ConfigValue configValue) {
        return () -> {
            if (configValue.valueType() == ConfigValueType.BOOLEAN) {
                boolean value = (boolean) configValue.unwrapped();
                return Optional.of(String.valueOf(value));
            }
            return Optional.empty();
        };
    }
}
```

- The `torrcOverrideFilePath` argument is optional, so it's better to use the `Optional<String>` data type.

`TorTransportConfig`:

```diff
@@ -51,8 +49,8 @@
                 (int) TimeUnit.SECONDS.toMillis(config.getInt("socketTimeout")),
                 config.getBoolean("testNetwork"),
                 parseDirectoryAuthorities(config.getList("directoryAuthorities")),
-                config.hasPath("torrcOverrideFilePath") ? config.getString("torrcOverrideFilePath") : "",
+                parseTorrcOverrideFilePath(dataDirPath, config),
                 config.getInt("sendMessageThrottleTime"),
                 config.getInt("receiveMessageThrottleTime"),
                 config.getBoolean("useExternalTor")
```

```diff
-    /**
-     * Parses a torrc-style override file into a map of key → list of values.
-     * Each non-blank, non-comment line is expected to have the form {@code Key Value}.
-     * Repeated keys (e.g. multiple {@code Bridge} lines) accumulate into a list so that
-     * all entries appear in the generated torrc.
-     */
-    public static Map<String, List<String>> parseTorrcOverrideFile(Path filePath) throws IOException {
-        Map<String, List<String>> result = new LinkedHashMap<>();
-        for (String line : Files.readAllLines(filePath)) {
-            String trimmed = line.strip();
-            if (trimmed.isEmpty() || trimmed.startsWith("#")) {
-                continue;
-            }
-            int spaceIndex = -1;
-            for (int i = 0; i < trimmed.length(); i++) {
-                if (Character.isWhitespace(trimmed.charAt(i))) {
-                    spaceIndex = i;
-                    break;
-                }
-            }
-            if (spaceIndex < 0) {
-                continue; // bare key with no value — skip
-            }
-            String key = trimmed.substring(0, spaceIndex);
-            String value = trimmed.substring(spaceIndex + 1).strip();
-            result.computeIfAbsent(key, k -> new ArrayList<>()).add(value);
+    private static Optional<Path> parseTorrcOverrideFilePath(Path dataDirPath,
+                                                             com.typesafe.config.Config config) {
+        if (!config.hasPath("torrcOverrideFilePath")) {
+            return Optional.empty();
+        }
+
+        String filePath = config.getString("torrcOverrideFilePath");
+        if (filePath.isBlank()) {
+            return Optional.empty();
         }
-        return result;
+
+        Path torrcOverridePath = Path.of(filePath);
+        return torrcOverridePath.isAbsolute() ? Optional.of(torrcOverridePath) :
+                Optional.of(dataDirPath.resolve(torrcOverridePath));
     }
 
     private static String getStringFromConfigValue(ConfigValue configValue, String key) {
@@ -136,13 +109,7 @@
     private final boolean isTestNetwork;
     private final Set<DirectoryAuthority> directoryAuthorities;
     private final Map<String, List<String>> torrcOverrides;
-    /**
-     * Optional path to a torrc-style file whose entries override {@code torrcOverrides}.
-     * Supports both absolute paths and paths relative to the data directory.
-     * When non-empty this file takes precedence over the inline {@code torrcOverrides} map.
-     * Leave empty (the default) to use {@code torrcOverrides} instead.
-     */
-    private final String torrcOverrideFilePath;
+    private final Optional<Path> torrcOverrideFilePath;
     private final int sendMessageThrottleTime;
     private final int receiveMessageThrottleTime;
     private final boolean useExternalTor;
@@ -155,7 +122,7 @@
                               boolean isTestNetwork,
                               Set<DirectoryAuthority> directoryAuthorities,
                               Map<String, List<String>> torrcOverrides,
-                              String torrcOverrideFilePath,
+                              Optional<Path> torrcOverrideFilePath,
                               int sendMessageThrottleTime,
                               int receiveMessageThrottleTime,
                               boolean useExternalTor) {
```

`TorService`:

```diff
@@ -430,9 +422,9 @@
     }
 
     private Path getExternalTorConfigPath() {
-        String torrcOverrideFilePath = transportConfig.getTorrcOverrideFilePath();
-        if (torrcOverrideFilePath != null && !torrcOverrideFilePath.isBlank()) {
-            Path overrideFilePath = resolveTorrcOverrideFilePath(transportConfig.getDataDirPath(), torrcOverrideFilePath);
+        Optional<Path> torrcOverrideFilePath = transportConfig.getTorrcOverrideFilePath();
+        if (torrcOverrideFilePath.isPresent()) {
+            Path overrideFilePath = torrcOverrideFilePath.get();
             if (Files.exists(overrideFilePath)) {
                 return overrideFilePath;
             }
@@ -485,22 +477,21 @@
         torInstaller.installIfNotUpToDate();
     }
 
-    private List<String> createTorrcConfigFile(Path dataDirPath, PasswordDigest hashedControlPassword) {
+    private void createTorrcConfigFile(Path dataDirPath, PasswordDigest hashedControlPassword) {
         TorrcClientConfigFactory torrcClientConfigFactory = TorrcClientConfigFactory.builder()
                 .isTestNetwork(transportConfig.isTestNetwork())
                 .dataDirPath(dataDirPath)
                 .hashedControlPassword(hashedControlPassword)
                 .build();
 
-        Map<String, List<String>> overrides = resolveOverrides(dataDirPath);
-        List<String> torrcConfigLines = torrcClientConfigFactory.torrcClientConfigLines(overrides);
+        Map<String, List<String>> torrcOverrideConfigs = resolveOverrides(dataDirPath);
+        Map<String, List<String>> torrcConfigMap = torrcClientConfigFactory.torrcClientConfigMap(torrcOverrideConfigs);
 
         Path torrcPath = dataDirPath.resolve("torrc");
         var torrcFileGenerator = new TorrcFileGenerator(torrcPath,
-                torrcConfigLines,
+                torrcConfigMap,
                 transportConfig.getDirectoryAuthorities());
         torrcFileGenerator.generate();
-        return torrcConfigLines;
     }
 
     /**
@@ -510,52 +501,19 @@
      * file path is empty or the file cannot be read.
      */
     private Map<String, List<String>> resolveOverrides(Path dataDirPath) {
-        String filePathStr = transportConfig.getTorrcOverrideFilePath();
-        if (filePathStr == null || filePathStr.isBlank()) {
+        Optional<Path> torrcOverrideFilePath = transportConfig.getTorrcOverrideFilePath();
+        if (torrcOverrideFilePath.isEmpty()) {
             return transportConfig.getTorrcOverrides();
         }
+
+        Path overrideFilePath = torrcOverrideFilePath.get();
         try {
-            Path overrideFilePath = resolveTorrcOverrideFilePath(dataDirPath, filePathStr);
-            return TorTransportConfig.parseTorrcOverrideFile(overrideFilePath);
+            return TorrcFileParser.parseTorrcOverrideFile(overrideFilePath);
+
         } catch (IOException e) {
             log.warn("Could not read torrcOverrideFilePath '{}', falling back to torrcOverrides. Error: {}",
-                    filePathStr, e.getMessage());
+                    overrideFilePath, e.getMessage());
             return transportConfig.getTorrcOverrides();
         }
     }
-
-    private Path resolveTorrcOverrideFilePath(Path dataDirPath, String torrcOverrideFilePath) {
-        Path overrideFilePath = Path.of(torrcOverrideFilePath);
-        if (!overrideFilePath.isAbsolute()) {
-            overrideFilePath = dataDirPath.resolve(overrideFilePath);
-        }
-        return overrideFilePath;
-    }
```

- Use dynamic path lookup so that Gradle handles directory setup and cross platform path lookup ("/", "\\")

`PackagePlugin`:

```kotlin
            val onionGraterDirectory: Directory = project.rootProject.layout
                .projectDirectory
                .dir("tor")
                .dir("tor")
                .dir("onion-grater")
            onionGraterDir.set(onionGraterDirectory)
```

- `prepare-tails.sh`
	- Better to use `cp` instead of `rsync` (`cp` does exist always)

- `TorControlReader`: Use `add` in line 92 because `offer` can silently fail (correct in line 110)


## Code Review Notes
- Check if all .conf files updated
	- [x] ./apps/api-app/src/main/resources/api_app.conf
	- [x] ./apps/desktop/desktop-app/src/main/resources/desktop.conf
	- [x] ./apps/node-monitor-web-app/src/main/resources/node_monitor.conf
	- [x] ./apps/oracle-node-app/src/main/resources/oracle_node.conf
	- [x] ./apps/seed-node-app/src/main/resources/seed_node.conf
	- [ ] ~~Looks outdated - ./network/network/src/integrationTest/resources/test.conf~~
- ConfigUtils.parseBooleanFlag edge-cases:
	- [x] null
	- [x] leading-trailing blanks
	- [x] 1, 0, -1
	- [x] true, True, tRue
	- [x] yes, Yes, yEs
	- [ ] "\u00A0true" (unicode whitespace) - fails but not important
- StringUtils.unquote (looks good!) - edge-cases:
	- [ ] null - leads to NullPointerException
	- In ConfigUtils.parseBooleanFlag you used String.trim to remove blanks, you could use it here too for consistency, but not needed
	- [x] Doesn't support escapes and nesting, but not needed
- Check official Tails marker files
	- Legacy/Marker file: /etc/amnesia_version
	- Recommended: /etc/os-release
		- *ID=tails
		- *NAME="Tails"

- Lookup `--app-content`

```
  --app-content <additional content>[,<additional content>...]
          A comma separated list of paths to files and/or directories
          to add to the application payload.
          This option can be used more than once.
```
- Check for `onion-grater-add`, we officially support Whonix and Bisq's profile is integrated upstream into the OS.
	- `onion-grater-add` missing in Tails but `onion-grater` running
		- `onion-grater` not in `PATH` but listening on port 951? - (`ss -tlpn`)
 			- find binary through `cmdline` in `/proc/<pid>/cmdline`
				- binary in `/usr/local/lib/`
				- cross-reference with Tails source code - https://gitlab.tails.boum.org/tails/tails/-/blob/stable/config/chroot_local-includes/usr/local/lib/onion-grater
	- Check included profiles: onion-share, Tor Browser, ...
		- Which package installed and why Bisq missing? - `dpkg -S` not working?
			- Cross-reference with `ls`
			- Check Tails source code: https://gitlab.tails.boum.org/tails/tails/-/tree/stable/config/chroot_local-includes/etc/onion-grater.d

[^1]: https://github.com/Whonix/onion-grater/pull/11

wodoro:
Keeping explicit AF_INET. `open(socketAddress)`reintroduces `::ffff:127.0.0.1` dual-stack, breaks onion-grater on Tails (the bug this PR fixes) 

Working on the rest

wodoro:
I have tried using the shipped profile and it doesn't work, its missing an important pattern: 
```
- pattern:       'net/listeners/socks'
```

also using `exe-paths` didnt work (with correct path or '*', tried both), but using `apparmor-profiles` with correct path does work

wodoro:
applied requested changes and rebased on main
testing again after rebase and after that it will be ready for review again

wodoro:
tested deb build on tails, inbound and outbound connections made successfully, offers visible, and everything seems to be working as expected
please review again

apps/desktop/desktop-app-launcher/package/linux/prepare_tails.sh
```diff
@@ -0,0 +1,94 @@
...
+        exit 1
+    fi
+    echo "Symlink already exists: $TARGET_LINK -> $CURRENT_TARGET"
+elif [ -d "$TARGET_LINK" ]; then
+    BACKUP_DIR="${TARGET_LINK}.pre-tails-migration.$(date +%Y%m%d%H%M%S)"
+    echo "Found existing non-persistent data at $TARGET_LINK; migrating into $DataDirectory ..."
+    # Use cp (always present) instead of rsync (not guaranteed). cp -a == -dR --preserve=all,
+    # preserving mode, ownership, timestamps, symlinks, hard links, ACLs and xattrs. The trailing
+    # "/." copies the directory contents (including dotfiles) into the existing $DataDirectory.
+    cp -a -- "$TARGET_LINK"/. "$DataDirectory"/
```
KimStrand: This overwrites matching files when the persistent destination already contains a Bisq profile. The backup below preserves the source directory, not the destination being overwritten, so existing identity and trade data can be lost.

Should we reject a populated destination and preserve both complete profiles before proceeding?

apps/desktop/desktop-app-launcher/package/linux/prepare_tails.sh
```diff
@@ -0,0 +1,94 @@
...
+    mv -- "$TARGET_LINK" "$BACKUP_DIR"
+    echo "Backed up original directory to $BACKUP_DIR"
+    chown -R amnesia:amnesia "$DataDirectory"
+    ln -s "$DataDirectory" "$TARGET_LINK"
+    echo "Created symlink: $TARGET_LINK -> $DataDirectory"
+elif [ -e "$TARGET_LINK" ]; then
+    echo "Error: $TARGET_LINK exists and is neither a symlink nor a directory; refusing to overwrite. Remove it manually and re-run." >&2
+    exit 1
+else
+    ln -s "$DataDirectory" "$TARGET_LINK"
```
KimStrand: This link is created only in Tails’ volatile home directory. After reboot, the persistent data survives but the default Bisq directory no longer points to it. Enabling Dotfiles alone does not register this link for restoration.

Should we persist the mapping through Dotfiles, or configure the launcher to use the persistent directory directly?

network/tor/tor/src/main/java/bisq/network/tor/TorrcOverrideConfigParser.java
```diff
@@ -0,0 +1,121 @@
...
+                        .toList();
+
+                return Optional.of(values);
+            }
+            return Optional.empty();
+        };
+    }
+
+    private static Optional<String> tryToParseNumber(ConfigValue configValue) {
+        if (configValue.valueType() == ConfigValueType.NUMBER) {
```
KimStrand, line 95: NUMBER values can unwrap to Long or Double, so this cast can throw ClassCastException during startup. For example, CircuitPriorityHalflife=1.5 reproduces the Double to Integer failure.

Should we convert the numeric value without narrowing it to int

network/tor/tor/src/main/java/bisq/network/tor/TorService.java
```diff
@@ -374,46 +387,65 @@ private Pair<String, Integer> getControlHostAndPort(Map<String, String> external
...
+            Path externalTorConfigPath = getExternalTorConfigPath();
+            activeExternalTorConfigPath = Optional.of(externalTorConfigPath);
+            externalTorConfigMap.clear();
+
+            readExternalTorConfig(externalTorConfigPath)
+                    .lines()
+                    .map(String::trim)
+                    .filter(line -> !line.isEmpty() && !line.startsWith("#"))
+                    .forEach(line -> {
+                        int firstSpaceIndex = line.indexOf(" ");
```
KimStrand: Reusing a torrc file through torrcOverrideFilePath fails with ordinary torrc syntax: a tab between ControlPort and 9051 causes the directive to be ignored, while “ControlPort 9051 # comment” produces NumberFormatException. Both cases were reproduced against this revision.

Should we handle whitespace and comments outside quoted values consistently when parsing external Tor settings?

wodoro:
addressed comments, rebased on main
need to manually test tails again

apps/desktop/desktop-app-launcher/package/linux/prepare_tails.sh
```diff
@@ -0,0 +1,96 @@
...
+systemctl restart onion-grater.service
+
+# Bisq2 uses Persistent Storage as its data directory whenever it is unlocked, so no symlink is
+# needed. Running amnesically is a supported choice, so a locked volume is a warning, not an error.
+if [ -d "$PERSISTENT_STORAGE_DIR" ]; then
+    DATA_DIR="$PERSISTENT_DATA_DIR"
+    mkdir -p "$DATA_DIR"
+
+    # Earlier versions wrote to the volatile home directory; migrate that data once.
+    if [ -L "$VOLATILE_DATA_DIR" ]; then
```
KimStrand, line 62: The previous helper linked ~/.local/share/Bisq2 to ~/Persistent/bisq/Bisq2. This removes that link without migrating its target, while the new default uses ~/Persistent/Bisq2. With that existing layout, the helper reports success but leaves the new directory empty, so Bisq2 no longer loads the saved identity and trade data. The original files remain at the old location. Should we preserve or migrate the existing profile before removing the link and switching paths?

wodoro: This is not shipped yet, so why do we need a migration?

KimStrand: sorry my mistake!

wodoro:
rebased on main
also tested the built deb pacakge on tail, it connects and receives data
I did not test permanent storage functionality (I plan to, but will take a while, as my setup does not support using it)

majesticsky:
Tested fd8a245 on Tails 7.13 (Persistent Storage, default features). Is reinstalling the .deb every session the intended flow? **Bisq doesn't survive a reboot, only its data does:**

- After reboot `/opt/bisq2` and `/etc/onion-grater.d/bisq.yml` are gone; `~/Persistent/Bisq2` is intact (same onion key, no re-onboarding).
- Additional Software can't keep it: `sudo apt install ./bisq2_*.deb` aborts in Tails' `asp-pre-apt` hook (`KeyError: ... no package named 'bisq2'`). `dpkg -i` works but isn't remembered.

Minor:
- The no-persistence popup says "with the Dotfiles feature", but the data dir is now `~/Persistent` (Persistent Folder). Works with Dotfiles off.
- That warning never reaches `bisq.log`: `evaluateTailsDataDirNonPersistent()` runs before `setupLogging()`.

Everything else works: onion-grater profile via postinst, auto `external_tor.config` (port 951), locked-storage popup, `prepare_tails.sh` with locked storage (warns, exits 0). **Inbound connections work**: the log shows `InboundConnection`s from 10 distinct peers, and an external SOCKS connect to the node's onion succeeds, so the loopback-bind issue @alvasw raised in [his review (2026-06-13)](https://github.com/bisq-network/bisq2/pull/4643#pullrequestreview-4491499650) doesn't reproduce on 7.13. I didn't publish an offer, so the maker flow itself is untested.

Not tested: Whonix, upgrade path, the volatile-dir migration in `prepare_tails.sh`.

wodoro:
Yes, installing .deb every session is the intended flow. I don't see the point of using Tails if anyone is going to just keep the installed programs, etc. 
I haven't get to trying the persistent storage myself, but if it's keeping its data across reboots where only `~/Persistent` survives, then it's the desired behavior.

wodoro:
Actually, I'll investigate a bit more, I think I forgot something
Thanks for the report!

wodoro:
799c248: migrate the legacy data dir in-app and drop prepare_tails.sh

- Popup wording: the no-persistence popup no longer mentions Dotfiles; the data dir is ~/Persistent, so Persistent Folder is what matters. It also says to unlock Persistent Storage at the Welcome Screen, since unlocking needs a Tails restart and not just a Bisq restart.
- Log ordering: the non-persistent data dir warning is now logged after logging is set up, so it reaches bisq.log.
- prepare_tails.sh is removed. Nothing ran it automatically, so its migration only reached users who found the script by hand. It was also wrong for the most likely legacy setup: data persisted with the Tails Dotfiles feature. Dotfiles symlinks each file into the home directory, and Bisq writes a store by renaming a temp file over it. That breaks the link, so a Dotfiles-backed data dir restores the same stale snapshot on every boot, and the script copied the symlinks instead of the data. The onion-grater profile is still installed by the deb postinst.
- The migration now runs in the app. On first start, if ~/Persistent/Bisq2 doesn't exist yet and ~/.local/share/Bisq2 has data, Bisq copies it over:
  - Symlinks are followed, so real files are copied. This covers a Dotfiles copy, a symlinked data dir, and symlinked subdirectories.
  - instance.lock and the stale tor/external_tor.config are skipped, so the Tails defaults (onion-grater, port 951) get written.
  - The copy goes into a sibling directory that is renamed into place atomically.
  - The source is never modified or deleted, and an existing data dir is never overwritten.
- Dotfiles cleanup popup: if the data came from a Dotfiles copy, the desktop app shows a one-time popup asking the user to delete that outdated copy.
- A stale javadoc describing an abort that no longer happens is removed.

2e96706: harden the migration

- Failure handling: a failed migration now stops startup with a localized dialog explaining why, instead of an unhandled exception with no UI. Bisq still refuses to start in that case: a fresh data dir would create a new identity and block the migration from being retried.
- Broken symlinks are skipped with a warning. They hold no data and would otherwise make every start fail.
- Crash safety: copied files and directories are synced before the rename, and the parent directory after it, since Tails users may pull the USB stick to shut down.
- Skipped files are matched by relative path (instance.lock, tor/external_tor.config), not by file name anywhere in the tree.
- Timestamps: copied files keep their last modified time.


Would be great to test on Tails:
- Whether the Dotfiles folder at /live/persistence/TailsData_unlocked/dotfiles is readable as amnesia. If it isn't, the cleanup popup just won't show.

alvasw:
> Tested [fd8a245](https://github.com/bisq-network/bisq2/commit/fd8a24525d447aa8ec1b6424328e6dc89278740f) on Tails 7.13 (Persistent Storage, default features). Is reinstalling the .deb every session the intended flow? **Bisq doesn't survive a reboot, only its data does:**
> 
>     * After reboot `/opt/bisq2` and `/etc/onion-grater.d/bisq.yml` are gone; `~/Persistent/Bisq2` is intact (same onion key, no re-onboarding).
> 
>     * Additional Software can't keep it: `sudo apt install ./bisq2_*.deb` aborts in Tails' `asp-pre-apt` hook (`KeyError: ... no package named 'bisq2'`). `dpkg -i` works but isn't remembered.
> 
> 
> Minor:
> 
>     * The no-persistence popup says "with the Dotfiles feature", but the data dir is now `~/Persistent` (Persistent Folder). Works with Dotfiles off.
> 
>     * That warning never reaches `bisq.log`: `evaluateTailsDataDirNonPersistent()` runs before `setupLogging()`.
> 
> 
> Everything else works: onion-grater profile via postinst, auto `external_tor.config` (port 951), locked-storage popup, `prepare_tails.sh` with locked storage (warns, exits 0). **Inbound connections work**: the log shows `InboundConnection`s from 10 distinct peers, and an external SOCKS connect to the node's onion succeeds, so the loopback-bind issue @alvasw raised in [his review (2026-06-13)](https://github.com/bisq-network/bisq2/pull/4643#pullrequestreview-4491499650) doesn't reproduce on 7.13. I didn't publish an offer, so the maker flow itself is untested.
> 
> Not tested: Whonix, upgrade path, the volatile-dir migration in `prepare_tails.sh`.

Thank you for reporting back

> Yes, installing .deb every session is the intended flow. I don't see the point of using Tails if anyone is going to just keep the installed programs, etc. I haven't get to trying the persistent storage myself, but if it's keeping its data across reboots where only `~/Persistent` survives, then it's the desired behavior.

We should support the default flow. Otherwise users will get confused why Bisq keeps disappearing. While testing this I was confused too and thought something is broken.

alvasw:
@wodoro Is this PR ready to review/merge?

wodoro:
Yes I think its ready for review

> We should support the default flow. Otherwise users will get confused why Bisq keeps disappearing. While testing this I was confused too and thought something is broken.

The default flow for tails users looks like this to me from what I understand from purpose of tails os: Spin up a new clean OS without anything installed, do your thing, dispose.

what do you suggest?

edit: have you tried installing using `apt install ./bisq.deb` to trigger Tails' Additional Software feature hooks ?

majesticsky:
Short answer: yes, it's readable, and the migration works on 2e96706.

Tested on Tails 7.13 with Dotfiles on, as amnesia, without sudo:

- `/live/persistence/TailsData_unlocked/dotfiles` is `amnesia:amnesia`, `drwx------`. `/live/persistence/TailsData_unlocked` and `/run/nosymfollow/live/persistence/TailsData_unlocked` are mountpoints of the same volume, so `PERSISTENCE_MOUNT` resolves fine.
- Tails creates real directories in `$HOME` and symlinks only the files, to the `/run/nosymfollow/...` path. The targets are readable as amnesia.

Migration run: data from the fd8a245 test moved into Dotfiles, `~/Persistent/Bisq2` absent, reboot, then the 2e96706 deb extracted and started as amnesia (not installed).

- The cleanup popup showed with the correct path.
- All non-backup files were copied; the 3 key/identity-related files I compared are byte-identical to the original.
- 33 old files under `backups/` were gone. They were all from the previous day, which matches the backup retention rules, so I think Bisq pruned them on first write. I didn't list the folder before starting Bisq, so this is inferred, not observed.
- No "Skipped the broken link" warnings in `bisq.log`.

PR #4659, opened 2026-04-14
chore/bring_connect_expire_revoke_sessions

rodvar:
- cherry picks https://github.com/bisq-network/bisq2/pull/4604 into main
 - implement revoke and disconnect connected device (bisq connect) both on Desktop UI and API
 - add the possibility to just expire current session (this forces a reconnect but it will be able to reconnect without repairing
 - apply CRAI suggestions
 - apply reviewer suggestions

## Summary by CodeRabbit

* **New Features**
  * Added ability to revoke connected API clients from settings with two options: revoke (removes all permissions and sessions) or invalidate session only.
  * Revoked clients are immediately disconnected from active connections.

* **Internationalization**
  * Added UI strings for client revocation actions.


HenrikJannsen (approved):
utACK

PR #4675, opened 2026-04-21
Feature/for mobile based on 2.1.10 ios hook decrypt push notification support

rodvar:
- contribs to https://github.com/bisq-network/bisq-mobile/issues/1230
- depends on https://github.com/bisq-network/bisq-relay/pull/16
- symmetric key iOS compatible encryption optional support (backward compatible) for relayed push notifications
- add mutable content support for relay notifications

## Summary by CodeRabbit

* **New Features**
  * AES‑GCM symmetric‑key encryption for mobile notifications.
  * Mutable‑content notifications allowing delivered notifications to be updated.

* **Improvements**
  * Relay payload format modernized to send Base64 payloads and platform-specific endpoints.
  * Device registration accepts an optional symmetric key and validates its Base64/length, returning clear 400 errors on invalid input.
  * Registration and device profiles now record presence of a symmetric key.

* **Tests**
  * Unit tests added to validate relay JSON payload generation.

### DIFF
--- a/bonded-roles/src/main/java/bisq/bonded_roles/mobile_notification_relay/MobileNotificationRelayClient.java
+++ b/bonded-roles/src/main/java/bisq/bonded_roles/mobile_notification_relay/MobileNotificationRelayClient.java
@@ -17,20 +17,25 @@
 
 package bisq.bonded_roles.mobile_notification_relay;
 
+import bisq.common.data.Pair;
+import bisq.common.json.JsonMapperProvider;
 import bisq.common.threading.ExecutorFactory;
 import bisq.network.NetworkService;
+import bisq.network.http.BaseHttpClient;
 import bisq.network.http.HttpRequestService;
 import bisq.network.http.HttpRequestServiceConfig;
 import bisq.network.http.HttpRequestUrlProvider;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
+import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutorService;
 
 @Slf4j
 public class MobileNotificationRelayClient extends HttpRequestService<MobileNotificationRelayClient.RequestData, String> {
-    private static final String SUCCESS = "success";
 
     private static ExecutorService getExecutorService() {
         return ExecutorFactory.newCachedThreadPool(MobileNotificationRelayClient.class.getSimpleName(),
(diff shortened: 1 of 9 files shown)


notifications/src/main/java/bisq/notifications/mobile/registration/MobileDeviceProfile.java
```diff
@@ -31,18 +31,25 @@ public final class MobileDeviceProfile implements PersistableProto {
     private final String publicKeyBase64;
     private final String deviceDescriptor;
     private final MobileDevicePlatform platform;
+    private final String symmetricKeyBase64; // nullable — when present, use AES-GCM instead of ECIES
```
HenrikJannsen, line 34: Better use Optional

HenrikJannsen (approved):
utACK

We prefer Optional over @Nullable in domain code. Matches also better with protobuf optional.

notifications/src/main/java/bisq/notifications/mobile/registration/MobileDeviceProfile.java
HenrikJannsen: @wodoro Could you review as well? I am not that familiar with the specifics here...

rodvar:
marking as draft to do some further changes (found an encoding issue preveing the circuit from working + will consider using the new bisq-relay API and stop using the deprecated one)

rodvar:
@devinbileck dev we are migrating to the newest relay API in this change 👍

rodvar:
squashed commits to align with this repo merging policy

bonded-roles/src/main/java/bisq/bonded_roles/mobile_notification_relay/MobileNotificationRelayClient.java
```diff
@@ -52,32 +57,84 @@ protected String parseResult(String json) {
...
+    /**
+     * Sends a push notification via the relay's v1 POST endpoint.
+     * Uses POST /v1/apns/device/{token} or /v1/fcm/device/{token} with a JSON body
+     * containing the Base64-encoded encrypted payload. This avoids the hex encoding
+     * round-trip of the legacy GET /relay endpoint which corrupts binary ciphertext.
+     */
     public CompletableFuture<Boolean> sendToRelayServer(boolean isAndroid,
-                                                        String deviceTokenHex,
-                                                        String encryptedMessageHex) {
-        RequestData requestData = new RequestData(isAndroid,
```
wodoro, line 64: Old code uses RequestData and request(RequestData) to do the request
is there any specific reason we are not using it anymore?

security/src/main/java/bisq/security/mobile_notifications/MobileNotificationEncryption.java
```diff
@@ -67,7 +73,52 @@ public static String encrypt(String publicKeyBase64, String message) throws Gene
...
+            byte[] keyBytes = Base64.decode(symmetricKeyBase64);
+            if (keyBytes.length != 32) {
+                throw new GeneralSecurityException("Symmetric key must be 256 bits (32 bytes), got " + keyBytes.length);
+            }
+
+            SecretKeySpec secretKey = new SecretKeySpec(keyBytes, "AES");
+
+            // Generate random nonce
+            byte[] nonce = new byte[GCM_NONCE_LENGTH];
+            SecureRandom secureRandom = new SecureRandom();
```
wodoro: better use `ByteArrayUtils#getRandomBytes(int)` (so that it doesn't create SecureRandom on every call)

security/src/main/java/bisq/security/mobile_notifications/MobileNotificationEncryption.java
```diff
@@ -67,7 +73,52 @@ public static String encrypt(String publicKeyBase64, String message) throws Gene
...
+            secureRandom.nextBytes(nonce);
+
+            Cipher cipher = Cipher.getInstance("AES/GCM/NoPadding");
```
wodoro: We also have `bisq.security.AesGcm`, please take a look to see if it can be reused here

security/src/main/java/bisq/security/mobile_notifications/MobileNotificationEncryption.java
```diff
@@ -67,7 +73,52 @@ public static String encrypt(String publicKeyBase64, String message) throws Gene
...
+            GCMParameterSpec gcmSpec = new GCMParameterSpec(GCM_TAG_LENGTH, nonce);
+            cipher.init(Cipher.ENCRYPT_MODE, secretKey, gcmSpec);
+
+            // Java's GCM mode appends the tag to the ciphertext
+            byte[] ciphertextWithTag = cipher.doFinal(message.getBytes(StandardCharsets.UTF_8));
+
+            // Output: nonce || ciphertext || tag (Java GCM already has ciphertext+tag together)
+            byte[] result = new byte[nonce.length + ciphertextWithTag.length];
```
wodoro: also can use `ByteArrayUtils#concat()` here I think for better readability

wodoro (changes requested):

wodoro:
@coderabbitai can you remember the utilities of the project like `ByteArrayUtils` ? they are hard to discover and easy to forget about
please remember to check available utilities of the project in every review so we can improve re-usability across the project

bonded-roles/src/main/java/bisq/bonded_roles/mobile_notification_relay/MobileNotificationRelayClient.java
rodvar: Yes, good question Wod.

The base class `HttpRequestService.request()` always calls client.get().
The new relay endpoint requires POST with a JSON body. The base class has no POST support in its request pipeline. We still extend HttpRequestService for its infrastructure (executorService, networkService, selectedProvider, userAgent, initialize()/shutdown() lifecycle), but bypass its request() for the actual HTTP call. The legacy getParam/parseResult/RequestData are kept to satisfy the abstract contract — removing them would require either modifying the base class (wider kaboom radius :P ) or stopping extending it (losing the infrastructure). We might want to create an issue from this to address in the near future?

security/src/main/java/bisq/security/mobile_notifications/MobileNotificationEncryption.java
rodvar: will be amend to the commit

security/src/main/java/bisq/security/mobile_notifications/MobileNotificationEncryption.java
rodvar: AesGcm.encrypt() does the cipher work but doesn't handle the nonce-prepend output format needed for iOS CryptoKit... We can use it for the cipher part and handle the output assembly ourselves with ByteArrayUtils.concat(), I'll do that in my commit amend 👍

security/src/main/java/bisq/security/mobile_notifications/MobileNotificationEncryption.java
rodvar: will be amend to the commit

bonded-roles/src/main/java/bisq/bonded_roles/mobile_notification_relay/MobileNotificationRelayClient.java
wodoro: since BaseHttpClient only supports post and get, my opinion is to modify the HttpRequestService, that way in future we still get to reuse it in case we need it again

rodvar: I agree and happy to do that, not sure if we want to do it in this PR or tackle it in a different one - @HenrikJannsen what are your thoughts?

wodoro: imo is it's better to do in this PR, so that the change is bundled with where it's used in the history

wodoro: HttpRequestService also takes care of shutting down the created httpclient cleanly

wodoro: on second thought, since it has retry logic, maybe it needs extra care for handling POST requests, if we decide to implement it

HenrikJannsen (approved):
utACK

HenrikJannsen:
I also think its better to move the encryption code to the security module, but can be done in a new PR.

bonded-roles/src/main/java/bisq/bonded_roles/mobile_notification_relay/MobileNotificationRelayClient.java
rodvar: @coderabbitai could you please create an issue with the discussed above and assign to me - will implement as a follow up shortly

PR #4697, opened 2026-04-27
feature/http_request_service_support_post_request

### LINKED ISSUE #4690: Refactor HttpRequestService to support POST requests
## Summary

Currently, `HttpRequestService.request()` only supports HTTP GET via `client.get()`. When `MobileNotificationRelayClient` was updated to use the relay's v1 POST endpoint (which requires a JSON body), it had to bypass the base class's request pipeline entirely and perform the HTTP POST manually inside `sendToRelayServer`.

The class still extends `HttpRequestService` to reuse its infrastructure (executor service, network service, provider selection, user agent, `initialize()`/`shutdown()` lifecycle), but the core `request()` method is not used for the actual HTTP call.

## Problem

- `HttpRequestService.request()` calls `client.get()` only — no POST support.
- `BaseHttpClient` already supports both `get()` and `post()`.
- `HttpRequestService` handles clean HTTP client shutdown; bypassing it risks resource leaks if not handled manually at every call site.
- Retry logic in `HttpRequestService` may need extra care when extended to POST (idempotency concerns).

## Proposed Solution

Extend `HttpRequestService` to support POST requests natively, so that callers like `MobileNotificationRelayClient` can leverage the full request pipeline (including retry logic and lifecycle management) for POST calls, not just GET.

## References

- PR: https://github.com/bisq-network/bisq2/pull/4675
- Discussion: https://github.com/bisq-network/bisq2/pull/4675#discussion_r3134268653
- Requested by: @rodvar

rodvar:
- resolves #4690 
 - Introduce HttpRequest descriptor (method, path, body, header, retr…yOnServerError, logPath) and replace the GET-only getParam() hook with buildRequest().
 - MobileNotificationRelayClient now uses the full pipeline (failover, retry, timeout, lifecycle) instead of bypassing it.       - POST 5xx retry is opt-in per descriptor; transport-level failures (connect/DNS/TLS) always retry regardless of method, restoring failover for unreachable providers. URL joining is normalized and device tokens are redacted from INFO logs via descriptor logPath.
 - added unit tests
 - tested with mobile relay circuit

## Summary by CodeRabbit

* **New Features**
  * Centralized HTTP request descriptor for standardized request handling and redacted logging.

* **Refactor**
  * Unified request building and execution across services with clearer retry behavior for safe vs. non-safe operations.
  * Mobile notification relay rewritten to use the unified HTTP pipeline and simplified public API, improving reliability.

* **Tests**
  * Added tests for request construction, redaction, retry policies, and notification relay behavior.

### DIFF
--- a/bonded-roles/src/main/java/bisq/bonded_roles/explorer/ExplorerService.java
+++ b/bonded-roles/src/main/java/bisq/bonded_roles/explorer/ExplorerService.java
@@ -22,6 +22,7 @@ import bisq.common.json.JsonMapperProvider;
 import bisq.common.network.TransportType;
 import bisq.common.threading.ExecutorFactory;
 import bisq.network.NetworkService;
+import bisq.network.http.HttpRequest;
 import bisq.network.http.HttpRequestService;
 import bisq.network.http.HttpRequestServiceConfig;
 import bisq.network.http.HttpRequestUrlProvider;
@@ -59,17 +60,17 @@ public class ExplorerService extends HttpRequestService<ExplorerService.RequestD
     }
 
     @Override
-    protected String getParam(HttpRequestUrlProvider provider, RequestData requestData) {
+    protected HttpRequest buildRequest(HttpRequestUrlProvider provider, RequestData requestData) {
         if (provider instanceof Provider explorerServiceProvider) {
             String apiPath = explorerServiceProvider.getApiPath();
             if (requestData instanceof TxRequestData txRequestData) {
                 String txPath = explorerServiceProvider.getTxPath();
                 String txId = txRequestData.getTxId();
-                return apiPath + "/" + txPath + "/" + txId;
+                return HttpRequest.get(apiPath + "/" + txPath + "/" + txId);
             } else if (requestData instanceof AddressRequestData addressRequestData) {
                 String addressPath = explorerServiceProvider.getAddressPath();
                 String address = addressRequestData.getAddress();
-                return apiPath + "/" + addressPath + "/" + address;
+                return HttpRequest.get(apiPath + "/" + addressPath + "/" + address);
             } else {
                 throw new IllegalArgumentException("requestData of unsupported type. requestData = " + requestData);
             }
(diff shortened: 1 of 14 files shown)


HenrikJannsen (approved):
utACK

@wodoro Could you review as well?

wodoro:
sure

wodoro (changes requested):
Discussed retry edge cases and helping APNs to dedupe over matrix

`ClearNetHttpClient` and `TorHttpClient` still may log the token
baseUrl contains the token right now and any 4xx/5xx error or debug log may dump the token (see line 100 in TorHttpClient for an example)

the rest looks good to me and the PR acks some concerns by adding TODOs for the future

rodvar:
> Discussed retry edge cases and helping APNs to dedupe over matrix
> 
> `ClearNetHttpClient` and `TorHttpClient` still may log the token baseUrl contains the token right now and any 4xx/5xx error or debug log may dump the token (see line 100 in TorHttpClient for an example)
> 
> the rest looks good to me and the PR acks some concerns by adding TODOs for the future

I think the test comment that lead to this discussion is misleading. We should still do at least one retry for the HTTP 5xx family errors, APNs and FCMs are known to be reliable in that sense - its better to retry that to have a skipped notification for this edge cases. I'll update the comment now.

during convs we explored possibilities using `collapse_key / apns-collapse-id` and that is definitely worth adding but does not concern this refactor, I'll create a bisq-mobile issue for this. (well done wod 👏 )

Finally, I saw the privacy token leak concern in base classes as well I was thinking of leaving it for another issue but I'll do it now with my review commit 💪

wodoro (approved):
Thanks for providing a more accurate summary of our conversations

thanks for going the extra mile to redact things in other classes as well
but, on second thought, It seems weird to me that we have to redact a baseUrl at all!
anything sensitive should not be touching the url (a best practice in rest api that I had forgotten)

let me know what you think

rodvar:
> Thanks for providing a more accurate summary of our conversations
> 
> thanks for going the extra mile to redact things in other classes as well but, on second thought, It seems weird to me that we have to redact a baseUrl at all! anything sensitive should not be touching the url (a best practice in rest api that I had forgotten)
> 
> let me know what you think

I agree, I've created this [issue](https://github.com/bisq-network/bisq-relay/issues/17) to solve this in the relay and then we can update the bisq2 codebase when that is done 👍 

For now I think the final result of this PR is good after our iterations - thank you

wodoro:
utACK

PR #4700, opened 2026-04-28
feat: implement closed trades api

wodoro:
## Summary by CodeRabbit

* **New Features**
  * Added a paginated closed-trades endpoint with sorting, ASC/DESC direction, role and state filtering, and free-text search.
  * Introduced compact paginated response format and pagination parameters (page/pageSize) plus richer closed-trade list payloads with pre-formatted display fields for efficient client use.
* **Tests**
  * Added extensive tests covering closed-trades querying, sorting, filtering, search, and pagination behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq2/pull/4700)


api/src/main/java/bisq/api/dto/presentation/open_trades/TradeItemPresentationDtoFactory.java
```diff
@@ -42,16 +46,29 @@ public static TradeItemPresentationDto create(BisqEasyTrade trade,
...
+                                                   UserProfile myUserProfile,
+                                                   UserProfile peersUserProfile,
+                                                   Optional<BisqEasyOpenTradeChannelDto> channelDto,
+                                                   ReputationService reputationService) {
         BisqEasyContract contract = trade.getContract();
         long date = contract.getTakeOfferDate();
 
         String directionalTitle = BisqEasyTradeFormatter.getDirectionalTitle(trade);
-        String offerId = channel.getBisqEasyOffer().getId();
-        String tradeId = trade.getId();
```
rodvar, line 50: Good on you for bringing this to the table wod once again. I checked the proposal on ClosedTradeListItemDto (very well documented btw 🙏 and these changes.

I believe removing the heavy fields would force a follow-up request for every open trade, regressing UX. Plus it's a breaking API change for existing clients (mobile + desktop's API consumer)..?

how about we take the opportunity to do the right long-term architectural fix using java record composition? something like:

```
public record TradePresentationCommon(
      UserProfileDto makerUserProfile,
      UserProfileDto takerUserProfile,
      String directionalTitle,
      String formattedDate,
      // ... the 17 overlapping fields
  ) {}

  public record TradeItemPresentationDto(
      TradePresentationCommon common,
      Optional<BisqEasyOpenTradeChannelDto> channel,
      BisqEasyTradeDto trade,
      // open-trade-specific extras
  ) {}

  public record ClosedTradeListItemDto(
      TradePresentationCommon common,
      TradeSlimDto trade,
      Optional<UserProfileDto> mediatorUserProfile,
      // closed-trade-specific extras
  ) {}
```

Sorry for the quick code might not be perfect, is just to communicate the idea better? Then we'll have a single source of truth for the overlapping 15+ fields. only cons of this approach is clients see a `common.directionalTitle` nesting instead of `directionalTitle` but I think that would be mild.

Hope this helps lmk!

wodoro: it depends, really. I did not find anywhere where the identity keys are used on client code, which means they are wasted on the wire (also probably a security issue to hand over the private keys like that)
there may be more like this, where it does not affect any functionality, but removing them saves memory and bandwidth

what I was thinking was maybe we can strike a balance by making the main presentation lighter, but still complete enough to not cause more requests

wodoro: and this would be a breaking change (older clients still expecting full Dto), so ... I'm not really sure

rodvar: I think the maintainability argument still holds..?

I would keep this slim `ClosedTradeListItemDto` as-is (it's already a partial fix for the security issue on the closed-trades path), and open a separate issue to address later on titled something like "Audit and remove sensitive identity/key material from BisqEasyTradeDto — security fix". That separates the security cleanup from the closed-trades feature so reviewers can prioritise it independently.   

What do you think?

wodoro: if we want to address security issue, and if it is considered urgent, then we can simply go for that and drop the heavy fields along with the private keys, as it is a breaking change already. then this slim class makes less sense, and the maintainability improves.

wodoro: the maintainability argument is simply (to me) that this extra Dto adds burden

rodvar: I'm sorry I think I missed the forest here:

TradeItemPresentationDto is being used as a list-item DTO but it carries detail-level data (full BisqEasyTradeDto with contract/identity/keys, full channel). The right split is:
                                                                                                                                
  - List endpoint → list-shaped DTO (small, paginated, no sensitive fields)                                                     
  - Detail endpoint → detail-shaped DTO (full trade, channel, etc.)
                                                                                                                                
Open-trades clients that need full data for the chat tab / detail view should fetch it via a GET /trades/{id} detail endpoint, not via the list endpoint. That's a normal API pattern; bisq2's just been mixing them.

So the question would be if we want to tackle the security issue now with a breaking change, or modify this PR a bit and report the issue to tackle later. 

From my convs with @HenrikJannsen  - there is no date for the next bisq2 desktop release (intentionally not release any time soon) so doing the breaking changes now will bring us to a dependency on a new desktop version for connect apps (can't really release)

For this reason I would suggest 2 options

 1) do my original suggestion of removing breaking changes in this PR and create an issue to address later on (when we know we are close to bisq2 release) OR
 2) A third path: make the heavy/sensitive fields nullable in the wire shape (additive-compatible), and stop populating them on the bisq2 side starting 2.1.11+. Mobile reads them as Optional/null. No version bump, no coordination — but the security fix is partial for Desktop as trusted node users(only kicks in once users upgrade their desktop which is acceptable, immediate for power users). 

Thoughts? :)

wodoro: Good points

so to summarize:
- remove breaking changes in this PR (if any. I need to double check but I don't think I broke anything)
- create issue to address the security issue (to fix later)
- create issue to address list vs detail (?)
- make heavy/sensitive fields nullable in dto for a future update
- bonus: mark heavy/sensitive fields `@transient` so we simply drop the memory overhead on mobile side immediately
- keep this slim Dto for closed trades till future updates land and make this obsolete

wodoro: @coderabbitai can you make both issues (security + list vs detail) on bisq-network/bisq-mobile ? (to not clutter here?)

api/src/main/java/bisq/api/rest_api/endpoints/trades/ClosedTradesQuery.java
```diff
@@ -0,0 +1,171 @@
...
+                || contains(item.takerUserProfile().nym(), needleLower)
+                || contains(item.bitcoinSettlementMethodDisplayString(), needleLower)
+                || contains(item.fiatPaymentMethodDisplayString(), needleLower)
+                || contains(item.formattedPrice(), needleLower)
+                || contains(String.valueOf(item.baseAmount()), needleLower)
+                || contains(String.valueOf(item.quoteAmount()), needleLower);
+    }
+
+    private static boolean contains(String value, String needleLower) {
+        return value.toLowerCase(Locale.ROOT).contains(needleLower);
```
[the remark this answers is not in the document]
wodoro: this goes against principles of bisq2: we only protect against null where it's mentioned it can be null, and we use Optional when it's possible to be null.

api/src/main/java/bisq/api/rest_api/pagination/PaginationParams.java
```diff
@@ -0,0 +1,49 @@
...
+                .map(s -> Math.min(s, MAX_PAGE_SIZE))
+                .orElse(DEFAULT_PAGE_SIZE);
+        return new PaginationParams(resolvedPage, resolvedSize);
+    }
+
+    public <T> PaginatedResponse<T> paginate(List<T> items) {
+        int total = items.size();
+        int totalPages = (int) Math.ceil((double) total / pageSize);
+        int fromIndex = Math.min((page - 1) * pageSize, total);
+        int toIndex = Math.min(fromIndex + pageSize, total);
```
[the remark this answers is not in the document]
wodoro: it will simply collapse to total, wont it ? why bring cast into this?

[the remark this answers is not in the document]
wodoro: why not just check it for < 0 and correct it? or even throw with better error ?

api/src/main/java/bisq/api/web_socket/domain/trades/ClosedTradesWebSocketService.java
```diff
@@ -0,0 +1,107 @@
...
+                send(item.dto(), ModificationType.ADDED);
+            }
+
+            @Override
+            public void onRemoved(Object element) {
+                // Closed trades are append-only in the domain. If this fires, the domain
+                // contract changed: implement REMOVED broadcast and update the client schema.
+                log.warn("Unexpected removal from closed trades collection: {}. " +
+                        "REMOVED broadcast not implemented.", element);
+            }
```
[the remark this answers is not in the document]
wodoro: the comment clearly says why. why suggest this?

api/src/main/java/bisq/api/rest_api/endpoints/trades/ClosedTradesQuery.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    public enum SortField {
+        // DATE sorts by the trade-completion timestamp when present (more meaningful for closed
+        // trades since users care about when the trade finished), falling back to the take-offer
+        // timestamp for trades that never reached a completion state (e.g. cancelled before any
+        // completion event was recorded). takeOfferDate is monotonic and always populated, so
+        // the fallback guarantees a stable ordering even for never-completed trades.
+        DATE(Comparator.comparingLong((ClosedTradeIndexedItem item) ->
+                item.dto().tradeCompletedDate().orElse(item.dto().trade().contract().takeOfferDate()))),
```
wodoro: Needs to be discussed. do we want to use tradeCompletedDate or not ?

api/src/main/java/bisq/api/dto/presentation/closed_trades/ClosedTradeIndexedItem.java
rodvar, line 1: nit - do we want package name to be open_trades for all the closetrade related stuff? otherwise please change package 👍

api/src/main/java/bisq/api/rest_api/endpoints/trades/ClosedTradesQuery.java
```diff
@@ -0,0 +1,179 @@
...
+        MARKET(Comparator.comparing(ClosedTradeIndexedItem::market, String.CASE_INSENSITIVE_ORDER)),
+        QUOTE_AMOUNT(Comparator.comparingLong(item -> item.dto().quoteAmount())),
+        BASE_AMOUNT(Comparator.comparingLong(item -> item.dto().baseAmount())),
+        ROLE(Comparator.comparing(ClosedTradeIndexedItem::formattedMyRole, String.CASE_INSENSITIVE_ORDER));
```
rodvar: would this cause problems with i18n? if so please use `tradeRole` instead?

api/src/main/java/bisq/api/dto/presentation/closed_trades/ClosedTradeListItemDtoFactory.java
```diff
@@ -0,0 +1,107 @@
...
+                mediatorUserProfile,
+                priceQuoteDto,
+                contract.getBaseSideAmount(),
+                contract.getQuoteSideAmount(),
+                bitcoinSettlementMethod,
+                bitcoinSettlementMethodDisplayString,
+                fiatPaymentMethod,
+                fiatPaymentMethodDisplayString,
+                peersReputationScoreDto,
+                Optional.ofNullable(trade.getPaymentAccountData().get()),
```
rodvar, line 81: trade.getPaymentAccountData() non nullable?

api/src/main/java/bisq/api/web_socket/domain/ClosedTradeItemsService.java
```diff
@@ -0,0 +1,94 @@
...
+        bisqEasyTradeService = tradeService.getBisqEasyTradeService();
+        reputationService = userService.getReputationService();
+    }
+
+    @Override
+    public CompletableFuture<Boolean> initialize() {
+        closedTradesPin = bisqEasyTradeService.getClosedTrades().addObserver(new CollectionObserver<>() {
+            @Override
+            public void onAdded(BisqEasyClosedTrade closedTrade) {
+                if (findListItem(closedTrade.trade().getId()).isEmpty()) {
```
rodvar, line 58: O(n) scan. Close trade adds are not that freq I would say but just noting it to consider a parallel Set<String> knownIds is O(1) and trivial maybe? Also affects onRemoved

api/src/main/java/bisq/api/web_socket/domain/trades/ClosedTradesWebSocketService.java
```diff
@@ -0,0 +1,106 @@
...
+    public CompletableFuture<Boolean> shutdown() {
+        if (closedTradesPin != null) {
+            closedTradesPin.unbind();
+            closedTradesPin = null;
+        }
+        return CompletableFuture.completedFuture(true);
+    }
+
+    @Override
+    public Optional<String> getJsonPayload() {
```
rodvar: clients are expected to paginate via REST and use the subscription only for ADDED/REMOVED deltas correct? maybe worth documenting this initial empty list return?

api/src/main/java/bisq/api/dto/presentation/closed_trades/ClosedTradeIndexedItem.java
rodvar: or maybe we just conslidate all in one single `trades` subpackage?

api/src/main/java/bisq/api/dto/presentation/closed_trades/ClosedTradeListItemDtoFactory.java
```diff
@@ -0,0 +1,107 @@
...
+        UserProfileDto peersUserProfileDto = DtoMappings.UserProfileMapping.fromBisq2Model(peersUserProfile);
+        UserProfileDto makerUserProfile = trade.isMaker() ? myUserProfileDto : peersUserProfileDto;
+        UserProfileDto takerUserProfile = trade.isMaker() ? peersUserProfileDto : myUserProfileDto;
+        Optional<UserProfileDto> mediatorUserProfile = contract.getMediator()
+                .map(DtoMappings.UserProfileMapping::fromBisq2Model);
+
+        PriceQuoteDto priceQuoteDto = DtoMappings.PriceQuoteMapping.fromBisq2Model(trade.getPriceQuote());
+
+        String bitcoinSettlementMethod = contract.getBaseSidePaymentMethodSpec().getPaymentMethodName();
+        String bitcoinSettlementMethodDisplayString = contract.getBaseSidePaymentMethodSpec().getShortDisplayString();
```
rodvar: do we need the *DisplayString fields? these travels in the json

api/src/main/java/bisq/api/web_socket/domain/trades/ClosedTradesWebSocketService.java
rodvar: looking at the mobile side impl I think that's the case right?

just wondering if we can avoid/is-it-justified the ws subscription vs for example a REST fetch + pull-to-refresh mobile pattern?

rodvar (changes requested):
overall a very solid PR and impl wod.

Just flagged some details to discuss/improve, the most important is the one related to the WS subs

api/src/main/java/bisq/api/rest_api/endpoints/trades/ClosedTradesQuery.java
rodvar: > Needs to be discussed. do we want to use tradeCompletedDate or not ?

thanks for bringing this up, its not just a code nit but a UX decision

tradeCompletedDate.orElse(takeOfferDate) matches the mental model of "what finished recently" for completed trades history. However we do need a tie breaker for trades that never completed successfully, so I would add a fallkback with `takeOfferDate` 👍 

Please do the necessary adaptations in mobile side

There is one more thing with this code, we need a tie breaker in sorting for 2 trades with the same single tradeCompletedDate otherwise would cause inconsistencies in pagination.

Suggestion sth like..

```
 DATE(Comparator
    .comparingLong((ClosedTradeIndexedItem item) ->                                                                           
        item.dto().tradeCompletedDate().orElse(item.dto().trade().contract().takeOfferDate()))
            .thenComparing(item -> item.dto().trade().id())),   
```

?

wodoro: I like the suggestion, will apply, but overall the chances of two trade having same millisecond timestamp is very low

rodvar: 💯 only mentioned because is a straighforward fix 👍

api/src/main/java/bisq/api/dto/presentation/closed_trades/ClosedTradeListItemDtoFactory.java
wodoro: sounds like there's an undocumented unspecified null case here. maybe need to be fixed in observable api later

api/src/main/java/bisq/api/web_socket/domain/ClosedTradeItemsService.java
wodoro: this is same as what BisqEasyHistoryController already does? maybe accept here and create another issue to address both later together ?

api/src/main/java/bisq/api/web_socket/domain/trades/ClosedTradesWebSocketService.java
wodoro: It is documented above the class definition
yes clients are expected paginate via rest, as it does not make sense to send everything over websocket subscribe, as the list may get huge over time.

the subscription works as a notifier and is cheap, of course we can avoid it. practically thinking about it, the client never has the both open trades and closed trades open at the same time, so most likely they will have fresh data every time they switch to the tab.

this may be useful for power users who use the api, but may be its even useless for them. I'm not really sure.

wodoro: if agreed we can drop websocket implementation completely and remove it client side

api/src/main/java/bisq/api/dto/presentation/closed_trades/ClosedTradeListItemDtoFactory.java
wodoro: no, thanks for spotting this, its not used on the client

api/src/main/java/bisq/api/dto/presentation/closed_trades/ClosedTradeListItemDtoFactory.java
rodvar: @coderabbitai please create an issue for this to address later on

api/src/main/java/bisq/api/web_socket/domain/ClosedTradeItemsService.java
rodvar: @coderabbitai please create an issue for this to address later on

api/src/main/java/bisq/api/web_socket/domain/trades/ClosedTradesWebSocketService.java
rodvar: I'm gonna do a full review on the trade history feature for mobile later today and update on this

rodvar: yes its agreed let's drop that - I'll do the change to finish this off as discussed whilst you are helping with Bisq1 🙏

rodvar:
gentle reminder this one is blocking [trade history feature](https://github.com/bisq-network/bisq-mobile/pull/1389) merge 🙏

HenrikJannsen (approved):
utACK

HenrikJannsen:
Please make PR on https://github.com/bisq-network/bisq2/tree/for-mobile-based-on-2.1.11 as well

PR #4701, opened 2026-04-28
chore/bring_http_request_service_support_post_request_to_main

rodvar:
- brings https://github.com/bisq-network/bisq2/pull/4697 to main
 - Introduce HttpRequest descriptor (method, path, body, header, retr…yOnServerError, logPath) and replace the GET-only getParam() hook with buildRequest().
 - MobileNotificationRelayClient now uses the full pipeline (failover, retry, timeout, lifecycle) instead of bypassing it.       - POST 5xx retry is opt-in per descriptor; transport-level failures (connect/DNS/TLS) always retry regardless of method, restoring failover for unreachable providers. URL joining is normalized and device tokens are redacted from INFO logs via descriptor logPath.
 - added unit tests
 - adressed reviewer suggestions: improved POST retry logic comments, fix token leaks in base classes as well
 - apply some CRAI suggestions


HenrikJannsen (approved):
utACK

PR #4702, opened 2026-04-28
chore/auto_generate_sync_to_main_pr_when_needed_on_active_mobile_branch

rodvar:
## Motivation

Sync to main policy switch based on dev remembering to do so its not working great, lots of commits are forgotten merged in the for mobile branch and not synced.
Using previous policy where bisq2 maintener needs to merge the for-mobile active branch back in before release also has its complications so came up with this idea to auto generate a PR on every commit merged to active for-mobile branch (active == latest)

 - Small tweaks might be needed after merge 🙏 
 - I will go through all the current `for-mobile-based-on-2.1.10` commits to see what hasn't been synced and send a PR (this change is for future merges won't help with already merged stuff)

## dev logs

 - add new GH workflow to do syncing of merged commits intoactive (last) for-mobile-based-on-* branch to the main branch
 - support for skipping based on adding a tag on the to-be-merged PR (`skip-sync-to-main` tag needs to be created by bisq2 maintainer)
 - merging this PR would be the actual real test (as we want this in main also for future new "for-mobile-based" branches to grab it

## Summary by CodeRabbit

* **Chores**
  * Added an automated sync that forwards merged PRs from the active release branch to main.
  * Honors an opt-out label only when applied by authorized maintainers.
  * Preserves assignees and requested reviewers, excludes specific opt-out labels, and skips merge commits.
  * Creates draft PRs with conflict markers on cherry-pick conflicts and posts guidance for resolution.
  * Serializes runs per source PR to avoid race conditions.

PR #4705, opened 2026-04-29
chore/bring_for_mobile_auto_sync_to_main_in

rodvar:
- Bring #4702 into main branch
 - add new GH workflow to do syncing of merged commits intoactive (la…st) for-mobile-based-on-* branch to the main branch
 - improve docs and robustness setup to avoid causing undesired effects
 - apply CRAI suggestions
 - Split the network call out, retry transient failures up to 3 times to avoid silent failure
 - add explicit skip when PR is closed with no merge
 - apply latest CRAI suggestion
 - pin dependencies to a exact commit instead of relying on the tag + updated docs


HenrikJannsen (approved):
utACK

PR #4746, opened 2026-05-19
Webcam resource cleanup

KimStrand:


HenrikJannsen (approved):
utACK

Maybe you can drop the other 2 commits as they are not related and then I can merge without dependency on those.

HenrikJannsen (approved):
utACK

PR #4747, opened 2026-05-20
Better storage persistence 

wodoro:
- Enforce a strict identifier format for MetaData.className.
- Restrict StorageService.getOrCreate* to store keys present in NetworkStorageWhiteList, with a permissive fallback when the list is empty so unit-test bootstrapping is not affected.


HenrikJannsen (approved):
utACK

PR #4800, opened 2026-06-03
Treat range amount with equal min/max as fixed

### LINKED ISSUE #4531: Range amount with same value should be treated like Fixed amount
If the min and max value of the range amount is the same the taker will get the amount selection but cannot change as the min/max is the same. We should convert in the offer creation such cases to a fixed amount.

balag3:
Fixes #4531

A range offer whose min equals max showed the taker an amount selector that couldn't be moved (min and max identical). Such offers should be fixed-amount offers.

## Change
- `AmountSpecFactory`: new `createBaseSideAmountSpec(long, long)` / `createQuoteSideAmountSpec(long, long)` overloads return a **fixed** spec when `min == max`, else a range spec. The existing `create*AmountSpec` range branches delegate to them, so the MuSig draft-workflow path is covered.
- `OfferAmountUtil`: the two price-recompute paths (`updateQuoteSideAmountSpecWithPriceSpec` / `updateBaseSideAmountSpecWithPriceSpec`) now route through the same overloads, so a price change can't reintroduce a single-value range.

Per @HenrikJannsen's confirmation on the issue, the normalization is centralized in `AmountSpecFactory`.

## Out of scope (by design)
- **Create-offer UI controllers** (`TradeWizardAmountController`, `MuSigCreateOfferAmountController`) already collapse `min == max` to fixed — left unchanged.
- **`fromProto()`** and **API `DtoMappings`** — deserialization / external input, not offer *creation*; not normalized here (avoids touching persisted/network data).

## Tests
`AmountSpecFactoryTest`: equal min/max → fixed, different → range, for both base and quote side. `./gradlew :offer:build` passes.

## Aside (not addressed here)
`OfferAmountUtil.updateBaseSideAmountSpecWithPriceSpec` appears to compute both min and max from `findQuoteSideFixedAmount` (looks like a copy-paste; max may be intended to use a max variant). Left untouched as unrelated to this issue — flagging in case it's a real bug worth a separate ticket.

## Summary by CodeRabbit

* **Refactor**
  * Refactored offer amount specification creation to use factory methods instead of direct instantiation, improving code consistency and maintainability across base-side and quote-side amount handling.

* **Tests**
  * Added comprehensive test coverage for the amount specification factory, verifying correct behavior for both fixed and range amount specifications with proper boundary testing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq2/pull/4800?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/offer/src/main/java/bisq/offer/amount/OfferAmountUtil.java
+++ b/offer/src/main/java/bisq/offer/amount/OfferAmountUtil.java
@@ -23,14 +23,13 @@ import bisq.common.monetary.Monetary;
 import bisq.common.util.MathUtils;
 import bisq.offer.Offer;
 import bisq.offer.amount.spec.AmountSpec;
+import bisq.offer.amount.spec.AmountSpecFactory;
 import bisq.offer.amount.spec.AmountSpecUtil;
 import bisq.offer.amount.spec.BaseSideAmountSpec;
 import bisq.offer.amount.spec.BaseSideFixedAmountSpec;
-import bisq.offer.amount.spec.BaseSideRangeAmountSpec;
 import bisq.offer.amount.spec.FixedAmountSpec;
 import bisq.offer.amount.spec.QuoteSideAmountSpec;
 import bisq.offer.amount.spec.QuoteSideFixedAmountSpec;
-import bisq.offer.amount.spec.QuoteSideRangeAmountSpec;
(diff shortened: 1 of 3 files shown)


HenrikJannsen (approved):
utACK

PR #4801, opened 2026-06-04
Sort mediation cases by last message

### LINKED ISSUE #3909: Order mediation cases so that ones with new messages are on top
The vertical ratio of the mediation view is so that most of the space is reserved for the chat, which makes sense, but, if there are several cases listed, and this is especially true for the closed cases, and you see the "new messages" numbered badge, you have to do a tedious exercise of scrolling up and down in a narrow panel until you catch that one case where the new messages come from.
Since, differently from Bisq1, cases in Bisq Easy are listed in a single row for both buyer and seller, it wouldn't create confusion if those ones where the new messages happened, were floated to the top so they are processed more easily.
Or at least you could add a sorting column for "number of new messages", whichever is easier on the coding side.

balag3:
Fixes #3909

Mediators with many cases had no quick way to see which ones were recently
active — they had to scroll a narrow panel hunting for them. This adds a
sortable **"Last message"** column to both mediator case tables, ordered by
the timestamp of the most recent chat message so the most recently active
cases surface on top.

## Change
- New **"Last message"** column (date + time) in both the Bisq Easy and
  MuSig mediator case tables, in `BisqEasyMediationTableView` /
  `MuSigMediationTableView`.
- **Default-sorted descending** so the most recently active cases appear on
  top; clicking any other column header takes over the sort (standard
  behaviour).
- Backed by observable `lastMessageDate` string/time properties on the list
  items; the value updates live via an observer on the channel's chat
  messages, and the cell is bound to it (same pattern as the existing
  per-trader badges).
- The per-trader maker/taker unread badges are unchanged and still indicate
  *who* has unread messages.

## Scope
- Applies to **both Bisq Easy and MuSig** mediator tables, open and closed
  case lists.
- Unit tests for the ordering comparator (`BY_LAST_MESSAGE_DATE`) on both
  `BisqEasyMediationCaseListItem` and `MuSigMediationCaseListItem`.

## Note on approach
The issue and earlier discussion leaned toward an unread-message **count**
column. I went with **last-message timestamp** instead because:
- It surfaces active cases regardless of read state, so the mediator tab
  auto-opening (and marking-read) the top case doesn't reshuffle the order
  or hide anything — the original count-based approach fought that
  auto-consume behaviour.
- It extends cleanly to MuSig, which has separate buyer/seller chats; "last
  message across the channel" needs no special chat-pair handling, so MuSig
  could be included rather than deferred.

The per-trader unread badges still cover the "who needs a reply" signal.
@suddenwhipvapor — happy to switch back to a count-based sort if you'd
prefer the literal unread-count column over recency.

## Release Notes

* **New Features**
  * Added "Last message" column to mediation case tables, displaying the timestamp of the most recent chat message
  * Mediation cases can now be sorted by last message date to prioritize recent activity

* **Tests**
  * Added verification tests for last message sorting functionality

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/authorized_role/mediator/bisq_easy/BisqEasyMediationCaseListItem.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/authorized_role/mediator/bisq_easy/BisqEasyMediationCaseListItem.java
@@ -44,6 +44,7 @@ import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
 
@@ -52,6 +53,9 @@ import java.util.Optional;
 @ToString
 @EqualsAndHashCode(onlyExplicitlyIncluded = true)
 public class BisqEasyMediationCaseListItem implements ActivatableTableItem, DateTableItem {
+    public static final Comparator<BisqEasyMediationCaseListItem> BY_LAST_MESSAGE_DATE =
+            Comparator.comparingLong(BisqEasyMediationCaseListItem::getLastMessageDate);
+
     @EqualsAndHashCode.Include
     private final BisqEasyMediationCase bisqEasyMediationCase;
     @EqualsAndHashCode.Include
@@ -66,7 +70,11 @@ public class BisqEasyMediationCaseListItem implements ActivatableTableItem, Date
     private final boolean isMakerRequester;
     private final Badge makersBadge = new Badge();
     private final Badge takersBadge = new Badge();
+    private long lastMessageDate;
+    private final StringProperty lastMessageDateString = new SimpleStringProperty("");
+    private final StringProperty lastMessageTimeString = new SimpleStringProperty("");
     private Pin changedChatNotificationPin;
+    private Pin chatMessagesPin;
     private Pin isClosedPin;
     private Long closeCaseDate = 0L;
     private final StringProperty closeCaseDateString = new SimpleStringProperty("");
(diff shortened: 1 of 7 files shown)


suddenwhipvapor:
> I raised this on the issue — options: keep as-is, or don't auto-open a case on load so unread stays visible until clicked. Marking ready once @suddenwhipvapor weighs in (happy to fold the chosen behaviour into this PR or do a follow-up).

I would personally prefer not having the top chat open automatically on tab open, @solomon1923 please also give your opinion

In general, I would personally prefer that also when a case is closed in the open cases tab, the view goes back to neutral without auto-opening the first case on top

balag3:
> > I raised this on the issue — options: keep as-is, or don't auto-open a case on load so unread stays visible until clicked. Marking ready once @suddenwhipvapor weighs in (happy to fold the chosen behaviour into this PR or do a follow-up).
> 
> I would personally prefer not having the top chat open automatically on tab open, @solomon1923 please also give your opinion
> 
> In general, I would personally prefer that also when a case is closed in the open cases tab, the view goes back to neutral without auto-opening the first case on top


  @suddenwhipvapor @solomon1923 ? I reworked this since your last comment, in two ways:

  **1. Sort by last-message *recency*, not unread *count*.**

  Instead of a "New messages" count column I went with a sortable **"Last message"** column ordered by the timestamp of the most recent chat message. Two
  reasons:

  - It surfaces active cases **regardless of read state**, so opening (and thereby marking-read) a case doesn't change its position or make it drop away ? the
  count-based version fought the auto-open behaviour you flagged below.
  - It extends cleanly to **MuSig**, whose separate buyer/seller chats made a single unread-count awkward ? so this PR now covers **both** Bisq Easy and MuSig
  tables (open + closed lists), rather than deferring MuSig.

  The per-trader maker/taker unread **badges** are unchanged, so the "who still needs a reply" signal is still there.

  **2. The auto-open concern is addressed in a separate PR.**
  
  You wrote:

  > I would personally prefer not having the top chat open automatically on tab open ? also when a case is closed in the open cases tab, the view goes back to
  neutral

  That's exactly what #4803 does (neutral on tab entry, de-focus on leave/close, no auto-select) ? which you tACK'd. With #4803 in, this PR's recency sort and
  the neutral view compose cleanly: cases are ordered by activity, and nothing is auto-opened/consumed.

suddenwhipvapor:
tACK

tested the ordering and now I can say this is how I actually had envisioned it, since even for MuSig the mediations chats will be in common between the two traders it will be good to go for that as well!

PR #4805, opened 2026-06-08
bugfix/any_network_active_chat_message_causing_relayed_notification_android_opted_in

rodvar:
- fixes https://github.com/bisq-network/bisq-mobile/issues/1464
 - implemented fix by improving notification filtering + docs + tests

## Summary by CodeRabbit

* **Improvements**
  * Mobile push eligibility now requires both message type and channel domain; unknown/null values are blocked and only specific trade-related channels are allowed while public/deprecated discussion domains are suppressed.
  * Trade push behavior: suppresses local/self-action notifications, preserves peer/received notifications, and refines payment-info exchange handling to avoid duplicate or misleading pushes.

* **Tests**
  * Added exhaustive tests asserting exact allowlists for message types and channel domains and pinned trade-state whitelist behavior to prevent regressions.


HenrikJannsen (approved):
utACK

PR #4810, opened 2026-06-08
Clear chat input when switching channels

### LINKED ISSUE #4808: Unsent chat text can carry over when switching chats and be sent to the wrong channel
## Problem

When a user types text into a chat input but does not send it, then switches to another chat/channel in the same chat area, the unsent text remains in the input field.

If the user then presses Enter or clicks send, the text is sent to the currently selected chat/channel, not necessarily the chat/channel where the text was originally typed.

## Why this matters

This can lead to accidental disclosure of private or trade-specific information to the wrong peer or channel. The risk is higher in areas where users switch between multiple chats, trades, mediation cases,  arbitration cases, public channels, or private chats.

## Steps to reproduce

1. Open a chat area with more than one selectable chat/channel.
2. Select chat A.
3. Type a message into the input field, but do not send it.
4. Switch to chat B.
5. Observe that the typed text is still present in the input field.
6. Press Enter or click the send button.

## Actual behavior

The draft text from chat A remains visible after switching to chat B and can be sent to chat B.

## Expected behavior

The application should prevent text typed for one chat/channel from being accidentally sent to another chat/channel.

## Notes

A related older issue exists: #2071, "Clear input fields in trade chat when switching trades". It is closed, but the current behavior appears broader than trade chat switching and still seems possible in the shared chat input flow.

balag3:
Fixes #4808

When text was typed into the chat input but not sent, then the user switched to another channel, the draft stayed in the shared input. Because the send path resolves the target channel at send time, that draft — and any pending reply citation — could then be sent to the **wrong** channel.

## Change
`ChatMessageContainerController.selectedChannelChanged()` now clears the unsent draft (input text **and** any pending citation) when the selected channel actually changes (`!previousChannel.equals(chatChannel)`). The guard means re-selecting the same channel (e.g. on tab re-activation) does **not** wipe an in-progress draft.

The input is a single shared component (`ChatMessageContainerController`, no subclasses) used by every chat area — public, private, trade, mediation, arbitration — so this one change covers all of them. The older #2071 only cleared the input after a successful send, never on a channel switch.

## Testing
Verified manually with two local clients (one on `main`, one with the fix):
- Type a draft in channel A → switch to B: `main` keeps the text (sendable to B); fixed clears it.
- Re-select the **same** channel: draft preserved (guard works).
- Reply-quote a message in A → switch to B: citation also cleared.

No unit test added: this is `UIThread`/service-bound controller logic and the desktop module has no controller test harness (UI logic is verified manually here). Happy to extract the clear-draft decision into a testable unit if preferred.

## Summary by CodeRabbit

* **Bug Fixes**
  * Unsent chat messages and draft content are now automatically cleared when switching between channels, providing a clean input slate. Any open citation/quote blocks are closed on channel change. After switching, channel visibility and user/profile state are correctly reapplied so the displayed conversation context matches the newly selected channel.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/ChatMessageContainerController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/ChatMessageContainerController.java
@@ -269,9 +269,18 @@ public class ChatMessageContainerController implements bisq.desktop.common.view.
 
     protected void selectedChannelChanged(ChatChannel<? extends ChatMessage> chatChannel) {
         UIThread.run(() -> {
+            ChatChannel<? extends ChatMessage> previousChannel = model.getSelectedChannel().get();
+            if (previousChannel != null && !previousChannel.equals(chatChannel)) {
+                clearChatInput();
+            }
             model.getSelectedChannel().set(chatChannel);
             applyUserProfileOrChannelChange();
         });
+    }
+
+    private void clearChatInput() {
+        model.getTextInput().set("");
+        citationBlock.close();
     }
 
     private void doSendMessage(String text) {


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/ChatMessageContainerController.java
```diff
@@ -269,11 +269,20 @@ private void applyUserProfileOrChannelChange() {
...
+            ChatChannel<? extends ChatMessage> previousChannel = model.getSelectedChannel().get();
+            if (previousChannel != null && !previousChannel.equals(chatChannel)) {
+                clearUnsentDraft();
+            }
             model.getSelectedChannel().set(chatChannel);
             applyUserProfileOrChannelChange();
         });
     }
 
+    private void clearUnsentDraft() {
```
KimStrand: minor: should we call it `clearChatInput()`? `clearUnsentDraft()` reads like we have a draft concept here

balag3: Done.

KimStrand:
utACK

PR #4812, opened 2026-06-09
Add bank accounts modeling for account musig rest api

salottodev:
## Release Notes

* **New Features**
  * Added support for creating National Bank payment accounts with configurable country, currency, holder details, and banking information.
  * Added support for creating Same Bank payment accounts with comprehensive account setup options including bank account type and national account identifiers.


HenrikJannsen (approved):
utACK

Issue #4712, opened 2026-04-30
Bisq2 did not find the payment onchain when the address is informed by the buyer with capital letters


MPins:
If the buyer inform the bitcoin address with capital letters the bisq2 does not find the payment on-chain.

balag3:
I would like to take this one.

balag3:
Fixed by #4816 (merged). It stayed open because the PR used "Fixes #4782, #4712". GitHub only auto-closes the issue directly after the keyword.

PR #4816, opened 2026-06-10
Canonicalize BTC address before on-chain match

### LINKED ISSUE #4782: Bisq2 fails to detect on-chain transaction when address includes bitcoin: prefix
Bisq2 accept a human readable prefix in the bitcoin address (like bitcoin:bc1q... ) and then does not find the transaction on chain

balag3:
Fixes #4782, #4712.

The buyer's BTC address is matched against the explorer's transaction output with an exact, case-sensitive `equals` in `StateMainChain3b`. Explorers return lowercase bech32, so an address entered with a `bitcoin:` prefix or as uppercase bech32 never matches — the payment shows as not found on chain.

Fix: canonicalize the address right before the comparison — new `BitcoinAddressValidation.canonicalize` strips the `bitcoin:` prefix (via the existing `BitcoinURIScheme`) and lowercases bech32 only; base58 is case-sensitive and left untouched.

This is a **match-time fix only** — it corrects detection regardless of how the address was entered, including trades already on the wire. Input-time normalization (storing it canonical in `BuyerState1a`) is intentionally out of scope, a possible follow-up.

Tests: `BitcoinAddressValidationTest.testCanonicalize` covers the prefix, uppercase bech32, both combined, and the base58 case-preservation guard.

## Bug Fixes
- Enhanced consistency of Bitcoin address validation when comparing transaction outputs to stored payment addresses, improving trade verification accuracy
- Improved handling of different Bitcoin address formats (Bech32 and base58) with proper canonicalization and Bitcoin URI extraction support

## Tests
- Added test coverage for Bitcoin address canonicalization across various address format types

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/StateMainChain3b.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/open_trades/trade_state/states/StateMainChain3b.java
@@ -24,6 +24,7 @@ import bisq.chat.bisq_easy.open_trades.BisqEasyOpenTradeChannel;
 import bisq.common.monetary.Coin;
 import bisq.common.util.ExceptionUtil;
 import bisq.common.util.StringUtils;
+import bisq.common.validation.BitcoinAddressValidation;
 import bisq.desktop.ServiceProvider;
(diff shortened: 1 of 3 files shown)


KimStrand:
utACK

HenrikJannsen (approved):
utACK

PR #4817, opened 2026-06-11
Inform user when chat peer is banned

### LINKED ISSUE #4769: Inform user when peer is banned
If a peer in a private chat or trade is banned, the messages are silently dropped. We should inform the not-banned user about the peer being banned and no messages are sent. The banned peer we do not inform.

balag3:
Fixes #4769.

When the peer in a private or trade chat is banned, their messages are silently dropped and the user gets no feedback — they just see their messages going nowhere. `PrivateChatChannelService.sendMessage` already blocks the send (it returns a failed future), it just doesn't surface anything.

This adds a local system message to the channel — *"Your message was not delivered to {0} because they have been banned."* — so the not-banned user knows what's happening. It's authored by the local user (otherwise `addMessage` would drop it as a banned sender), only added locally so the banned peer is never told. Both private and trade chats go through the same `sendMessage`, so one hook covers both, and it renders via the existing `PROTOCOL_LOG_MESSAGE` system-message box (same style as "X left that chat").

**Mediation fan-out** (raised by @KimStrand): in a mediated trade the send fans out to each trader and the mediator, and sends to non-banned participants still go out. The notice is therefore recipient-specific — it only claims that delivery *to the named banned receiver* failed, so it stays accurate when another participant did receive the message. De-duplication is per banned peer: each banned receiver is named once, repeated send attempts don't stack duplicate notices, and a regular message in between resets the suppression so a later failed send notifies again. (The alternative — aggregating the group-send results into one summary notice — would need failure-classification plumbing across both trade channel services; happy to do that as a follow-up if preferred.)

The de-duplication guard (`alreadyNotifiedAboutBannedPeer`) is unit-tested. The end-to-end behaviour was verified manually with a three-instance setup — two traders plus a mediator, ban check forced on: the sender sees one notice per banned receiver, the banned peer sees nothing, and resends don't stack.

Screenshot from 2026-06-11 16-36-17
Screenshot from 2026-06-11 16-34-31
Screenshot from 2026-06-11 16-28-45

## Summary by CodeRabbit

* **New Features**
  * Added a notification when attempting to send a message to a banned peer, clearly indicating the peer has been banned and the message was not delivered.
  * Implemented deduplication to prevent duplicate notifications.

### DIFF
--- a/chat/src/main/java/bisq/chat/priv/PrivateChatChannelService.java
+++ b/chat/src/main/java/bisq/chat/priv/PrivateChatChannelService.java
@@ -36,9 +36,12 @@ import bisq.persistence.PersistableStore;
 import bisq.user.UserService;
 import bisq.user.identity.UserIdentity;
 import bisq.user.profile.UserProfile;
+import com.google.common.annotations.VisibleForTesting;
 import lombok.extern.slf4j.Slf4j;
 
 import javax.annotation.Nullable;
+import java.util.Collection;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.Optional;
@@ -100,6 +103,7 @@ public abstract class PrivateChatChannelService<
             return CompletableFuture.failedFuture(new RuntimeException());
         }
         if (isPeerBanned(receiver)) {
+            addPeerBannedSystemMessage(channel, myUserIdentity, receiver, date);
             return CompletableFuture.failedFuture(new RuntimeException("Peer is banned"));
         }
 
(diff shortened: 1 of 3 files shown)


KimStrand:
The notice text says “Your message was not delivered,” but mediated trade sends still fan out to other participants. In BisqEasyOpenTradeChannelService and MuSigOpenTradeChannelService, a banned receiver returns a failed future, while non-banned peer/mediator/arbitrator sends are still started. So if only one recipient is banned, the user can see a whole-message failure notice even though another participant received it. We could make the text recipient-specific, e.g. “not delivered to {0}”, or aggregate group-send results before adding the notice.

balag3:
> The notice text says “Your message was not delivered,” but mediated trade sends still fan out to other participants. In BisqEasyOpenTradeChannelService and MuSigOpenTradeChannelService, a banned receiver returns a failed future, while non-banned peer/mediator/arbitrator sends are still started. So if only one recipient is banned, the user can see a whole-message failure notice even though another participant received it. We could make the text recipient-specific, e.g. “not delivered to {0}”, or aggregate group-send results before adding the notice.

Good point. Went with the recipient-specific text: the notice now reads "Your message was not delivered to {0} because they have been banned.".
The dedup is per banned peer now too, a fan-out with several banned receivers names each once, without stacking on resends.
Also attached some screnshots of the different scenarios (private, offer chat, mediation chat)

KimStrand:
utACK

HenrikJannsen (approved):
ACK

PR #4839, opened 2026-06-18
feature/relayed_notifications_deep_linking_support

rodvar:
- add support for including tradeId in relayed-notifications to enable secure and precise deep-linking to trade on notification tap
 - add AI generated related tests

## Summary by CodeRabbit

* **New Features**
  * Mobile notifications now support trade-specific deep linking, enabling direct navigation to relevant trades from push notifications.
  * Notification payload structure enhanced to include trade identifiers for improved routing on mobile clients.

* **Tests**
  * Expanded test coverage for notification JSON serialization and backward compatibility with legacy client payloads.

PR #4842, opened 2026-06-22
chore/bring_push_notifications_deep_linking_support_to_main

rodvar:
- bring #4839  to main
 - add support for including tradeId in relayed-notifications to enab…le secure and precise deep-linking to trade on notification tap


HenrikJannsen (approved):
utACK

PR #4846, opened 2026-06-23
feature/dockerised_distrib_for_trusted_node

rodvar:
- contribs to https://github.com/bisq-network/bisq-mobile/issues/366

### Motivation

 - to distribute bisq2 api in popular Bitcoin-first app stores we need to be able to run it in a docker container. This is a first crucial step to achieve this goal. For devs this is useful, now images of the api can be added to a local docker

### dev logs

 - added docker setup for image generation of the api module (multi-staged, bundled tor, single data/ mount, non-root user)
 - added proper ignores and code comments
 - fully tested locally, can pair to the generated QR code and use the apps normally
 - expose default bisq-mobile websocket port 8090 for now (to be configurable in next PRs)
 - add README.md for docker + TODOs / FIXMEs for next steps
 
Screenshot 2026-06-23 at 11 44 17
Screenshot 2026-06-23 at 11 34 26

## Release Notes

* **New Features**
  * Added Docker container support for the API app as a headless trusted node with persistent `/data`.
  * Exposes port **8090** and configures pairing TTL (default **86400s**).
  * Supports switching between bundled Tor and external Tor via environment settings; pairing QR output is written to the data directory.

* **Documentation**
  * Added a Docker README with build/run instructions, environment variable options, and operational notes (including memory guidance).

* **Chores**
  * Added a `.dockerignore` to reduce Docker build context and speed up builds.

### DIFF
--- a/apps/api-app/docker/Dockerfile
+++ b/apps/api-app/docker/Dockerfile
@@ -1,0 +1,85 @@
+# syntax=docker/dockerfile:1
+#
+# Dockerised Bisq2 trusted node (api-app) — headless daemon for mobile clients.
+#
+# Build context MUST be the bisq2 repository root, e.g. from the repo root:
+#   docker build -f apps/api-app/docker/Dockerfile -t bisq2-api .
+#
+# Currently linux/amd64 only.
+# TODO(multi-arch): add linux/arm64 once the release pipeline does multi-arch builds.
+
+# ── Build stage: compile the api-app distribution from source ──────────────────
+# Self-contained: the image build runs the Gradle build itself, so no host build
+# step is required and the result is reproducible from the repo alone.
+#
+# Must be an Azul Zulu JDK 21: bisq2's Gradle toolchain pins vendor=AZUL
+# (e.g. :apps:desktop:bi2p), so a non-Azul JDK makes Gradle try to auto-download
+# Zulu — which fails (and picks the wrong arch) inside the container. This mirrors
+# the host's sdkman toolchain (.sdkmanrc -> java=21.x.fx-zulu).
+# TODO(build-speed): revisit this build strategy near the finish line — once a
+# release pipeline produces the distribution, this can switch to copying a
+# pre-built `installDist` artifact instead of rebuilding from source each time.
+#
+# Pinned by digest so the build is reproducible from the repo alone (the floating
+# :21 tag moves). Bump the digest when intentionally updating the build JDK.
+FROM azul/zulu-openjdk:21@sha256:fcf711e8fbcb7aa1371e1f831e5114c677e6e9134d9fdd2ee866c7d1027846e2 AS build
+WORKDIR /src
+COPY . .
+# installDist builds the api-app launcher plus every sibling module it depends on.
+# We deliberately do NOT run the full `build` task — packaging the daemon does not
+# need the whole repository's test suite.
+# auto-download=false forces Gradle to use the image's Azul JDK (auto-detected via
+# JAVA_HOME) rather than attempting to provision a toolchain (which fails in-container).
+# The --mount cache keeps the Gradle home (downloaded deps + compiled module output)
+# across builds, so a source change recompiles incrementally instead of from scratch.
+RUN --mount=type=cache,target=/root/.gradle \
+    ./gradlew --no-daemon \
+    -Dorg.gradle.java.installations.auto-download=false \
+    :apps:api-app:installDist
+
+# ── Runtime stage: minimal JRE + system Tor + the built distribution ───────────
+# Azul Zulu JRE to match the build JDK vendor (and bisq1's runtime), pinned by digest
+# for reproducibility. Bump the digest when intentionally updating the runtime JRE.
+FROM azul/zulu-openjdk:21-jre@sha256:97defd96b4ff0e00e3bce1ce66a3bc5a3d69dc05df04fc55030dc6f3b6376771
+
+LABEL org.opencontainers.image.title="Bisq2 trusted node (api-app)" \
+      org.opencontainers.image.description="Headless Bisq2 node exposing the api-app REST/WebSocket API, e.g. for mobile clients" \
+      org.opencontainers.image.source="https://github.com/bisq-network/bisq2" \
+      org.opencontainers.image.licenses="AGPL-3.0-only"
+
+# bisq2's bundled Tor mode on Linux launches the SYSTEM tor binary
+# (TorService#getTorBinaryPath -> EmbeddedTorProcess.getSystemTorPath()), so it must
+# be installed for the default (self-contained) Tor mode to work.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends tor \
+    && rm -rf /var/lib/apt/lists/*
+
+# Non-root runtime user. All persistent state lives under the /data mount.
+RUN useradd --system --create-home --home-dir /home/bisq --shell /usr/sbin/nologin bisq \
+    && mkdir -p /data \
+    && chown -R bisq:bisq /data
+
+# The trailing slash copies the distribution's CONTENTS (bin/, lib/) into
+# /opt/bisq2-api/, so the launcher lands at /opt/bisq2-api/bin/api-app.
+COPY --from=build /src/apps/api-app/build/install/api-app/ /opt/bisq2-api/
+# Fail the build early if the launcher isn't where the entrypoint expects it.
+RUN test -x /opt/bisq2-api/bin/api-app
+COPY apps/api-app/docker/entrypoint.sh /usr/local/bin/entrypoint.sh
+RUN chmod +x /usr/local/bin/entrypoint.sh
+
+USER bisq
+WORKDIR /opt/bisq2-api
+
+# Single mount for all persistent state (identity, db, tor, pairing) so a host-level
+# backup of this one directory captures everything.
+ENV BISQ_DATA_DIR=/data
+VOLUME ["/data"]
+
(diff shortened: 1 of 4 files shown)


apps/api-app/docker/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Setting `TOR_CONTROL_PORT` makes the entrypoint write `${BISQ_DATA_DIR}/tor/external_tor.config`
+and enable external-Tor mode (`TOR_SKIP_LAUNCH=1`).
+
+> FIXME(external-tor): bisq2's external-Tor mode talks to Tor's **control port** (with
+> cookie auth) and discovers the SOCKS port from it — it does not just consume a SOCKS
+> port. `TorSocksProxyFactory` also hardcodes `127.0.0.1`, so the external Tor must be
+> reachable on the container's loopback. Reusing a host/shared Tor needs its control
+> port + auth cookie exposed on loopback inside the container; this is wired up when
+> integrating with the target self-hosting platform's bundled Tor.
```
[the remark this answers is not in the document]
rodvar: oops I didn't commit these, let me fix now in my next commit

apps/api-app/docker/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+# We deliberately do NOT run the full `build` task — packaging the daemon does not
+# need the whole repository's test suite.
+# auto-download=false forces Gradle to use the image's Azul JDK (auto-detected via
+# JAVA_HOME) rather than attempting to provision a toolchain (which fails in-container).
+RUN ./gradlew --no-daemon \
+    -Dorg.gradle.java.installations.auto-download=false \
+    :apps:api-app:installDist
+
+# ── Runtime stage: minimal JRE + system Tor + the built distribution ───────────
+FROM eclipse-temurin:21-jre
```
HenrikJannsen: Why not use azul/zulu JRE here? 

We also should pin the exact version used in Bisq 2 to be in line with upcoming reproducible builds

apps/api-app/docker/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+
+LABEL org.opencontainers.image.title="Bisq2 trusted node (api-app)" \
```
HenrikJannsen, line 45: Do we want to limit it to mobile use case or keep it more generic as API?

HenrikJannsen (approved):
utACK

HenrikJannsen:
@wodoro Can you review as well?

apps/api-app/docker/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+# step is required and the result is reproducible from the repo alone.
+#
+# Must be an Azul Zulu JDK 21: bisq2's Gradle toolchain pins vendor=AZUL
+# (e.g. :apps:desktop:bi2p), so a non-Azul JDK makes Gradle try to auto-download
+# Zulu — which fails (and picks the wrong arch) inside the container. This mirrors
+# the host's sdkman toolchain (.sdkmanrc -> java=21.x.fx-zulu).
+# TODO(build-speed): revisit this build strategy near the finish line — once a
+# release pipeline produces the distribution, this can switch to copying a
+# pre-built `installDist` artifact instead of rebuilding from source each time.
+FROM azul/zulu-openjdk:21 AS build
```
wodoro: Pin base image by digest for the "reproducible from the repo alone" goal: `:21` is a floating tag, so builds aren't reproducible. use `FROM azul/zulu-openjdk:21@sha256:...`

apps/api-app/docker/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+WORKDIR /src
+COPY . .
```
wodoro, line 27: `COPY . .` invalidates the build-cache layer on any file change, forcing a full Gradle recompile. The Gradle cache mount (see gradlew line) absorbs most of the cost; flagging for awareness.

apps/api-app/docker/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+# installDist builds the api-app launcher plus every sibling module it depends on.
+# We deliberately do NOT run the full `build` task — packaging the daemon does not
+# need the whole repository's test suite.
+# auto-download=false forces Gradle to use the image's Azul JDK (auto-detected via
+# JAVA_HOME) rather than attempting to provision a toolchain (which fails in-container).
+RUN ./gradlew --no-daemon \
+    -Dorg.gradle.java.installations.auto-download=false \
+    :apps:api-app:installDist
```
wodoro: Add a BuildKit Gradle cache mount. `syntax=docker/dockerfile:1` already enables it. Without it every source change re-downloads all deps and recompiles every module:
```
RUN --mount=type=cache,target=/root/.gradle ./gradlew --no-daemon \
          -Dorg.gradle.java.installations.auto-download=false \
          :apps:api-app:installDist
```

apps/api-app/docker/Dockerfile
wodoro: Same as build stage: pin runtime base by digest for reproducibility
also as Henrik has already pointed out, lets match the jdk and bisq1 and use `azul/zulu-openjdk:21-jre` (latest hash at the moment is `49e8056e23f5`, so `azul/zulu-openjdk:21-jre@sha256:49e8056e23f5b1dae71991dd8b4751740b91b1e260f56b129ba935e98c6ec878`)

apps/api-app/docker/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+      org.opencontainers.image.description="Headless Bisq2 trusted node (api-app) for mobile clients" \
+      org.opencontainers.image.source="https://github.com/bisq-network/bisq2" \
+      org.opencontainers.image.licenses="AGPL-3.0-only"
+
+# bisq2's bundled Tor mode on Linux launches the SYSTEM tor binary
+# (TorService#getTorBinaryPath -> EmbeddedTorProcess.getSystemTorPath()), so it must
+# be installed for the default (self-contained) Tor mode to work.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends tor \
+    && rm -rf /var/lib/apt/lists/*
```
wodoro, line 55: `apt-get update` + unpinned `tor` is non-deterministic across builds, so it should be pinned probably

apps/api-app/docker/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+COPY apps/api-app/docker/entrypoint.sh /usr/local/bin/entrypoint.sh
+RUN chmod +x /usr/local/bin/entrypoint.sh
+
+USER bisq
+WORKDIR /opt/bisq2-api
+
+# Single mount for all persistent state (identity, db, tor, pairing) so a host-level
+# backup of this one directory captures everything.
+ENV BISQ_DATA_DIR=/data
+VOLUME ["/data"]
```
wodoro, line 76: `VOLUME` with no `-v` at run time creates an anonymous volume each `docker run`, which pile up orphaned. Document always passing `-v` in the README.

apps/api-app/docker/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+
+# WebSocket API. Bound to loopback inside the container by default; the host is
+# expected to forward/publish it as appropriate for the deployment.
+EXPOSE 8090
```
wodoro, line 83: Optional: add a HEALTHCHECK so orchestrators detect a hung daemon, e.g. `HEALTHCHECK CMD curl -f http://127.0.0.1:8090/ || exit 1`

apps/api-app/docker/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Notes
+
+- Recommended `mem_limit`: 2g; advertise ~4GB free RAM.
```
wodoro: 2g or 4g? a bit confusing

wodoro (changes requested):
Did a quick review with AI help, posted the ones I agree with

apps/api-app/docker/Dockerfile
rodvar: I think we are good as is in this case, debian repos serve `only the current version` - pinning it risk getting 404s issues

we could pin to a debian snapshot but I suggest we leave it as is

apps/api-app/docker/Dockerfile
rodvar: curl isn't in the JRE base, and 8090 is WebSocket-only, this would need a real health endpoint. Its a good idea but let's keep it in the TODO for now 🙏  (we could reuse the version endpoint we already use as a health check in mobile)

apps/api-app/docker/README.md
rodvar: thank you I'll reword sth lik mem_limit ≈ 2g, host should have ~4GB free

apps/api-app/docker/Dockerfile
rodvar: I'll make it a bit more generic + give the example for bisq-mobile

apps/api-app/docker/Dockerfile
rodvar: 💯 thank you will address along with wod's comments

rodvar:
using azul adds 100MB+ to the image 

Screenshot 2026-06-24 at 15 59 12

I tried to lower it trying different artifacts but this is the best one - we can accept this in favour of using our most preferred vendor 👍

wodoro (approved):
utACK

rodvar:
@HenrikJannsen we are ready now 💪 💪 💪

HenrikJannsen (approved):
utACK

PR #4849, opened 2026-06-24
Use O(1) lookups for trade item services

### LINKED ISSUE #4722: perf: Replace O(n) linear scan with O(1) Set lookup in ClosedTradeItemsService (and BisqEasyHistoryController)
## Summary

In `ClosedTradeItemsService` (introduced in PR #4700), the `onAdded` and `onRemoved` handlers use a linear `findListItem` scan over the items list to check for existing trade IDs. This is O(n) per operation.

A parallel `Set<String> knownIds` would make these lookups O(1) and is trivial to implement.

The same pattern exists in `BisqEasyHistoryController`, so both should be addressed together for consistency.

## Suggested Fix

Add a `Set<String> knownIds` field alongside the `ObservableArray<ClosedTradeIndexedItem> items` list. Maintain the set in sync with the list on add, remove, and clear events. Replace the linear scan with a `knownIds.contains(tradeId)` check.

## References

- PR #4700 (comment): https://github.com/bisq-network/bisq2/pull/4700#discussion_r3186420723
- Requested by: @rodvar

balag3:
## What

Replaces the `O(n)` `findListItem` stream scans used to dedup and look up trade items by id with `O(1)` lookups.

- **`BisqEasyHistoryController`** keeps a `Set<String>` of known trade ids for the existence check on add.
- **`OpenTradeItemsService`** keeps a `Map<String, item>` so the by-id lookups it also does on update and remove are `O(1)`. All list mutations go through small `addItem`/`removeItem`/`clearItems` helpers so the map can't drift from the `items` list.

## Scope

The original target `ClosedTradeItemsService` no longer exists on `main` (it was dropped before #4700 merged). After checking with @rodvar, the agreed scope is `BisqEasyHistoryController` (Set) + `OpenTradeItemsService` (Map, since it retrieves items by id, not just checks existence). The two larger desktop controllers with the same pattern (`BisqEasyOpenTradesController`, `MuSigPendingTradesController`) are intentionally left as a follow-up — small lists, and not worth the churn here.

## Tests

Adds `OpenTradeItemsServiceTest` covering the add/dedup, remove and clear paths, so the items list and the id index can't fall out of sync. `BisqEasyHistoryController` is a UI controller with no test precedent, so it's covered by inspection only.

Fixes #4722

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of open trades so items stay consistent across updates, removals, and resets.
  * Fixed duplicate entries in trade history and made cleared/re-added trades appear correctly.
* **Quality**
  * Added automated coverage for trade item syncing and history deduplication.

### DIFF
--- a/api/src/main/java/bisq/api/web_socket/domain/OpenTradeItemsService.java
+++ b/api/src/main/java/bisq/api/web_socket/domain/OpenTradeItemsService.java
@@ -56,6 +56,7 @@ public class OpenTradeItemsService implements Service {
     @Nullable
     private Pin channelsPin, tradesPin;
     private final Map<String, Pin> isInMediationPinMap = new HashMap<>();
+    private final Map<String, TradeItemPresentationDto> itemByTradeId = new HashMap<>();
 
     public OpenTradeItemsService(ChatService chatService,
                                  TradeService tradeService,
@@ -124,7 +125,7 @@ public class OpenTradeItemsService implements Service {
         }
         isInMediationPinMap.values().forEach(Pin::unbind);
         isInMediationPinMap.clear();
-        items.clear();
+        clearItems();
         return CompletableFuture.completedFuture(true);
     }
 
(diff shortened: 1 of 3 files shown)


HenrikJannsen (approved):
utACK

Issue #4114, opened 2025-12-04
Navigate to private chat fails sometimes


HenrikJannsen:
To navigate to a screen in a tab view fails sometimes. This is problematic when we navigate to private chats and the user ends up in the public discussion chat. If the user does not pay attention he might share private data in the public chat, thinking he is in the private chat.

This bug is caused in the navigation framework.

iMac7:
@HenrikJannsen is this still valid? can't reproduce 🤔

HenrikJannsen:
Yes, but it happens only sometimes. It is likely some bug in the Navigation framework code.
Its happens also with other tabs when you are not on the first tab and then navigate back to the main target sometimes it jumps to the first tab instead the prev. selected one. That is likely the same bug.

balag3:
I spent some time tracing this one.. and quite a lot of tokens :)  Here's what's going on.

  ## Root cause

  Navigation pushes a target out to whatever host controllers are registered for it: `Navigation.navigateTo(target)` walks `navigationControllers` and calls
  `processNavigationTarget(target)` on each host on the target's path.

  The Chat tab host (`CommonChatTabController`, registered under `NavigationTarget.CHAT`) only gets created the first time you open Chat ?
  `ContentController.createController(CHAT)` builds it on demand. So if you're somewhere else and jump straight into a private chat, that host doesn't exist
  yet.
  
  When `navigateTo(CHAT_PRIVATE)` runs in that state, the loop reaches `CONTENT`, which creates the Chat host on the spot. The host registers itself while the
  loop is still running, but by then the loop has already passed the slot where it would have visited it, so `CHAT_PRIVATE` never gets handed to the Chat host.
  The host comes up in `onActivateInternal` with nothing to go on and opens its default tab, which for the discussion domain is the public `CHAT_DISCUSSION`
  (`CommonChatTabModel.getDefaultNavigationTarget()`). You end up in public even though you asked for a private chat.

  Why it only happens sometimes: `navigationControllers` is a `ConcurrentHashMap` keyed on the `NavigationTarget` enum, and enum `hashCode()` is
  identity-based, so iteration order shuffles between JVM runs. Some runs the loop circles back and reaches the freshly-registered Chat host (works); other
  runs it doesn't (you land in public).
  
  ## Reproducing it

  You have to open a private chat while the Chat screen is cold ? you're not already in Chat. A peer's profile card ? "Send private message" is the quickest
  way in.

[...]

PR #4852, opened 2026-06-26
Don't record history for automatic tab selection

### LINKED ISSUE #4821: Automatic initial tab selection pushes a navigation history entry
This came up in the review discussion of #4809. It's pre-existing behaviour on
 `main` (not introduced by that PR), so we agreed to handle it as a follow-up.

 When a tab host activates, it automatically selects its default (or persisted)
 tab. That selection goes through `TabController.onTabSelected`, which calls
 `Navigation.navigateTo(target)` and that appends to the navigation history,
 exactly as if the user had clicked the tab. The same happens when a deeper
 target arrives and the host corrects the tab selection.

 The effect is easiest to see when opening a private chat from outside the Chat
 screen while the Chat host isn't created yet (e.g. first use this session).
 Traced in #4809:

 1. `navigateTo(CHAT_PRIVATE)` -> history: `CHAT_PRIVATE`
 2. The lazily created Chat host activates and selects its default tab ->
    history: `CHAT_PRIVATE, CHAT_DISCUSSION`
 3. The dispatch reaches the host with the deep target and re-selects the tab ->
    history: `CHAT_PRIVATE, CHAT_DISCUSSION, CHAT_PRIVATE`

 The user lands in the private chat as intended, but pressing back now goes to
 the public Discussion channel instead of where they actually came from.

 I'd suggest only recording history entries for user-initiated tab clicks: the
 automatic selections (initial/default selection and deep-target correction)
 would still dispatch the navigation, just without touching the history - e.g.
 a `navigateTo` variant that skips the history append, used by the automatic
 paths in `TabController`. Suppressing the default selection while a deeper
 target is in flight was already considered and rejected in #4809, since it
 would reintroduce the global in-flight dispatch state that PR removed.

balag3:
## Problem

Fixes #4821. A tab host auto-selects its default or persisted tab when it activates, and a deep-target dispatch later corrects that selection. Both paths went through `Navigation.navigateTo`, which records a navigation-history entry as if the user had clicked the tab.

In practice, opening a private chat from outside the Chat screen (the lazy Chat-host flow from #4114) records a spurious `CHAT_DISCUSSION` entry, the cold host's default tab, before correcting to `CHAT_PRIVATE`. A `back()` from there would land on the Discussion channel rather than the user's actual origin.

## Fix

Record a history entry only for user-initiated tab clicks. Automatic selections (the initial/persisted tab and deep-target corrections) now go through a new `Navigation.navigateToWithoutAddingToHistory`, which dispatches and updates the current target but leaves the history untouched, mirroring the existing `navigateTo(target, data)` overload.

`TabController` routes the user-click path and the two automatic paths through separate named methods (`onTabSelected` and `selectTabWithoutHistory`) sharing a `selectTabButton` helper, which avoids a boolean flag argument. `AuthorizedRoleView` auto-selects the first visible role tab directly (from view-attach and a role-list listener, never a user click), so it points at the same no-history method.

## Note

The history list is currently maintained on every navigation but not consumed by any UI control, since there is no back or forward affordance wired up today. This change keeps the history accurate for whenever one is added, and is the follow-up agreed in #4114.

## Testing

- New unit test `NavigationControllerTest.navigateToWithoutAddingToHistoryDoesNotAffectBackNavigation`: a user navigation and an automatic selection both dispatch to the host, but `back()` returns the user-navigated target, proving the automatic selection never entered the history. Reuses the existing `RecordingController` harness.
- Full `:apps:desktop:desktop:test` green (82 tests).
- Verified live by reading the running app's `Navigation.history` before and after the fix: without it the history contains the spurious `CHAT_DISCUSSION` entry (plus a duplicate `MU_SIG_OFFERBOOK` from the MuSig tab host's auto-select); with it, every entry corresponds to a real user navigation.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/common/view/Navigation.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/common/view/Navigation.java
@@ -72,6 +72,11 @@ public class Navigation {
         currentNavigationTarget.set(navigationTarget);
     }
 
+    public static void navigateToWithoutAddingToHistory(NavigationTarget navigationTarget) {
+        dispatch(navigationTarget, Optional.empty());
+        currentNavigationTarget.set(navigationTarget);
+    }
+
     // Dispatches the target to each registered host along its navigation path, from the root down to
     // the target. Iterating the path in order (rather than the unordered controllers map) processes
     // parents before children, so a host that lazily creates a child host on its path registers that
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/common/view/TabController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/common/view/TabController.java
@@ -39,25 +39,36 @@ public abstract class TabController<T extends TabModel> extends NavigationContro
     public void onActivateInternal() {
         super.onActivateInternal();
 
-        onTabSelected(model.getNavigationTarget());
+        selectTabWithoutHistory(model.getNavigationTarget());
     }
 
     @Override
     protected void onNavigationTargetApplied(NavigationTarget navigationTarget, Optional<Object> data) {
         if (model.getSelectedTabButton().get() != null &&
                 navigationTarget != model.getSelectedTabButton().get().getNavigationTarget()) {
-            onTabSelected(navigationTarget);
+            selectTabWithoutHistory(navigationTarget);
         }
     }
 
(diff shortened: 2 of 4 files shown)


KimStrand:
utACK

HenrikJannsen (approved):
utACK

PR #4859, opened 2026-07-03
Sync `rpc.proto` with bisq-musig & add Custom Payout protos

stejbac:
Sync the `Musig` gRPC service with the latest Rust changes, introducing the methods `SignCustomPayoutTx` and `CustomCloseTrade` to make a custom payout from the Deposit outputs, for a generic mediated trade closure. The first method takes the requisite feerate (which needs to be agreed with the trade peer in advance, or say just use the `NonceSharesRequest.preparedTxFeeRate` parameter agreed at the trade start) and the seller's payout amount before tx fee (from which the buyer's payout amount is inferred), and returns a Custom Payout PSBT with a single multisig signature for each of its two inputs. This can then be sent to the peer, who also calls the first method, followed by the second, taking the received PSBT as input, combining it internally and broadcasting the resultant signed Custom Payout Tx. (The first trader's Rust service should then pick it up from the network and automatically close the trade, but the `Musig` service currently only mocks broadcasts, so this hasn't been implemented yet.)

Currently, the custom payout construction is "single shot", in the sense that once the parameters (payout split & feerate) have been passed to the first method, it won't accept any subsequent readjustment (though it should be possible to change this).

Also implement Java protos for each of the four new message types: `CustomCloseTradeRequest`, `CustomCloseTradeResponse`, `CustomPayoutPsbtRequest` and `CustomPayoutPsbt`.

## Summary by CodeRabbit

* **New Features**
  * Added support for signing custom payout transactions.
  * Added a new trade-closing flow that lets users finalize a trade using a custom payout transaction.
  * Introduced exchanging payout details as PSBT payloads, including transaction IDs and buyer/seller payout amounts.
* **Refactor**
  * Marked contractual transaction ID message handling as immutable for consistency.

### DIFF
--- a/trade/src/main/java/bisq/trade/mu_sig/messages/grpc/ContractualTxIds.java
+++ b/trade/src/main/java/bisq/trade/mu_sig/messages/grpc/ContractualTxIds.java
@@ -25,7 +25,7 @@ import lombok.ToString;
 @Getter
 @ToString
 @EqualsAndHashCode
-public class ContractualTxIds implements Proto {
+public final class ContractualTxIds implements Proto {
     private final String depositTxId;
     private final String buyersWarningTxId;
     private final String sellersWarningTxId;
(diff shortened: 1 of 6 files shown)


KimStrand (commented):
Should we make the four new classes final?

stejbac:
> Should we make the four new classes final?

Yes, I think so, for consistency with the other proto classes in the `grpc` package. It looks like I also missed `final` when I added the `ContractualTxIds` proto in an earlier PR.

KimStrand:
utACK

Issue #4410, opened 2026-02-18
Round Fiat amounts at create offer


HenrikJannsen:
Currently when creating a MuSig offer with range amounts very close it can happen to have a slightly different amount. It should be rounded to the fiat value and only if that value is different it becomes a range amount.
I could not further reproduce it but i had one offer with such.
There is also not rounded values in the offer display.

<img width="394" height="172" alt="Image" src="https://github.com/user-attachments/assets/db429cc7-0b1f-4c0d-8b66-09721fd0a2d4" />

fenlark:
On the "not rounded values in the offer display" part of this issue:

The range vs fixed side is handled in #4865 (a range whose min and max are the same fiat value now collapses to a fixed offer, so it no longer shows as a range with two near-identical values).

For the display itself: the MuSig review, offer details and take-offer screens already round fiat to the display precision (via `AmountFormatter.formatQuoteAmount`, which uses low precision for `Fiat`). The only place showing full-precision fiat is the offerbook list (`MuSigOfferListItem`, `useLowPrecision=false`), and BisqEasy's list does the same (`OfferbookListItem`, also `false`), so it looks like a deliberate shared convention rather than a MuSig-specific bug.

Do you want the offerbook list to round fiat too? That would be a cross-cutting change across BisqEasy and MuSig, so I left it out of #4865. Happy to do it as a follow-up if that is the intent.

PR #4865, opened 2026-07-09
Decide offer amount range on rounded fiat value

fenlark:
When creating a MuSig offer on a fiat market, the range vs fixed amount decision compared the base side amounts (sats). Two quote side amounts that are equal once rounded to the fiat display precision can still map to base side amounts that differ by a few sats due to price conversion rounding, so a range offer was created when the fiat min and max are effectively the same.

Decide on the rounded quote side value instead: a range only remains when the fiat amounts actually differ, otherwise the offer collapses to a fixed amount. Non-fiat markets keep the exact base side comparison.

Extract the decision to a package-private static isFixedAmount, unit tested for equal, near-equal and differing fiat amounts, non-fiat markets and missing quote amounts.

Part of #4410

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how fixed vs. range offers are detected, especially for BTC/fiat markets.
  * Prevented offers from being shown as ranges when fiat-side values only differ due to display rounding.
  * Kept exact behavior for non-fiat markets and when quote-side values are unavailable.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/create_offer/amount_and_price/amount/MuSigCreateOfferAmountController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/create_offer/amount_and_price/amount/MuSigCreateOfferAmountController.java
@@ -46,6 +46,7 @@ import bisq.settings.CookieKey;
 import bisq.settings.SettingsService;
 import bisq.user.identity.UserIdentityService;
 import bisq.user.reputation.ReputationService;
+import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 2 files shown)


fenlark:
Relation to #4800 (already merged): it collapses an exact `min == max` base amount in `AmountSpecFactory`. #4410 is the case where the amounts are equal once rounded to the fiat value but still differ by a few sats on the base side, which the exact == misses.

The rounded fiat comparison needs the quote side amounts, which on a fiat market `AmountSpecFactory` doesn't get (base side sats only), so I did it in the controller. Centralizing would mean widening the factory to also take the quote side amounts and touching its callers.

KimStrand (commented):
utACK

PR #4727, opened 2026-05-11
feat: Add Polygon USDC stablecoin support to Bisq Easy

FeatureSpitter:
## Motivation

Bisq 2 supports BTC/fiat trading via Bisq Easy, but there is growing demand for **KYC-free BTC-to-stablecoin trading** — especially driven by the rise of Polymarket and on-chain prediction markets, where users need to move between BTC and stablecoins without centralized intermediaries.

Adding stablecoin support to Bisq Easy lets users:
- Trade BTC for USDC, USDT, DAI, and other stablecoins directly peer-to-peer
- Settle on-chain across Polygon, Ethereum, Tron, Solana, and BNB Smart Chain
- Avoid KYC entirely — no exchange accounts needed

## What was already there

The Bisq team had built a comprehensive **stablecoin domain model** that was not yet wired into the Bisq Easy trading protocol:

- **7 stablecoins**: USDT, USDC, DAI, FDUSD, TUSD, USDP, GUSD
- **4 networks**: Ethereum (ERC-20), Tron (TRC-20), BNB Smart Chain (BEP-20), Solana (SPL)
- **11 payment rails** across these coins and networks
- Full `StableCoin`, `StableCoinPaymentRail`, `StableCoinPaymentMethod`, `StableCoinPaymentMethodSpec` types
- `Market.isBtcStableCoinMarket()` detection
- `StableCoinRepository` with coin metadata

None of this was integrated into Bisq Easy's offer creation, trade wizard, settlement flow, or account management.

## What this PR adds

**This PR adds the Polygon network, the USDC_POLYGON rail, and then performs the entire Bisq Easy integration** — unlocking ALL pre-existing stablecoins (plus the new USDC on Polygon) for trading.

image

image

### New infrastructure
- `POLYGON` added to `StableCoin.Network` enum
- `USDC_POLYGON` added to `StableCoinRepository` and `StableCoinPaymentRail`

### Core integration

- **Generalized quote-side payment handling** — refactored `PaymentMethodSpec`, `Market`, and trade flows to support stablecoins alongside fiat on the quote side (Protobuf `oneof`, models, UI)
- **StableCoinAccount type** — new account/payload with EVM address validation, creation wizard with network selection, and full Protobuf persistence
- **GUI/UX stablecoin polish** — replaced fiat-specific labels throughout trade wizard, offer details, seller states, payment method validation, and waiting animations with stablecoin-aware variants; fixed account wizard null controller and model reset bugs; moved hardcoded strings to i18n

[...]

PR #4867, opened 2026-07-10
Require explicit precision on crypto assets

fenlark:
## Problem

bisq2 derives a crypto amount's precision from a hardcoded switch in
`Coin.derivePrecision(code)`: XMR to 12, BSQ to 2, everything else to 8.
`CryptoAsset` carries only code and name, so an asset cannot state its
own precision. A coin with fewer than 8 decimals added to the repository
(for example USDC, which has 6) therefore silently gets precision 8, and
its trade amount carries digits the chain cannot send.

Nothing sub-8 is listed today except BSQ (handled), so this is latent.
But #4727 wires stablecoin trading (USDC, USDT, DAI) into Bisq Easy,
where a USDC quote amount would be built at precision 8 while USDC has
6 decimals.

The same class of problem in bisq1 is fixed in bisq-network/bisq#7960.
There the trade volume has to be rounded because the `Altcoin` type is
fixed at 8 decimals; here a correct per-asset precision is enough.

## Change

- Add a required `precision` to `CryptoAsset`, so a coin cannot be added
  without stating it. Repository entries now declare it (BTC, LTC and
  the rest = 8, XMR = 12, BSQ = 2).
- `Coin.derivePrecision(code)` reads the precision from
  `CryptoAssetRepository` instead of the hardcoded switch, which is
  removed. Custom (unlisted) coins keep the default 8.
- A `Coin`'s low display precision is `min(precision, 4)`, replacing
  the `BSQ` special case so it never exceeds the coin's own precision.

A `Coin` already stores its value at `10^-precision`, so a correct
precision makes the amount sendable by construction and no rounding
sites are needed. Precision is re-derived from the code via the
repository, not serialized, so there is no proto or persistence change.

## Open questions

- bisq1 effectively caps altcoin precision at 8 (its `Altcoin` storage
  exponent), XMR included. bisq2 already exceeds that (XMR = 12), so
  this treats precision as a per-asset property with a `[0, 12]` ceiling
  rather than a hard 8 cap. If a hard cap (say <= 8 except XMR) is the
  intended policy, the ceiling in the `CryptoAsset` constructor should
  be tightened instead. Which is intended?
- Stablecoins and CBDCs are also `DigitalAsset` subtypes and fall
  through `derivePrecision` to the same default 8 (their codes are not
  fiat, e.g. `USDC`, `eCNY`). They are not fixed in this PR because
  whether a stablecoin trade is denominated in the token (needs
  precision) or in its fiat peg (precision 4) is undecided. Related:
  #4727.
- `Fiat` is a separate but related case: a flat precision 4 / low
  precision 2 for every currency that ignores the currency's own
  fraction digits (EUR = 2, JPY = 0), so a fiat amount can show more
  decimals than the payment rail supports. Out of scope for this PR.

## Tests

`CoinPrecisionTest` covers that precision comes from the repository
(XMR = 12, BSQ = 2, BTC = 8, unknown = 8), that low precision never
exceeds precision, that a sub-8 asset keeps its precision, and that an
out-of-range precision is rejected.

## Summary by CodeRabbit

* **New Features**
  * Added configurable decimal precision for crypto assets.
  * Supported assets now use their defined precision for amounts and trading values.
  * Custom assets default to 8 decimal places, while unknown assets safely fall back to the same precision.
  * Precision values are validated to remain within the supported range.

* **Bug Fixes**
  * Improved handling of assets with fewer than 8 decimal places.
  * Low-precision amount display is now capped appropriately.

### DIFF
--- a/common/src/main/java/bisq/common/asset/CryptoAsset.java
+++ b/common/src/main/java/bisq/common/asset/CryptoAsset.java
@@ -23,22 +23,34 @@ import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.ToString;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 @EqualsAndHashCode(callSuper = true)
 @ToString(callSuper = true)
 public final class CryptoAsset extends DigitalAsset {
 
     @Getter
+    @EqualsAndHashCode.Exclude
+    private final int precision;
+    @Getter
     private transient final Validation addressValidation;
     @Getter
     private transient final boolean supportAutoConf;
 
-    // For custom cryptoCurrencies not listed in Bisq
+    // For custom cryptoCurrencies not listed in Bisq. Their native precision is unknown, so we
+    // fall back to 8. Listed assets must state their precision explicitly (see the constructor
+    // below) so a coin with fewer than 8 decimals cannot silently use 8 and carry unsendable digits.
     public CryptoAsset(String code) {
-        this(code, code);
+        this(code, code, 8);
     }
 
-    public CryptoAsset(String code, String name) {
+    public CryptoAsset(String code, String name, int precision) {
         super(code, name);
+        // precision is the number of decimals Bisq trades the coin at; it must be <= the coin's
+        // native on-chain decimals (fewer is safe, coarser rounding). 12 (XMR) is the highest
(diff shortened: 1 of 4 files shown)


KimStrand:
utACK

PR #4869, opened 2026-07-10
Avoid exception if no price found for a market

### LINKED ISSUE #4271: Exception if no price found for a market
In case the user had Bulgaria as default locale, we use BGN as default market and try to find a market price for that, which fails.
Bisq Easy offer domain and UI rely on existing market price. There is no easy fix for that. We should add a filter before applying any market to ensure that no market is selected which has not a market price available. 

To reproduce the issue one can set the default locale to `new Locale("bg", "BG")` in LocaleRepository and use the code base before https://github.com/bisq-network/bisq2/pull/4270 was merged (or revert that commit for testing). Then when opening the trade wizard an exception is thrown.

neminatha:
fixes #4271

### To Test

1. Temporarily add the following override to `FiatCurrencyRepository.java` for testing. Replace the country code `"DE"` with your local country code. This associates the legacy currency `"BGN"` with that country code:

  common/src/main/java/bisq/common/asset/FiatCurrencyRepository.java#L101
```java
if (countryCode.equals("DE")) {
    return new FiatCurrency("BGN");
}
```

2. Start the application and create a test user.

3. Stop the application and make a copy of the data directory. This copy can be reused for subsequent test runs.

4. Revert the temporary change.

5. Start the application using the copied data directory.

6. Create one or more test offers:
   - Bisq Easy (direct or assisted)
   - MuSig Offerbook
   - MuSig My Offers tab
   - Create a MuSig crypto offer and switch in the wizard from crypto to the fiat offers

**Expected result:** No exception is thrown when no market price is available.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved market selection in the trade wizard, including more reliable default selection.
  * Prevented offer creation when no market is selected or when market pricing is unavailable.
  * Added clear warnings explaining why offer creation cannot continue.
  * Prevented unpriced markets from updating offer details or proceeding to the create-offer screen.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/trade_wizard/direction_and_market/TradeWizardDirectionAndMarketController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/trade_wizard/direction_and_market/TradeWizardDirectionAndMarketController.java
@@ -19,6 +19,7 @@ package bisq.desktop.main.content.bisq_easy.trade_wizard.direction_and_market;
 
 import bisq.bisq_easy.BisqEasyTradeAmountLimits;
 import bisq.bonded_roles.market_price.MarketPriceService;
+import bisq.chat.ChatChannel;
 import bisq.chat.ChatMessage;
 import bisq.chat.bisq_easy.offerbook.BisqEasyOfferbookChannel;
 import bisq.chat.bisq_easy.offerbook.BisqEasyOfferbookChannelService;
@@ -42,6 +43,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.fxmisc.easybind.EasyBind;
 import org.fxmisc.easybind.Subscription;
 
+import javax.annotation.Nullable;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Consumer;
@@ -101,15 +103,6 @@ public class TradeWizardDirectionAndMarketController implements Controller {
         applyShowReputationInfo();
 
         model.getSearchText().set("");
-        if (model.getSelectedMarket().get() == null) {
-            // Use selected public channel or if private channel is selected we use any of the public channels for
-            // setting the default market
-            Optional.ofNullable(bisqEasyOfferbookSelectionService.getSelectedChannel().get())
-                    .filter(channel -> channel instanceof BisqEasyOfferbookChannel)
-                    .map(channel -> (BisqEasyOfferbookChannel) channel)
-                    .map(BisqEasyOfferbookChannel::getMarket)
-                    .ifPresent(market -> model.getSelectedMarket().set(market));
-        }
 
         model.getListItems().setAll(MarketRepository.getAllFiatMarkets().stream()
                 .filter(market -> marketPriceService.getMarketPriceByCurrencyMap().containsKey(market))
@@ -126,11 +119,7 @@ public class TradeWizardDirectionAndMarketController implements Controller {
                             .map(ChatMessage::getAuthorUserProfileId)
                             .distinct()
                             .count();
-                    TradeWizardDirectionAndMarketView.ListItem item = new TradeWizardDirectionAndMarketView.ListItem(market, numOffersInChannel, numUsersInChannel);
-                    if (market.equals(model.getSelectedMarket().get())) {
-                        model.getSelectedMarketListItem().set(item);
-                    }
-                    return item;
+                    return new TradeWizardDirectionAndMarketView.ListItem(market, numOffersInChannel, numUsersInChannel);
                 })
                 .collect(Collectors.toList()));
 
@@ -146,6 +135,16 @@ public class TradeWizardDirectionAndMarketController implements Controller {
                 );
             }
         });
+
+        // Use the market from the selected public channel, or fall back to the service's default channel,
+        // or the first available market list item
+        findMarketListItem(bisqEasyOfferbookSelectionService.getSelectedChannel().get())
+                .or(() -> bisqEasyOfferbookChannelService.getDefaultChannel().flatMap(this::findMarketListItem))
+                .or(this::findFirstMarketListItem)
+                .ifPresentOrElse(
+                        this::onMarketListItemClicked,
+                        () -> log.error("No market list item found to select.")
+                );
     }
 
     @Override
@@ -226,5 +225,22 @@ public class TradeWizardDirectionAndMarketController implements Controller {
     private void showReputationInfoOverlay() {
         navigationButtonsVisibleHandler.accept(false);
         model.getShowReputationInfo().set(true);
+    }
+
+    private Optional<TradeWizardDirectionAndMarketView.ListItem> findMarketListItem(@Nullable ChatChannel<?> channel) {
+        if (channel instanceof BisqEasyOfferbookChannel) {
+            return findMarketListItem(((BisqEasyOfferbookChannel) channel).getMarket());
+        }
+        return Optional.empty();
+    }
+
+    private Optional<TradeWizardDirectionAndMarketView.ListItem> findMarketListItem(Market market) {
+        return model.getListItems().stream()
+                .filter(item -> item.getMarket().equals(market))
+                .findAny();
+    }
+
+    private Optional<TradeWizardDirectionAndMarketView.ListItem> findFirstMarketListItem() {
+        return model.getListItems().stream().findFirst();
     }
 }
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/create_offer/direction_and_market/MuSigCreateOfferDirectionAndMarketController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/create_offer/direction_and_market/MuSigCreateOfferDirectionAndMarketController.java
@@ -34,10 +34,13 @@ import lombok.extern.slf4j.Slf4j;
 import org.fxmisc.easybind.EasyBind;
 import org.fxmisc.easybind.Subscription;
 
+import javax.annotation.Nullable;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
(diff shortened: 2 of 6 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/listing/MuSigOfferbookController.java
```diff
@@ -321,7 +317,19 @@ void onSelectMarketItem(MuSigMarketItem marketItem) {
...
     void onCreateOffer() {
         MuSigMarketItem marketItem = model.getSelectedMarketItem().get();
-        checkArgument(marketItem != null, "No selected market item");
+
+        if (marketItem == null ) {
+            new Popup().information(Res.get("muSig.offer.listing.createOffer.noSelectedMarketItem"))
+                    .closeButtonText(Res.get("confirmation.ok"))
+                    .show();
+            return;
+        } else if (!marketItem.hasMarketPrice()) {
```
KimStrand, line 329: This guard only protects offer creation from the offerbook. MuSigMyOffersController.onCreateOffer() navigates directly to the create-offer wizard with the stored/default market. If that market has no price, MuSigCreateOfferAmountController.applyQuoteSideMinMaxRange() still reaches orElseThrow(), reproducing the original exception. Could the validation or fallback be enforced at the create-wizard boundary, or added to every entry point?

common/src/main/java/bisq/common/asset/FiatCurrencyRepository.java
```diff
@@ -98,6 +98,12 @@ public static FiatCurrency getCurrencyByCountryCode(String countryCode, Locale l
         try {
             checkArgument(!StringUtils.isEmpty(countryCode), "countryCode must not be null or empty");
 
+//            //  FIXME(remove): FOR TEST ONLY !!!
```
KimStrand: Could we remove this commented test-only override before merging? Keeping a DE→BGN hook in production code is unrelated to the fix and risks being enabled accidentally. The reproduction setup would be better documented in the PR description or implemented in test code.

KimStrand (changes requested):

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/create_offer/direction_and_market/MuSigCreateOfferDirectionAndMarketController.java
```diff
@@ -198,7 +198,12 @@ private void initializeMarketSelection() {
     }
 
     private void updateSelectedMarket(Market market) {
-        if (market != null) {
+        if (market != null ) {
+            if (!marketPriceService.hasMarketPrice(market)) {
+                log.warn("There is no market price available for the selected market: {}, ignored. This can happen if a currency becomes legacy.", market);
+                return;
+            }
+
```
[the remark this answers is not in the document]
neminatha: That's intentional. I wanted to keep the impact on the code as minimal as possible. The added code doesn't change the program's original behavior when a market price exists. However, if no market price exists, selecting this market is prevented here—which would otherwise lead to an exception...

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/listing/MuSigOfferbookController.java
neminatha: Thanks! I actually overlooked that.

I’ve now implemented it in MuSigMyOffersController.onCreateOffer().

I also added a guard in the wizard so it’s no longer possible to select a market without a price. This could previously happen when switching from the crypto list to the fiat list.

common/src/main/java/bisq/common/asset/FiatCurrencyRepository.java
neminatha: Okay. I removed the temporary test override from the production code and updated the PR description.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/create_offer/direction_and_market/MuSigCreateOfferDirectionAndMarketController.java
```diff
@@ -198,7 +198,12 @@ private void initializeMarketSelection() {
     }
 
     private void updateSelectedMarket(Market market) {
-        if (market != null) {
+        if (market != null ) {
+            if (!marketPriceService.hasMarketPrice(market)) {
```
KimStrand: I think there is still a reachable state inconsistency here. Suppose the selected market is a priced crypto market, while muSigLastSelectedFiatMarket is now a legacy market without a price. Selecting Fiat first updates selectedMarketTypeListItem and rebuilds the list with fiat markets, but updateSelectedMarket() then returns, leaving selectedMarket on the crypto market. setSelectedMarketAndListItem() also still persists the rejected fiat market.

Because the direction buttons advance immediately, the user can dismiss the popup and continue with the stale crypto market after selecting the Fiat category. Could a failed update either restore the previous market type or choose a priced fallback, and only persist the market after successful validation?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/create_offer/direction_and_market/MuSigCreateOfferDirectionAndMarketController.java
```diff
@@ -164,56 +166,70 @@ void onMarketTypeListItemSelected(MarketTypeListItem listItem) {
     }
 
     private void updateWithLastSelectedOrDefaultMarket(MarketType marketType) {
-        if (marketType == MarketType.FIAT) {
-            setSelectedMarketAndListItem(settingsService.getMuSigLastSelectedFiatMarket().get());
-        } else {
-            setSelectedMarketAndListItem(settingsService.getMuSigLastSelectedOtherMarket().get());
+        @Nullable Market market = switch (marketType) {
```
KimStrand: Could we remove @Nullable from this local variable to match the surrounding repository style?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/create_offer/direction_and_market/MuSigCreateOfferDirectionAndMarketController.java
```diff
@@ -164,56 +166,70 @@ void onMarketTypeListItemSelected(MarketTypeListItem listItem) {
...
+            log.warn("There is no market price available for the stored '{}' market, setting '{}' market.", market, fallback);
+            market = fallback;
         }
+
+        if (market != null) {
+            setSelectedMarketAndListItem(market);
+        }
+    }
+
+    private @Nullable Market getFallbackMarket(MarketType marketType) {
```
KimStrand: Could this return Optional<Market> and be renamed findFallbackMarket()? The stream already produces an Optional, but orElse(null) converts it back into nullable state. Returning the Optional directly would match the repository's nullability convention and the existing findMarketListItem()/findFirstMarketListItem() pattern.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/create_offer/direction_and_market/MuSigCreateOfferDirectionAndMarketController.java
```diff
@@ -164,56 +166,70 @@ void onMarketTypeListItemSelected(MarketTypeListItem listItem) {
...
+            settingsService.setMuSigLastSelectedFiatMarket(market);
+        } else {
+            settingsService.setMuSigLastSelectedOtherMarket(market);
         }
     }
 
     private void initializeMarketSelection() {
-        Market market = settingsService.getSelectedMuSigMarket().get();
-        if (market != null) {
+        @Nullable Market market = settingsService.getSelectedMuSigMarket().get();
```
KimStrand: The settings observable can legitimately contain null, but I couldn't find @Nullable used on local variables elsewhere in the repository. Is the annotation needed here, or would the explicit null handling below be sufficient?

Also, maketType should be spelled marketType.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/trade_wizard/direction_and_market/TradeWizardDirectionAndMarketController.java
```diff
@@ -146,6 +135,16 @@ public void onActivate() {
                 );
             }
         });
+
+        // Use the market from the selected public channel, or fall back to the service's default channel,
+        // or the first available market list item
+        findMarketListItem(bisqEasyOfferbookSelectionService.getSelectedChannel().get())
+                .or(() -> findMarketListItem(bisqEasyOfferbookChannelService.getDefaultChannel().orElse(null)))
```
KimStrand: getDefaultChannel() already returns an Optional, but orElse(null) unwraps it here only for findMarketListItem() to wrap the result again. Could this remain an Optional using something like getDefaultChannel().flatMap(this::findMarketListItem)? That would keep null handling limited to the genuinely nullable selected-channel observable.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/create_offer/direction_and_market/MuSigCreateOfferDirectionAndMarketController.java
neminatha: good point.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/create_offer/direction_and_market/MuSigCreateOfferDirectionAndMarketController.java
neminatha: You're right. I'm now using “Optional<Market>” for nullable local variables.

PR #4878, opened 2026-07-20
Add NETWORK_INFO websocket topic for network state

salottodev:
## Summary by CodeRabbit

* **New Features**
  * Added a new WebSocket subscription topic to stream real-time network information snapshots.
  * Network updates now include connection metadata (id, address, outbound/seed, timestamp), Tor running state, local address/key id when available, and whether initial data synchronization has completed.
* **Tests**
  * Updated subscription service tests to wire the new network dependency for the added topic.

### DIFF
--- a/api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
+++ b/api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
@@ -23,10 +23,12 @@ import bisq.api.web_socket.domain.BaseWebSocketService;
 import bisq.api.web_socket.subscription.ModificationType;
 import bisq.api.web_socket.subscription.Subscriber;
 import bisq.api.web_socket.subscription.SubscriberRepository;
+import bisq.api.web_socket.subscription.SubscriptionRequest;
 import bisq.api.web_socket.subscription.Topic;
 import bisq.common.network.Address;
 import bisq.common.network.TransportType;
 import bisq.common.observable.Pin;
+import bisq.common.util.StringUtils;
 import bisq.network.NetworkService;
 import bisq.network.identity.NetworkId;
 import bisq.network.p2p.ServiceNode;
@@ -46,6 +48,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.stream.Collectors;
 
 /**
  * Pushes the local node's network state (peer count, data-sync flag, tor status, own address, key id and
@@ -94,14 +97,21 @@ public class NetworkInfoWebSocketService extends BaseWebSocketService {
     }
 
     @Override
+    public void validate(SubscriptionRequest request) {
+        if (StringUtils.toOptional(request.getParameter()).isPresent()) {
+            throw new IllegalArgumentException("NETWORK_INFO does not support a subscription parameter");
+        }
+    }
+
+    @Override
     public CompletableFuture<Boolean> initialize() {
         findPrimaryServiceNode().ifPresent(serviceNode -> {
             observedServiceNode = serviceNode;
             serviceNode.getDefaultNode().addListener(nodeListener);
 
             serviceNode.getInventoryService()
-                    .map(InventoryService::getNumPendingInventoryRequests)
-                    .ifPresent(observable -> pins.add(observable.addObserver(numPending -> onChange())));
+                    .map(InventoryService::getInitialInventoryRequestsCompleted)
+                    .ifPresent(observable -> pins.add(observable.addObserver(completed -> onChange())));
         });
 
         return CompletableFuture.completedFuture(true);
@@ -125,7 +135,9 @@ public class NetworkInfoWebSocketService extends BaseWebSocketService {
     }
 
     private void onChange() {
-        Set<Subscriber> subscribers = subscriberRepository.findSubscribers(topic, Optional.empty());
+        Set<Subscriber> subscribers = subscriberRepository.findSubscribers(topic).values().stream()
+                .flatMap(Set::stream)
+                .collect(Collectors.toSet());
         if (subscribers.isEmpty()) {
             return;
         }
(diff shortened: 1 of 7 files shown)


api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
```diff
@@ -0,0 +1,194 @@
...
+                .orElse(null);
+
+        // We traverse the active connections only once, so that numConnections and the connection list
+        // are consistent with each other. Connections can come and go while we build the payload.
+        List<ConnectionDto> connections = defaultNode
+                .map(node -> toConnectionDtos(node, peerGroupService))
+                .orElseGet(List::of);
+        int numConnections = connections.size();
+
+        boolean allDataReceived = primaryServiceNode
```
KimStrand, line 164: allDataReceived is true before any sync has happened

Deriving this from numPendingInventoryRequests == 0 gives the wrong state in two common cases:
   • At startup the pending counter is 0 before any inventory request was ever made, so a fresh node with zero peers reports allDataReceived=true.
   • After losing all connections the counter also sits at 0, so clients keep seeing "synced" even though sync was lost.

The desktop app uses the dedicated InventoryService.getInitialInventoryRequestsCompleted() observable for exactly this semantic (see BannerNotificationController), and it resets correctly when initial requests are incomplete.
```suggestion
        boolean allDataReceived = primaryServiceNode
                     .flatMap(ServiceNode::getInventoryService)
                     .map(inventoryService -> inventoryService.getInitialInventoryRequestsCompleted().get())
                     .orElse(false);
   ```

File: api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java [hunk 2]
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    @Override
+    public CompletableFuture<Boolean> initialize() {
+        findPrimaryServiceNode().ifPresent(serviceNode -> {
+            observedServiceNode = serviceNode;
+            serviceNode.getDefaultNode().addListener(nodeListener);
+
+            serviceNode.getInventoryService()
+                    .map(InventoryService::getNumPendingInventoryRequests)
```
KimStrand: numPendingInventoryRequests is 0 before any request was ever made and flips after the pending counter updates within the same completion, so observing it pushes stale sync state. Suggestion would be to use getInitialInventoryRequestsCompleted() instead.

   ```java
     serviceNode.getInventoryService()
             .map(InventoryService::getInitialInventoryRequestsCompleted)
             .ifPresent(observable -> pins.add(observable.addObserver(completed -> onChange())));
   ```

api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
```diff
@@ -0,0 +1,194 @@
...
+        return CompletableFuture.completedFuture(true);
+    }
+
+    @Override
+    public Optional<String> getJsonPayload() {
+        return toJson(buildNetworkInfoDto());
+    }
+
+    private void onChange() {
+        Set<Subscriber> subscribers = subscriberRepository.findSubscribers(topic, Optional.empty());
```
KimStrand: Subscribers who subscribed with a parameter are stored in a parameter-specific bucket, so findSubscribers(topic, Optional.empty()) never reaches them. They get the initial snapshot but no live updates

api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
salottodev: @KimStrand Thanks for the detailed review! I was following the pattern used in the Node Android app. I'll update the implementation here and make the fix in the Node app as well.

api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
salottodev: Good catch, fixed it two ways:

  1. NETWORK_INFO doesn't take a parameter, so it shouldn't accept one. Added a validate() that rejects any request with a parameter and returns a clear error to the client, instead of silently creating a
  broken subscription. So nobody ends up in a parameter-specific bucket in the first place. This follows the same validate() pattern already used in other WebSocket services like AlertNotificationsWebSocketService and TradeRestrictingAlertWebSocketService.
  2. As a safety net, onChange() now sends to all buckets for the topic (findSubscribers(topic)) instead of just the empty one. Even if a subscriber somehow lands in another bucket, they'll still get live
  updates.

  So now a stray parameter gets rejected up front, and the update push reaches everyone regardless.

salottodev:
@KimStrand I will submit the update soon.

api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
```diff
@@ -0,0 +1,205 @@
...
+
+    private NetworkInfoDto buildNetworkInfoDto() {
+        Optional<ServiceNode> primaryServiceNode = Optional.ofNullable(observedServiceNode);
+        Optional<Node> defaultNode = primaryServiceNode.map(ServiceNode::getDefaultNode);
+        PeerGroupService peerGroupService = primaryServiceNode
+                .flatMap(ServiceNode::getPeerGroupManager)
+                .map(PeerGroupManager::getPeerGroupService)
+                .orElse(null);
+
+        // We traverse the active connections only once, so that numConnections and the connection list
```
KimStrand: The comment still refers to the removed numConnections field. Suggested replacement:
// Snapshot the currently active connections while building the payload.

KimStrand:
The payload is logged twice through the generic WebSocket send path and includes network data. Since this logging predates this PR and affects all WebSocket topics, could it make sense to open a follow-up issue to review whether we should log only non-sensitive metadata at INFO, and only once, instead of logging the full JSON payload at both layers?

KimStrand (commented):
utACK

salottodev:
> The payload is logged twice through the generic WebSocket send path and includes network data. Since this logging predates this PR and affects all WebSocket topics, could it make sense to open a follow-up issue to review whether we should log only non-sensitive metadata at INFO, and only once, instead of logging the full JSON payload at both layers?

@coderabbitai Good point, and thanks for catching that this is broader than this PR. The full JSON payload is logged twice at INFO through the shared send path: once in `BaseWebSocketService.send(...)` and again in `Subscriber.send(...)`. Since this lives in the generic WebSocket layer, it affects all topics, not just `NETWORK_INFO`.

Could you open a follow-up issue to track reviewing this? The gist:

- **Log once, not twice** — the two INFO lines are redundant.
- **Log only non-sensitive metadata at INFO** — e.g. topic, modificationType, subscriber id, sequence number (and maybe payload size), while moving the full JSON body down to DEBUG/TRACE.

Motivation: some payloads (like `NETWORK_INFO`, which includes the node's own address, key id, and peer connection details) are more sensitive than others, and full payloads currently land in INFO logs on every update. Keeping this out of the current PR since it's a cross-cutting logging change unrelated to the feature.

PR #4879, opened 2026-07-21
Expose bootstrap trade limits and API capabilities

rodvar:
- contribs to https://github.com/bisq-network/bisq-mobile/issues/1260
 - network feature currently needs this in as well (new WS TOPIC as per #4878
 - new config and capabilities APIs for bisq-mobile to fetch static data and new-features running node support
 - this allows dynamic check against running node coping with lower release pace of Bisq2 Desktop for both, REST APIs and WS TOPICs dependency
 - added docs and AI generated tests

## Summary by CodeRabbit

- **New Features**
  - Added public configuration endpoints for trade amount limits and API capabilities.
  - Exposed supported API features and application version metadata.
  - Added unauthenticated access to configuration information.

- **Tests**
  - Added coverage validating configuration responses, feature mappings, endpoint reachability, and trade limit accuracy.

### DIFF
--- a/api/src/main/java/bisq/api/rest_api/endpoints/config/ConfigRestApi.java
+++ b/api/src/main/java/bisq/api/rest_api/endpoints/config/ConfigRestApi.java
@@ -18,9 +18,11 @@
 package bisq.api.rest_api.endpoints.config;
 
 import bisq.api.dto.DtoMappings;
+import bisq.api.dto.config.ApiCapabilitiesDto;
 import bisq.api.dto.config.TradeAmountLimitsDto;
 import bisq.api.rest_api.endpoints.RestApiBase;
 import bisq.bisq_easy.BisqEasyTradeAmountLimits;
+import bisq.common.application.ApplicationVersion;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.media.Content;
 import io.swagger.v3.oas.annotations.media.Schema;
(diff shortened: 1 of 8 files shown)


api/src/main/java/bisq/api/rest_api/endpoints/config/ConfigRestApi.java
```diff
@@ -0,0 +1,117 @@
...
+ *     <li>static: they change only between bisq2 core/api versions, never per user or at runtime;</li>
+ *     <li>global: identical for every peer on the network;</li>
+ *     <li>small: cheap enough for a client to fetch once and cache.</li>
+ * </ul>
+ * Anything user-specific, runtime-mutable, large, or that belongs to a single feature's flow does
+ * NOT go here — give it its own feature endpoint. Keeping this endpoint honest avoids turning it
+ * into a kitchen-sink blob that couples unrelated features' release cadence and cache invalidation.
+ */
+@Slf4j
+@Path("/config")
```
KimStrand, line 61: /config endpoints always return 403 when authorization is required.

RestPermissionMapping has no matching rule, the authorization filter (RestApiAuthorizationFilter) therefore rejects both new endpoints before execution

rodvar: its working as-is. Your comment might lead to discovering a missing security feature in the handover from WS calls or simply is not applying for debug..? ... I'm checking and will push a commit soon either way to make it unauth or fix any issue

rodvar: I think we can just make config @AllowUnauthenticated which skips authentication, so /config is readable without a session. 

The only thing exposed is the trade constants (public knowledge) and the node's API version + feature list (mild version fingerprinting) — and the API is already behind the onion/LAN binding, so I judged that acceptable. If you'd rather it require a session but no feature permission, the alternative I thinks is a new Permission.CONFIG force-added to every grant (enum + protobuf + grant-logic churn). But I think the former is better. I'll send a fix soon let me know your thoughts.

KimStrand (commented):
utACK

PR #4885, opened 2026-07-23
Fix: Bisq Easy trades can get stuck when a protocol message arrives out of order

rodvar:
- resolves bisq-network/bisq-mobile#1622

## Motivation

Reported in [bisq-network/bisq-mobile#1622](https://github.com/bisq-network/bisq-mobile/issues/1622): sellers occasionally find that **"Mark Payment Sent" / "Payment Received" stop advancing the trade**. The payment goes through and chat still works, but the trade state is stuck, forcing both traders to finish the trade manually via chat.

As I noted in the issue comments, this turned out to be a **structural bisq2-core issue, not a Bisq Connect one** — it can affect any bisq2 seller (the original reporter was on Bisq 2 desktop 2.1.11). It is intermittent and rare, but disruptive when it hits, and it is a **very important fix overall**: core-embedded apps (Bisq 2 desktop, Bisq Node mobile) can work around it with a restart, but for Bisq Connect the workaround means restarting the **trusted node**.

## Problem

A Bisq Easy trade is driven by a finite-state machine (FSM). Phase 1 has two independent steps that arrive as separate, unordered messages: the seller **sends account data**, and the seller **receives the buyer's Bitcoin address**. The next step — the buyer's **"fiat sent" confirmation** — is only accepted from the single state where *both* phase-1 steps are already done.

When the "fiat sent" message arrives **before** phase 1 has fully settled (a realistic race over Tor/mailbox, where message ordering isn't guaranteed and a fast Connect buyer fires messages back-to-back) the FSM finds no matching transition and parks the message in an **in-memory queue**, retrying it only after some *later* successful transition. If no such transition happens while the app is running, the message is stranded and the trade stays stuck — even though the network layer already ACKed it as "delivered" and stopped resending it.

**Why restarting sometimes worked around it:** on startup, bisq2 re-feeds previously received messages through the FSM, which re-applies the stranded message once the state is correct. But this recovery was *incidental* — the FSM queue itself was in-memory only, so whether a restart healed the trade depended on the message still being around to re-feed. That's why the issue sometimes cleared on its own overnight and sometimes never did.

## Solution

The fix makes the recovery **structural instead of incidental**, at the FSM / trade-state layer. Three parts:

**1. Persist the FSM's out-of-order buffer, drain it on load.** The pending-event queue (and processed-events set) is now serialized with the trade — the framework even shipped an unused restore constructor built for exactly this, which we activate. Out-of-order messages are no longer silently lost at shutdown, and on load each restored trade drains its queue once against the restored state (mirrored for BisqEasy and MuSig). The `{state, eventQueue}` pair is captured as **one atomic snapshot** under the model lock, so the async persistence write can't tear it (which could double-apply or silently drop an event on restore). Reaching a final state clears the queue through an **unthrottled persist path** — the ordinary rate-limited persist() could silently drop that wipe, leaving sensitive queued messages on disk.

**2. Close the startup handoff window.** The live message listener is now registered **before** the startup replay. With the old order there was a window where a message could be ACKed and removed from the mailbox without ever reaching the FSM — lost for good. Register-first closes that structurally, at the cost of possible duplicate delivery (live + replay), which is expected and handled: once a protocol exists a duplicate is absorbed by the FSM (processed-events filter, forward-only transitions); for the one case where no protocol exists yet (a duplicate take-offer request) the trade id is now **single-sourced** (`BisqEasyTrade#createId`) so duplicates route to the existing protocol, and creation itself is **check-then-act atomic** under a creation lock — two concurrent deliveries can no longer split-brain into two trade/protocol instances for the same id.

**3. Make the persist actually durable.** The above only helps if the queue reliably reaches disk, so the persistence layer got hardened along the way: write failures are propagated instead of swallowed, dirty-state is derived from a requested/persisted **generation pair** (a failed or dropped persist can't be cleared by an older write completing), snapshot capture + write scheduling run under one lock so writes reach the serializer in order, a **write-id guard** ensures a stale queued write can never overwrite a newer one, and shutdown waits (bounded) for the in-flight write and falls back to a direct synchronous write if it doesn't confirm.

**Why the drain is safe** (it re-applies messages, so this matters): drained events go **directly through the FSM** — never the public message entry point — so no handler re-runs, no peer traffic, no duplicate ACKs. The drain also re-checks what `onMessage()` would check for a live message: an emergency **halt / min-version alert defers** the drain (events stay queued for a later restart — and the alert observer is registered before the restore so this check sees real state on cold start), a **banned sender's** queued events are dropped, and each trade drains in isolation so one failing trade can't block restoring the rest.

### Notes

 - earlier revisions of this PR also shipped a live recovery pass (reconnect-debounced + periodic); it was dropped during review in favour of the leaner structural fix above → recovery is now **deterministic on the next transition or restart** instead of incidental, and the window that stranded messages in the first place is closed
 - this solution doesn't require any UI changes neither mobile nor desktop → solution entirely at the FSM / trade-state / persistence layer
 - logs to look for: `Re-feeding a processed message at startup was rejected …`, `Deferring the queued-event drain for restored trade …`, `Removed queued event(s) from a banned sender …`

### Dev commit logs

 - persist FSM event queue + processed events so out-of-order messages survive restart; drain once on load (BisqEasy + MuSig)
 - atomic `{state, eventQueue}` snapshot under the model lock; final-state queue wipe routed through an unthrottled persist
 - register the live listener before the startup replay (closes a message-loss handoff window); duplicates absorbed by the FSM
 - take-offer duplicate safety: single-sourced trade id + creation lock (no split-brain trade/protocol on concurrent delivery)
 - drain guards mirror onMessage(): halt/min-version defers, banned-sender events dropped, per-trade isolation
 - persistence hardening: write-failure propagation, generation-pair dirty tracking, ordered snapshot/write scheduling, write-id guard against stale overwrites, bounded shutdown write + direct fallback
 - added AI generated tests (FSM, trade restore/drain, persistence write guards)

### DIFF
--- a/common/src/main/java/bisq/common/fsm/Fsm.java
+++ b/common/src/main/java/bisq/common/fsm/Fsm.java
@@ -61,7 +61,18 @@ public abstract class Fsm<M extends FsmModel> {
     abstract protected void configTransitions();
 
     public <E extends Event> void handle(E event) {
-        synchronized (this) {
+        // Synchronize on the model rather than on this Fsm instance: the model (see FsmModel#getStateAndEventQueueSnapshot)
+        // must be lockable directly by callers that only hold a reference to the model - e.g. Trade#getTradeBuilder
+        // during persistence - without needing a reference to this Fsm/TradeProtocol. Locking on the model gives
+        // both sides the same monitor, so a persistence snapshot of {state, eventQueue} can never be taken mid-transition.
+        synchronized (model) {
+            // Tracks whether this call reached a final state, so the finally block below can route the mandatory
+            // post-transition persist() call through persistOnFinalState() instead of the normal persist(). This
+            // matters for privacy/retention (#1622 follow-up): a final state clears eventQueue/processedEvents
+            // in-memory a few lines below, and that wipe must reach disk promptly - persistOnFinalState() bypasses
+            // any write-rate limiting a subclass may apply to the ordinary persist(), so the wipe is not silently
+            // dropped and left stranded on disk for an indefinite time (see RateLimitedPersistenceClient#persistNow).
+            boolean reachedFinalState = false;
             try {
                 checkNotNull(event, "event must not be null");
                 State currentState = model.getState();
@@ -95,14 +106,11 @@ public abstract class Fsm<M extends FsmModel> {
                     if (targetState.isFinalState()) {
                         model.processedEvents.clear();
                         model.eventQueue.clear();
+                        reachedFinalState = true;
                     } else {
                         model.processedEvents.add(eventClass);
                         // Apply all pending events to see if any of those match our current state.
-                        // If an exception is thrown by the processed pending event it will get thrown to the
-                        // caller. This would be a different triggering event as the event which cause
-                        // the exception (the one from the queue).
-                        // Clone set to avoid ConcurrentModificationException
-                        new HashSet<>(model.getEventQueue()).forEach(this::handle);
+                        drainEventQueue();
                     }
                 } else {
                     log.info("We did not find a transition with state {} and event {}. " +
@@ -128,12 +136,53 @@ public abstract class Fsm<M extends FsmModel> {
                 // We throw the exception to allow specific error handling to the implementation class.
                 throw fsmException;
             } finally {
-                persist();
+                if (reachedFinalState) {
+                    persistOnFinalState();
+                } else {
+                    persist();
+                }
             }
         }
     }
 
     protected abstract void persist();
+
+    /**
+     * Called instead of {@link #persist()} exactly once, when a transition reaches a final state. The default
+     * implementation just delegates to {@link #persist()}; override to bypass any write-rate limiting the
+     * concrete {@link #persist()} implementation applies (see {@code bisq.persistence.RateLimitedPersistenceClient}).
+     * <br/>
+     * Reaching a final state clears {@code eventQueue}/{@code processedEvents} in-memory a few lines above in
+     * {@link #handle(Event)}; those may hold sensitive, trade-specific network messages. Without a guaranteed,
+     * unthrottled flush here, a rate-limited persist() call could be silently dropped, leaving the (already
+     * in-memory-wiped) sensitive data sitting on disk for an unbounded time - potentially until the process's
+     * next unrelated persist() call, or its next clean shutdown, neither of which is guaranteed to happen promptly
+     * (or at all, on a crash/kill -9).
+     */
+    protected void persistOnFinalState() {
+        persist();
+    }
+
+    /**
+     * Re-attempts to handle every event currently sitting in the event queue against the current state.
+     * <br/>
+     * This is called automatically after any successful transition (see {@link #handle(Event)}), but it is
+     * also safe - and sometimes necessary - to call explicitly, e.g. once right after a model has been restored
+     * from persisted data: the event queue itself is not guaranteed to be persisted for every {@link FsmModel}
+     * subclass, so events which arrived out of order before a restart may otherwise never be re-applied because
+     * no further live transition ever occurs for that trade.
+     * <br/>
+     * Safe to call when the queue is empty (no-op) and safe to call multiple times. Exceptions from handling a
+     * queued event behave exactly as for a live event: each event goes through the concrete {@link #handle(Event)}
+     * implementation, so whether an {@link FsmException} propagates to the caller or is absorbed is up to the
+     * subclass (e.g. the trade protocols swallow it and surface an error state instead).
+     */
+    public void drainEventQueue() {
+        synchronized (model) {
+            // Clone set to avoid ConcurrentModificationException as handle() mutates model.eventQueue.
+            new HashSet<>(model.getEventQueue()).forEach(this::handle);
+        }
+    }
 
     public TransitionBuilder<M> addTransition() {
         return new TransitionBuilder<>(this);
--- a/common/src/main/java/bisq/common/fsm/Fsm.java
+++ b/common/src/main/java/bisq/common/fsm/Fsm.java
@@ -61,11 +61,15 @@ public abstract class Fsm<M extends FsmModel> {
     abstract protected void configTransitions();
 
     public <E extends Event> void handle(E event) {
-        // Synchronize on the model rather than on this Fsm instance: the model (see FsmModel#getStateAndEventQueueSnapshot)
-        // must be lockable directly by callers that only hold a reference to the model - e.g. Trade#getTradeBuilder
-        // during persistence - without needing a reference to this Fsm/TradeProtocol. Locking on the model gives
-        // both sides the same monitor, so a persistence snapshot of {state, eventQueue} can never be taken mid-transition.
-        synchronized (model) {
+        // Lock on the model's explicit lock rather than on this Fsm instance: the model (see
+        // FsmModel#getStateAndEventQueueSnapshot) must be lockable directly by callers that only hold a reference
+        // to the model - e.g. Trade#getTradeBuilder during persistence - without needing a reference to this
+        // Fsm/TradeProtocol. Sharing the same lock object gives both sides the same mutual exclusion, so a
+        // persistence snapshot of {state, eventQueue} can never be taken mid-transition. Unlike
+        // getStateAndEventQueueSnapshot()'s bounded tryLock, a live transition takes the lock unbounded - its
+        // correctness cannot be short-circuited by a timeout.
+        model.lock.lock();
+        try {
             // Tracks whether this call reached a final state, so the finally block below can route the mandatory
             // post-transition persist() call through persistOnFinalState() instead of the normal persist(). This
             // matters for privacy/retention (#1622 follow-up): a final state clears eventQueue/processedEvents
@@ -78,13 +82,21 @@ public abstract class Fsm<M extends FsmModel> {
                 State currentState = model.getState();
                 checkNotNull(currentState, "currentState must not be null");
                 if (currentState.isFinalState()) {
-                    log.warn("We have reached the final state and do not allow further state transition. New event was: {}", event);
+                    // Render the class only, never the event instance: the Fsm layer is generic and handles
+                    // domain events (e.g. BisqEasyAccountDataMessage / MuSig SendAccountPayloadMessage) whose
+                    // generated toString() can contain private payment-account data. This log line - like the
+                    // checkArgument message and the error log below - can end up persisted as trade error info
+                    // and included in the error report sent to the peer, so it must never carry event content.
+                    log.warn("We have reached the final state and do not allow further state transition. New event class was: {}", event.getClass().getSimpleName());
                     return;
                 }
                 log.info("Start transition from currentState {}", currentState);
                 Class<? extends Event> eventClass = event.getClass();
                 var transitionMapEntriesForEvent = findTransitionMapEntriesForEvent(eventClass);
-                checkArgument(!transitionMapEntriesForEvent.isEmpty(), "No transition found for given event " + event);
+                // Class name only - see the final-state log above for why the event instance itself must never
+                // be rendered here. This message reaches FsmException's cause chain and from there the trade's
+                // persisted error info / peer error report.
+                checkArgument(!transitionMapEntriesForEvent.isEmpty(), "No transition found for given event class " + eventClass.getSimpleName());
                 Optional<Transition> transition = findTransition(currentState, transitionMapEntriesForEvent);
                 if (transition.isPresent()) {
                     State targetState = transition.get().getTargetState();
@@ -124,7 +136,8 @@ public abstract class Fsm<M extends FsmModel> {
                             .forEach(e -> model.eventQueue.add(event));
                 }
             } catch (Exception exception) {
-                log.error("Error at handling {}.", event, exception);
+                // Class name only - see the comment at the final-state log above.
+                log.error("Error at handling event of class {}.", event.getClass().getSimpleName(), exception);
                 FsmException fsmException = new FsmException(exception, event);
                 // In case of an exception we fire the FsmErrorEvent to trigger an error state.
                 // We apply that only if the event which triggered the exception was not the FsmErrorEvent itself
@@ -142,6 +155,8 @@ public abstract class Fsm<M extends FsmModel> {
                     persist();
                 }
             }
+        } finally {
+            model.lock.unlock();
         }
     }
 
@@ -178,9 +193,14 @@ public abstract class Fsm<M extends FsmModel> {
      * subclass (e.g. the trade protocols swallow it and surface an error state instead).
      */
     public void drainEventQueue() {
-        synchronized (model) {
+        // Unbounded, like handle() - ReentrantLock is reentrant, so the nested lock() calls inside each
+        // handle(event) below (same thread) are free.
+        model.lock.lock();
+        try {
             // Clone set to avoid ConcurrentModificationException as handle() mutates model.eventQueue.
             new HashSet<>(model.getEventQueue()).forEach(this::handle);
+        } finally {
+            model.lock.unlock();
         }
     }
 
--- a/persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
+++ b/persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
@@ -39,7 +39,16 @@ public abstract class RateLimitedPersistenceClient<T extends PersistableStore<T>
     // write can't hang JVM shutdown indefinitely.
     private static final long SHUTDOWN_WRITE_AWAIT_MS = 2000;
 
+    // lastWrite/lastWriteGeneration are only ever read or written under scheduleLock (including the failure-path
+    // rollback below) - the volatile on lastWrite predates that and is now redundant, but harmless; left as-is to
+    // keep this fix minimal.
     private volatile long lastWrite;
+    // Tags lastWrite with the generation that last set it, so the failure-path rollback (see the completion
+    // handler in persist()/persistNow()) can tell whether it's still safe to restore the PREVIOUS lastWrite value,
+    // or whether a newer call has since moved lastWrite past it (in which case rolling back would clobber
+    // legitimate, newer throttle state). A plain timestamp equality check would work for the common case but is
+    // vulnerable to two calls landing in the same millisecond; the generation counter is unique by construction.
+    private long lastWriteGeneration;
     // Throttle heuristic only: with the generation pair below, CORRECTNESS no longer depends on this flag,
     // so its benign races (two threads may both observe it false) cost at most an extra write.
     private volatile boolean writeInProgress;
@@ -88,27 +97,65 @@ public abstract class RateLimitedPersistenceClient<T extends PersistableStore<T>
             if (tooFast || writeInProgress) {
                 return CompletableFuture.completedFuture(false);
             }
+            long previousLastWrite = lastWrite;
             lastWrite = System.currentTimeMillis();
+            lastWriteGeneration = myGeneration;
             writeInProgress = true;
-            CompletableFuture<Boolean> future = getPersistence()
-                    .persistAsync(getPersistableStore().getClone())
-                    .handle((nil, throwable) -> {
-                        writeInProgress = false;
-                        boolean success = throwable == null;
-                        if (success) {
-                            // The snapshot was taken after the bump to myGeneration, so it covers everything up
-                            // to it. max() keeps persistedGeneration monotone; it can never claim a generation a
-                            // newer request has already moved past.
-                            persistedGeneration.accumulateAndGet(myGeneration, Math::max);
-                        }
-                        return success;
-                    });
-            return future;
+            return scheduleWrite(myGeneration, previousLastWrite);
         }
     }
 
     protected long getMaxWriteRateInMs() {
         return 1000;
+    }
+
+    /**
+     * Submits the actual write and wires up its completion. Called from within scheduleLock by persist()/
+     * persistNow(), but the returned future's completion handler itself runs later, off the persistence
+     * executor thread, OUTSIDE scheduleLock (persistAsync() only submits and returns immediately - the write
+     * itself, and this handler, run asynchronously).
+     */
+    private CompletableFuture<Boolean> scheduleWrite(long myGeneration, long previousLastWrite) {
+        return getPersistence()
+                .persistAsync(getPersistableStore().getClone())
+                .handle((nil, throwable) -> {
+                    writeInProgress = false;
+                    boolean success = throwable == null;
+                    if (success) {
+                        // The snapshot was taken after the bump to myGeneration, so it covers everything up
+                        // to it. max() keeps persistedGeneration monotone; it can never claim a generation a
+                        // newer request has already moved past.
+                        persistedGeneration.accumulateAndGet(myGeneration, Math::max);
+                    } else {
+                        // A failed write must not consume the rate-limit budget: without this, lastWrite already
+                        // reflects "we just wrote" even though nothing reached disk, so every ORDINARY persist()
+                        // call within the next getMaxWriteRateInMs() window is silently dropped (tooFast) instead
+                        // of retrying. The generation pair still correctly reports the store dirty (isDropped()),
+                        // but nothing actively retries it until either an unrelated call happens to land outside
+                        // that window, or an explicit persistNow()/the shutdown fallback runs - for a store with
+                        // no further natural persist() traffic that can be an unbounded wait. Rolling lastWrite
+                        // back to what it was before this attempt makes the very next ordinary persist() call
+                        // retry through the normal path, exactly as if this failed attempt had never happened.
+                        rollBackLastWriteIfStillOurs(myGeneration, previousLastWrite);
+                    }
+                    return success;
+                });
+    }
+
+    /**
+     * Undoes this call's lastWrite update, but ONLY if no newer call has since moved lastWrite past it - this
+     * completion handler runs outside scheduleLock (see scheduleWrite), concurrently with any new persist()/
+     * persistNow() call that already took the lock and legitimately advanced lastWrite/lastWriteGeneration past
+     * ours; unconditionally overwriting lastWrite here would clobber that newer call's throttle state and let a
+     * write happen sooner than intended for state this failed write never even covered. Guarded by the unique
+     * generation counter, not a timestamp equality check, since two calls can land in the same millisecond.
+     */
+    private void rollBackLastWriteIfStillOurs(long myGeneration, long previousLastWrite) {
+        synchronized (scheduleLock) {
+            if (lastWriteGeneration == myGeneration) {
+                lastWrite = previousLastWrite;
+            }
+        }
     }
 
     /**
@@ -124,19 +171,14 @@ public abstract class RateLimitedPersistenceClient<T extends PersistableStore<T>
     public CompletableFuture<Boolean> persistNow() {
         synchronized (scheduleLock) {
             long myGeneration = requestedGeneration.incrementAndGet();
+            long previousLastWrite = lastWrite;
             lastWrite = System.currentTimeMillis();
+            lastWriteGeneration = myGeneration;
             writeInProgress = true;
-            CompletableFuture<Boolean> future = getPersistence()
-                    .persistAsync(getPersistableStore().getClone())
-                    .handle((nil, throwable) -> {
-                        writeInProgress = false;
-                        boolean success = throwable == null;
-                        if (success) {
-                            persistedGeneration.accumulateAndGet(myGeneration, Math::max);
-                        }
-                        return success;
-                    });
-            return future;
+            // Same rollback-on-failure contract as persist() (see scheduleWrite): persistNow() bypasses the
+            // throttle for ITSELF, but still updates lastWrite, so a failed persistNow() must not silently
+            // throttle away the NEXT ordinary persist() call either.
+            return scheduleWrite(myGeneration, previousLastWrite);
         }
     }
 
--- a/persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
+++ b/persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
@@ -18,10 +18,12 @@
 package bisq.persistence;
 
 import bisq.common.util.StringUtils;
-import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * This implementation of PersistenceClient drops persist requests if they happen too frequently.
@@ -33,10 +35,31 @@ import java.util.concurrent.CompletableFuture;
  */
 @Slf4j
 public abstract class RateLimitedPersistenceClient<T extends PersistableStore<T>> implements PersistenceClient<T> {
+    // Max time the shutdown hook waits for its queued fallback write to confirm. Bounded so a stuck
+    // write can't hang JVM shutdown indefinitely.
+    private static final long SHUTDOWN_WRITE_AWAIT_MS = 2000;
+
     private volatile long lastWrite;
-    @Getter
-    private volatile boolean dropped;
+    // Throttle heuristic only: with the generation pair below, CORRECTNESS no longer depends on this flag,
+    // so its benign races (two threads may both observe it false) cost at most an extra write.
     private volatile boolean writeInProgress;
+    // Dirty state is DERIVED from this generation pair instead of a mutable boolean flag:
+    //  - requestedGeneration is bumped on EVERY persist()/persistNow() entry, before the store snapshot is
+    //    taken, so a write scheduled at generation N covers all state up to N.
+    //  - persistedGeneration advances (monotonically, via max) only when a write confirms, to the generation
+    //    its snapshot covered.
+    //  - dirty <=> requestedGeneration > persistedGeneration.
+    // This removes the check-then-act races a boolean flag had (an older write's completion could wipe the
+    // dirty mark of a newer, never-written request), and it makes failed writes keep the store dirty for
+    // free: a failure simply never advances persistedGeneration, so the shutdown fallback retries it.
+    private final AtomicLong requestedGeneration = new AtomicLong();
+    private final AtomicLong persistedGeneration = new AtomicLong();
+    // Serializes generation bump -> snapshot capture -> handoff to the persistence executor, so snapshots
+    // reach the (write-serializing) executor in generation order. Without it, an older call could capture
+    // its snapshot, stall, and schedule AFTER a newer call's write - leaving stale state on disk while the
+    // generation pair reports clean. Only an in-memory clone and an executor submit happen under this lock,
+    // never disk I/O.
+    private final Object scheduleLock = new Object();
 
     public RateLimitedPersistenceClient() {
         //todo (Critical) check if we want to use ShutdownHook here
@@ -46,22 +69,41 @@ public abstract class RateLimitedPersistenceClient<T extends PersistableStore<T>
         }));
     }
 
+    /**
+     * True while the in-memory store holds state which no confirmed write has covered yet — because a request
+     * was throttled away, a write is still in flight, or a write failed. Consulted by the shutdown hook to
+     * decide whether the fallback synchronous write is needed.
+     */
+    public boolean isDropped() {
+        return requestedGeneration.get() > persistedGeneration.get();
+    }
+
     @Override
     public CompletableFuture<Boolean> persist() {
-        boolean tooFast = System.currentTimeMillis() - lastWrite < getMaxWriteRateInMs();
-        dropped = tooFast || writeInProgress;
-        if (dropped) {
-            return CompletableFuture.completedFuture(false);
-        } else {
+        synchronized (scheduleLock) {
+            // Bump BEFORE the throttle decision and before any snapshot: if this request gets dropped, the store
+            // is dirty by construction (requested > persisted) until some later write covers this generation.
+            long myGeneration = requestedGeneration.incrementAndGet();
+            boolean tooFast = System.currentTimeMillis() - lastWrite < getMaxWriteRateInMs();
+            if (tooFast || writeInProgress) {
+                return CompletableFuture.completedFuture(false);
+            }
             lastWrite = System.currentTimeMillis();
             writeInProgress = true;
-            dropped = false;
-            return getPersistence()
+            CompletableFuture<Boolean> future = getPersistence()
                     .persistAsync(getPersistableStore().getClone())
                     .handle((nil, throwable) -> {
                         writeInProgress = false;
-                        return throwable == null;
+                        boolean success = throwable == null;
+                        if (success) {
+                            // The snapshot was taken after the bump to myGeneration, so it covers everything up
+                            // to it. max() keeps persistedGeneration monotone; it can never claim a generation a
+                            // newer request has already moved past.
+                            persistedGeneration.accumulateAndGet(myGeneration, Math::max);
+                        }
+                        return success;
                     });
+            return future;
         }
     }
 
@@ -69,10 +111,87 @@ public abstract class RateLimitedPersistenceClient<T extends PersistableStore<T>
         return 1000;
     }
 
-    private void persistOnShutdown() {
-        if (dropped) {
-            dropped = false;
-            getPersistence().persist(getPersistableStore().getClone());
+    /**
+     * Persists immediately, bypassing the write-rate limiter that {@link #persist()} applies.
+     * <br/>
+     * Reserved for privacy/retention-critical writes where a throttled-and-dropped {@link #persist()} call could
+     * leave sensitive data on disk for an unbounded time - e.g. a completed trade's pending FSM events, which are
+     * cleared in-memory the instant the trade reaches a final state (see {@code bisq.common.fsm.Fsm#handle}), but
+     * whose on-disk copy has no other guaranteed opportunity to catch up: the next unrelated persist() call may
+     * never come (a completed trade has no further transitions), and the shutdown-hook flush this class registers
+     * is best-effort only - it never runs on a JVM crash or {@code kill -9}.
+     */
+    public CompletableFuture<Boolean> persistNow() {
+        synchronized (scheduleLock) {
+            long myGeneration = requestedGeneration.incrementAndGet();
+            lastWrite = System.currentTimeMillis();
+            writeInProgress = true;
+            CompletableFuture<Boolean> future = getPersistence()
+                    .persistAsync(getPersistableStore().getClone())
+                    .handle((nil, throwable) -> {
+                        writeInProgress = false;
+                        boolean success = throwable == null;
+                        if (success) {
+                            persistedGeneration.accumulateAndGet(myGeneration, Math::max);
+                        }
+                        return success;
+                    });
+            return future;
+        }
+    }
+
+    // Package-private for testing
+    void persistOnShutdown() {
+        // An in-flight or failed write leaves requested > persisted, so this single check covers "write
+        // still pending", "write failed" and "request throttled away" alike.
+        if (!isDropped()) {
+            return;
+        }
+        try {
+            // Route the fallback through the SAME serialized executor as every other write (persistNow
+            // bumps the generation, captures under scheduleLock and submits): queued behind any still-
+            // pending write, it can never be overwritten by an older snapshot landing later. Bounded so
+            // a congested queue can't hang JVM shutdown.
+            boolean success = persistNow().get(SHUTDOWN_WRITE_AWAIT_MS, TimeUnit.MILLISECONDS);
+            if (success) {
+                return;
+            }
+            // The queued write failed (e.g. disk error) - retry directly below.
+        } catch (TimeoutException e) {
+            // The queued write may never run: the executor uses daemon threads, which the JVM stops
+            // right after the shutdown hooks return. Fall through to the direct write - it is durable
+            // before this hook returns, and Persistence's write-id guard makes it safe: if the stale
+            // queued write does wake up later it is SKIPPED, not applied over the newer state.
+            log.warn("The queued shutdown write did not confirm within {} ms; writing directly on the shutdown thread.",
+                    SHUTDOWN_WRITE_AWAIT_MS);
+        } catch (Exception e) {
+            // Submission itself failed (e.g. the executor is already shut down) - direct write below.
+            log.warn("Submitting the shutdown fallback write failed; falling back to a direct write.", e);
+        }
+        // Generation, snapshot AND the write ticket are all captured under scheduleLock - the identical
+        // atomicity contract the async paths have (their ticket is assigned inside persistAsync, invoked
+        // under this same lock). Reserving the ticket outside the lock would open a gap where a concurrent
+        // persistNow() captures a NEWER snapshot yet ends up with an OLDER ticket, making the guard skip
+        // the newer write as stale and falsely report it clean. Only the disk write stays outside the lock.
+        long myGeneration;
+        T snapshot;
+        long writeId;
+        synchronized (scheduleLock) {
+            myGeneration = requestedGeneration.get();
+            snapshot = getPersistableStore().getClone();
+            writeId = getPersistence().reserveWriteId();
+        }
+        try {
+            // Durable the moment this returns (shares the store's write monitor with the executor task, so
+            // it cannot interleave with an in-flight write of this store). If a write with a NEWER ticket
+            // already landed by now, this is skipped - correct: that ticket's snapshot was captured after
+            // ours under the same lock, so strictly newer state is already on disk.
+            getPersistence().persist(snapshot, writeId);
+            persistedGeneration.accumulateAndGet(myGeneration, Math::max);
+        } catch (Exception e) {
+            // Persistence.persist() propagates write failures (so callers can't mistake failure for
+            // durability); on the shutdown-hook thread there is nothing left to do but log.
+            log.error("Direct write at shutdown failed.", e);
         }
     }
 }
--- a/trade/src/main/java/bisq/trade/Trade.java
+++ b/trade/src/main/java/bisq/trade/Trade.java
@@ -17,6 +17,7 @@
 
 package bisq.trade;
 
+import bisq.common.fsm.Event;
 import bisq.common.fsm.FsmModel;
 import bisq.common.fsm.State;
 import bisq.common.monetary.Monetary;
@@ -26,6 +27,7 @@ import bisq.common.observable.ReadOnlyObservable;
 import bisq.common.proto.PersistableProto;
 import bisq.contract.Contract;
 import bisq.identity.Identity;
+import bisq.network.p2p.message.EnvelopePayloadMessage;
 import bisq.offer.Direction;
 import bisq.offer.Offer;
 import bisq.security.DigestUtil;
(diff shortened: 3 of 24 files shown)


trade/src/main/java/bisq/trade/Trade.java
```diff
@@ -141,9 +161,64 @@ protected bisq.trade.protobuf.Trade.Builder getTradeBuilder(boolean serializeFor
...
         Optional.ofNullable(getTradeProtocolFailure()).ifPresent(e -> builder.setTradeProtocolFailure(e.toProtoEnum()));
         Optional.ofNullable(getPeersTradeProtocolFailure()).ifPresent(e -> builder.setPeersTradeProtocolFailure(e.toProtoEnum()));
+        // Only network messages are serializable; local, user-triggered events (e.g. BisqEasyConfirmFiatSentEvent)
+        // have no proto representation and are never at risk of being lost across a restart since they are only
+        // ever created while the app is already running.
+        getEventQueue().stream()
+                .filter(EnvelopePayloadMessage.class::isInstance)
+                .map(event -> ((EnvelopePayloadMessage) event).toProto(serializeForHash))
+                .forEach(builder::addPendingFsmEvents);
+        getProcessedEvents().forEach(eventClass -> builder.addProcessedEventClasses(eventClass.getName()));
```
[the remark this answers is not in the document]
rodvar: I think this is cosmetic, Trade  has noequals/hashCode (reference equality), is a locally-persisted PersistableProto there is really no impact as far as I'm concerned?

trade/src/main/java/bisq/trade/bisq_easy/BisqEasyTradeService.java
```diff
@@ -456,6 +501,9 @@ private BisqEasyProtocol createAndAddTradeProtocol(BisqEasyTrade trade) {
         }
         trade.setProtocolVersion(tradeProtocol.getVersion());
         tradeProtocolById.put(id, tradeProtocol);
+        if (isRestoredTrade) {
+            tradeProtocol.drainEventQueue();
+        }
```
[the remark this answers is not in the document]
rodvar: resolved

rodvar:
@KimStrand please review 🙏

KimStrand (commented):
I have a few concerns about the proposed implementation:
- The tests verify the new recovery mechanism, but do not appear to reproduce the original failure through the real application lifecycle. Without a test that fails on the unmodified target branch, it is difficult to confirm the root cause.
- The periodic recovery path appears to send messages directly into the FSM instead of through the normal onMessage validation path. This could allow messages previously rejected because of a trading halt, minimum-version requirement, or banned user to be applied later.
- Persisting mutable FSM queues asynchronously without the same synchronization as protocol processing looks race-prone.
- Persisting generic FSM events and restoring them through class names adds considerable complexity and compatibility risk.

I think a good direction could be:
- Add a deterministic regression test reproducing the real failure.
- If the test confirms that a message can be missed during startup, a good first fix could be to close that gap directly by registering the listener and collecting previously processed messages in one operation
- If tests also demonstrate message loss during normal runtime, we chould address that separately with a durable pending inbox that survives failures and restarts

Since this touches message delivery, persisted trade state, and protocol transitions, I consider it a sensitive area. A missed or incorrectly replayed message can leave a trade stuck or cause duplicate side effects, so I think we should proceed carefully with a reproducible test and a narrowly scoped first fix.

rodvar:
> I have a few concerns about the proposed implementation:
> 
> * The tests verify the new recovery mechanism, but do not appear to reproduce the original failure through the real application lifecycle. Without a test that fails on the unmodified target branch, it is difficult to confirm the root cause.
> * The periodic recovery path appears to send messages directly into the FSM instead of through the normal onMessage validation path. This could allow messages previously rejected because of a trading halt, minimum-version requirement, or banned user to be applied later.
> * Persisting mutable FSM queues asynchronously without the same synchronization as protocol processing looks race-prone.
> * Persisting generic FSM events and restoring them through class names adds considerable complexity and compatibility risk.
> 
> I think a good direction could be:
> 
> * Add a deterministic regression test reproducing the real failure.
> * If the test confirms that a message can be missed during startup, a good first fix could be to close that gap directly by registering the listener and collecting previously processed messages in one operation
> * If tests also demonstrate message loss during normal runtime, we chould address that separately with a durable pending inbox that survives failures and restarts
> 
> Since this touches message delivery, persisted trade state, and protocol transitions, I consider it a sensitive area. A missed or incorrectly replayed message can leave a trade stuck or cause duplicate side effects, so I think we should proceed carefully with a reproducible test and a narrowly scoped first fix.

💯 agreed Kim - thanks for the careful review. I'll work on each item and update soon 🙏

rodvar:
Thanks @KimStrand - the concerns are valid and I've reworked the approach around them (see last commit with test):

  - Root cause now has a real-lifecycle regression test. It drives the actual message→FSM→persist→restore path via initialize()/onMessage() (no direct FSM calls), and fails on the unmodified base with the exact "stuck one step short" state (…RECEIVED_BTC_ADDRESS instead of …RECEIVED_FIAT_SENT_CONFIRMATION), passing only with the fix. Confirms the queued out-of-order message is lost on restart ✅ 
  - Splitting the change into the proven fix + the speculative one:
    - Keep: persist the FSM eventQueue + drain-on-restore - the minimal, test-backed fix for the confirmed bug. I'll also fix the unsynchronized persistence snapshot race you flagged.
    - Cut: the periodic/reconnect recovery pass. Agreed it's over-scoped and unsafe as written; it re-applies messages via direct handle(), bypassing the halt/min-version/banned-user checks in onMessage(), and it's the source of most of the class-name-based event persistence complexity. No test demonstrates the runtime-loss scenario it targets.
    - Refile as a separate issue - a durable pending-inbox for received-but-unapplied messages, gated on a test that actually reproduces runtime loss, and designed with validation intact rather than a raw handle() sweep.
    
Working on this now.

KimStrand:
Thanks for taking the time to rework this and address the earlier concerns. I appreciate the work that has already gone into investigating the issue and improving the solution. The revised approach is a meaningful improvement and addresses several earlier concerns, although a few important questions still need to be resolved. Before we implement too much in this sensitive area, I would appreciate @HenrikJannsen perspective on the root cause and the preferred architectural direction.

1. Review of the current changes

Removing the periodic/reconnect recovery path resolves my main concern about bypassing normal validation and potentially repeating protocol side effects. Removing persisted Java class names is also a good  simplification.

The new lifecycle test is useful, but it currently proves a narrower case: an event is queued, the node restarts before the prerequisite message arrives, and the queued event would previously be lost. It does not yet confirm that this was the cause of the originally reported issue. The restart is simulated with a direct toProto()/fromProto() round trip, so the real asynchronous persistence path is not exercised.

I also think the persistence synchronization concern remains. CopyOnWriteArraySet prevents concurrent iteration failures, but it does not make the FSM state and pending queue one atomic snapshot. Persistence could still read the state before a transition and the queue after it.

Since the change is implemented in the generic Trade model and also affects MuSig trades, that scope should either be covered by tests or narrowed to Bisq Easy.

2. A simple first fix if the issue is at startup

If a regression test confirms that messages can be missed during startup, I think the most direct fix would be to close the gap between reading previously processed messages and registering the listener.

The network service could provide one atomic operation that registers the listener and captures the messages that arrived before registration. After that snapshot is returned, the service can replay it, with message-ID deduplication. This gives a clear guarantee: every message is either included in the startup snapshot or delivered to the registered listener.

That would be a relatively small and focused first fix without introducing persistence of generic FSM internals.

3. Longer-term improvement

The stronger long-term solution would be a durable per-trade inbox:
  - Authenticate and validate an incoming message.
  - Persist it by message ID before acknowledging responsibility for it.
  - Process each trade serially.
  - Keep messages pending when the current state cannot accept them.
  - Atomically persist the new FSM state and inbox result.
  - Retry pending messages after relevant transitions and startup.
  - Ensure handlers and outgoing side effects are idempotent.

This would address startup gaps, runtime failures, restarts and duplicate delivery through one explicit model.

I think we need to be especially careful here because this sits between message delivery and trade state. Once a message is acknowledged, the peer may stop resending it, so losing it can leave a trade stuck. At the same time, processing a message twice could repeat protocol actions. We also need to ensure that the trade state and pending messages are saved consistently, so a restart cannot restore an invalid combination.

Given the scope and sensitivity of these changes, I think we should get @HenrikJannsen input on the root cause and preferred architectural direction before taking the implementation further.

trade/src/main/proto/trade.proto
```diff
@@ -85,9 +85,12 @@ message Trade {
...
   repeated network.EnvelopePayloadMessage pendingFsmEvents = 15;
-  // Fully qualified class names of FSM events which have already triggered a transition at least
-  // once. Used to avoid re-queueing stale duplicates after a restore. See bisq.common.fsm.FsmModel.
-  repeated string processedEventClasses = 16;
+  // Field 16 (processedEventClasses: fully qualified class names of already-processed FSM events,
+  // resolved back via Class.forName()) was reserved so the field number is not silently reused -
+  // this field never shipped in a release (it was introduced and removed again within the same
+  // unreleased #1622 fix), so no wire-compatibility concern actually applies here, but reserving
+  // costs nothing and avoids relying on that being true forever.
+  reserved 16;
```
HenrikJannsen: As it was never deployed it does not make sense to pollute the protobuf. AI was a bit too motivated ;-)

HenrikJannsen:
I looked quickly over the PR and the description.

The issue with not persisted eventQueue can lead to a stuck trade in case the user close the app before the required messages arrives. So I think that makes sense to fix. 
One thing we should take care of is that pending events do not stay persisted longer as needed, specially not when a trade is completed as it might contain private data which gets deleted after some time in the trade. But I guess by persisting the trade after a successful application of a persisted event, that issues should be resolved. Still there could be a race that we persist, but then receive the missing message and apply it, thus when applying it after restart the state change has already happened and the message is ignored. 

At runtime there should not be a way that a trade get stuck by that as the eventQueue is iterated at each new arriving message. The only possible reason that it still get stuck might be some threading issue. If that can be reproduced by a test then I think the better approach is to try to fix the threading issue.

rodvar:
thanks guys 🙏 went through everything carefully. I think this splits cleanly: the persisted-eventQueue fix stands on its own for the shutdown/restart case, and the runtime/threading question is a separate beast - I've filed it as a follow-up (#4896) so we don't block this on the bigger durable-inbox design.

For this PR fixes, I reckon if I get these 3 in then we should be able to land it comfortably:

1) **Make the persisted snapshot atomic**. Right now handle()/drainEventQueue() mutate under synchronized(this), but Trade#toProto() reads the eventQueue off-thread without that lock - the CopyOnWriteArraySet keeps each read crash-safe but doesn't give a consistent {state + eventQueue} snapshot, which is exactly the "persist state before the transition, queue after" race you both flagged. I'll take the state+queue snapshot under the fsm lock so persistence can't capture a torn pair (no double-apply, no drop on restore). processedEvents isn't persisted, so only those two need to move together. I'll also make the regression test round-trip through the real persist/restore path rather than a bare toProto/fromProto.

2) **Guarantee pending events don't outlive the trade**. handle() already clears the eventQueue on the final state, so completed trades wipe their queued messages - I'll make sure we persist right after that clear so it actually lands on disk, and add a test asserting persisted pendingFsmEvents is empty once the trade completes. That covers the private-data-after-completion point.

3) **Sort out the MuSig scope**. The persistence lives in the generic Trade model so MuSig picks it up too, but only BisqEasyTradeService drains on restore today. I'll mirror the drain-on-restore in MuSigTradeService + a restore test (or narrow the persistence to Bisq Easy) so the generic change isn't half-covered.

I'm working on this and will update you guys as soon as I get it in.

And on the runtime case - agreed Henrik, from what I've seen at runtime the queue drains on every arriving message, so a genuine runtime stuck points at a threading issue rather than persistence. I'll chase a runtime repro in #4896 and, if it reproduces, fix the threading path directly instead of building the durable inbox speculatively 👍 .

rodvar:
> thanks guys 🙏 went through everything carefully. I think this splits cleanly: the persisted-eventQueue fix stands on its own for the shutdown/restart case, and the runtime/threading question is a separate beast - I've filed it as a follow-up (#4896) so we don't block this on the bigger durable-inbox design.
> 
> For this PR fixes, I reckon if I get these 3 in then we should be able to land it comfortably:
> 
> 1. **Make the persisted snapshot atomic**. Right now handle()/drainEventQueue() mutate under synchronized(this), but Trade#toProto() reads the eventQueue off-thread without that lock - the CopyOnWriteArraySet keeps each read crash-safe but doesn't give a consistent {state + eventQueue} snapshot, which is exactly the "persist state before the transition, queue after" race you both flagged. I'll take the state+queue snapshot under the fsm lock so persistence can't capture a torn pair (no double-apply, no drop on restore). processedEvents isn't persisted, so only those two need to move together. I'll also make the regression test round-trip through the real persist/restore path rather than a bare toProto/fromProto.
> 2. **Guarantee pending events don't outlive the trade**. handle() already clears the eventQueue on the final state, so completed trades wipe their queued messages - I'll make sure we persist right after that clear so it actually lands on disk, and add a test asserting persisted pendingFsmEvents is empty once the trade completes. That covers the private-data-after-completion point.
> 3. **Sort out the MuSig scope**. The persistence lives in the generic Trade model so MuSig picks it up too, but only BisqEasyTradeService drains on restore today. I'll mirror the drain-on-restore in MuSigTradeService + a restore test (or narrow the persistence to Bisq Easy) so the generic change isn't half-covered.
> 
> I'm working on this and will update you guys as soon as I get it in.
> 
> And on the runtime case - agreed Henrik, from what I've seen at runtime the queue drains on every arriving message, so a genuine runtime stuck points at a threading issue rather than persistence. I'll chase a runtime repro in #4896 and, if it reproduces, fix the threading path directly instead of building the durable inbox speculatively 👍 .

I've pushed these, still testing but looking good so far. Please have a look to make sure we are all in the same page 🙏

rodvar:
I confirm once again this PR solves the described issue. If a do trades in a very fast fashion:

 - using Bisq Desktop (current prod 2.1.11 release) -> I get stucked every single time
 - using Bisk Desktop (build from this branch) -> ✅ 
 - Also, using connect with the node running this code never gets the trade stuck on mobile ✅ 
 
If you guys agree I would like to get this in for this weekend mobile releases. We might also want to consider releasing this fairly soon for Desktop? 
Please let me know if any other concerns should be addressed - thanks for the help 🙏

trade/src/main/proto/trade.proto
KimStrand: can be removed, right?

trade/src/main/java/bisq/trade/Trade.java
```diff
@@ -126,24 +148,64 @@ protected Trade(C contract,
...
         Optional.ofNullable(getPeersErrorStackTrace()).ifPresent(builder::setPeersErrorStackTrace);
         Optional.ofNullable(getTradeProtocolFailure()).ifPresent(e -> builder.setTradeProtocolFailure(e.toProtoEnum()));
         Optional.ofNullable(getPeersTradeProtocolFailure()).ifPresent(e -> builder.setPeersTradeProtocolFailure(e.toProtoEnum()));
+        // Only network messages are serializable; local, user-triggered events (e.g. BisqEasyConfirmFiatSentEvent)
+        // have no proto representation and are never at risk of being lost across a restart since they are only
+        // ever created while the app is already running.
+        snapshot.eventQueue().stream()
+                .filter(EnvelopePayloadMessage.class::isInstance)
+                .map(event -> ((EnvelopePayloadMessage) event).toProto(serializeForHash))
+                .forEach(builder::addPendingFsmEvents);
```
KimStrand, line 185: Could we define the cleanup lifecycle for these persisted messages? This stores the complete envelope payload, which can include BisqEasyAccountDataMessage or MuSig SendAccountPayloadMessage. A stuck trade may never reach a final state, so the queue may never be cleared. The existing redaction logic does not inspect pendingFsmEvents, and MuSig redaction is still a TODO. This means private account data could remain persisted indefinitely. I think pending events should at least be removed when a trade is archived or reaches the existing retention threshold.

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
```diff
@@ -69,6 +69,28 @@ protected long getMaxWriteRateInMs() {
...
+     * leave sensitive data on disk for an unbounded time - e.g. a completed trade's pending FSM events, which are
+     * cleared in-memory the instant the trade reaches a final state (see {@code bisq.common.fsm.Fsm#handle}), but
+     * whose on-disk copy has no other guaranteed opportunity to catch up: the next unrelated persist() call may
+     * never come (a completed trade has no further transitions), and the shutdown-hook flush this class registers
+     * is best-effort only - it never runs on a JVM crash or {@code kill -9}.
+     */
+    public CompletableFuture<Boolean> persistNow() {
+        lastWrite = System.currentTimeMillis();
+        writeInProgress = true;
+        dropped = false;
```
KimStrand: Could persistNow() keep the store marked dirty until its asynchronous write completes successfully, and could the shutdown path await or synchronously replace an in-progress write? Currently the dirty flag is cleared when the write is scheduled, so shutdown cannot distinguish “safely written” from “still waiting to be written.”

trade/src/main/java/bisq/trade/Trade.java
rodvar: easy fix - will push soon

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
rodvar: great catch, I'll fix on next push

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
```diff
@@ -56,20 +64,65 @@ public CompletableFuture<Boolean> persist() {
...
+        // races the still-in-flight write the shutdown hook must treat the data as not-yet-safely-written. `dropped`
+        // is cleared only once the write completes successfully, and the hook additionally awaits `writeFuture`.
+        dropped = true;
+        CompletableFuture<Boolean> future = getPersistence()
+                .persistAsync(getPersistableStore().getClone())
+                .handle((nil, throwable) -> {
+                    writeInProgress = false;
+                    boolean success = throwable == null;
+                    if (success) {
+                        dropped = false;
```
KimStrand: Thanks for addressing the retention and shutdown concerns. I think there is still one concurrency case here: while persistNow() is in progress, a newer persist() call can be dropped because writeInProgress is true, setting dropped = true. When the older persistNow() completes, this line clears that flag again. Shutdown could then believe everything was saved even though the newer state was never written.

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
```diff
@@ -56,20 +64,65 @@ public CompletableFuture<Boolean> persist() {
...
+        // Keep the store marked dirty until the async write actually confirms. Unlike persist() we do NOT clear
+        // `dropped` optimistically: persistNow() is used for privacy/retention-critical writes, so if a shutdown
+        // races the still-in-flight write the shutdown hook must treat the data as not-yet-safely-written. `dropped`
+        // is cleared only once the write completes successfully, and the hook additionally awaits `writeFuture`.
+        dropped = true;
+        CompletableFuture<Boolean> future = getPersistence()
+                .persistAsync(getPersistableStore().getClone())
+                .handle((nil, throwable) -> {
+                    writeInProgress = false;
+                    boolean success = throwable == null;
```
KimStrand: Does this future reliably report a disk-write failure? PersistableStoreReaderWriter.write() currently catches and logs serialization and I/O exceptions, so throwable could be null even when the data was not written. In that case we would clear the dirty flag without having successfully persisted the cleanup. Could the write failure be propagated here?

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
rodvar: let me have a look, will push changes if agreed

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
rodvar: mmm... let me think about this one, I'll add changes in or reply back soon

rodvar:
Just a heads-up , feedback keeps coming in about this problem hitting users - this time a direct report reach for `Bisq Easy (mobile) 0.9.0` ☝️ 

In this case, restarting is not even helping to solve the issue - I instructed the user to backup profile and do a full wipe out to see if that helps

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
rodvar: I see, the root problem is that a single boolean can't distinguish "my write landed" from "everything is clean."... Edge but possible - I think we can fix it cleanly using generation counter, I'm working on it

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
rodvar: easy we can just do a rethrow and its safer 👍  - will add in now

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
```diff
@@ -51,28 +65,83 @@ public CompletableFuture<Boolean> persist() {
...
         } else {
             lastWrite = System.currentTimeMillis();
             writeInProgress = true;
             dropped = false;
-            return getPersistence()
+            CompletableFuture<Boolean> future = getPersistence()
                     .persistAsync(getPersistableStore().getClone())
                     .handle((nil, throwable) -> {
                         writeInProgress = false;
                         return throwable == null;
```
KimStrand: Now that write failures are propagated, an ordinary persist() failure reaches this callback. However, it only returns false; the store remains marked clean. Since the future is already completed and dropped is false, persistOnShutdown() will not perform its fallback write. Could we mark the state dirty here and add a test verifying that a failed ordinary persist is retried during shutdown?

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
KimStrand: Thanks, the generation counter handles the original case when those operations occur in sequence. I think there may still be a small concurrency gap because setting dropped and incrementing dirtyGeneration are separate operations. A newer persist() could enter the dropped branch, while the older persistNow() completion still reads the previous generation and then clears dropped. Would it make sense to protect the dirty-state transitions with one lock, or derive the dirty state entirely from requested and successfully persisted generations? The persistence executor serializes the file writes, but this bookkeeping can  still run on different caller and completion threads.

trade/src/main/java/bisq/trade/bisq_easy/BisqEasyTradeService.java
```diff
@@ -456,6 +467,20 @@ private BisqEasyProtocol createAndAddTradeProtocol(BisqEasyTrade trade) {
...
         tradeProtocolById.put(id, tradeProtocol);
+        if (isRestoredTrade) {
+            // Isolate per trade: drainEventQueue() re-applies queued events and can raise an FsmException. The
+            // trade is already created and registered above, so we keep it regardless. Without this guard a
+            // single failing trade would escape the persistableStore.getTrades().forEach(...) loop in
+            // initialize() and block restoring every subsequent trade. A failed drain here means the trade
+            // stays stuck until it is manually looked at - there is no periodic or reconnect-triggered safety
+            // net to retry it.
+            try {
+                tradeProtocol.drainEventQueue();
```
KimStrand: Could we check the current trading-halt, minimum-version and banned-user state before draining restored events? The message passed validation when it was originally queued, but those conditions may have changed before restart. Since drainEventQueue() applies it directly through the FSM, startup could otherwise process an event that onMessage() would now reject. The same seems to apply to MuSig.

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
rodvar: I agree - working on it

I think your 2nd suggestion is spot on

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
rodvar: 💯  - will push soon

rodvar:
@KimStrand done , I've also rebased the branch with latest commits on for-mobile.. 👍 
Thanks for all the help

trade/src/main/java/bisq/trade/bisq_easy/BisqEasyTradeService.java
```diff
@@ -179,6 +177,24 @@ public void onCleared() {
...
+                .flatMap(service -> service.getProcessedEnvelopePayloadMessages().stream())
+                .forEach(message -> {
+                    // Per-message isolation: with the alert observer registered above, onMessage()'s
+                    // halt/min-version guards can now legitimately throw during this startup replay (they
+                    // could not before - the observer used to be registered only after this loop). A rejected
+                    // replayed message must not abort initialize() and skip every subsequent message.
+                    try {
+                        onMessage(message);
+                    } catch (Exception e) {
+                        log.warn("Re-feeding a processed message at startup was rejected (e.g. trading halted " +
```
KimStrand, line 205: Could we avoid logging the full message here? Some replayed messages, such as BisqEasyAccountDataMessage, include private payment-account data in their generated toString(). During a halt or minimum-version rejection, that data could therefore be written to the logs.

Logging only the message type and identifiers should be sufficient. The same applies to the corresponding MuSig catch, where SendAccountPayloadMessage contains the account payload.

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
KimStrand: Thanks, the generation counters address the dirty-state race and failed-write retry. I think one write-ordering race may still remain because generation assignment, snapshot capture and asynchronous scheduling do not necessarily happen in the same order.

An older call can capture snapshot v1 and pause before scheduling it. A newer call can then capture and write v2, after which the older call schedules and writes v1. persistedGeneration remains at 2, so isDropped() returns false even though the stored state is now v1.

trade/src/main/java/bisq/trade/bisq_easy/BisqEasyTradeService.java
rodvar: great catch, fix incoming

rodvar: missed this one https://github.com/bisq-network/bisq2/pull/4885#discussion_r3690007114 , I'll work it out in my next push as well

trade/src/main/java/bisq/trade/bisq_easy/BisqEasyTradeService.java
```diff
@@ -179,6 +177,24 @@ public void onCleared() {
...
+                                "by an emergency alert). messageType={}", message.getClass().getSimpleName(), e);
+                    }
+                });
+        networkService.addConfidentialMessageListener(this);
```
KimStrand: Thanks, persisting the FSM queue covers messages that have already reached the trade protocol. I think there is still a startup handoff gap before that, though.

A message can be processed after this replay has taken its snapshot but before the listener is registered. In that case it is neither included in the replay nor delivered live. The network layer may still ACK it and remove it from the mailbox, but it never reaches the FSM queue, so pendingFsmEvents cannot recover it.

Could listener registration and replay be made atomic, with a test that delivers a message during that handoff? This also seems like a good focused first fix for the startup issue.

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
```diff
@@ -46,33 +71,110 @@ public RateLimitedPersistenceClient() {
...
+            // Both captured under scheduleLock so a concurrent caller can't interleave a newer capture; the
+            // disk write itself stays outside the lock.
+            long myGeneration;
+            T snapshot;
+            synchronized (scheduleLock) {
+                myGeneration = requestedGeneration.get();
+                snapshot = getPersistableStore().getClone();
+            }
+            try {
+                getPersistence().persist(snapshot);
```
KimStrand: Thanks, the scheduleLock addresses the ordering between normal asynchronous persistence requests. I think the synchronous shutdown fallback can still be overtaken by an older write that is already queued on the
  persistence executor.

It seems reproducable: an older v1 write was queued, the shutdown fallback wrote v2 after timing out, and the queued write then wrote v1 over it. isDropped() still returned false because persistedGeneration had already advanced to 2.

Could we ensure that the shutdown fallback cannot be overwritten by any previously submitted write?

trade/src/main/java/bisq/trade/Trade.java
```diff
@@ -126,24 +148,64 @@ protected Trade(C contract,
...
+                EnvelopePayloadMessage envelopePayloadMessage = EnvelopePayloadMessage.fromProto(envelopePayloadMessageProto);
+                if (envelopePayloadMessage instanceof Event event) {
+                    pendingEvents.add(event);
+                } else {
+                    log.warn("Persisted pending FSM event does not implement Event and is dropped. envelopePayloadMessage={}",
+                            envelopePayloadMessage);
+                }
+            } catch (Exception e) {
+                log.warn("Could not resolve a persisted pending FSM event. It will be dropped and not re-applied " +
+                        "after restore. proto={}", envelopePayloadMessageProto, e);
```
KimStrand: Could we avoid logging the complete message or protobuf here? A pending event can contain Bisq Easy payment-account data or a MuSig account payload, so a restoration failure could write that private data to the logs.

Logging only safe metadata, such as the message type or protobuf oneof case, should be sufficient.

trade/src/main/java/bisq/trade/bisq_easy/BisqEasyTradeService.java
rodvar: this is a pre-existing issue (you can verify is also in main branch)

I will still take the opportunity and address it now, just flagging to not over-extend this important PR description fix too much (We can always create issues for improvement to address shortly)

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
rodvar: thank you, please check my next push

trade/src/main/java/bisq/trade/Trade.java
rodvar: trivial and valid thank you, will address now

rodvar:
@KimStrand all done

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
KimStrand: Thanks, routing the shutdown fallback through the serialized executor fixes the stale-overwrite ordering issue.

It seemas that the timeout path still leaves a shutdown-durability gap, though. The persistence executor uses daemon threads. If persistNow().get(...) times out, the shutdown hook returns while the newest snapshot is only queued. Once the shutdown hooks finish, the JVM may stop that executor before the queued write runs, so the state this fallback is intended to preserve can still be lost.

The new test keeps the JVM alive after the hook returns and only then completes olderWriteGate, allowing both queued writes to finish. A real JVM shutdown does not guarantee that additional time.

Should we ensure that the newest snapshot has actually been written before the shutdown hook returns?

rodvar: I see.. It took me some time to figure out this one.

My last fix eliminated the stale-overwrite but with a trade-off that made the timeout case strictly worse than the old code..

The thing is, there is a design tension in this component:

 - `Wait unbounded` vs  `Return after timeout` (my current solution) vs `Sync-write on the hook thread after timeout` (would bring back the mini-race risk)

I think we can solve it by adding a sort of "monotonic write id" to the Persistance layer.. I'm working on this, will update / add tests promptly as well

rodvar: I'm gonna re-push soon. I think after all this work we are doing together this PR is not just a fix but a substantial improvement to the overall components involved. Thanks for the help Kim 🙏

trade/src/main/java/bisq/trade/bisq_easy/BisqEasyTradeService.java
KimStrand: Thanks, registering the listener first closes the message-loss side of the startup handoff.

It seems that the handoff is not duplicate-safe yet, though. A message can be added to processedEnvelopePayloadMessages after listener registration but before the replay snapshot is taken. It would then be delivered live and also included in the replay.

For ordinary trade messages the FSM can absorb that duplicate. A BisqEasyTakeOfferRequest, and similarly MuSig SetupTradeMessage_A, is different because it creates and stores the trade and protocol before entering the FSM. Two concurrent deliveries could both pass findProtocol()/tradeExists(), create separate trade models, and race the store and tradeProtocolById updates. This could leave the active protocol using a different model from the persisted trade, as well as repeat handler side effects.

The current handoff test only covers the message-loss case by returning an empty replay snapshot. Could we make this handoff atomic or otherwise duplicate-safe and add a deterministic test for a take-offer message delivered through both paths?

rodvar: please check now - I think we should be good now , please let me know.

KimStrand (commented):
utACK

rodvar:
@coderabbitai full review

rodvar:
Updated description to reflect the final state of the solution since the reviewing process ended up improving it substantially

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
```diff
@@ -46,33 +69,129 @@ public RateLimitedPersistenceClient() {
...
+        }
+        try {
+            // Route the fallback through the SAME serialized executor as every other write (persistNow
+            // bumps the generation, captures under scheduleLock and submits): queued behind any still-
+            // pending write, it can never be overwritten by an older snapshot landing later. Bounded so
+            // a congested queue can't hang JVM shutdown.
+            boolean success = persistNow().get(SHUTDOWN_WRITE_AWAIT_MS, TimeUnit.MILLISECONDS);
+            if (success) {
+                return;
+            }
```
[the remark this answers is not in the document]
rodvar: The window is real but **unclosable** - a throttled persist() can land right after any re-check too; closing it needs producer quiescence at shutdown, and losing a final-instant mutation is equivalent to the process dying 1ms earlier... The hook is documented best-effort (never runs on kill -9). Not worth churning the most-reviewed code in the PR post-utACK.

rodvar:
@HenrikJannsen please do not merge this yet - I found some issues when testing with connect apps I want to make sure this PR is not responsible for first. I'll update on this ASAP

rodvar:
@KimStrand sorry Kim, I found a potential hazard causing a JVM indefinite lock and addressed in the [commit](https://github.com/bisq-network/bisq2/pull/4885/commits/708a660cb39255677042dcda24b2bee31178ff04) above  - I left it as a separate commit so we can decide easily if we add / change / reject . Please review and if you utACK I'll amend it and give the green light for Henrik. Thank you 🙏

KimStrand:
Thanks for catching this and keeping the change separate. I reviewed commit 708a660. Bounding the snapshot lock prevents the shared persistence executor from waiting indefinitely, which addresses the hazard you found.

I still found one retry gap in the timeout path, as well as two concerns around persisted events. I’ll add those inline with concrete scenarios.

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
```diff
@@ -46,33 +69,129 @@ public RateLimitedPersistenceClient() {
...
-        } else {
+        synchronized (scheduleLock) {
+            // Bump BEFORE the throttle decision and before any snapshot: if this request gets dropped, the store
+            // is dirty by construction (requested > persisted) until some later write covers this generation.
+            long myGeneration = requestedGeneration.incrementAndGet();
+            boolean tooFast = System.currentTimeMillis() - lastWrite < getMaxWriteRateInMs();
+            if (tooFast || writeInProgress) {
+                return CompletableFuture.completedFuture(false);
+            }
             lastWrite = System.currentTimeMillis();
```
KimStrand, line 101: Could we guarantee recovery after a snapshot timeout through the normal persistence path?

Both persist() and persistNow() update lastWrite before asynchronous serialization. If serialization then times out, the generation remains dirty, but a following ordinary persist() within the one-second rate- limit window returns false and does not schedule a later retry. The generation pair records the missed write, but it does not actively retry it.

Seems like this is reproducable: the snapshot timed out, the next ordinary persist() returned false, and isDropped() remained true. The current test retries with a second explicit persistNow(), which bypasses the rate limiter and therefore does not cover this path.

Could we add coverage showing that the normal persistence path recovers after the timeout without requiring another explicit persistNow()?

common/src/main/java/bisq/common/fsm/Fsm.java
```diff
@@ -128,13 +140,61 @@ public <E extends Event> void handle(E event) {
...
+     * implementation, so whether an {@link FsmException} propagates to the caller or is absorbed is up to the
+     * subclass (e.g. the trade protocols swallow it and surface an error state instead).
+     */
+    public void drainEventQueue() {
+        // Unbounded, like handle() - ReentrantLock is reentrant, so the nested lock() calls inside each
+        // handle(event) below (same thread) are free.
+        model.lock.lock();
+        try {
+            // Clone set to avoid ConcurrentModificationException as handle() mutates model.eventQueue.
+            new HashSet<>(model.getEventQueue()).forEach(this::handle);
```
KimStrand, line 201: Before replaying persisted network events through handle(), could we make sure its generic error paths do not render the complete event?

handle() currently logs the full event when the state is final and when handling fails, and also includes it in the “No transition found” exception message. BisqEasyAccountDataMessage and MuSigSendAccountPayloadMessage use generated toString() methods that include the payment-account data or account payload.

This means a replay failure can still expose private data in the logs. In the checkArgument case, that data can also become part of the persisted error information and the error report sent to the peer. The service-specific logging cleanup does not cover these generic FSM paths.

Could these paths use only safe metadata such as the event type, trade ID and message ID?

rodvar:
@KimStrand amended to the same separate commit please review 🙏

rodvar:
> @KimStrand amended to the same separate commit please review 🙏

sorry it didn't got pushed - now it is

persistence/src/main/java/bisq/persistence/RateLimitedPersistenceClient.java
KimStrand: Thanks, this now covers the sequential timeout -> ordinary persist case. I found one remaining overlap case in the rollback:

persistNow A records t1, then persistNow B records t2 and remembers t1. If both writes fail, A cannot roll back because B now owns lastWriteGeneration. B then restores t1, but t1 also belongs to a failed attempt. The next ordinary persist() is therefore still rejected as tooFast.

This seems to be reproducable with two overlapping failed persistNow() futures followed by one ordinary persist(); the ordinary call returned false and the store stayed dirty. Overlapping persistNow() calls are possible, for example when multiple trades reach a final state close together.

Could we make the failure handling correct for overlapping immediate writes as well, and add that case as a regression test?

This is blocking because it can recreate the original retry gap under concurrency.

common/src/main/java/bisq/common/fsm/Fsm.java
```diff
@@ -116,7 +136,8 @@ public <E extends Event> void handle(E event) {
                             .forEach(e -> model.eventQueue.add(event));
                 }
             } catch (Exception exception) {
-                log.error("Error at handling {}.", event, exception);
+                // Class name only - see the comment at the final-state log above.
+                log.error("Error at handling event of class {}.", event.getClass().getSimpleName(), exception);
```
KimStrand, line 140: The class-only logging addresses the sensitive-data concern, thanks. One small null-handling regression remains here: handle() deliberately validates event with checkNotNull(), but this catch then calls event.getClass(). If event is null, this throws a new raw NullPointerException, masking “event must not be null” and bypassing the configured FsmException/error-state handling.

Can be confirmed with a focused handle(null) test. Could the safe event-class rendering also handle null, with a regression test for this case?

common/src/main/java/bisq/common/fsm/Fsm.java
```diff
@@ -61,11 +61,15 @@ protected Fsm(M model) {
...
-        synchronized (model) {
+        // Lock on the model's explicit lock rather than on this Fsm instance: the model (see
+        // FsmModel#getStateAndEventQueueSnapshot) must be lockable directly by callers that only hold a reference
+        // to the model - e.g. Trade#getTradeBuilder during persistence - without needing a reference to this
+        // Fsm/TradeProtocol. Sharing the same lock object gives both sides the same mutual exclusion, so a
+        // persistence snapshot of {state, eventQueue} can never be taken mid-transition. Unlike
+        // getStateAndEventQueueSnapshot()'s bounded tryLock, a live transition takes the lock unbounded - its
+        // correctness cannot be short-circuited by a timeout.
+        model.lock.lock();
+        try {
```
HenrikJannsen, line 72: Why using 2 try blocks?

common/src/main/java/bisq/common/fsm/Fsm.java
```diff
@@ -178,9 +193,14 @@ protected void persistOnFinalState() {
...
     public void drainEventQueue() {
-        synchronized (model) {
+        // Unbounded, like handle() - ReentrantLock is reentrant, so the nested lock() calls inside each
+        // handle(event) below (same thread) are free.
+        model.lock.lock();
+        try {
             // Clone set to avoid ConcurrentModificationException as handle() mutates model.eventQueue.
             new HashSet<>(model.getEventQueue()).forEach(this::handle);
+        } finally {
+            model.lock.unlock();
```
HenrikJannsen, line 203: This would unlock before the persist code in 
```
finally {
                if (reachedFinalState) {
                    persistOnFinalState();
                } else {
                    persist();
                }
            }
```

trade/src/main/java/bisq/trade/mu_sig/MuSigTradeService.java
```diff
@@ -600,30 +640,67 @@ public MusigGrpc.MusigStub getMusigAsyncStub() {
...
+            MuSigTrade trade = new MuSigTrade(contract, isBuyer, false, myIdentity, offer, sender, receiver);
+
+            AccountPayload<? extends PaymentMethod<?>> accountPayload = findMyAccount(trade).orElseThrow().getAccountPayload();
+            trade.getMyself().setAccountPayload(accountPayload);
+
+            String tradeId = trade.getId();
+            checkArgument(findProtocol(tradeId).isEmpty(), "We received the MuSigTakeOfferRequest for an already existing protocol");
+            checkArgument(!tradeExists(tradeId), "A trade with that ID exists already");
+
+            awaitCreationTestSeam(tradeId);
```
HenrikJannsen, line 674: Would be good to avoid that as only a test aspect.

HenrikJannsen:
Thanks for the work on that. 
I reviewed only lightly. I think the PR is doing too much. 

We should separate the difference changes:
- trade protocol threading changes
- other trade protocol changes
- persistence threading changes
- FMS handling of pending events with the applications in the protocols

It touches several fundamental areas and we should break those changes down to reduce risks.  

I would also avoid packing the isRestoredTrade in `createAndAddTradeProtocol` and would use a separate method instead and call that at the initialize. after createAndAddTradeProtocol as it concerns a different aspect as create and add protocol.

For the persistence changes I would prefer to get also a review from @alvasw . I have not looked too close but if feels like substantial changes. 

To avoid endless dependency chains here, I would suggest to keep the threading and persistence issues separate. 
The changes in the trade domain seem less risky. The threading issues in FMS as well.

I would recommend:
- PR for FSM threading improvements
- PR with event queue persistence and application (accepting open threading issues in the other layers)
- PR with threading improvements in trade domain
- PR with other improvements in trade domain
- PR with threading improvements in persistence - this feels the most critical to me and requires very careful review

rodvar:
OK, I've started work on this. Is going to take a while, just a heads up on how things are gonna go so everyone is on the same page:

I'm gonna split this PR into a total of 5 (labeled A - E)

 1. A → B (in that order for review) unblocks the actual bug that this PR fixes
 2. C, D and E can proceed in parallel afterwards, C is very important for the full UX-top experience; E on its own careful timeline (persistence). 

Here is a brief on each:

  - **PR A**: Merge "FSM threading" with the FSM plumbing the fix needs
  - **PR B**: the verifiable **bugfix** post A merge, and only it (trade/ + trade.proto, depends on A): persist {state, queue, processed} + drain-on-restore for BisqEasy and MuSig. The disk round-trip regression test lives here.
  - **PR C**: live recovery pass (reconnect-debounced + periodic re-apply): it's the no-restart enhancement that I consider very important to do before releasing the fix, depends on B.
  - **PR D**: trade-domain error handling (persisted trade error + peer report, rollback-on-error at service level).
  - **PR E**: the persistence rework (generation counters, write-failure propagation, shutdown fallback, scheduleLock ordering). This is the alvasw-review PR, B's thin hook gets replaced when this lands.

After all these are lodged I'll close this one.
I'm working on A first let me know if you have any comments please  - thanks for the help 🙏

KimStrand:
@rodvar @HenrikJannsen 
Thanks for all the work on this, and thanks Henrik for the feedback. I agree that this is a very sensitive area because it crosses networking, message acknowledgement, FSM state transitions, and persistence.

Before continuing with the implementation or splitting it into several PRs, I think we should first take a step back and understand the concrete issue thoroughly. We should document:
- Under which exact circumstances the trade gets stuck.
- Whether it happens during startup, runtime, or both.
- How it can be reproduced reliably.
- What the expected and actual behavior are.
- Which component is responsible for losing or retaining the message.

Based on that, we can describe two possible directions:

- A small and focused short-term fix that reduces the immediate risk.
- A long-term solution that addresses the underlying architectural problem.

We should then write a short specification for both solutions, including their guarantees, responsibilities, failure cases, and acceptance tests. Once we agree on those specifications, we can decide how to split the work into PRs and review each implementation against the agreed behavior.

I think this would give us a clear direction and help us avoid continuing the current implementation–review–change cycle.

rodvar:
> @rodvar @HenrikJannsen Thanks for all the work on this, and thanks Henrik for the feedback. I agree that this is a very sensitive area because it crosses networking, message acknowledgement, FSM state transitions, and persistence.
> 
> Before continuing with the implementation or splitting it into several PRs, I think we should first take a step back and understand the concrete issue thoroughly. We should document:
> 
> * Under which exact circumstances the trade gets stuck.
> * Whether it happens during startup, runtime, or both.
> * How it can be reproduced reliably.
> * What the expected and actual behavior are.
> * Which component is responsible for losing or retaining the message.
> 
> Based on that, we can describe two possible directions:
> 
> * A small and focused short-term fix that reduces the immediate risk.
> * A long-term solution that addresses the underlying architectural problem.
> 
> We should then write a short specification for both solutions, including their guarantees, responsibilities, failure cases, and acceptance tests. Once we agree on those specifications, we can decide how to split the work into PRs and review each implementation against the agreed behavior.
> 
> I think this would give us a clear direction and help us avoid continuing the current implementation–review–change cycle.

---

**DISCLAIMER**: This is a long response so grab your beverage of preference and come back to read :) — I tweaked my words using AI in an attempt to make it as digestable as possible, please ask questions if something doesn't make sense

---

Thank you Kim for taking a step back to discuss this, and Henrik for the earlier feedback.

First, a bit of framing. I know this repo reviewers generally don't manually test PRs (utACK vs ACK) — not a criticism, just a fact — but if you do run this branch against the repro described in the comments, you'll see it fixes a real issue. It probably got convoluted because we folded a lot of hardening in alongside the core fix, and I understand the wish to slow down. But I want to be very clear about the stakes: these are real problems hitting production users TODAY. Trades get stuck and users end up "finishing in chat", which makes Bisq Easy feel not robust — with good reason. It affects Desktop and Connect (it's especially visible in the mobile apps), so this isn't a mobile-only concern. That's why I'm pushing hard here: users are living with a protocol bug while we discuss it.

That said, I fully agree with writing the problem and the solutions down before continuing — and I think we're closer to alignment than it looks. Here's the concrete issue, already understood and verified during this work, mapped to your questions:

### WHAT/WHERE THE TRADE GETS STUCK

Bisq Easy phase-1 has two independent, unordered sub-steps: the seller sends account data, and the seller receives the buyer's BTC-address message. But the fiat-sent transition is single-source-state — it only fires from the fully-settled "...SELLER_RECEIVED_BTC_ADDRESS" state (BisqEasySellerAsMakerProtocol). If BisqEasyConfirmFiatSentMessage arrives while phase-1 is only half-settled, Fsm.handle() parks it in model.eventQueue, and that queue only drains on a later successful transition. If none comes, the message is stranded and the trade freezes at phase-1. The buyer-as-taker path has the identical gate, and mu_sig shares the same Trade/FsmModel base — so **it's systemic, not a one-off.**

### STARTUP OR RUNTIME? BOTH

It strands at runtime and persists across startup. eventQueue and processedEvents are not persisted (Trade.toProto serializes only the state name; the 3-arg FsmModel restore ctor is currently dead code). The "restart sometimes fixes it" behavior is incidental, not a guarantee: it only works if the message is still in local mailbox storage for ConfidentialMessageService to replay, and that replay buffer is also in-memory-only and cleared on shutdown.

### EXPECTED VS ACTUAL

**Expected**: fiat-sent is eventually applied and the trade advances. 
**Actual**: it sits in the ephemeral queue and the trade freezes until an unrelated later transition or an incidental restart-replay happens to drain it.

### WHICH COMPONENT LOSES/RETAINS THE MESSAGE

FsmModel.eventQueue retains it in memory but loses it on restart because Trade.toProto never serializes it. Two neighbors compound this: the ConfidentialMessageService replay buffer (in-memory, incidental) and — the deeper one — the ACK layer marking the message delivered the instant it's decrypted, regardless of whether the FSM actually consumed it. Delivery-success != domain-success, with no path back. I'd call that a second, more fundamental architectural gap, likely its own effort since it touches network ACK semantics.

### REPRODUCTION (one honest caveat)

It's deterministic in a controlled test (deliver fiat-sent before phase-1 settles; we have an e2e repro in BisqEasyTradeTest). In the wild it's a timing race — the peer's fiat-sent has to land before the BTC-address message is processed — so it's hard to force on demand, but the exact preconditions are known.

### TWO DIRECTIONS

- Short-term, focused fix: persist eventQueue (filtered to proto-serializable messages) + processedEvents, wire the dead restore ctor, and drain once when a restored trade's protocol is rebuilt. This makes restart-recovery deterministic instead of incidental. Small and verifiable.
- Long-term, architectural: the FSM threading correctness, the persistence rework (the generation-counter/write-failure work), and a live recovery pass so no restart is needed at all — plus honestly documenting the ACK-vs-domain-consumption gap as the boundary of this effort.

### WHY THE PROPOSED SPLIT ALREADY MATCHES THIS

The A-E split we discussed with Henrik is exactly this decomposition, just phrased as PRs:

- PR A (common/fsm): the persistable queue + atomic snapshot machinery the fix stands on.
- PR B: the short-term fix above — persist {state, queue, processed} + drain-on-restore, BisqEasy + mu_sig, with the duplicate-safe replay handoff you flagged. This is the verifiable bugfix and closes the restart-heal path.
- PR C: the live recovery pass (no-restart self-heal) — the long-term "never gets stuck" piece.
- PR D: trade-domain error handling.
- PR E: the persistence threading rework (generation counters, write-failure propagation), where your two still-open races land, for careful review (alvasw).

So we're not proposing to keep barreling ahead — [A](#4921) is already parked awaiting review, and the split isolates the small fix (B) from the architectural pieces precisely so each can be reviewed against agreed behavior. I'm very happy to write the short spec you're describing (guarantees, responsibilities, failure cases, acceptance tests for both directions) so we lock the behavior before resuming the split work (next is B) — and if that spec adjusts the split, then great.

My only ask: **let's keep the momentum**. There's a live protocol bug affecting users while we align, so I'd like to converge on the spec quickly and get the short-term fix (B) moving, with the architectural pieces following behind it.

Do you agree with this direction? Anything above you'd like me to expand or correct?
Thanks again.

HenrikJannsen:
Should we close that or at least mark it as Draft?

rodvar:
Closed superseeded by PRs above doing the split of the work

PR #4890, opened 2026-07-24
chore: sync #4878 to main — Add NETWORK_INFO websocket topic for network state

salottodev:
Manual sync of #4878 from `for-mobile-based-on-2.1.11` to `main` (with conflict resolution).

## Summary by CodeRabbit

* **New Features**
  * Added a `NETWORK_INFO` WebSocket subscription topic.
  * Subscribers can receive real-time network status, including data synchronization, Tor availability, local identity details, and active connections.
  * Connection details include peer address, direction, seed status, and connection time.
  * Updates are delivered automatically as network conditions change.
* **Tests**
  * Added coverage for network information updates, subscription delivery, role changes, and service shutdown.

### DIFF
--- a/api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
+++ b/api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
@@ -1,0 +1,204 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.api.web_socket.domain.network;
+
+import bisq.api.dto.network.ConnectionDto;
+import bisq.api.dto.network.NetworkInfoDto;
+import bisq.api.web_socket.domain.BaseWebSocketService;
+import bisq.api.web_socket.subscription.ModificationType;
+import bisq.api.web_socket.subscription.Subscriber;
+import bisq.api.web_socket.subscription.SubscriberRepository;
+import bisq.api.web_socket.subscription.SubscriptionRequest;
+import bisq.api.web_socket.subscription.Topic;
+import bisq.common.network.Address;
+import bisq.common.network.TransportType;
+import bisq.common.observable.Pin;
+import bisq.common.util.StringUtils;
+import bisq.network.NetworkService;
+import bisq.network.identity.NetworkId;
+import bisq.network.p2p.ServiceNode;
+import bisq.network.p2p.message.EnvelopePayloadMessage;
+import bisq.network.p2p.node.CloseReason;
+import bisq.network.p2p.node.Connection;
+import bisq.network.p2p.node.Node;
+import bisq.network.p2p.services.data.inventory.InventoryService;
+import bisq.network.p2p.services.peer_group.PeerGroupManager;
+import bisq.network.p2p.services.peer_group.PeerGroupService;
+import lombok.extern.slf4j.Slf4j;
+
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.stream.Collectors;
+
+/**
+ * Pushes the local node's network state (peer count, data-sync flag, tor status, own address, key id and
+ * the active connections) to subscribers. The values are read from {@link Node} / {@link ServiceNode} /
+ * {@link PeerGroupService} / {@link InventoryService}, giving clients that reach the node over the API the
+ * same network view an embedded-core node has.
+ */
+@Slf4j
+public class NetworkInfoWebSocketService extends BaseWebSocketService {
+    // Priority order used to pick the primary node whose address, key id and connections we expose.
+    // We expose a single default node; for a node app this is typically the only supported transport,
+    // so the first match wins.
+    private static final List<TransportType> TRANSPORT_PRIORITY = List.of(TransportType.TOR, TransportType.I2P, TransportType.CLEAR);
+
+    private final NetworkService networkService;
+    private final Node.Listener nodeListener;
+    private final Set<Pin> pins = new CopyOnWriteArraySet<>();
+    @Nullable
+    private volatile ServiceNode observedServiceNode;
+
+    public NetworkInfoWebSocketService(SubscriberRepository subscriberRepository,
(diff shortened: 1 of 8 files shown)


api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
```diff
@@ -0,0 +1,204 @@
...
+    @Override
+    public CompletableFuture<Boolean> initialize() {
+        findPrimaryServiceNode().ifPresent(serviceNode -> {
+            observedServiceNode = serviceNode;
+            serviceNode.getDefaultNode().addListener(nodeListener);
+
+            serviceNode.getInventoryService()
+                    .map(InventoryService::getInitialInventoryRequestsCompleted)
+                    .ifPresent(observable -> pins.add(observable.addObserver(completed -> onChange())));
+        });
```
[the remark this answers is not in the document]
salottodev: All the other websocket services in this module return completedFuture(true) unconditionally, none of them returns a failed or incomplete future. They don't need to, because their dependencies always exist (getChannels(), getTrades(), getItems()).

Retrying until findServiceNode(...) becomes available wouldn't help here: ServiceNodesByTransport creates a ServiceNode for every supported transport in its constructor, which runs inside NetworkService's constructor. So the map is already full before any initialize() runs, and it's only cleared on shutdown, it never goes from empty to present.

Since TRANSPORT_PRIORITY covers all three TransportType values, the Optional is only empty when no transports are configured at all. And a node that exists but isn't running yet is already covered: we listen on getDefaultNode() and onStateChange pushes an update once it starts.

api/src/main/java/bisq/api/web_socket/subscription/Topic.java
```diff
@@ -29,4 +29,5 @@ public enum Topic {
     NUM_USER_PROFILES,
     ALERT_NOTIFICATIONS,
     TRADE_RESTRICTING_ALERT,
+    NETWORK_INFO,
```
KimStrand, line 32: Please advertise this topic through the capability registry introduced by #4879 and synced to main in #4887, which is this PR’s base. ApiFeature explicitly reserves "network-info" for when this topic ships, but currently exposes only CLOSED_TRADES. Capability-gated clients will therefore interpret NETWORK_INFO as unsupported. Please add NETWORK_INFO("network-info") to ApiFeature and extend ConfigRestApiTest to verify both the advertised key and its backing Topic.NETWORK_INFO.

api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    private List<ConnectionDto> toConnectionDtos(Node node, @Nullable PeerGroupService peerGroupService) {
+        List<ConnectionDto> connections = new ArrayList<>();
+        node.getAllActiveConnections().forEach(connection ->
+                connections.add(new ConnectionDto(
+                        connection.getId(),
+                        connection.getPeerAddress().getFullAddress(),
+                        connection.isOutboundConnection(),
+                        peerGroupService != null && peerGroupService.isSeed(connection),
```
KimStrand, line 226: The emitted seed value can become stale. Runtime seed additions and removals mutate PeerGroupService’s seed set, but this service only emits updates for node connection, disconnection, state, and inventory events. An active peer can therefore be reclassified without subscribers receiving an update. Please make seed-registry changes observable here, including cleanup during shutdown, or otherwise ensure reclassification triggers onChange().

api/src/main/java/bisq/api/web_socket/subscription/Topic.java
salottodev: https://github.com/bisq-network/bisq2/pull/4901 is merged, so ApiFeature.NETWORK_INFO("network-info") is already on main. I've rebased this PR so its base includes it. This PR adds the backing Topic.NETWORK_INFO. https://github.com/bisq-network/bisq2/pull/4902 extends ConfigRestApiTest to assert both the advertised "network-info" key and its backing Topic.NETWORK_INFO. I'll sync it to main right after this one is merged.

api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
salottodev: Thanks, you're right that a reclassification won't reach subscribers on its own. I looked into wiring this up and couldn't find a way to do it as things stand.

**There's no seed-change hook to subscribe to.** `PeerGroupService.seedNodeAddresses` ([`PeerGroupService.java:109`]) is a plain `CopyOnWriteArraySet<Address>`, and nothing along the mutation chain emits an event:

`AuthorizedBondedRolesService` → `NetworkService.addSeedNodeAddressByTransport` → `ServiceNodesByTransport` → `ServiceNode` → `PeerGroupManager` → `PeerGroupService`

The only `Observable` in the `peer_group` package is `PeerGroupManager.state`, which is unrelated.

**I checked how Bisq 2 desktop handles it, and it doesn't.** `ConnectionListItem` ([`ConnectionListItem.java:74-81`]) resolves `peerGroupService.isSeed(peerAddress)` once in the constructor into a `final String peer` field — not a property, never re-evaluated. A row keeps its original "Seed node"/"Gossip node" label for the connection's whole lifetime. So there's no established pattern here I could follow.

What would you recommend?

KimStrand: Could you please check AuthorizedBondedRolesService.Listener? It seems that its callbacks are invoked after NetworkService updates the seed set, and SubscriptionService can access it through bondedRolesService.getAuthorizedBondedRolesService(). It looks like NetworkInfoWebSocketService could register there in initialize(), call onChange() for seed-role additions and removals, and unregister in shutdown().

rodvar: @KimStrand please let's try to do the modifications on the original `for-mobile` PRs, the chore/sync-to-main PRs are ment to be simply to sync after the changes were accepted.

Otherwise this breaks the for-mobile --> main flow and we need to be careful to not loose stuff 🙏 

Salotto will make sure we don't loose this review changes in this case 🙏

api/src/test/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketServiceTest.java
```diff
@@ -0,0 +1,100 @@
...
+        WebSocket webSocket = subscribe(subscriberRepository);
+
+        // Authorized data which is not a seed node role must not trigger an update
+        listener.onAuthorizedDataAdded(authorizedBondedRoleData(BondedRoleType.MEDIATOR));
+        verify(webSocket, never()).send(anyString());
+
+        // Adding and removing a seed node role reclassifies existing connections, so both must trigger an update
+        listener.onAuthorizedDataAdded(authorizedBondedRoleData(BondedRoleType.SEED_NODE));
+        listener.onAuthorizedDataRemoved(authorizedBondedRoleData(BondedRoleType.SEED_NODE));
+        verify(webSocket, times(2)).send(anyString());
```
KimStrand: This test currently fails because Subscriber.send() dispatches through its executor, so the immediate times(2) verification can run before both WebSocket sends complete. Please use verify(webSocket, timeout(1000).times(2)).send(anyString()), as already used in SubscriptionServiceTest. Alternatively, mock SubscriberRepository to return a mocked Subscriber and verify subscriber.send(...) twice, keeping the test synchronous and avoiding a timeout.

KimStrand (commented):
utACK

PR #4893, opened 2026-07-24
feature/telebirr_ethiopian_payment_support

rodvar:
Screenshot 2026-07-28 at 11 16 58
Screenshot 2026-07-28 at 11 14 01
Screenshot 2026-07-28 at 11 13 41

### Offer in current Bisq 2.1.11 Desktop is shown as custom method 👍 
Screenshot 2026-07-28 at 11 15 13

 - contribs to https://github.com/bisq-network/bisq-mobile/issues/1487
 - added Telebirr payment method to bisq2 traditional payment rail to upport Ethiopian Market with most used digital payment method
 - added telebirr asset and i18n
 - (saved) account creation not supported for now

## Summary

* **New Features**
  * Added Telebirr (ETB) as a new Ethiopian fiat payment rail, including localized labels and an image, with 24-hour trade duration and low chargeback risk.
* **Bug Fixes**
  * Updated fiat payment method listings (desktop and API) to only show rails supported by classic account creation, avoiding unsupported/ghost selections.
  * MuSig screens now use the standard (account-backed) payment-method set and ignore stale cookie selections.
* **Limitations**
  * Telebirr remains Bisq Easy–only and is not implemented in classic account/payment-method detail views.
* **Tests**
  * Added regression tests for filtering rules and the fiat payment-methods endpoint response.


rodvar:
@KimStrand

KimStrand (commented):
Telebirr is a valid Bisq Easy rail but not an account-backed rail. Currently, account consumers use the unrestricted rail list: this causes the payment-account REST endpoint to return HTTP 500 and exposes an unusable option in MuSig.

We could centralize the distinction, e.g. supportsStandardPaymentAccount(), and use it in the REST endpoint, MuSig picker, and account-creation UI, while keeping Telebirr available in Bisq Easy:

eg:

public static boolean supportsStandardAccountCreation(FiatPaymentRail rail) {
      return switch (rail) {
          case CASH_APP, CUSTOM, TELE_BIRR -> false;
          default -> true;
      };
  }

@HenrikJannsen any thoughts about the suggestion?

rodvar:
> Telebirr is a valid Bisq Easy rail but not an account-backed rail. Currently, account consumers use the unrestricted rail list: this causes the payment-account REST endpoint to return HTTP 500 and exposes an unusable option in MuSig.
> 
> We could centralize the distinction, e.g. supportsStandardPaymentAccount(), and use it in the REST endpoint, MuSig picker, and account-creation UI, while keeping Telebirr available in Bisq Easy:
> 
> eg:
> 
> public static boolean supportsStandardAccountCreation(FiatPaymentRail rail) { return switch (rail) { case CASH_APP, CUSTOM, TELE_BIRR -> false; default -> true; }; }
> 
> @HenrikJannsen any thoughts about the suggestion?

I agree with this Kim, also I confirm it addresses the REST 500 + MuSig, and note the `getPaymentMethods` split so Bisq Easy availability is preserved.
Predicate is low-risk and strictly better, so I'd implement it and let Henrik bless naming/placement rather than wait blocked.

KimStrand (commented):
The REST fix and centralized account-capability distinction look good. One issue remains in the actual MuSig create-offer flow: `MuSigCreateOfferPaymentController.onActivate()` still populates payment methods
  through `PaymentMethodUtil.getPaymentMethods()`. For fiat currencies this uses the Bisq Easy list, so `TELE_BIRR` is still displayed for ETB.

Because Telebirr has no standard payment account, selecting it leads to the no-account/create-account flow, where Telebirr is unavailable.

Full MuSig support would require a proper Telebirr account

rodvar:
> The REST fix and centralized account-capability distinction look good. One issue remains in the actual MuSig create-offer flow: `MuSigCreateOfferPaymentController.onActivate()` still populates payment methods through `PaymentMethodUtil.getPaymentMethods()`. For fiat currencies this uses the Bisq Easy list, so `TELE_BIRR` is still displayed for ETB.
> 
> Because Telebirr has no standard payment account, selecting it leads to the no-account/create-account flow, where Telebirr is unavailable.
> 
> Full MuSig support would require a proper Telebirr account

thank you @KimStrand  - I thought this wouldn't be a problem because MuSig is a feature-flagged, isn't it?

Just in case, to avoid any issues or concerns on this PR, let me push one more change to make sure Telebirr is skipped for MuSig for now (and we can revisit once MuSig is getting closer to the finish line)

rodvar:
done @KimStrand  - added a test also to clearly prove the change 👍

KimStrand (commented):
utACK

PR #4894, opened 2026-07-25
Add connection metrics to NETWORK_INFO payload

salottodev:
## Summary by CodeRabbit

* **New Features**
  * Network information now includes per-connection performance metrics for active connections.
  * When available, round-trip time (RTT) is shown; otherwise it’s omitted until measured.
  * Connection totals are added, including sent/received bytes and sent/received message counts.

### DIFF
--- a/api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
+++ b/api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
@@ -17,8 +17,8 @@
 
 package bisq.api.web_socket.domain.network;
 
+import bisq.api.dto.mappings.network.ConnectionMetricsDtoMapping;
 import bisq.api.dto.network.ConnectionDto;
-import bisq.api.dto.network.ConnectionMetricsDto;
 import bisq.api.dto.network.NetworkInfoDto;
 import bisq.api.web_socket.domain.BaseWebSocketService;
 import bisq.api.web_socket.subscription.ModificationType;
@@ -37,7 +37,6 @@ import bisq.network.p2p.message.EnvelopePayloadMessage;
 import bisq.network.p2p.node.CloseReason;
 import bisq.network.p2p.node.Connection;
 import bisq.network.p2p.node.Node;
-import bisq.network.p2p.node.network_load.ConnectionMetrics;
 import bisq.network.p2p.services.data.inventory.InventoryService;
 import bisq.network.p2p.services.peer_group.PeerGroupManager;
 import bisq.network.p2p.services.peer_group.PeerGroupService;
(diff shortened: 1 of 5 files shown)


api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
```diff
@@ -190,10 +192,23 @@ private List<ConnectionDto> toConnectionDtos(Node node, @Nullable PeerGroupServi
...
         return connections;
     }
 
+    private ConnectionMetricsDto toConnectionMetricsDto(ConnectionMetrics metrics) {
+        // averageRtt is 0 until a round-trip is measured (handshake / request-response); treat that as "unmeasured".
+        // Null-check the raw average, not the rounded value: a measured sub-millisecond rtt still rounds to 0.
+        double rttRaw = metrics.getAverageRtt();
+        return new ConnectionMetricsDto(
+                rttRaw > 0 ? Math.round(rttRaw) : null,
+                metrics.getSentBytes(),
```
rodvar: heads up: these getters (getSentBytes / getReceivedBytes / getNumMessagesSent/Received) sum over plain java.util.TreeMaps in ConnectionMetrics that get structurally mutated (computeIfAbsent adds a new per-minute bucket) on the connection i/o thread - worth making those per-minute maps thread-safe (or snapshotting under a lock) in ConnectionMetrics. getAverageRtt is fine, rttList is a CopyOnWriteArrayList 🙏

api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
```diff
@@ -190,10 +192,23 @@ private List<ConnectionDto> toConnectionDtos(Node node, @Nullable PeerGroupServi
                         connection.getPeerAddress().getFullAddress(),
                         connection.isOutboundConnection(),
                         peerGroupService != null && peerGroupService.isSeed(connection),
-                        connection.getCreated())));
+                        connection.getCreated(),
+                        toConnectionMetricsDto(connection.getConnectionMetrics()))));
         return connections;
     }
 
+    private ConnectionMetricsDto toConnectionMetricsDto(ConnectionMetrics metrics) {
```
HenrikJannsen: We use mapper classes normally... Why not here as well?

HenrikJannsen (changes requested):

rodvar:
@salottodev please check the review comments above - we can't release network feature backend until this gets merged 🙏

KimStrand (commented):
-

api/src/main/java/bisq/api/web_socket/domain/network/NetworkInfoWebSocketService.java
```diff
@@ -190,9 +191,10 @@ private List<ConnectionDto> toConnectionDtos(Node node, @Nullable PeerGroupServi
                         connection.getPeerAddress().getFullAddress(),
                         connection.isOutboundConnection(),
                         peerGroupService != null && peerGroupService.isSeed(connection),
-                        connection.getCreated())));
+                        connection.getCreated(),
+                        ConnectionMetricsDtoMapping.fromBisq2Model(connection.getConnectionMetrics()))));
```
KimStrand, line 195: Will existing NETWORK_INFO subscribers receive updates when only these metrics change? The payload seems to refresh for connection, node-state, and inventory events, but not for traffic, so these values may remain unchanged until another event triggers a snapshot.

A new subscription or unrelated network-state change would receive fresh values; the issue is updates during an existing stable connection.

salottodev: Good catch, you're right, that's the gap. Let me explain the reasoning.

We deliberately didn't hook metrics into the update path. onChange() sends a full REPLACE snapshot of the entire connection list, and connection metrics change on every message sent or received. Pushing on metric changes would turn every single P2P message into a full-topic broadcast to every subscriber — traffic amplification proportional to node activity, which we especially don't want over Tor.

I discussed this with @rodvar and we landed on the current behaviour intentionally: for the mobile client this data isn't important enough to justify a continuous push over Tor. It's diagnostic detail, not something the app acts on, so we took the trade-off of values that lag in exchange for not pushing a lot of data over the websocket.

What do you think?

KimStrand: Makes sense too me

KimStrand:
utACK

PR #4899, opened 2026-07-28
Replace socket based instance lock with a file lock

wodoro:
similar to instance lock from https://github.com/bisq-network/bisq/pull/7924

The instance lock derived a port from the data directory hash and bound a ServerSocket to it. That has 3 defects:

- The port range (49152-65535) overlaps the Linux ephemeral port range (32768-60999 by default), so an unrelated outbound connection can hold the port and Bisq refuses to start.
- Different data directories can hash to the same port, which also gives a false positive.
- ServerSocket binds to all interfaces, so we expose a listening port to the network for no reason. `setReuseAddress(false)` after bind is a no-op.

We now take an OS level lock on an instance.lock file inside the data directory. The OS releases it when the process exits, including at a crash or kill, and it is scoped to exactly the resource we protect. The owner PID is written to the file for diagnostics.

If another instance holds the lock we throw `AnotherInstanceRunningException`, which Executable turns into a log entry and a message on stderr. The desktop app overrides the handler and shows an AWT dialog, as JavaFX is not launched at that point. If the locking mechanism itself is unusable we fail open and continue without lock, so that an unwritable lock file cannot brick startup.

The check moved after the Res setup so the message can be localized. It still runs before any service touches the data directory.

## Summary by CodeRabbit

* **New Features**
  * Added OS-level data directory locking (`instance.lock`) to prevent multiple concurrent instances.
  * “Already running” messaging now includes an optional owning process ID; desktop may show a warning dialog (when available).
  * Improved instance-lock failure popups with guidance and the `checkInstanceLock=false` override.
  * Backups exclude the instance lock file; directory copy can now skip specific filenames.
* **Bug Fixes**
  * Startup now fails fast if exclusive lock verification can’t be completed (fail-closed).
* **Tests**
  * Strengthened lock ownership (PID) reporting tests and added filename-skip coverage for directory copying.


application/src/main/java/bisq/application/InstanceLock.java
```diff
@@ -0,0 +1,155 @@
...
+ * <p>
+ * We keep strong references to the {@link FileChannel} and the {@link FileLock}. If they got
+ * garbage collected the lock would be released silently.
+ */
+@Slf4j
+public class InstanceLock implements AutoCloseable {
+    private static final String LOCK_FILE_NAME = "instance.lock";
+    // We lock a region beyond any content of the lock file instead of the whole file. On Windows file locks are
+    // mandatory, thus a lock on the region which holds the PID would prevent any other process from reading the
+    // file. That would break the PID diagnostics as well as any code which copies the data directory, like the
```
KimStrand: The manual backup currently copies `instance.lock` via `ResourcesController.onBackup()` → `FileMutatorUtils.copyDirectory()`.

Java documents that closing another channel to the same file may release all locks held by the JVM:
  https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/channels/FileLock.html

This means using the in-app Backup feature disables single-instance protection for the rest of that session.

application/src/main/java/bisq/application/ApplicationService.java
```diff
@@ -264,22 +268,21 @@ protected void setupLogging(Path appDataDirPath) {
...
-        }));
+        Path appDataDirPath = config.getAppDataDirPath();
+        instanceLock = new InstanceLock(appDataDirPath);
+        try {
+            if (!instanceLock.tryLock()) {
+                throw new AnotherInstanceRunningException(config.getAppName(),
+                        appDataDirPath,
+                        instanceLock.readOwnerPid());
+            }
+        } catch (IOException e) {
```
KimStrand, line 281: Do we really want to change the previous fail-closed behavior here?

Before this PR, a failure to acquire the instance lock aborted startup. With this catch, an inaccessible lock file or unsupported filesystem allows the application to continue without protection, potentially letting two instances access the same persisted and wallet state.

application/src/main/java/bisq/application/InstanceLock.java
wodoro: Added file names to skip instance.lock to avoid this issue

application/src/main/java/bisq/application/ApplicationService.java
wodoro: The reason was that on some envs where the data is on network storage, file lock may have failed, making the app unusable

but the app already had a solution as config to ignore it, so I made it fail-closed again.

wodoro:
Thanks for the review @KimStrand , applied fixes

KimStrand (commented):
utACK

PR #4903, opened 2026-07-31
Add permessage-deflate to the WebSocket API, and harden the WebSocket-REST proxy

wodoro:
as parts of improvements related to https://github.com/bisq-network/bisq-mobile/issues/1558

WebSocket compression                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                    
Adds RFC 7692 permessage-deflate support to the Grizzly WebSocket stack, which does not implement it and in fact rejects any frame with a reserved bit set. PerMessageDeflateFilter is installed directly below the WebSocketFilter and works on the wire bytes only — inbound it inflates and hands plain frames upwards, outbound it deflates what the WebSocketFilter just serialized — so everything above it sees an uncompressed connection.                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                    
Controlled by a new api.server.websocketCompressionEnabled key, defaulting to on. Compression is only ever used if the client offers it; an offer with parameters we cannot honour is declined and the connection stays uncompressed.                                                                             
                                                                                                                                                                                                                                                                                                                    
Notable properties:                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                    
  - Always negotiated with client_no_context_takeover / server_no_context_takeover, with a fresh Deflater/Inflater per message. Carrying the window across messages would let a length observer probe secrets from an earlier message (CRIME); resetting per message means each message's compressed size depends only on itself.                                                                                                                                                                                                                                                                                                   
  - Inbound compressed input and inflated output are both bounded, so a peer cannot use a zip bomb to drive allocation.                                                                                                                                                                                             
  - Outbound fragments are never compressed. Several client stacks fail to inflate a compressed message split across frames, so a fragment goes out uncompressed instead.                                                                                                                                           
  - A frame we cannot parse is forwarded as-is, so enabling compression can never fail a send that would otherwise have succeeded.                                                                                                                                                                                  
  - Protocol errors close with an RFC 6455 close frame (1002, or 1009 for size limits) rather than dropping the socket silently.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added optional WebSocket message compression, negotiated automatically when supported.
  * Improved WebSocket-to-REST forwarding with safer path validation and IPv6 support.
* **Security**
  * Redacted session credentials from logs and blocked client-supplied identity headers from forwarding.
* **Bug Fixes**
  * Improved handling of fragmented, oversized, malformed, and unsupported WebSocket messages.
* **Tests**
  * Added comprehensive coverage for compression, forwarding validation, credential protection, and protocol errors.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


api/src/main/java/bisq/api/web_socket/rest_api_proxy/WebSocketRestApiService.java
```diff
@@ -145,6 +172,73 @@ private CompletableFuture<WebSocketRestApiResponse> sendToRestApiServer(WebSocke
...
+     */
+    /**
+     * Drops the headers the client sent, so that nothing downstream can read them. Clients still send
+     * their session and client id here, which the node used to trust; the identity now comes from the
+     * authenticated upgrade request and what the message carries is discarded before anything sees it.
+     */
+    static void stripClientSuppliedHeaders(WebSocketRestApiRequest request) {
+        request.takeHeaders();
+    }
+
```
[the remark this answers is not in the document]
wodoro: fixed in [a622793](https://github.com/bisq-network/bisq2/pull/4903/commits/a6227937690dedeccfee77fa2771089c162e8385)

rodvar:
@KimStrand

api/src/main/java/bisq/api/web_socket/compression/PerMessageDeflateFilter.java
```diff
@@ -0,0 +1,662 @@
...
+            // The handshake response is the only point where we know the upgrade actually succeeded,
+            // so compression is switched on exactly when we confirm the extension to the client.
+            if (!state.enabled
+                    && httpContent.getHttpHeader() instanceof HttpResponsePacket response
+                    && response.getStatus() == HttpStatus.SWITCHING_PROTOCOLS_101.getStatusCode()) {
+                // Replaces rather than appends: Grizzly may have written the extensions a
+                // WebSocketApplication declares, but it only echoes their names and cannot apply any of
+                // them, so permessage-deflate is the only one this response can honestly confirm.
+                response.setHeader(SEC_WEBSOCKET_EXTENSIONS, NEGOTIATED_EXTENSION);
+                state.enabled = true;
```
rodvar, line 167: question - if the first upgrade request on a connection offers deflate but the upgrade itself fails (WebSocketFilter answers non-101), the state stays attached to the connection. If the client then retries the upgrade on the same connection *without* offering the extension, `onHandshakeRequest` early-returns because state is already present, and this branch would confirm permessage-deflate on the new 101 even though that request never offered it. Probably hard to hit (does grizzly even keep the connection alive after a failed upgrade?), but clearing the state when a non-101 response to a handshake goes out would close it cheaply.

api/src/main/java/bisq/api/web_socket/rest_api_proxy/WebSocketRestApiService.java
```diff
@@ -145,6 +172,64 @@ private CompletableFuture<WebSocketRestApiResponse> sendToRestApiServer(WebSocke
...
+        // required, whereas a forwarded request has to be safe regardless of that setting.
+        URI_VALIDATOR.validate(uri);
+        return uri;
+    }
+
+    /**
+     * Fails rather than forwarding the request without an identity: this is the only place the identity
+     * is established, so a connection we cannot resolve it for must not reach the REST API at all.
+     */
+    private static void forwardAuthenticatedIdentity(WebSocket webSocket, HttpRequest.Builder requestBuilder) {
```
rodvar, line 216: nit - the javadoc says it fails rather than forwarding without an identity, but it only throws when the upgrade request can't be resolved..? - if the upgrade request simply has no session/client headers it forwards happily with none. Either soften the doc or also check the headers, whichever matches the intent.

api/src/main/java/bisq/api/web_socket/WebSocketConnectionHandler.java
```diff
@@ -85,14 +86,16 @@ public void onClose(WebSocket webSocket, DataFrame frame) {
 
     @Override
     public void onMessage(WebSocket webSocket, String message) {
-        log.info("Received message {}", message);
+        // Redacted: a client released before the node took the identity from the handshake puts its
+        // session id into the payload, and this line would otherwise write it to disk for every message.
+        log.info("Received message {}", JsonUtil.redactCredentials(message));
```
rodvar: nit - this logs and regex-redacts every single message at info level. Now that there's a regex on the hot path anyway, maybe worth dropping to debug so production doesn't pay it per message?

api/src/test/java/bisq/api/web_socket/rest_api_proxy/WebSocketRestApiServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Test
+    void keepsTheQuery() throws Exception {
+        assertThat(WebSocketRestApiService.resolveRestApiUri(REST_SERVER_URL, "/api/v1/offers?market=BTC-EUR"))
+                .hasToString("https://127.0.0.1:8090/api/v1/offers?market=BTC-EUR");
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {
```
rodvar, line 44: nit - could you add a literal traversal case like `/api/v1/../../doc/v1/`? `URI.normalize()` resolves it to `/doc/v1/` with the host unchanged, so it sails past the host comparison and whether it's rejected rests entirely on `URI_VALIDATOR` - would be good to pin that with a test next to the encoded `%2e%2e` one either way.

api/src/main/java/bisq/api/web_socket/compression/PerMessageDeflateFilter.java
```diff
@@ -0,0 +1,662 @@
...
+            payloadLength = 0;
+            for (int i = 0; i < 8; i++) {
+                payloadLength = payloadLength << 8 | buffer.get(cursor + i) & 0xFFL;
+            }
+            cursor += 8;
+            if (payloadLength >= 0 && payloadLength <= 0xFFFF) {
+                throw new ProtocolError("Payload length is not minimally encoded");
+            }
+        }
+        if (payloadLength < 0 || payloadLength > MAX_FRAME_PAYLOAD_SIZE) {
```
rodvar: nit - a 64-bit length with the MSB set is a protocol violation per RFC 6455 (bit must be 0), but here it lands in MessageTooLarge -> close 1009; 1002 would be the honest close code for that one. Total nitpick, feel free to ignore.

api/src/main/java/bisq/api/web_socket/rest_api_proxy/WebSocketRestApiRequest.java
```diff
@@ -39,24 +40,32 @@ public class WebSocketRestApiRequest implements WebSocketMessage {
...
+     * further — not into a forwarded request, and not into a log line via {@link #toString()}. The node
+     * takes the identity of the call from the authenticated upgrade request instead, so what the
+     * message carries is discarded. Called once, right after deserialization.
+     */
+    void clearHeaders() {
+        // Jackson deserializes into the field, so emptying it is what actually removes them. Emptied
+        // rather than nulled: the generated getter stays safe to call for anything added later.
+        headers = Map.of();
+    }
 }
```
rodvar: nit - missing newline at end of file

apps/api-app/src/main/resources/api_app.conf
```diff
@@ -27,6 +27,8 @@ application {
         server={
             restEnabled=false
             websocketEnabled=true
+            # RFC 7692 permessage-deflate. Negotiated only if the client offers it.
+            websocketCompressionEnabled=true
```
rodvar, line 31: 👏 well done adding in all relevant files

rodvar (commented):
on a first code-only pass only found nits - looking really good wod 💪 
Also added a question, please review
Hope this helps 🙏

api/src/main/java/bisq/api/web_socket/compression/PerMessageDeflateFilter.java
wodoro: Grizzly closes after every failed upgrade in any test I tried to produce the bug, so the fix is defensive rather than a bug fix. but I still added the guard, just in case if it somehow did not.

api/src/main/java/bisq/api/web_socket/rest_api_proxy/WebSocketRestApiService.java
wodoro: updated the docs

wodoro:
Thank you @rodvar for the review, applied the requested changes

edit: just saw the CRAI comments now, addressing

api/src/main/java/bisq/api/web_socket/compression/PerMessageDeflateFilter.java
```diff
@@ -0,0 +1,690 @@
...
+            String name = (separator < 0 ? parameter : parameter.substring(0, separator))
+                    .trim().toLowerCase(Locale.ROOT);
+            Optional<String> value = separator < 0
+                    ? Optional.empty()
+                    : Optional.of(unquote(parameter.substring(separator + 1).trim()));
+            switch (name) {
+                case "client_no_context_takeover", "server_no_context_takeover" -> {
+                    // We apply both unconditionally
+                }
+                case "client_max_window_bits", "server_max_window_bits" -> {
```
KimStrand: client_max_window_bits should not be handled like server_max_window_bits. A client value below 15 is a hint about the client’s compressor and may be ignored by the server; a 15-bit Inflater can decode streams produced with a smaller window. This currently declines compression negotiation for valid offers. Should we handle the parameters directionally? reject unsupported server window sizes, but validate and accept or ignore client values from 8 through 15.

wodoro: correct, pushed a fix, thanks for the review

KimStrand (commented):
utACK

PR #4914, opened 2026-08-10
feature/new_permissions_support_for_node_and_connect_upgrades

rodvar:
- upgrades support for permissions schema for endpoints
 - without this we can't release any new features requiring new permissions so this is very importante for mobile

## Summary by CodeRabbit

* **New Features**
  * Added support for granting all eligible permissions while preserving explicit selections.
  * Improved permission persistence and compatibility across versions.
  * Authorization errors now return structured 403 responses identifying the required permission.
  * QR code updates now replace existing files more safely.

* **Bug Fixes**
  * Prevented interrupted QR code writes from leaving invalid files.
  * Excluded sensitive permissions from automatic grants.

* **Documentation**
  * Clarified REST permission mapping and security considerations.

* **Tests**
  * Added coverage for permissions, authorization, persistence compatibility, and QR code replacement.

### DIFF
--- a/api/src/main/java/bisq/api/access/permissions/Permission.java
+++ b/api/src/main/java/bisq/api/access/permissions/Permission.java
@@ -20,27 +20,87 @@ package bisq.api.access.permissions;
 import bisq.common.proto.ProtoEnum;
 import lombok.Getter;
 
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.LinkedHashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+
 /**
  * The id must not be changed as it is used for the serialisation.
+ * <p>
+ * {@code autoGrantable} encodes the standard-vs-sensitive distinction: a grantAll grant expands
+ * only to auto-grantable permissions (see {@link PermissionSet}), so already-paired clients gain
+ * future NORMAL features automatically but can never silently acquire a security-sensitive
+ * capability.
+ * <p>
+ * SECURE BY DEFAULT: the single-argument constructor marks a permission SENSITIVE
+ * ({@code autoGrantable = false}). A new permission is therefore excluded from grantAll unless it
+ * explicitly opts in via {@code STANDARD}, making "is this safe to auto-grant?" a conscious
+ * decision at declaration time. Forgetting to decide fails closed — the permission requires an
+ * explicit per-device grant rather than silently joining grantAll.
  */
 public enum Permission implements ProtoEnum {
-    TRADE_CHAT_CHANNELS(0),
-    EXPLORER(1),
-    MARKET_PRICE(2),
-    OFFERBOOK(3),
-    PAYMENT_ACCOUNTS(4),
-    REPUTATION(5),
-    SETTINGS(6),
-    TRADES(7),
-    USER_IDENTITIES(8),
-    USER_PROFILES(9),
-    MOBILE_DEVICES(10);
+    TRADE_CHAT_CHANNELS(0, Kind.STANDARD),
+    EXPLORER(1, Kind.STANDARD),
+    MARKET_PRICE(2, Kind.STANDARD),
+    OFFERBOOK(3, Kind.STANDARD),
+    PAYMENT_ACCOUNTS(4, Kind.STANDARD),
+    REPUTATION(5, Kind.STANDARD),
+    SETTINGS(6, Kind.STANDARD),
+    TRADES(7, Kind.STANDARD),
+    USER_IDENTITIES(8, Kind.STANDARD),
+    USER_PROFILES(9, Kind.STANDARD),
+    MOBILE_DEVICES(10, Kind.STANDARD);
+
+    /** Grant classification. Named so a declaration reads as an explicit security decision. */
+    public enum Kind {
+        /** Auto-grantable: covered by grantAll, gained automatically by already-paired clients. */
+        STANDARD,
+        /** Security-sensitive: never covered by grantAll, always needs an explicit per-device grant. */
+        SENSITIVE
+    }
 
     @Getter
     private final int id;
+    private final Kind kind;
 
+    // Sensitive by default: an id-only declaration is treated as SENSITIVE so a forgotten
+    // classification fails closed. Standard permissions must opt in explicitly with Kind.STANDARD.
     Permission(int id) {
+        this(id, Kind.SENSITIVE);
+    }
+
+    Permission(int id, Kind kind) {
         this.id = id;
+        this.kind = kind;
+    }
+
+    public boolean isAutoGrantable() {
+        return kind == Kind.STANDARD;
+    }
+
+    /**
+     * All permissions a grantAll grant expands to — the "standard" set. Sensitive permissions
+     * (autoGrantable = false) are excluded by construction and must be granted explicitly.
+     * <p>
+     * Ordered by {@code id} into an insertion-ordered set: this feeds the grantAll expansion
+     * serialized by {@code PermissionSet.getBuilder}, and an unspecified iteration order (as a
+     * plain immutable Set has) would make the proto's repeated field non-deterministic across
+     * runs, breaking serializeForHash stability. Set semantics are unchanged — equality stays
+     * order-independent, so promotion/folding comparisons are unaffected.
+     */
+    public static Set<Permission> autoGrantable() {
+        // Unmodifiable AND insertion-ordered: PermissionSet caches this in a static field and
+        // returns it directly from getPermissions(), so a mutable set would let any caller alter
+        // the grantAll expansion for every client (and, once sensitive permissions exist, add
+        // one). LinkedHashSet keeps the id order needed for serializeForHash determinism.
+        LinkedHashSet<Permission> ordered = Arrays.stream(values())
+                .filter(Permission::isAutoGrantable)
+                .sorted(Comparator.comparingInt(Permission::getId))
+                .collect(Collectors.toCollection(LinkedHashSet::new));
+        return Collections.unmodifiableSet(ordered);
     }
 
     @Override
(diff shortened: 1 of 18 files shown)


rodvar:
@KimStrand

api/src/main/java/bisq/api/access/persistence/ApiAccessStore.java
```diff
@@ -71,13 +74,37 @@ public static ApiAccessStore fromProto(bisq.api.protobuf.ApiAccessStore proto) {
...
                 .collect(Collectors.toMap(Map.Entry::getKey,
                         e -> ClientProfile.fromProto(e.getValue())));
-        Map<String, Set<Permission>> permissionsByClientId = proto.getPermissionsByClientIdMap().entrySet().stream()
+        boolean isLegacyStore = proto.getPermissionsSchemaVersion() < PERMISSIONS_SCHEMA_VERSION;
+        Map<String, PermissionSet> permissionsByClientId = proto.getPermissionsByClientIdMap().entrySet().stream()
                 .collect(Collectors.toMap(Map.Entry::getKey,
-                        e -> PermissionSet.fromProto(e.getValue()).getPermissions()));
+                        e -> {
+                            PermissionSet permissionSet = PermissionSet.fromProto(e.getValue());
+                            return isLegacyStore ? migrateLegacyEntry(e.getKey(), permissionSet) : permissionSet;
```
KimStrand: A version-0 entry is not sufficient evidence that it represents grant-all. The existing pairing API can persist permission subsets. Additionally, if a version-1 explicit subset is rewritten by a pre-PR node, that node drops the new schema fields; after re-upgrading, this line treats it as legacy and promotes it to grantAll. That silently turns restricted access into full API access. Please preserve ambiguous unversioned subsets or introduce a marker that survives downgrade, and add a new → old → new persistence test.

bisq2-startos
```diff
@@ -0,0 +1 @@
+../bisq2-startos-fork
```
KimStrand: not needed I think :)

KimStrand (commented):
bisq2

bisq2-startos
rodvar: sorry, I'll replace it with the symlink gitignore change if that's ok (same as the umbrel one)

api/src/main/java/bisq/api/access/persistence/ApiAccessStore.java
rodvar: great catch Kim, thank you.

You're right on both. I'm changing the approach to be evidence-based instead of stamp-based: on load we only promote an entry to grantAll if it already contains every permission of the running version.

I'm gonna also reply a better structured 403 error so the mobiel side can react better for this cases offering the user to repair to get the benefits of a node upgrade.

+ I'll add the requested tests -thanks!

api/src/main/java/bisq/api/access/persistence/ApiAccessStore.java
```diff
@@ -71,13 +77,39 @@ public static ApiAccessStore fromProto(bisq.api.protobuf.ApiAccessStore proto) {
...
+     * re-read by a binary that has since gained permissions is NOT promoted and that client
+     * keeps the old set (fails closed; re-pairing restores full access).
+     */
+    private static PermissionSet promoteIfFullGrant(String clientId, PermissionSet permissionSet) {
+        if (permissionSet.isGrantAll()) {
+            return permissionSet;
+        }
+        if (permissionSet.getPermissions().containsAll(EnumSet.allOf(Permission.class))) {
+            log.info("Promoting full permission set of client {} to grantAll", clientId);
+            return PermissionSet.grantAll();
```
KimStrand, line 144: Should we persist this promotion as part of the migration? `fromProto()` currently changes the entry to grantAll only in memory. `readPersisted()` applies that value but does not write it back, and ApiAccessStoreService does not persist after loading.

In the version 1 (5 permissions) → version 2 (this PR, still 5) → version 3 (adds new permission) rollout, version 2 recognizes 5 of 5 and promotes the client in memory, but the disk can remain the original five-permission list. Version 3 then restarts, reads 5 of 6, keeps the client restricted, and requires re-pairing.

Should we persist the store when an entry is promoted and add a test proving that the migrated grantAll marker survives a restart.

KimStrand (commented):
I think the usability goal is good, but grantAll should ideally mean “all standard permissions” rather than every Permission value added forever.

Having all permissions today is reasonable evidence that a client should receive future normal Bisq Connect features, but it should not automatically grant future security-sensitive capabilities.

I recommend encoding that distinction in the permission model now, for example, auto-grantable versus sensitive, and expanding grantAll only to auto-grantable permissions. Sensitive permissions should always require explicit per-device approval.

api/src/main/java/bisq/api/access/persistence/ApiAccessStore.java
rodvar: thanks Kim I'm already working on both fixes incoming soon!

api/src/main/java/bisq/api/access/permissions/Permission.java
```diff
@@ -38,9 +50,33 @@ public enum Permission implements ProtoEnum {
 
     @Getter
     private final int id;
+    @Getter
+    private final boolean autoGrantable;
 
     Permission(int id) {
```
KimStrand, line 71: Could we avoid defaulting new permissions to auto-grantable? A future sensitive permission added using the same one-argument constructor as every existing constant would silently join grantAll, and everyCurrentPermissionIsAutoGrantable would still pass.

That makes adding a permission an explicit security decision.

api/src/test/java/bisq/api/access/persistence/ApiAccessStoreServiceTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        // Legacy-shaped entry with the full standard set: loading promotes it to grantAll.
+        bisq.api.protobuf.PermissionSet legacyFullGrant = bisq.api.protobuf.PermissionSet.newBuilder()
+                .addAllPermissions(Permission.autoGrantable().stream().map(Permission::toProtoEnum).toList())
+                .build();
+        ApiAccessStore promotedOnLoad = ApiAccessStore.fromProto(bisq.api.protobuf.ApiAccessStore.newBuilder()
+                .putPermissionsByClientId("legacy-client", legacyFullGrant)
+                .build());
+
+        service.onPersistedApplied(promotedOnLoad);
```
KimStrand: The production write-back fixes the reported migration issue, thanks. However, this test calls onPersistedApplied() directly without first applying promotedOnLoad to the service's persistable store. persist() therefore snapshots the empty default store, so the test passes even though the migrated entry is not present in the written value.

Please exercise readPersisted(), or apply the loaded store first, capture the ApiAccessStore passed to persistAsync(), reload it, and assert that legacy-client is still grantAll. Also please stub getStorePath() on the persistence mock; :api:test currently finishes with two shutdown-hook NullPointerExceptions from these test instances.

api/src/main/proto/api.proto
```diff
@@ -38,6 +38,10 @@ enum Permission {
 
 message PermissionSet {
   repeated Permission permissions = 1;
+  // When true the client is granted every permission the node knows, including ones added
+  // after pairing. The expanded list is still written to `permissions` so nodes running a
+  // version without this field keep working after a downgrade.
+  bool grantAll = 2;
```
KimStrand, line 44: grantAllAutoGrantablePermissions instead of grantAll?

api/src/main/java/bisq/api/access/permissions/Permission.java
```diff
@@ -38,9 +50,33 @@ public enum Permission implements ProtoEnum {
...
+     * All permissions a grantAll grant expands to — the "standard" set. Sensitive permissions
+     * (autoGrantable = false) are excluded by construction and must be granted explicitly.
+     * <p>
+     * Ordered by {@code id} into an insertion-ordered set: this feeds the grantAll expansion
+     * serialized by {@code PermissionSet.getBuilder}, and an unspecified iteration order (as a
+     * plain immutable Set has) would make the proto's repeated field non-deterministic across
+     * runs, breaking serializeForHash stability. Set semantics are unchanged — equality stays
+     * order-independent, so promotion/folding comparisons are unaffected.
+     */
+    public static Set<Permission> autoGrantable() {
```
KimStrand, line 94: This method now returns a mutable LinkedHashSet. PermissionSet caches that set statically and returns it directly from getPermissions(), so any caller can modify the grantAll expansion globally for every client. Once sensitive permissions exist, a caller could even add one to that shared set.

Please preserve the deterministic order while returning an unmodifiable set, for example by wrapping the LinkedHashSet with Collections.unmodifiableSet, and add a mutation test.

api/src/main/java/bisq/api/access/permissions/Permission.java
rodvar: 💯  - thanks for this I was planning to implement and forgot last minute

api/src/main/proto/api.proto
rodvar: can we accept this one as is? if I change just this it. creates inconsistency with the rest of the code we should change all of them

api/src/test/java/bisq/api/access/persistence/ApiAccessStoreServiceTest.java
rodvar: yeah I agree this is weak, I'll fix it using AI

api/src/main/java/bisq/api/access/permissions/Permission.java
rodvar: my bad I think I added a regression bug here - fixing it now. I'll improve the tests as well to avoid doing this again thank you

api/src/main/proto/api.proto
KimStrand: Yes, that's fair. Let's keep grantAll consistently. Please update the proto comments to match the implemented semantics: grantAll covers all current and future auto-grantable permissions, but never sensitive permissions. The ApiAccessStore comment should likewise say that promotion occurs only when the stored set exactly matches the running version's auto-grantable permission set.

rodvar: done

KimStrand (commented):
utACK

PR #4922, opened 2026-08-12
Fail PriceQuote conversions on long overflow

### LINKED ISSUE #4910: PriceQuote amount conversions wrap silently on overflow
The amount conversions in `PriceQuote` end in `BigDecimal.longValue()`, which silently wraps when the result does not fit into a long instead of failing:

- `PriceQuote.from(...)` (`common/src/main/java/bisq/common/monetary/PriceQuote.java:145`)
- `toQuoteSideMonetary(...)` (`:162`)
- `toBaseSideMonetary(...)` (`:182`)

So an amount that overflows when converted at a given price comes out as a plausible-looking but wrong value rather than being rejected. For example a base-side amount around 2^60 sats converted at a normal BTC/USD price wraps into a small, valid-looking quote amount.

This is reachable from the create-offer and Bisq Easy amount flows, which convert amounts through these methods (`OfferAmountUtil`, `TradeAmountLimits`, and the Bisq Easy / MuSig `AmountSelectionController`s).

Suggested fix: use a checked conversion (`longValueExact`) so an overflowing conversion throws instead of wrapping, and reject such amounts at the input boundary. The take-offer flow already does this with exact conversion variants; the shared `PriceQuote` methods still wrap for every other caller.

balag3:
Fixes #4910

The three amount conversions in `PriceQuote` ended in `BigDecimal.longValue()`, which silently wraps on long overflow: an amount that overflows when converted at a given price came out as a plausible-looking but wrong value instead of being rejected. Reachable from offer display and formatting, the create/take wizards, and one wire-facing site (the maker-side take-offer amount validation, where both the contract amounts and, for a fixed price spec, the price quote itself come verbatim from the peer's request).

### Fix

Seven commits, small to large blast radius (5–7 came out of review):

1. **Fail `PriceQuote` conversions on long overflow** — `longValueExact` with the rounding made explicit (`setScale(0, DOWN)` preserves the truncation on the quote side, the scale-0 `HALF_UP` divide preserves the rounding on the base side). In-range results are unchanged, including negative values; overflow throws `ArithmeticException`. Tests pin the rounding semantics and assert the overflow behavior of all three methods.
2. **Reject take offer requests with overflowing amounts** — the maker-side conversion is wrapped and rethrown as `IllegalArgumentException` with a clean message, matching how the other failed amount validations in that handler are reported (this also covers the already-reachable division by zero from a peer-supplied fixed price of value 0). The new test drives `verify()` to the conversion with a real fixed price spec and a `Long.MAX_VALUE` quote amount.
3. **Hide offers whose amounts cannot be resolved** — `BisqEasyOfferbookMessageService.isValid` now requires the offer's amounts to resolve on both sides (present and non-overflowing) before the direction-dependent checks; buy offers previously short-circuited the only path that touched the conversions. Rendering and matching paths that stream network offers without the validity gate get the same skip-on-failure treatment: profile-card overview totals, the MuSig offerbook list (no message validity chain exists there), the trade wizard's amount matching validators, the select-offer list construction, and the min/lowest amount evaluation in both `TradeAmountLimits` classes (whose highest-amount evaluation already caught per offer). Without this, one hostile offer would throw during list rendering on the JavaFX thread for every viewer.
4. **Ignore overflowing conversions in the amount selection components** — the passive-side setters convert once and skip the update on overflow, so the paired displays never show a wrapped value and always change together.
5. **Isolate offer mapping failures in the REST offer list** — one offer whose DTO cannot be built no longer fails the whole listing.
6. **Retry offers after market prices arrive** — the two Bisq Easy list controllers re-check offers that were skipped because their market price was not there yet.
7. **Propagate offer validity changes to all consumers** — `BisqEasyOfferbookMessageService` publishes one offer validity revision (market prices, ignore list, ban list, a maker's reputation score) and every consumer that caches results of `isValid` observes it: the market badges, the dashboard count, the `NUM_OFFERS` web socket topic and both list controllers, which now reconcile in both directions (invalid offers are removed together with their processed id, newly valid ones added; queued callbacks after deactivation or a channel switch are ignored). The sellers reputation service drops its negative cache on price changes (the required score is price-derived), never caches a result computed before an invalidation, treats a USD-overflowing amount as insufficient instead of throwing out of `isValid`, and observes the reputation score map instead of the equality-gated score change observable.

### Testing

- New unit tests: `PriceQuoteTest` (overflow in both sign directions for all three methods, truncation/HALF_UP semantics pinning incl. negatives), `BisqEasyTakeOfferRequestHandlerTest` (hostile request rejected at `verify()`), `BisqEasyOfferbookMessageServiceTest` (overflowing offers invalid in both directions, missing-market-price offers invalid, convertible offers valid). The lenient overflow try/catch in `MonetaryTest` is tightened into an assertion.
- Commits 6–7: observer-driven tests through the real service and a real market price map — `OfferbookListControllerTest` and `ChatMessagesListControllerTest` (offer before price, price change making the conversion overflow incl. re-add, price removal, ignore/un-ignore, revision after deactivation, a failing list item not blocking the others), `NumOffersWebSocketServiceTest`, `MarketChannelItemTest`, the revision sources in `BisqEasyOfferbookMessageServiceTest`, and `BisqEasySellersReputationBasedTradeAmountServiceTest` (price invalidation, in-flight check not re-caching a stale result, consecutive score changes, USD overflow).
- All module suites green: common, offer, bisq-easy, trade, api, mu-sig, desktop.
- The UI guards (wizard validators, list-item skips, amount-selection setters) have no controller test harness and are verified by inspection.

Manual A/B on a local CLEAR-net setup, with one client publishing an offer whose base-side amount overflows a long when converted to the quote side:

- On `main`, the offer renders in the offerbook with the silently wrapped amount (a negative fiat value), and the receiving client counts it as a valid offer.

Screenshot from 2026-08-12 11-24-55

- On this branch, the same offer is hidden from the offerbook — the market shows one fewer offer and there is no error popup or log error, only a debug line.

 
Screenshot from 2026-08-12 11-25-20

An honest offer created and taken across the same setup is unaffected.

### Remarks

- Failures on the wire path are classified like the handler's existing amount validations (`UNKNOWN` failure with a clean message). A dedicated `TradeProtocolFailure` value for malformed amounts would be a separate wire-visible change; happy to file it as a follow-up.
- Pre-existing and left out of commit 7 (not consumers of the cached validity): `OffersWebSocketService` and the REST offer count/list work on raw offers without a validity filter; the profile card and the trade wizard build one-shot snapshots at open; a maker profile arriving after or removed before its offer message is not re-evaluated. Happy to file follow-ups.
- One adjacent pre-existing issue is left out of scope: list items calling `PriceUtil.findPercentFromMarketPrice(...).orElseThrow()` throw `NoSuchElementException` for fixed-price offers in markets without a current market price (a different exception, so the guards here do not catch it). I can file it as a separate issue. Relatedly, `FixPriceSpec.verify()` is empty, so a peer can supply a zero or negative fixed price; the resulting division by zero is now contained at the conversion boundaries by this PR, but the root cause would be better fixed by validating the spec.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented currency-conversion overflows from interrupting amount selection, offer lists, profiles, API responses, and trade validation.
  * Invalid or unconvertible offers are now safely rejected or excluded while valid offers remain available.
  * Preserved previously displayed amounts when conversions cannot be completed.
  * Improved handling of empty, invalid, excessive, and malformed amount values.
  * Offerbooks and offer counts now refresh when market prices, reputation, or offer validity changes.

* **Tests**
  * Added coverage for conversion rounding, overflow handling, market-price updates, API listings, offerbook refreshes, and malformed take-offer requests.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListController.java
@@ -38,6 +38,7 @@ import bisq.user.identity.UserIdentityService;
 import bisq.user.profile.UserProfile;
 import bisq.user.profile.UserProfileService;
 import bisq.user.reputation.ReputationService;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import javafx.beans.property.ReadOnlyBooleanProperty;
 import lombok.Getter;
@@ -46,6 +47,7 @@ import org.fxmisc.easybind.EasyBind;
 import org.fxmisc.easybind.Subscription;
 
 import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
@@ -61,8 +63,9 @@ public class OfferbookListController implements bisq.desktop.common.view.Control
     private final UserIdentityService userIdentityService;
     private final BisqEasyOfferbookMessageService bisqEasyOfferbookMessageService;
     private Pin showBuyOffersPin, showOfferListExpandedSettingsPin, offerMessagesPin, showMyOffersOnlyPin,
-            userIdentityPin, userProfileIdWithScoreChangePin;
+            userIdentityPin, userProfileIdWithScoreChangePin, offerValidityRevisionPin;
     private Subscription showBuyOffersFromModelPin, activeMarketPaymentsCountPin, showMyOffersOnlyFromModelPin;
+    private int channelBindingGeneration;
 
     public OfferbookListController(ServiceProvider serviceProvider) {
         settingsService = serviceProvider.getSettingsService();
@@ -78,6 +81,11 @@ public class OfferbookListController implements bisq.desktop.common.view.Control
 
     public ReadOnlyBooleanProperty getShowOfferListExpanded() {
         return model.getShowOfferListExpanded();
+    }
+
+    @VisibleForTesting
+    OfferbookListModel getModel() {
+        return model;
     }
 
     @Override
@@ -97,6 +105,8 @@ public class OfferbookListController implements bisq.desktop.common.view.Control
         showMyOffersOnlyFromModelPin = EasyBind.subscribe(model.getShowMyOffersOnly(), showMyOffersOnly -> updatePredicate());
         userIdentityPin = userIdentityService.getSelectedUserIdentityObservable().addObserver(userIdentity -> UIThread.run(this::updatePredicate));
         userProfileIdWithScoreChangePin = reputationService.getUserProfileIdWithScoreChange().addObserver(userProfileId -> UIThread.run(this::updatePredicate));
+        offerValidityRevisionPin = bisqEasyOfferbookMessageService.getOfferValidityRevision().addObserver(revision ->
+                UIThread.run(this::reconcileSelectedChannelOffers));
     }
 
     @Override
@@ -110,11 +120,14 @@ public class OfferbookListController implements bisq.desktop.common.view.Control
         activeMarketPaymentsCountPin.unsubscribe();
         if (offerMessagesPin != null) {
             offerMessagesPin.unbind();
+            offerMessagesPin = null;
         }
         showMyOffersOnlyPin.unbind();
         showMyOffersOnlyFromModelPin.unsubscribe();
         userIdentityPin.unbind();
         userProfileIdWithScoreChangePin.unbind();
+        offerValidityRevisionPin.unbind();
+        offerValidityRevisionPin = null;
     }
 
     private void disposeAndClearOfferbookListItems() {
@@ -135,23 +148,13 @@ public class OfferbookListController implements bisq.desktop.common.view.Control
         }
         disposeAndClearOfferbookListItems();
         model.getChatMessageIds().clear();
+        int bindingGeneration = ++channelBindingGeneration;
         offerMessagesPin = channel.getChatMessages().addObserver(new CollectionObserver<>() {
             @Override
             public void onAdded(BisqEasyOfferbookMessage offerbookMessage) {
                 UIThread.runOnNextRenderFrame(() -> {
-                    if (!model.getChatMessageIds().contains(offerbookMessage.getId()) &&
-                            offerbookMessage.hasBisqEasyOffer()) {
-                        if (bisqEasyOfferbookMessageService.isValid(offerbookMessage)) {
-                            userProfileService.findUserProfile(offerbookMessage.getAuthorUserProfileId())
-                                    .ifPresent(authorUserProfile -> {
-                                        OfferbookListItem item = new OfferbookListItem(offerbookMessage,
-                                                authorUserProfile,
-                                                reputationService,
-                                                marketPriceService);
-                                        model.getOfferbookListItems().add(item);
-                                        model.getChatMessageIds().add(offerbookMessage.getId());
-                                    });
-                        }
+                    if (isShowing(channel, bindingGeneration)) {
+                        maybeAddOfferbookMessage(offerbookMessage);
                     }
                 });
             }
@@ -159,7 +162,8 @@ public class OfferbookListController implements bisq.desktop.common.view.Control
             @Override
             public void onRemoved(Object element) {
                 UIThread.runOnNextRenderFrame(() -> {
-                    if (element instanceof BisqEasyOfferbookMessage && ((BisqEasyOfferbookMessage) element).hasBisqEasyOffer()) {
+                    if (isShowing(channel, bindingGeneration) &&
+                            element instanceof BisqEasyOfferbookMessage && ((BisqEasyOfferbookMessage) element).hasBisqEasyOffer()) {
                         BisqEasyOfferbookMessage offerMessage = (BisqEasyOfferbookMessage) element;
                         Optional<OfferbookListItem> toRemove = model.getOfferbookListItems().stream()
                                 .filter(item -> item.getBisqEasyOfferbookMessage().getId().equals(offerMessage.getId()))
@@ -176,11 +180,63 @@ public class OfferbookListController implements bisq.desktop.common.view.Control
             @Override
             public void onCleared() {
                 UIThread.runOnNextRenderFrame(() -> {
-                    disposeAndClearOfferbookListItems();
-                    model.getChatMessageIds().clear();
+                    if (isShowing(channel, bindingGeneration)) {
+                        disposeAndClearOfferbookListItems();
+                        model.getChatMessageIds().clear();
+                    }
                 });
             }
         });
+    }
+
+    private boolean isShowing(BisqEasyOfferbookChannel channel, int bindingGeneration) {
+        return offerMessagesPin != null &&
+                bindingGeneration == channelBindingGeneration &&
+                model.getChannel().get() == channel;
+    }
+
+    private void reconcileSelectedChannelOffers() {
+        BisqEasyOfferbookChannel channel = model.getChannel().get();
+        if (offerValidityRevisionPin == null || !isShowing(channel, channelBindingGeneration)) {
+            return;
+        }
+
+        List<OfferbookListItem> invalidItems = model.getOfferbookListItems().stream()
+                .filter(item -> !bisqEasyOfferbookMessageService.isValid(item.getBisqEasyOfferbookMessage()))
+                .toList();
+        invalidItems.forEach(item -> {
+            item.dispose();
+            model.getOfferbookListItems().remove(item);
+            model.getChatMessageIds().remove(item.getBisqEasyOfferbookMessage().getId());
+        });
+
+        channel.getChatMessages().forEach(this::maybeAddOfferbookMessage);
+    }
+
+    private void maybeAddOfferbookMessage(BisqEasyOfferbookMessage offerbookMessage) {
+        if (model.getChatMessageIds().contains(offerbookMessage.getId()) ||
+                !offerbookMessage.hasBisqEasyOffer() ||
+                !bisqEasyOfferbookMessageService.isValid(offerbookMessage)) {
+            return;
+        }
+
+        addOfferbookMessage(offerbookMessage);
+    }
+
+    private void addOfferbookMessage(BisqEasyOfferbookMessage offerbookMessage) {
+        userProfileService.findUserProfile(offerbookMessage.getAuthorUserProfileId())
+                .ifPresent(authorUserProfile -> {
+                    try {
+                        OfferbookListItem item = new OfferbookListItem(offerbookMessage,
+                                authorUserProfile,
+                                reputationService,
+                                marketPriceService);
+                        model.getOfferbookListItems().add(item);
+                        model.getChatMessageIds().add(offerbookMessage.getId());
+                    } catch (RuntimeException e) {
+                        log.warn("Could not create the list item for offer message {}", offerbookMessage.getId(), e);
+                    }
+                });
     }
 
     void onToggleOfferList() {
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/dashboard/top_panel/DashboardTopPanel.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/dashboard/top_panel/DashboardTopPanel.java
@@ -74,7 +74,7 @@ public class DashboardTopPanel {
         private final UserProfileService userProfileService;
         private final BisqEasyOfferbookChannelService bisqEasyOfferbookChannelService;
         private final BisqEasyOfferbookMessageService bisqEasyOfferbookMessageService;
-        private Pin selectedMarketPin, marketPricePin, getNumUserProfilesPin;
+        private Pin selectedMarketPin, marketPricePin, offerValidityRevisionPin, getNumUserProfilesPin;
         private final Set<Pin> channelsPins = new HashSet<>();
         private boolean allowUpdateOffersOnline;
 
@@ -99,6 +99,8 @@ public class DashboardTopPanel {
             channelsPins.addAll(bisqEasyOfferbookChannelService.getChannels().stream()
                     .map(channel -> channel.getChatMessages().addObserver(this::updateOffersOnline))
                     .collect(Collectors.toSet()));
+            offerValidityRevisionPin = bisqEasyOfferbookMessageService.getOfferValidityRevision()
+                    .addObserver(revision -> updateOffersOnline());
 
             // We trigger a call of updateOffersOnline for each channel when registering our observer. But we only want one call,
             // so we block execution of the code inside updateOffersOnline to only call it once.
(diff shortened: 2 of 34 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/listing/MuSigOfferbookController.java
```diff
@@ -136,12 +136,23 @@ public void onAdded(MuSigOffer muSigOffer) {
...
+                                    accountService);
+                        } catch (ArithmeticException e) {
+                            // The item formats the offer amounts at construction; an offer whose
+                            // amounts overflow at the current price is not listed instead of
+                            // failing on the JavaFX thread.
+                            log.warn("Skipping offer {}: its amounts cannot be converted at the current market price",
+                                    offerId);
+                            return;
+                        }
+                        model.getMuSigOfferListItems().add(item);
```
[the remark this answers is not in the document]
balag3: The skip only fires on overflow, which means the amount is far larger than any real offer (well past total BTC supply). It would only become "valid" after a huge market-price swing, and then we'd be showing a malformed offer. Skipping it for good is what we want here, so I'll leave this as is.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/overview/ProfileCardOverviewController.java
```diff
@@ -124,7 +124,15 @@ private long getTotalBaseOfferAmount(String userProfileId, Predicate<BisqEasyOff
...
-                .flatMap(offer -> OfferAmountUtil.findBaseSideMaxOrFixedAmount(marketPriceService, offer).stream())
+                .flatMap(offer -> {
+                    try {
+                        return OfferAmountUtil.findBaseSideMaxOrFixedAmount(marketPriceService, offer).stream();
+                    } catch (ArithmeticException e) {
+                        // An offer whose amounts overflow at the current price cannot contribute
+                        // to the total; it is hidden from the offer lists as invalid.
+                        return java.util.stream.Stream.<bisq.common.monetary.Monetary>empty();
+                    }
+                })
```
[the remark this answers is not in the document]
balag3: Fair point, though that's a separate long-addition overflow, not the PriceQuote wrap this PR is about. It only bites on already-malformed offers summing past Long.MAX, and the worst case is a wrong number in a display-only total, no crash. I'd rather keep this PR scoped and handle the aggregate separately if we want to.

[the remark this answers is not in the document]
balag3: I'll open it myself thx.

bisq-easy/src/main/java/bisq/bisq_easy/BisqEasyOfferbookMessageService.java
```diff
@@ -95,8 +100,32 @@ public boolean isValid(BisqEasyOfferbookMessage message) {
...
                 isNotIgnored(message) &&
                 (isTextMessage(message) ||
-                        isBuyOffer(message) ||
-                        hasSellerSufficientReputation(message));
+                        (hasResolvableAmounts(message) &&
+                                (isBuyOffer(message) || hasSellerSufficientReputation(message))));
+    }
+
+    // An offer whose amounts cannot be resolved on both sides - because a conversion overflows
+    // at its price, or because the required market price is missing and the conversions come
```
KimStrand, line 183: This also rejects a MarketPriceSpec when its market price is merely unavailable yet. That state is transient: MarketPriceRequestService.initialize() starts the request asynchronously and returns immediately. OfferbookListController and ChatMessagesListController observe chat-message changes, not market-price availability, so an offer skipped here is not reconsidered when the price arrives; it remains hidden for the active channel until a rebind or message re-add. OfferbookListItem's own price observer cannot help because the item is never constructed.

Should we distinguish arithmetic overflow from temporary price unavailability and trigger reevaluation when the relevant market price becomes available. A regression test could cover “offer arrives first, market price arrives later.”

common/src/main/java/bisq/common/monetary/PriceQuote.java
```diff
@@ -176,10 +181,12 @@ public Monetary toBaseSideMonetary(Monetary quoteSideMonetary) {
...
                         "parameter quoteSideMonetary=" + quoteSideMonetary + "\n" +
                         "this.quoteSideMonetary" + this.quoteSideMonetary);
+        // Explicit scale 0 keeps the HALF_UP rounding the plain divide applied;
+        // longValueExact makes an overflowing conversion fail instead of silently wrapping.
         long value = BigDecimal.valueOf(quoteSideMonetary.value)
                 .movePointRight(baseSideMonetary.precision)
-                .divide(BigDecimal.valueOf(this.value), RoundingMode.HALF_UP)
-                .longValue();
+                .divide(BigDecimal.valueOf(this.value), 0, RoundingMode.HALF_UP)
+                .longValueExact();
```
KimStrand, line 189: This new ArithmeticException is not isolated in REST offer-list rendering. OfferbookRestApi.findOffer() maps raw channel messages through OfferItemPresentationDtoFactory, whose formatBaseAmount() path reaches this conversion.

For example, a QuoteSideFixedAmountSpec(Long.MAX_VALUE) at a fixed price of 0.0001 USD/BTC throws here. That aborts the stream, and the endpoint's outer catch returns HTTP 500 for the entire market. Chat ingress does not reject this arithmetic condition, so a peer can trigger it.

Should we isolate conversion failures per offer, as OffersWebSocketService already does, or filter through BisqEasyOfferbookMessageService. Add a REST test containing one valid offer and one overflowing offer, asserting that the valid offer is still returned.

balag3: Fixed in ae8d6369a4 — the REST offer list maps each offer independently now, so one offer that can't be converted is skipped instead of 500-ing the whole market. Added a test with one valid and one overflowing offer asserting the valid one is still returned.

bisq-easy/src/main/java/bisq/bisq_easy/BisqEasyOfferbookMessageService.java
balag3: Good point — the desktop re-evaluation on price arrival needs the exception-safe list item from #4928 and also touches the pre-existing list/tab observer lifecycle in these controllers, so I'll do it as a focused follow-up rather than a partial fix here.

KimStrand: Thanks. The added service test does not resolve this issue: it manually calls isValid() again after changing the mocked price, but no production consumer performs that second call.

OfferbookListController and ChatMessagesListController still react only to chat-message changes. PR #4928 updates already-constructed list items, so it cannot restore an offer filtered out before construction.

As a result, a valid offer received before its market price remains hidden until a channel rebind or message mutation. Should we add the price-triggered re-evaluation here, or stop treating temporary price absenc as invalid?

balag3: Added price-change retries in both controllers. They now re-check only offers that were previously skipped, so offer-first/price-later no longer leaves a valid offer hidden.

KimStrand: The two list controllers now handle the transition, but the underlying issue is not limited to those controllers.

MarketChannelItem and NumOffersWebSocketService recompute getOffers(channel).count() only when chat messages change. In the offer-first/price-later sequence, both cache a count excluding the offer; price arrival changes isValid() but emits no chat event, so the market badge and NUM_OFFERS WebSocket remain stale even after the offer appears in the lists. DashboardController and DashboardTopPanel have the same trigger mismatch.

Should we propagate price driven validity changes to these cached consumers as well, preferably through one shared observable, and add regression coverage for at least NUM_OFFERS.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListController.java
```diff
@@ -183,6 +179,50 @@ public void onCleared() {
...
+    @VisibleForTesting
+    static void reprocessOffers(BisqEasyOfferbookChannel channel,
+                                Set<String> processedMessageIds,
+                                Predicate<BisqEasyOfferbookMessage> validator,
+                                Consumer<BisqEasyOfferbookMessage> processor) {
+        channel.getChatMessages().stream()
+                .filter(BisqEasyOfferbookMessage::hasBisqEasyOffer)
+                .filter(message -> !processedMessageIds.contains(message.getId()))
+                .filter(validator)
+                .forEach(processor);
```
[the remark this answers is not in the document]
KimStrand: Confirmed. Amount specs are peer-controlled and only require a positive value, so a base-side MarketPriceSpec amount can convert successfully at one market price and overflow after a later price update. Because processedMessageIds bypasses validation for displayed entries, both controllers retain that now-unconvertible offer indefinitely.

Should we reconcile in both directions on price changes: dispose and remove displayed items that fail isValid(), remove their IDs, and add previously skipped items that now pass? Adding a valid-to-invalid regression case test could also make sense. The current test directly invokes reprocessOffers, so removing the actual market price observer would still leave it green.

bisq-easy/src/main/java/bisq/bisq_easy/BisqEasyOfferbookMessageService.java
balag3: Confirmed. Fixed in: a419871fba — one shared validity revision in BisqEasyOfferbookMessageService, observed by the counts (badge, dashboard, NUM_OFFERS with a test).

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListController.java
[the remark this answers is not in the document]
balag3: Confirmed. Fixed in: a419871fba — both controllers reconcile in both directions on the revision; observer-driven tests cover invalid→valid and valid→invalid.

bisq-easy/src/main/java/bisq/bisq_easy/BisqEasyOfferbookMessageService.java
```diff
@@ -36,15 +47,72 @@ public class BisqEasyOfferbookMessageService implements Service {
...
+                    incrementOfferValidityRevision();
+                }
+            }
+
+            @Override
+            public void putAll(Map<? extends String, ? extends Long> map) {
+                // Fired at registration with all known scores; one revision instead of one per score.
+                incrementOfferValidityRevision();
+            }
+        });
```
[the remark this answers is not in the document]
balag3: Added remove/clear handling in both observers in 176f517ea4. Note that a bulk update already invalidated the cache: the default `putAll` delegates to `put` per key. Today only `put` reaches this map (the removals you point at are on the per-source services' own maps), so this is defensive.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListController.java
```diff
@@ -176,13 +180,65 @@ public void onRemoved(Object element) {
...
+        userProfileService.findUserProfile(offerbookMessage.getAuthorUserProfileId())
+                .ifPresent(authorUserProfile -> {
+                    try {
+                        OfferbookListItem item = new OfferbookListItem(offerbookMessage,
+                                authorUserProfile,
+                                reputationService,
+                                marketPriceService);
+                        model.getOfferbookListItems().add(item);
+                        model.getChatMessageIds().add(offerbookMessage.getId());
+                    } catch (RuntimeException e) {
```
KimStrand, line 236: This catch appears to make the retry path leak partially constructed OfferbookListItems. A fixed-price offer can pass isValid() without a market quote, but the constructor registers its market price observer before updatePriceSpecAsPercent().orElseThrow() fails. Since ObservableHashMap.addObserver() installs the observer before invoking it, construction aborts without returning an item that can be disposed. The message ID also remains unrecorded, so later validity revisions can retry construction and install additional unreachable observers.

balag3: Confirmed. Fixed in: 3208fa28c8

apps/desktop/desktop/src/test/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListItemTest.java
```diff
@@ -0,0 +1,112 @@
...
+        when(marketPrice.getPriceQuote()).thenReturn(PriceQuote.fromFiatPrice(90_000, "USD"));
+        marketPriceByCurrencyMap.put(MARKET, marketPrice);
+
+        WaitForAsyncUtils.asyncFx(() -> {
+            OfferbookListItem item = new OfferbookListItem(message, senderUserProfile, reputationService, marketPriceService);
+            item.dispose();
+        });
+        WaitForAsyncUtils.waitForFxEvents();
+
+        verify(marketPriceByCurrencyMap, times(1)).addObserver(any(Runnable.class));
```
KimStrand: The test name says dispose() removes the observer, but this assertion only proves that one observer was registered; it would still pass if dispose() were a no-op. Since this change is specifically about observer lifecycle, could we either verify that a map update after dispose() no longer invokes the callback, or rename the test to match what it covers?

balag3: Confirmed. Fixed in: c8bf62572e

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/dashboard/top_panel/DashboardTopPanel.java
```diff
@@ -99,6 +99,8 @@ public void onActivate() {
             channelsPins.addAll(bisqEasyOfferbookChannelService.getChannels().stream()
                     .map(channel -> channel.getChatMessages().addObserver(this::updateOffersOnline))
                     .collect(Collectors.toSet()));
+            offerValidityRevisionPin = bisqEasyOfferbookMessageService.getOfferValidityRevision()
+                    .addObserver(revision -> updateOffersOnline());
```
KimStrand, line 103: The validity revision can be emitted from market-price or reputation worker threads, while allowUpdateOffersOnline is written by onActivate()/onDeactivate() on the JavaFX thread. updateOffersOnline() reads that plain boolean before UIThread.run(), so this is a data race: a refresh can observe stale state, and queued work can update the model after deactivation. Should we move the guard inside the UIThread.run() lambda so both the lifecycle check and model mutation run on the FX thread? The same pattern occurs in DashboardController and MuSigDashboardTopPanel.

balag3: Confirmed. Fixed in: 4e33e055e2.

KimStrand:
utACK

PR #4924, opened 2026-08-12
Add MuSig custom payout persistence and P2P messages

KimStrand:
## Summary by CodeRabbit

* **New Features**
  * Added support for sharing and storing custom payout transaction data between trade parties.
  * Added support for recording custom close-trade responses.
  * Added mediation rejection status and dedicated rejection messaging for MuSig trades.
  * Added secure network messaging for custom payout details and mediation outcomes.

* **Bug Fixes**
  * Improved validation for malformed, incomplete, oversized, or conflicting payout data.
  * Prevented incompatible payout and mediation states from being recorded.
  * Improved protection of sensitive transaction data during storage and transmission.

### DIFF
--- a/support/src/main/proto/support.proto
+++ b/support/src/main/proto/support.proto
@@ -122,6 +122,12 @@ message MuSigMediationResultAcceptanceMessage {
   bool mediationResultAccepted = 3;
 }
 
+message MuSigMediationResultRejectionMessage {
+  string tradeId = 1;
+  network.identity.NetworkId senderNetworkId = 2;
+  bytes mediationResultHash = 3;
+}
+
 enum MuSigMediationIssueType {
   MUSIGMEDIATIONISSUETYPE_UNSPECIFIED = 0;
   MUSIGMEDIATIONISSUETYPE_PEER_CONTRACT_HASH_MISMATCH = 1;
--- a/support/src/test/java/bisq/support/mediation/mu_sig/MuSigMediationResultRejectionMessageTest.java
+++ b/support/src/test/java/bisq/support/mediation/mu_sig/MuSigMediationResultRejectionMessageTest.java
@@ -1,0 +1,81 @@
+/*
+ * This file is part of Bisq.
+ *
+ * Bisq is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Bisq is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with Bisq. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bisq.support.mediation.mu_sig;
+
+import bisq.network.p2p.message.ExternalNetworkMessage;
+import bisq.user.profile.UserProfile;
+import com.google.protobuf.Any;
+import org.junit.jupiter.api.Test;
+
+import static bisq.support.mu_sig.MuSigRequestSizeTestFixtures.TRADE_ID;
+import static bisq.support.mu_sig.MuSigRequestSizeTestFixtures.createUserProfile;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+class MuSigMediationResultRejectionMessageTest {
+
+    @Test
+    void roundTripsThroughNetworkMessageResolver() {
+        byte[] mediationResultHash = createMediationResultHash();
+        MuSigMediationResultRejectionMessage message = createMessage(mediationResultHash);
+
+        Any any = Any.pack(message.resolveValueProto(false));
+        ExternalNetworkMessage restored =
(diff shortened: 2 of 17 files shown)


HenrikJannsen:
@stejbac Could you review?

support/src/test/java/bisq/support/mediation/mu_sig/MuSigMediationResultRejectionMessageTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    void roundTripsThroughNetworkMessageResolver() {
+        MuSigMediationResultRejectionMessage message = createMessage(new byte[20]);
+
+        Any any = Any.pack(message.resolveValueProto(false));
+        ExternalNetworkMessage restored =
+                MuSigMediationResultRejectionMessage.getNetworkMessageResolver().fromAny(any);
+
+        assertThat(restored).isEqualTo(message);
```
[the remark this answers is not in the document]
KimStrand: fixed

trade/src/main/java/bisq/trade/mu_sig/messages/network/MuSigCustomPayoutPsbtMessage.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.Arrays;
+import java.util.Objects;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+@Slf4j
+@ToString(callSuper = true)
+@Getter
+public final class MuSigCustomPayoutPsbtMessage extends MuSigTradeMessage {
+    public static final int MAX_PSBT_SIZE = 10_000;
```
stejbac, line 37: FWIW, I just measured the actual PSBT size and it's 1263 bytes with a single trader's signatures (that is, the PSBTs which are actually exchanged), 1525 bytes with both traders' signatures and 757 bytes after finalisation of both fully signed inputs. It's plausible more metadata could be added to the PSBT at a later date to simplify or otherwise improve the signing logic, and the size depends on the output address types (which are currently always P2TR), so there should probably be some room for growth, but its unlikely to ever approach 10,000 bytes.

support/src/main/proto/support.proto
```diff
@@ -122,6 +122,12 @@ message MuSigMediationResultAcceptanceMessage {
   bool mediationResultAccepted = 3;
 }
 
+message MuSigMediationResultRejectionMessage {
+  string tradeId = 1;
+  network.identity.NetworkId senderNetworkId = 2;
+  bytes mediationResultHash = 3;
+}
```
stejbac, line 129: I'm a bit confused about the separate roles of `MuSigMediationResultRejectionMessage` and the existing `MuSigMediationResultAcceptanceMessage`, when the latter already has a boolean `mediationResultAccepted` field which could be set to false. Also, the latter doesn't have a `mediationResultHash` field -- isn't it meant to?

trade/src/main/proto/trade.proto
```diff
@@ -224,6 +235,8 @@ message MuSigTradeParty {
   optional common.ByteArray peersOutputPrvKeyShare = 12;
   optional account.AccountPayload accountPayload = 13;
   optional bool mediationResultAccepted = 14;
+  optional MuSigCustomPayoutPartyData customPayoutData = 15;
+  bool mediationResultRejected = 16;
```
stejbac, line 239: Related to the above comment: Under what circumstances would the optional boolean `mediationResultAccepted` field be set to false, as opposed to left empty? Isn't it redundant to have a separate `mediationResultRejected` field (or maybe neither field should be optional, for consistency)?

stejbac (commented):
I couldn't spot any errors in the actual Java code itself, though I notice it's doing a fair bit of defensive copying of byte arrays (obstructing the use of Lombok) in the DTOs, with tests added for that, which adds quite a lot of boilerplate -- I assume that's AI generated. Even though it's safer, it does break the conventions followed in most of the existing DTOs.

support/src/main/proto/support.proto
KimStrand: Thanks for the review!. The PR is currently in a transitional state. The intended flow is that sending a MuSigCustomPayoutPsbtMessage represents acceptance, and that message already contains the mediation result hash.
  
Rejection remains explicit because receiving no PSBT could also mean that the peer is offline or unresponsive. I’m already working on replacing the old acceptance path, so MuSigMediationResultAcceptanceMessage and mediationResultAccepted will be removed.

trade/src/main/proto/trade.proto
KimStrand: You’re right: in the intended flow, mediationResultAccepted = false will no longer have a distinct meaning. I kept the existing acceptance state temporarily because I’m implementing the flow step by step: first the custom-payout persistence and P2P messages, then the signing flow, and finally removal of the obsolete acceptance path. Once that is complete, acceptance will be represented by the custom payout PSBT, explicit rejection by mediationResultRejected, and neither will mean that the trader has not responded yet.

trade/src/main/java/bisq/trade/mu_sig/messages/network/MuSigCustomPayoutPsbtMessage.java
KimStrand: Thanks for measuring this. I agree that 10,000 bytes is unnecessarily high. Would a limit of 4,096 bytes make sense?

stejbac: Yes, I think that would make sense.

PR #4928, opened 2026-08-13
Tolerate missing market price in offer list items

### LINKED ISSUE #4923: Offerbook list items crash on fixed-price offers when no market price is available
`OfferbookListItem` and `ProfileCardOfferListItem` compute the price-deviation percentage in their constructor with:

```java
priceSpecAsPercent = PriceUtil.findPercentFromMarketPrice(marketPriceService, bisqEasyOffer).orElseThrow();
```

- `OfferbookListItem` (`apps/desktop/desktop/.../bisq_easy/offerbook/offerbook_list/OfferbookListItem.java:125`)
- `ProfileCardOfferListItem` (`apps/desktop/desktop/.../user/profile_card/offers/ProfileCardOfferListItem.java:121`)

For a fixed-price offer, `findPercentFromMarketPrice` returns the deviation only when a market price is available for the offer's market; otherwise it returns `Optional.empty()`. The `orElseThrow()` then throws `NoSuchElementException`. Because the list item is built on the JavaFX thread while rendering the offerbook side table and the profile card offers tab, a fixed-price offer in a market without a current market price takes the view down with an uncaught exception.

This is specific to fixed-price offers — market-price and floating offers return a percentage without needing a market price. It is reachable whenever the price feed does not cover a market, or transiently before prices have loaded.

The MuSig equivalent, `MuSigOfferListItem`, already handles this with `ifPresent(...)` and degrades gracefully, so the Bisq Easy list items could follow the same pattern (skip the percentage or show a placeholder) instead of throwing.

Noticed while working on #4910 (silent overflow in `PriceQuote`); this is a separate `NoSuchElementException` on a different path, not covered by that fix.

balag3:
Fixes #4923

## Problem

Opening the Bisq Easy offerbook list or a maker's profile card offers tab crashes with `NoSuchElementException` when it contains a fixed-price offer in a market without a market price: both list items resolve the percent with `PriceUtil.findPercentFromMarketPrice(...).orElseThrow()`, which is empty for a fixed price without a market quote.

## Fix

Both items keep the percentage as `Optional<Double>`: while it cannot be derived the cell shows the localized `N/A` and the tooltip still shows the fixed price; the row recovers through the existing market-price observer when a price arrives. The formatted percentage and tooltip are `StringProperty`s bound by the table cells, so an already-rendered row updates on recovery instead of waiting for cell recycling.

## Tests

New `OfferbookListItemTest` and `ProfileCardOfferListItemTest`: the crash cases reproduced the exact exception before the fix, recovery on price arrival is asserted through a label bound to the property, and a control case covers the normal percent path.

## Manual testing

Local CLEAR-net setup: Alice with market prices creates a fixed-price BTC/USD offer; Bob runs with the market price service disabled (`application.bondedRoles.marketPrice.enabled=false`) and a fresh data dir.

- Bob without a market price: the offer renders in the offer list and in Alice's profile card with N/A in the price column and the fixed price in its tooltip, no error popup, no `NoSuchElementException` in the log (on main the list item construction throws).

Screenshot from 2026-08-18 19-31-19
Screenshot from 2026-08-18 19-31-28

- Bob restarted with market prices, same data dir: the same row shows the percentage.

Screenshot from 2026-08-18 14-47-43

## Summary by CodeRabbit

- **Bug Fixes**
  - Percentage prices and tooltips now update when market-price data becomes available or changes.
  - Offers without market-price data no longer show incorrect percentages or cause errors.
  - Reused and refreshed offer entries retain accurate price displays.
  - Offers without percentage prices appear last when sorting.
  - Price lists automatically resort when percentage data becomes available.

- **Tests**
  - Added coverage for missing, delayed, and initially available market-price scenarios, including sorting behavior.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListItem.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListItem.java
@@ -73,12 +73,12 @@ public class OfferbookListItem {
     private final long totalScore;
     private final Pin marketPriceByCurrencyMapPin;
     private final long offerAgeInDays;
+    private final StringProperty formattedPercentagePrice = new SimpleStringProperty();
+    private final StringProperty priceTooltipText = new SimpleStringProperty();
     // Empty while the offer has no resolvable percentage (a fixed-price offer without a
     // market price yet); comparators place these after known values instead of treating
     // them as an exactly-at-market 0%.
     private OptionalDouble priceSpecAsPercent = OptionalDouble.empty();
-    private final StringProperty formattedPercentagePrice = new SimpleStringProperty();
-    private final StringProperty priceTooltipText = new SimpleStringProperty();
 
     public OfferbookListItem(BisqEasyOfferbookMessage bisqEasyOfferbookMessage,
                              UserProfile senderUserProfile,
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListItem.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListItem.java
@@ -48,6 +48,7 @@ import lombok.extern.slf4j.Slf4j;
 
 import java.util.Comparator;
 import java.util.List;
+import java.util.OptionalDouble;
 import java.util.stream.Collectors;
 
 @Slf4j
@@ -72,7 +73,10 @@ public class OfferbookListItem {
     private final long totalScore;
     private final Pin marketPriceByCurrencyMapPin;
     private final long offerAgeInDays;
-    private double priceSpecAsPercent;
+    // Empty while the offer has no resolvable percentage (a fixed-price offer without a
+    // market price yet); comparators place these after known values instead of treating
+    // them as an exactly-at-market 0%.
+    private OptionalDouble priceSpecAsPercent = OptionalDouble.empty();
     private final StringProperty formattedPercentagePrice = new SimpleStringProperty();
     private final StringProperty priceTooltipText = new SimpleStringProperty();
 
@@ -130,7 +134,7 @@ public class OfferbookListItem {
         // price arrives; the table cells bind to them so visible rows follow.
         PriceUtil.findPercentFromMarketPrice(marketPriceService, bisqEasyOffer)
                 .ifPresent(percent -> {
-                    priceSpecAsPercent = percent;
+                    priceSpecAsPercent = OptionalDouble.of(percent);
                     formattedPercentagePrice.set(PercentageFormatter.formatToPercentWithSignAndSymbol(percent));
                     priceTooltipText.set(PriceSpecFormatter.getFormattedPriceSpecWithOfferPrice(bisqEasyOffer.getPriceSpec(), marketPriceService, bisqEasyOffer));
                 });
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListItem.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListItem.java
@@ -48,7 +48,7 @@ import lombok.extern.slf4j.Slf4j;
 
 import java.util.Comparator;
 import java.util.List;
-import java.util.OptionalDouble;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 @Slf4j
@@ -78,7 +78,7 @@ public class OfferbookListItem {
     // Empty while the offer has no resolvable percentage (a fixed-price offer without a
     // market price yet); comparators place these after known values instead of treating
     // them as an exactly-at-market 0%.
-    private OptionalDouble priceSpecAsPercent = OptionalDouble.empty();
+    private Optional<Double> priceSpecAsPercent = Optional.empty();
 
     public OfferbookListItem(BisqEasyOfferbookMessage bisqEasyOfferbookMessage,
                              UserProfile senderUserProfile,
@@ -134,7 +134,7 @@ public class OfferbookListItem {
         // price arrives; the table cells bind to them so visible rows follow.
         PriceUtil.findPercentFromMarketPrice(marketPriceService, bisqEasyOffer)
                 .ifPresent(percent -> {
-                    priceSpecAsPercent = OptionalDouble.of(percent);
+                    priceSpecAsPercent = Optional.of(percent);
                     formattedPercentagePrice.set(PercentageFormatter.formatToPercentWithSignAndSymbol(percent));
                     priceTooltipText.set(PriceSpecFormatter.getFormattedPriceSpecWithOfferPrice(bisqEasyOffer.getPriceSpec(), marketPriceService, bisqEasyOffer));
                 });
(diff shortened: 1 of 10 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListItem.java
```diff
@@ -122,9 +125,15 @@ boolean isBuyOffer() {
     }
 
     private void updatePriceSpecAsPercent() {
-        priceSpecAsPercent = PriceUtil.findPercentFromMarketPrice(marketPriceService, bisqEasyOffer).orElseThrow();
-        formattedPercentagePrice = PercentageFormatter.formatToPercentWithSignAndSymbol(priceSpecAsPercent);
-        priceTooltipText = PriceSpecFormatter.getFormattedPriceSpecWithOfferPrice(bisqEasyOffer.getPriceSpec(), marketPriceService, bisqEasyOffer);
+        // A fixed-price offer in a market without a market price has no percent to derive;
+        // the properties keep their previous values and refresh through the observer once a
+        // price arrives; the table cells bind to them so visible rows follow.
+        PriceUtil.findPercentFromMarketPrice(marketPriceService, bisqEasyOffer)
```
KimStrand: A missing percentage is currently represented by the primitive default `0.0` here. Both price-column comparators consume `priceSpecAsPercent`, so a fixed-price offer without a market quote is displayed with a blank percentage but sorted as though it were exactly at market. The new tests currently codify that sentinel value.

Should we model the percentage as explicitly unavailable and update both comparators to place unknown percentages consistently, such as after known values? The same issue applies to `ProfileCardOfferListItem`.

balag3: Fixed in 4a72fbc029 — the percentage is now modelled as explicitly unavailable (OptionalDouble), unknowns sort after known values in both directions, and a resolving percentage re-sorts its row. Same for ProfileCardOfferListItem.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListModel.java
```diff
@@ -32,7 +32,11 @@
 @Getter
 class OfferbookListModel implements bisq.desktop.common.view.Model {
     private final Set<String> chatMessageIds = new HashSet<>();
-    private final ObservableList<OfferbookListItem> offerbookListItems = FXCollections.observableArrayList();
+    // Extractor on the formatted percentage: it is set together with the sortable
+    // percentage, so a market price that resolves a previously empty percentage emits a
+    // list-update event and the SortedList re-sorts the row to its new position.
+    private final ObservableList<OfferbookListItem> offerbookListItems =
+            FXCollections.observableArrayList(item -> new javafx.beans.Observable[]{item.getFormattedPercentagePrice()});
```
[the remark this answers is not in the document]
KimStrand: Should we close this as not applicable under the current model? The only percentage that changes after construction is a fixed-price percentage, and PriceUtil.getPercentageToMarketPrice() rounds it to four fractional digits—the same resolution PercentageFormatter renders as two decimal percentage points. MarketPriceSpec and FloatPriceSpec values are immutable for the row, so the numeric comparator key cannot change while the observed formatted property remains equal.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListView.java
```diff
@@ -403,19 +404,34 @@ private void configOffersTableView() {
         tableView.getColumns().add(userProfileColumn);
         tableView.getSortOrder().add(userProfileColumn);
 
+        BisqTableColumn<OfferbookListItem>[] priceColumnHolder = new BisqTableColumn[1];
```
KimStrand: Should we build `priceColumn` first and set its comparator afterward instead of using a raw generic array as a mutable holder? `new BisqTableColumn[1]` introduces an unchecked-conversion warning, and this holder pattern does not appear elsewhere in the desktop code. Calling `priceColumn.setComparator(...)` after `build()` would also remove the `Supplier` indirection. The same applies to `ProfileCardOffersView`.

KimStrand (commented):
Other minor cleanup:
  - Sort the Optional/OptionalDouble imports.
  - Replace fully qualified Java/JavaFX types with imports.
  - Keep final fields before mutable fields in both list items.
  - Consider @VisibleForTesting for the package-private comparator helper.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListView.java
balag3: Done — both columns are built first and get their comparator afterward. The raw holders and Supplier indirection are gone, along with the minor import/field-order cleanup.

KimStrand (commented):
Should we update or remove the Remarks section now that the list extractors trigger re-sorting when the percentage changes? It still says live price updates never re-sort an already-sorted table, which no longer matches the implementation or the Fix section.

balag3:
right, thats stale, removed that section

apps/desktop/desktop/src/test/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListItemTest.java
```diff
@@ -0,0 +1,125 @@
...
+        boundLabel.textProperty().bind(item.getFormattedPercentagePrice());
+
+        MarketPrice marketPrice = mock(MarketPrice.class);
+        when(marketPrice.getPriceQuote()).thenReturn(PriceQuote.fromFiatPrice(40_000, "USD"));
+        when(marketPriceService.findMarketPrice(MARKET)).thenReturn(Optional.of(marketPrice));
+        marketPriceByCurrencyMap.put(MARKET, marketPrice);
+        WaitForAsyncUtils.waitForFxEvents();
+
+        assertThat(item.getPriceSpecAsPercent()).hasValue(0.25);
+        assertThat(item.getFormattedPercentagePrice().get()).isNotNull();
```
KimStrand: Should we assert the recovered display value itself here, or at least assert that it is no longer data.na? isNotNull() followed by equality with the same bound property would still pass if formattedPercentagePrice remained N/A, so this does not currently protect the visible recovery behavior described by the PR.

KimStrand (commented):
apps/desktop/desktop/src/main/java/bisq/desktop/main/content/user/profile_card/offers/ProfileCardOffersController.java:60 is outside the diff:

Should we align the profile-card offers lifecycle with the other list controllers, or explicitly track this as a follow-up? onDeactivate() disposes every row’s market-price pin but leaves the rows in the model, while onActivate() does not recreate or rebind them. Because the child controller is cached, switching away from Offers and back reuses disposed rows, so an N/A percentage can no longer recover and ordinary live percentage updates stop. setUserProfile() can also replace rows without first disposing their old pins.

This is pre-existing, but it directly weakens the PR’s recovery claim. The usual pattern is visible in apps/desktop/desktop/src/main/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListController.java:120: dispose and clear together.

apps/desktop/desktop/src/test/java/bisq/desktop/main/content/bisq_easy/offerbook/offerbook_list/OfferbookListItemTest.java
balag3: Confirmed. Fixed in: 5751c8bb4c

balag3:
Confirmed. Fixed in: e17d6bbf4b

KimStrand:
utACK

PR #4929, opened 2026-08-14
bugfix/android_sequenced_collection_crash_inventory_request_polic

rodvar:
### Motivation

with latest release Bisq Easy (mobile) 0.11.0 Google Play warns us that it could crash on Android 15 devices because of the code modified here (usage of certain list apis)

### Analysis

that code has been running in production in Android 15 devices as far as we are aware and haven't caused the warned issue... So we proceeded with the release anyways, but we propose this change to avoid future warnings and potential crashes in some devices (Android fragmentation).

### Conservative Solution

 - used JdkFacade to keep existing code for the rest of Bisq2 but change the usage of firstOf for Android, conservatively protecting from crashes in Android 15 running devices

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved list handling for inventory requests across supported Java and Android environments.
  * Added compatibility for Android versions that do not support newer list APIs.
  * Improved peer authorization token selection across supported platforms.
  * Preserved existing candidate-selection behavior while improving cross-platform reliability.
  * Reduced the risk of platform-specific issues when processing inventory data and selecting peers.

### DIFF
--- a/network/network/src/main/java/bisq/network/p2p/node/authorization/AuthorizationService.java
+++ b/network/network/src/main/java/bisq/network/p2p/node/authorization/AuthorizationService.java
@@ -17,7 +17,6 @@
 
 package bisq.network.p2p.node.authorization;
 
-import bisq.common.facades.FacadeProvider;
 import bisq.network.p2p.message.EnvelopePayloadMessage;
 import bisq.network.p2p.node.Feature;
 import bisq.network.p2p.node.authorization.token.equi_hash.EquiHashTokenService;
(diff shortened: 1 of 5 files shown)


rodvar:
@KimStrand this is a short easy one :)

rodvar:
after upgrading R8 I realised some more changes are needed - working on it

rodvar:
done 👍

network/network/src/main/java/bisq/network/p2p/node/authorization/AuthorizationService.java
```diff
@@ -139,7 +140,7 @@ static AuthorizationTokenType selectAuthorizationTokenType(List<AuthorizationTok
         return myPreferredAuthorizationTokenTypes.stream()
                 .filter(peersAuthorizationTokenTypes::contains)
                 .findFirst()
-                .orElseGet(peersAuthorizationTokenTypes::getFirst);
+                .orElseGet(() -> FacadeProvider.getJdkFacade().getFirst(peersAuthorizationTokenTypes));
```
KimStrand: When no preferred token type matches, this now calls FacadeProvider.getJdkFacade(). Pure callers and existing unit tests do not initialize that global, causing NullPointerException: jdkFacade is not set.
Using peersAuthorizationTokenTypes.get(0) directly is Android-compatible and avoids the new dependency.

KimStrand (commented):
utACK

PR #4921, opened 2026-08-11
feature/fsm_threading_improvements

rodvar:
This is the first of a series of 5 PRs as explained in https://github.com/bisq-network/bisq2/pull/4885#issuecomment-5248651679

 - "PR A" extracted from #4885
 - implemented overall hardening of FSM state machine threading.
 - SnapshotLockTimeoutException — new, with javadoc explaining the must-propagate contract.
 - added AI generated tests

## Summary by CodeRabbit

* **Improvements**
  * Improved state-machine concurrency and synchronization.
  * Added reliable snapshots of current state and queued events.
  * Added tools to clear, drain, or selectively remove queued events.
  * Improved final-state persistence handling.
  * Enhanced error reporting while avoiding exposure of event details.
* **Bug Fixes**
  * Fixed event-queue restoration and draining behavior.
  * Preserved informative handling for null events and lock timeouts.
* **Tests**
  * Added regression coverage for queue management, snapshots, persistence, timeouts, and error handling.

PR #4931, opened 2026-08-14
Bugfix/fsm out of order message causing stuck trades

### LINKED ISSUE #4946: Out-of-order trade message parked in the FSM event queue is lost on restart
Split out of #4885 per review feedback on #4933, which contributes the failing proof test.

### What happens

When a trade protocol message arrives out of order (e.g. "payment sent" before the Bitcoin-address message), the FSM parks it in its in-memory event queue until the enabling state is reached. That queue is not persisted. On restart the parked message disappears — while the network layer long since ACKed it, so the sender's auto-resends never fire and `MessageDeliveryStatus` reads delivered.

When the enabling message then arrives after the restart, it is processed normally, but the trade halts one state short of where it should be — permanently, since the parked message is unrecoverable from either side.

This is the inverse of the intuitive expectation that a restart "fixes" a stuck trade: for this mechanism a restart is precisely what loses the message.

### Repro

`BisqEasyTradeStuckByOutOfOrderMessageProofTest#parkedOutOfOrderMessageSurvivesRestartSoTradeCompletesWhenEnablingMessageArrives` (#4933) — fails on main: the restored trade stops one state too early.

### Direction

The parked queue must survive restart: persist it with the trade store, and drain it against the restored state on load. #4921 landed supporting plumbing; the fix should keep transitions single-source-of-state so the drained queue cannot race live message delivery. The failing test is the acceptance criterion.

### Impact

Root cause of trades stuck mid-protocol observed in the field (bisq-network/bisq-mobile#1622), where restarts left trades stranded one state behind. Sibling issue: the startup handoff window loss (posted separately).

rodvar:
- ~~resolves https://github.com/bisq-network/bisq-mobile/issues/1622~~ -> this will be added to PR "C"
 - resolves #4946
 - depends on #4921 
 - ported #4933 (with the unresolved one disabled for now until PR C lands) into this solution
 - code extracted from already working PR #4885 as per PR splitting description (this is labelled as "PR B" dependent on "PR A")
 - added comments are meaningful but if reviewers don't agree happy to review and trim/erase

## New Features
- Restored trades retain pending messages and automatically process them when required information becomes available.
- Trade state and queued messages are saved consistently, including at final state.
- Added safer management of queued trade events.

## Bug Fixes
- Removed queued messages from banned senders.
- Deferred processing during emergency halts or version restrictions, with automatic recovery afterward.
- Isolated trade restoration failures so one trade does not block others.
- Improved recovery for out-of-order messages.
- Improved persistence reliability after temporary save failures.
- Prevented incomplete data from replacing valid saved state.

### DIFF
--- a/common/src/main/java/bisq/common/fsm/FsmModel.java
+++ b/common/src/main/java/bisq/common/fsm/FsmModel.java
@@ -24,18 +24,53 @@ import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Predicate;
 
 @Slf4j
 @ToString
 @EqualsAndHashCode
 public class FsmModel {
+    /**
+     * Bound for {@link #getStateAndEventQueueSnapshot()}'s {@code tryLock}. Chosen short enough that the single,
+     * JVM-wide {@code Persistence} executor thread (see {@code bisq.persistence.Persistence#EXECUTOR}) can never
+     * be stuck for long on one trade's live transition - which would otherwise queue up every other store's
+     * writes behind it - yet long enough to comfortably clear the lock's normal, fast holders (a transition's
+     * state mutation plus a non-blocking persist() submission). A timed-out snapshot fails this write cycle
+     * rather than risk a torn read; RateLimitedPersistenceClient's generation pair keeps the store dirty so the
+     * next persist()/shutdown fallback retries - by which time a merely-slow (not stuck) handler has typically
+     * released the lock.
+     */
+    static final long SNAPSHOT_LOCK_TIMEOUT_MS = 500;
+
     private final Observable<State> state = new Observable<>();
 
-    // Package visibility for access from Fsm mutating the collections 
-    final Set<Event> eventQueue = new HashSet<>();
-    final Set<Class<? extends Event>> processedEvents = new HashSet<>();
+    // Explicit lock replacing the previous intrinsic-monitor use (synchronized(model) in Fsm#handle()/
+    // drainEventQueue(), synchronized methods below): ReentrantLock, unlike synchronized, offers a bounded
+    // tryLock(timeout), which getStateAndEventQueueSnapshot() below needs. Every party that used to synchronize
+    // on the model instance must now go through this SAME lock object - Fsm#handle()/drainEventQueue() included -
+    // or mutual exclusion between a live transition and a persistence-facing read silently breaks.
+    @EqualsAndHashCode.Exclude
+    @ToString.Exclude
+    final ReentrantLock lock = new ReentrantLock();
+
+    // Package visibility for access from Fsm mutating the collections.
+    // CopyOnWriteArraySet rather than a plain HashSet, as defense-in-depth for any reader that bypasses
+    // getStateAndEventQueueSnapshot() and calls getEventQueue()/getProcessedEvents() directly without acquiring
+    // the shared lock field above (Fsm#handle()/Fsm#drainEventQueue() run under that same lock - see
+    // getStateAndEventQueueSnapshot() below for why). Concurrently iterating a plain HashSet while another thread
+    // structurally mutates it is undefined behavior (ConcurrentModificationException or a torn read). Both sets
+    // are always small (a handful of pending FSM events per trade at most) and read far more often than mutated,
+    // which is exactly CopyOnWriteArraySet's sweet spot; it also gives every such reader (including
+    // Fsm#drainEventQueue()'s defensive copy-before-iterate, which a plain HashSet copy does not actually make
+    // safe under concurrent mutation) consistent snapshot-iterator semantics for free, without requiring any
+    // external synchronization. The primary, recommended mechanism for a cross-thread, all-fields-consistent read
+    // (e.g. Trade#getTradeBuilder during persistence) is getStateAndEventQueueSnapshot(), not these getters.
+    final Set<Event> eventQueue = new CopyOnWriteArraySet<>();
+    final Set<Class<? extends Event>> processedEvents = new CopyOnWriteArraySet<>();
 
     public FsmModel(State initialState) {
         if (initialState == null) {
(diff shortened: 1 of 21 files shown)


common/src/main/java/bisq/common/fsm/FsmModel.java
```diff
@@ -73,4 +108,102 @@ public Set<Event> getEventQueue() {
...
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            log.warn("Interrupted while waiting up to {}ms to acquire the FsmModel lock during serialization; " +
+                    "write cycle will retry. modelClass={}", SNAPSHOT_LOCK_TIMEOUT_MS, getClass().getSimpleName());
+            throw new SnapshotLockTimeoutException(this, SNAPSHOT_LOCK_TIMEOUT_MS, e);
+        }
+        if (!acquired) {
+            log.warn("FsmModel lock not acquired within {}ms during serialization - a live transition is likely " +
+                    "holding it; write cycle will retry. modelClass={}", SNAPSHOT_LOCK_TIMEOUT_MS, getClass().getSimpleName());
+            throw new SnapshotLockTimeoutException(this, SNAPSHOT_LOCK_TIMEOUT_MS);
```
KimStrand, line 157: PersistableStoreReaderWriter.write() catches this exception and returns normally, so persistAsync() treats the write as successful and no retry is retained. Should we instead let the timeout escape and preserve a dirty/retry state, including during shutdown, so the stale snapshot is not treated as successfully persisted?

trade/src/main/java/bisq/trade/bisq_easy/protocol/BisqEasyProtocol.java
```diff
@@ -58,6 +58,16 @@ protected void persist() {
...
     }
 
+    @Override
+    protected void persistOnFinalState() {
+        // Thin hook for now: delegates to the same rate-limited persist() as every other transition. A final
+        // state's in-memory clear of eventQueue/processedEvents (see Fsm#handle) is therefore not yet guaranteed
+        // to reach disk promptly - RateLimitedPersistenceClient does not expose an unthrottled write path on this
+        // branch. This will be swapped to an unthrottled write (see Fsm#persistOnFinalState's javadoc) once that
+        // lands.
+        persist();
```
KimStrand: This still uses the rate-limited persistence path, so the final-state cleanup can be dropped when another write is running or occurred recently. Should we instead use a forced/unthrottled persistence path here, and in MuSigProtocol, to ensure the final state and cleared queue reach disk?

common/src/main/java/bisq/common/fsm/FsmModel.java
```diff
@@ -73,4 +108,102 @@ public Set<Event> getEventQueue() {
...
+     * {@link #getStateAndEventQueueSnapshot()}, mirroring the final-state cleanup in {@link Fsm#handle(Event)}.
+     * Takes the lock unbounded, like {@link Fsm#handle(Event)}: this is called from the trade services, not from
+     * the persistence-serialization path, so there is no shared-executor-stall concern to bound against here.
+     * <br/>
+     * Used by the trade data-retention/redaction pass: the queue can hold sensitive account-data network messages
+     * (e.g. {@code BisqEasyAccountDataMessage} / MuSig {@code SendAccountPayloadMessage}) which are persisted with
+     * the trade. A stuck trade may never reach a final state to clear them, so they must be scrubbed once the trade
+     * passes the redaction threshold, otherwise the persisted copy would linger on disk indefinitely.
+     */
+    public void clearEventQueue() {
```
KimStrand, line 183: Now that pending network events are persisted, should we include this queue in the production retention cleanup? Otherwise old trades can retain account-data messages after the normal payment-account field has been redacted. We would also need to avoid the Bisq Easy early return when only the queue changed and implement the equivalent MuSig cleanup.

trade/src/main/java/bisq/trade/bisq_easy/BisqEasyTradeService.java
```diff
@@ -456,6 +472,43 @@ private BisqEasyProtocol createAndAddTradeProtocol(BisqEasyTrade trade) {
...
+            // The queued events passed onMessage()'s guards when they were originally received, but those
+            // conditions may have changed before the restart. Draining applies them directly through the FSM,
+            // so re-check here what onMessage() would check for a live message:
+            //  - Trading halt / min-version (global, temporary): defer the drain entirely and KEEP the events
+            //    queued - they persist and drain on a later restart once the emergency alert is lifted.
+            //    Deliberately no throw: the catch below would mislabel it "stuck until investigated".
+            //  - Banned sender (per event): DROP the event, mirroring onMessage() ignoring a banned sender's
+            //    live message. Kept queued it would get applied after a ban-lift restart - something
+            //    onMessage() would never have allowed.
+            if (haltTrading || isMinVersionForTradingViolated()) {
```
KimStrand, line 448: Should we drain these queued events when onRemoved()/onCleared() lifts the restriction, rather than waiting for a restart? If no subsequent live transition arrives, the restored trade otherwise remains stuck after trading resumes. The equivalent MuSig drain would need to run on its executor.

KimStrand, line 448: Should we drain these queued events when onRemoved()/onCleared() lifts the restriction, rather than waiting for a restart? If no subsequent live transition arrives, the restored trade otherwise remains stuck after trading resumes. The equivalent MuSig drain would need to run on its executor.

KimStrand (commented):
One broader readability point: this change adds many comments that restate what the code is doing, which makes the control flow harder to scan and risks the comments becoming outdated. Should we trim them down and keep only comments that explain non-obvious reasons, constraints, or trade-offs?

common/src/main/java/bisq/common/fsm/FsmModel.java
```diff
@@ -73,4 +109,110 @@ public Set<Event> getEventQueue() {
...
+
+    /**
+     * Immutable {state, eventQueue} pair captured atomically by {@link #getStateAndEventQueueSnapshot()}.
+     */
+    public record StateAndEventQueue(State state, Set<Event> eventQueue) {
+        public StateAndEventQueue {
+            // Enforce the documented immutability in the type itself rather than relying on caller discipline;
+            // no-op when the caller already passes an immutable set (Set.copyOf returns it unchanged).
+            eventQueue = Set.copyOf(eventQueue);
+        }
```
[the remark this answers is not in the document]
rodvar: I think we skip this - The finding is technically true as an observation but describes a threat with no mutator, and the proposed fix is a codebase-wide convention change smuggled into this PR.

rodvar:
@KimStrand thank you for your review once again. I've addressed most of the comments above (including CRAI one) in this PR except the ones I'm describing in this comment as I believe they belong to their own follow up PR:

 1. `Unthrottled persistOnFinalState()` - needs a forced write path on RateLimitedPersistenceClient, which touches every store client in the system, not just trades. The swallow fix above narrows the window: a dropped final write now leaves dropped=true, so the shutdown fallback picks it up.
 2. `Retention cleanup of the persisted queue` - needs skipping the Bisq Easy early return when only the queue changed, plus the MuSig equivalent. Exposure is bounded meanwhile: a queued account-data message only outlives redaction on a trade that never reaches a final state. clearEventQueue() javadoc now says explicitly it's not wired yet.
 3. `Drain on halt/version-restriction lift` - belongs with the follow-up PR that re-enables the disabled #4933 case (restart drain covers correctness until then; the halt path is rare and short-lived).

I hope we can agree on this so we can first focus on merging this one first. I can add this follow up to the next PR that will address #4947 💪

KimStrand:
Thanks. The snapshot-timeout change addresses my first comment.

I do not think #4947 tracks the other items, since that issue only covers the startup listener-handoff race. The shutdown fallback also only covers a clean shutdown, while RateLimitedPersistenceClient explicitly excludes crashes and forced termination. A stuck trade that never reaches a final state is also exactly the case covered by the existing retention cleanup, rather than an unrelated edge case.

Should we address the forced final-state write and queue retention in this PR before enabling persistence of account-data events? If draining after an alert is lifted should be deferred, should we create a dedicated tracking issue for it rather than attach it to #4947?

common/src/test/java/bisq/common/fsm/FsmTest.java
```diff
@@ -158,6 +170,341 @@ void testOutOfOrderEvents() {
...
+                .from(MockState.S1)
+                .on(MockEvent2.class)
+                .run(MockEventHandler.class)
+                .to(MockState.S2);
+
+        fsm.handle(new MockEvent2(model, "queued"));
+        assertEquals(MockState.INIT, fsm.getModel().getState());
+        assertEquals(1, model.getEventQueue().size());
+
+        // "Restart": rebuild the model purely from the persisted snapshot (state + queue + processedEvents),
```
KimStrand: This comment refers to Trade#processedEventsFromProto, but no such method exists and production deliberately restores processedEvents with Set.of(). In this test getProcessedEvents() is empty, so passing it does not test restoration either. Should we make the fixture mirror production with Set.of() and trim this explanation accordingly?

rodvar:
> Thanks. The snapshot-timeout change addresses my first comment.
> 
> I do not think #4947 tracks the other items, since that issue only covers the startup listener-handoff race. The shutdown fallback also only covers a clean shutdown, while RateLimitedPersistenceClient explicitly excludes crashes and forced termination. A stuck trade that never reaches a final state is also exactly the case covered by the existing retention cleanup, rather than an unrelated edge case.
> 
> Should we address the forced final-state write and queue retention in this PR before enabling persistence of account-data events? If draining after an alert is lifted should be deferred, should we create a dedicated tracking issue for it rather than attach it to #4947?

great thanks Kim, I've overstated that on 4947. I decided to address it here; will let you know when its ready - thank you
The other comment is easy fix will address as well.

HenrikJannsen:
Can you resolve the conflicts?

rodvar:
will address conflicts early next week

rodvar:
@KimStrand @HenrikJannsen conflicts addressed - PR is ready again 👍

rodvar:
rebased to for mobile 2.1.13 branch

PR #4932, opened 2026-08-14
fix(api): decline malformed deflate offers

wodoro:
follow up on #4907 and #4903

-----

RFC 7692 requires a server to decline a permessage-deflate offer that repeats a parameter, or that gives a value to client_no_context_takeover or server_no_context_takeover, both of which are flags. Neither was checked, so such offers were accepted.

The header was also split on "," and ";" without regard for RFC 6455 quoted parameter values, in which those characters are data. That could read one offer as several, letting the value of an unrelated extension pose as a bare permessage-deflate offer the client never made, and we would then compress frames it never agreed to decompress. A single pass that tracks quoting replaces both splits and declines a header holding a quote that is never closed.

Declining leaves the connection uncompressed rather than failing the handshake, and no extension header is sent, so a client that offered something malformed falls back to plain frames.

Also carries the earlier follow-ups on this branch: the handshake response branches in handleWrite folded into one, a note on buffer disposal in deflateFrame, and a field order fix in ConnectionState.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket compression negotiation for quoted and malformed parameters.
  * Rejected duplicate, empty, and invalid extension parameters more reliably.
  * Ensured connection state is cleaned up when negotiation fails.
  * Correctly updates compression settings after successful upgrades.

* **Tests**
  * Expanded coverage for invalid offers, quoted values, duplicate parameters, and resource cleanup.

### DIFF
--- a/api/src/main/java/bisq/api/web_socket/compression/PerMessageDeflateFilter.java
+++ b/api/src/main/java/bisq/api/web_socket/compression/PerMessageDeflateFilter.java
@@ -39,9 +39,11 @@ import org.glassfish.grizzly.websockets.ProtocolError;
 import java.io.ByteArrayOutputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
+import java.util.Set;
 import java.util.zip.DataFormatException;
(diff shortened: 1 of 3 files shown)


KimStrand (commented):
utACK

PR #4938, opened 2026-08-16
Validate MuSig take offer requests on the maker side

### LINKED ISSUE #4927: MuSig maker side is missing the take-offer validations that Bisq Easy has
While working on the take-offer use case for #4883 I ran into a few maker-side gaps in the MuSig trade protocol. They all pre-exist on main and none of the affected files are touched by the take-offer work, so I'm filing them separately rather than mixing them into that series. Listing them together because they share one theme: the Bisq Easy maker handlers already have these checks and the MuSig handlers don't.

1. The maker builds its trade straight from the taker's contract and never validates it against the local offer. `MuSigTradeService.handleMuSigTakeOfferMessage` constructs the trade from `message.getContract()`, and `SetupTradeMessage_A_Handler` then compares `message.getContract()` with `trade.getContract()`, which is the same object, so the contract equality check cannot fail. Amounts, price, rail and mediator are all taker-authored. Example: a sell offer of 0.1 BTC for 10,000 USD can be taken with a contract stating 1 BTC for 1 USD; the maker signs it and `getTradeAmount()` feeds the 1 BTC to musigd. The Bisq Easy maker handler validates the taker amounts against the local offer; the MuSig handler has no analog.

2. The maker never checks that the offer is still its own live offer. Deactivating or removing an offer does not stop a peer with a cached copy from taking it, since `makerCreatesProtocol` only requires the embedded maker identity and account option to resolve. Bisq Easy looks the offer up locally and rejects when it is gone.

3. The contract signature is not bound to the claimed taker identity. `MuSigContractVerifier` verifies against the public key supplied inside `ContractSignatureData`, and nothing compares the network-level sender with `contract.taker.networkId`. A sender can submit a contract naming a different profile as taker, signed with the sender's own key. The Bisq Easy handler checks both sender-to-contract identity and key-to-NetworkId equality.

4. Trades are persisted before the signature is verified. `addTrade`, persistence and protocol creation all run before `verify`; a failed signature moves the trade to an error state but does not remove it. The take date is part of the trade id and any date inside the accepted window passes, so one peer can persist an unbounded number of durable failed trades from a single cached offer.

5. `NonceSharesRequestUtil` hard-codes the deposit (30,000 sats) and the trade fee (5,000 sats), so the review screen's percentage-based deposit and the fee display cannot match what the protocol actually requests. The fee side of this is tied to the open fee schedule decision; the deposit side has the same shape.

None of this blocks dev testing of the make/take flow (the backend is still mocked), but since MuSig is pre-release this looks like the right time to close the gaps on the protocol side. Should these be filed as separate issues, or is this covered by planned protocol work? Happy to take some of them if wanted.

balag3:
Adds the maker-side take-offer validations the MuSig protocol was missing, so a
`SetupTradeMessage_A` is fully checked before any trade, protocol or contact-list
entry is created. This covers the validation gaps from the issue (items 1-3, and
the persist-before-verify concern); the trade fee / deposit hardcoding is left out
of scope and tracked separately.

The maker now validates the request in `MuSigTradeService.handleMuSigTakeOfferMessage`
before `makerCreatesProtocol`, and drops it on failure. The checks live in a new
`MuSigTakeOfferRequestValidator`:

- **Identity** — the message sender equals the contract's taker network id; the
  sender and contract-signature public keys are the taker's; the signature verifies;
  the receiver is the offer's maker; the message trade id equals the id derived from
  the contract.
- **Offer liveness** — the embedded offer must be one of the maker's currently
  activated offers and equal to the retained offer (compared by its serialize-for-hash
  form, which also covers the currency names that `equals` omits). No fallback to an
  existing trade, so a cached copy of a removed offer cannot be replayed. The contract's
  protocol type and party roles are pinned, and its take date cannot predate the offer.
- **Economics** — positive amounts; the contract price spec equals the offer's; a
  present taker account-payload commitment; payment-method-spec membership; the
  amount-spec-pinned side matches the offer exactly and the free side is within the
  configured price tolerance in both directions (an inflated favorable side cannot slip
  past the rail cap); absolute and payment-rail USD limits, valued with exact arithmetic
  against a single fresh market-price snapshot; and the deterministic mediator and
  arbitrator selections, compared over the whole profile.

Two follow-up commits address the creation path:

- Creation is serialized under a lock so two concurrent same-id requests cannot race a
  second trade/protocol into the registries or throw on the network thread, and the
  peer is added to the contact list only after the handler has verified and processed
  the request (best-effort, so it cannot fail the trade).
- A review pass tightened three residuals: the rail cap is valued on the actual USD
  obligation for USD-quoted markets, the dispute-agent comparison covers the profile's
  version metadata, and the maker identity/account lookups drop cleanly (requiring an
  active identity) instead of throwing.

### Commits

- Validate take offer identity before trade creation
- Reject takes of offers that are not live own offers
- Move MuSig trade limits policy to the offer module
- Validate take offer economics on the maker side
- Serialize maker-side take offer trade creation
- Harden take offer validation from review findings

### Testing

`MuSigTakeOfferRequestValidator` has unit tests covering each check and its bypass
attempts (identity/signature binding, offer liveness and equality, price tolerance in
both directions and both price regimes, USD/rail limits and overflow, dispute-agent
forgery, take date, currency-name and role forgery). `:trade`, `:mu-sig`, `:contract`
and `:common` suites pass. The `MuSigTradeService` / handler changes (the gate wiring,
the creation lock and the contact-add relocation) are verified by inspection, matching
how the equivalent Bisq Easy handler wiring is covered.

### Known limitations / follow-ups

These are pre-existing and bounded; I kept them out to keep the change focused:

- A taker can open more than one trade against a single offer by varying `takeOfferDate`
  (a different trade id each time). The creation lock only prevents same-id races. Bounding
  this needs a product decision — MuSig currently allows retaking an offer — so I did not
  want to enforce a one-take-per-offer rule here. **How would you like takes-per-offer
  bounded given retake is allowed?**
- The rail-cap obligation is valued exactly only for USD-quoted markets; other fiat and
  altcoin quotes keep the tolerance-bounded (~5%) approximation until a fiat/USD conversion
  is added.
- `MuSigContract.marketPrice`, the offer version strings and `protocolVersion` are
  taker-supplied and hash-relevant but not consumed by any calculation; they are persisted
  as-is rather than canonicalized.
- `takeOfferDate` is only required to be at or after the offer date; the generic date
  bound still allows up to ~2h in the future.
- `PubKeyShares.verify()` is still a TODO on main.

### Remarks

- The "Validate take offer economics" commit message says the dispute-agent selections are
  compared "by their full serialized form"; the final code compares the profile field by
  field (equals plus the three version fields), which is equivalent. The last commit
  implements that precisely.

Fixes #4927

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive validation for MuSig offer requests, including identities, offers, payment details, prices, limits, and dispute-agent assignments.
  * Added activated-offer lookup, payment-rail-specific trade limits, and sanitized logging.
  * Improved handling of invalid, duplicate, ignored, and concurrent trade requests.

* **Bug Fixes**
  * Prevented contact-list update failures from interrupting accepted trades.
  * Improved rejection reporting for invalid trade requests.

* **Tests**
  * Added coverage for validation, trade limits, offer handling, concurrency, and rejection reporting.

### DIFF
--- a/trade/src/main/java/bisq/trade/mu_sig/MuSigTradeService.java
+++ b/trade/src/main/java/bisq/trade/mu_sig/MuSigTradeService.java
@@ -389,18 +389,25 @@ public final class MuSigTradeService extends RateLimitedPersistenceClient<MuSigT
             // must not leave a trade, a protocol entry or a contact-list entry behind. The
             // handler keeps its own check as defense in depth.
             log.warn("Dropping a take offer request from an ignored taker. tradeId={}",
-                    StringUtils.truncate(message.getTradeId().replaceAll("[\\p{Cntrl}\\u2028\\u2029]", "?"), 60));
+                    StringUtils.sanitizeForLog(message.getTradeId()));
             return;
         }
         try {
             // The request is validated before any trade is created or persisted; a rejected
             // request must not leave a failed trade, a protocol entry or a contact-list entry.
             MuSigTakeOfferRequestValidator.validateIdentity(serviceProvider.getContractService(), message);
+            MuSigTakeOfferRequestValidator.validateTakerProfileKnown(userProfileService, message);
             MuSigTakeOfferRequestValidator.validateOffer(
                     serviceProvider.getOfferService().getMuSigOfferService().getMyMuSigOffersService(), message);
+            MuSigTakeOfferRequestValidator.validateEconomics(
+                    serviceProvider.getBondedRolesService().getMarketPriceService(),
+                    settingsService,
+                    muSigTraderMediationService,
+                    muSigTraderArbitrationService,
+                    message);
         } catch (TradeProtocolException e) {
             log.warn("Dropping an invalid MuSig take offer request. tradeId={}",
-                    StringUtils.truncate(message.getTradeId().replaceAll("[\\p{Cntrl}\\u2028\\u2029]", "?"), 60));
+                    StringUtils.sanitizeForLog(message.getTradeId()));
             return;
         }
         MuSigContract muSigContract = message.getContract();
--- a/trade/src/main/java/bisq/trade/mu_sig/MuSigTradeService.java
+++ b/trade/src/main/java/bisq/trade/mu_sig/MuSigTradeService.java
@@ -397,19 +397,20 @@ public final class MuSigTradeService extends RateLimitedPersistenceClient<MuSigT
             return;
         }
         try {
+            MuSigOffer claimedOffer;
             try {
                 // The request is validated before any trade is created or persisted; a rejected
                 // request must not leave a failed trade, a protocol entry or a contact-list entry.
                 MuSigTakeOfferRequestValidator.validateIdentity(serviceProvider.getContractService(), message);
                 MuSigTakeOfferRequestValidator.validateTakerProfileKnown(userProfileService, message);
-                MuSigTakeOfferRequestValidator.validateOffer(
-                        serviceProvider.getOfferService().getMuSigOfferService().getMyMuSigOffersService(), message);
                 MuSigTakeOfferRequestValidator.validateEconomics(
                         serviceProvider.getBondedRolesService().getMarketPriceService(),
                         settingsService,
                         muSigTraderMediationService,
                         muSigTraderArbitrationService,
                         message);
+                claimedOffer = MuSigTakeOfferRequestValidator.validateOffer(
+                        serviceProvider.getOfferService().getMuSigOfferService().getMyMuSigOffersService(), message);
             } catch (TradeProtocolException e) {
                 throw e;
             } catch (RuntimeException e) {
@@ -421,7 +422,11 @@ public final class MuSigTradeService extends RateLimitedPersistenceClient<MuSigT
                         e);
             }
             MuSigContract muSigContract = message.getContract();
-            makerCreatesProtocol(muSigContract, message.getTradeId(), message.getSender(), message.getReceiver())
+            makerCreatesProtocol(muSigContract,
+                    claimedOffer,
+                    message.getTradeId(),
+                    message.getSender(),
+                    message.getReceiver())
                     .ifPresent(protocol -> handleMuSigTradeMessage(message, protocol));
         } catch (TradeProtocolException e) {
             log.warn("Dropping an invalid MuSig take offer request. tradeId={}",
@@ -718,15 +723,20 @@ public final class MuSigTradeService extends RateLimitedPersistenceClient<MuSigT
     // The maker has added the salted account id to the AccountOptions.
     // We will use the payment method chosen by the taker to determine which account we had assigned to that offer.
     public Optional<Account<? extends PaymentMethod<?>, ?>> findMyAccount(MuSigTrade trade) {
+        return findMyAccount(trade, trade.getOffer());
+    }
+
+    private Optional<Account<? extends PaymentMethod<?>, ?>> findMyAccount(MuSigTrade trade,
+                                                                           MuSigOffer claimedOffer) {
         MuSigContract contract = trade.getContract();
-        MuSigOffer offer = contract.getOffer();
         PaymentMethod<?> selectedPaymentMethod = contract.getNonBtcSidePaymentMethodSpec().getPaymentMethod();
         Set<Account<? extends PaymentMethod<?>, ?>> matchingAccounts = serviceProvider.getAccountService().getAccounts(selectedPaymentMethod);
-        Set<AccountOption> accountOptions = OfferOptionUtil.findAccountOptions(offer.getOfferOptions());
+        Set<AccountOption> accountOptions = OfferOptionUtil.findAccountOptions(claimedOffer.getOfferOptions());
         return accountOptions.stream()
                 .filter(accountOption -> accountOption.getPaymentMethod().equals(selectedPaymentMethod))
                 .map(AccountOption::getSaltedAccountId)
-                .flatMap(saltedAccountId -> OfferOptionUtil.findAccountFromSaltedAccountId(matchingAccounts, saltedAccountId, offer.getId()).stream())
+                .flatMap(saltedAccountId -> OfferOptionUtil.findAccountFromSaltedAccountId(
+                        matchingAccounts, saltedAccountId, claimedOffer.getId()).stream())
                 .findAny();
     }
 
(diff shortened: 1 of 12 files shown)


trade/src/main/java/bisq/trade/mu_sig/MuSigTradeService.java
```diff
@@ -654,27 +685,59 @@ public MusigGrpc.MusigStub getMusigAsyncStub() {
...
+        // Verification will happen in the MuSigTakeOfferRequestHandler.
+        // tradeId is the validated message trade id (== the id derived from the contract).
+        synchronized (tradeCreationLock) {
+            // A duplicate or concurrent request for the same trade id must not race a second trade
+            // or protocol into the registries; drop it cleanly instead of throwing on the network
+            // notify thread. The duplicate check runs before the identity/account lookups so a
+            // duplicate that arrives after the account or identity was removed still drops cleanly
+            // rather than throwing. Resuming a persisted INIT trade on a resend is a separate
+            // concern (see the take-offer follow-ups).
+            if (findProtocol(tradeId).isPresent() || tradeExists(tradeId)) {
```
KimStrand, line 758: This deduplicates only the exact tradeId, while Trade.createId includes the taker-controlled takeOfferDate. A known taker could therefore submit multiple otherwise-valid Setup A requests for the same offer using different timestamps, creating and persisting a separate trade/protocol for each one.

Should we add an atomic admission constraint independent of takeOfferDate, for example, allowing at most one pending setup per (offerId, takerNetworkId), with explicit retry, resume, and cleanup semantics?

trade/src/test/java/bisq/trade/mu_sig/messages/network/handler/maker/MuSigTakeOfferRequestValidatorTest.java
```diff
@@ -0,0 +1,938 @@
...
+
+        assertEconomicsRejected(contract, TradeProtocolFailure.MEDIATORS_NOT_MATCHING);
+    }
+
+    @Test
+    void mediatorWithForgedProfileMetadataIsRejected() {
+        // Same network id as the maker's selection, but a forged nickname: the full-profile
+        // comparison rejects it so the maker never signs an altered dispute-agent profile.
+        NetworkId mediatorNetworkId = mediatorProfile.getNetworkId();
+        UserProfile forged = org.mockito.Mockito.mock(UserProfile.class);
```
KimStrand: mediatorProfile and forged are separate Mockito mocks, so UserProfile.equals returns false by object identity regardless of the forged nickname. This means the test would still pass if the nickname specific validation were removed.

Should we use otherwise equal real UserProfile instances here and vary only the nickname, similar to the following avatar-version test?

trade/src/test/java/bisq/trade/mu_sig/messages/network/handler/maker/MuSigTakeOfferRequestValidatorTest.java
```diff
@@ -0,0 +1,938 @@
...
+        org.mockito.Mockito.when(forged.getNetworkId()).thenReturn(mediatorNetworkId);
+        org.mockito.Mockito.when(forged.getUserName()).thenReturn("forged");
+        MuSigOffer offer = createOffer(new BaseSideFixedAmountSpec(2_000_000L));
+        MuSigContract contract = createContract(offer, 2_000_000L, 10_000_000L, offer.getPriceSpec(),
+                Optional.of(forged), Optional.of(arbitratorProfile));
+
+        assertEconomicsRejected(contract, TradeProtocolFailure.MEDIATORS_NOT_MATCHING);
+    }
```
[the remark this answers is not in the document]
balag3: Agreed — the test was passing because of mock identity rather than the nickname difference. Fixed with otherwise-equal real profiles in 0b3df69adb.

trade/src/main/java/bisq/trade/mu_sig/MuSigTradeService.java
balag3: Good point. This is a deliberate scope decision and is called out in the PR body: the lock protects an exact trade ID, while varying `takeOfferDate` still allows multiple setups for the same offer and taker.

I avoided introducing a permanent one-trade-per-offer rule because MuSig currently allows retakes. My plan is to address this in a separate issue with one pending setup per `(offerId, takerNetworkId)` and explicit retry, resume, and cleanup semantics. Does that separation and scope sound right to you, or would you prefer it handled in this PR?

trade/src/test/java/bisq/trade/mu_sig/messages/network/handler/maker/MuSigTakeOfferRequestValidatorTest.java
balag3: Agreed — replaced the mocks with otherwise-equal real profiles that differ only in nickname. Fixed in 0b3df69adb.

trade/src/main/java/bisq/trade/mu_sig/MuSigTradeService.java
```diff
@@ -381,9 +388,82 @@ public void onMessage(EnvelopePayloadMessage envelopePayloadMessage) {
...
+                log.warn("Unexpected failure while validating a MuSig take offer request. tradeId={}",
+                        StringUtils.sanitizeForLog(message.getTradeId()), e);
+                throw new TradeProtocolException(
+                        "The maker has rejected the take offer request because the offer is not available anymore.",
+                        TradeProtocolFailure.OFFER_NOT_AVAILABLE,
+                        e);
+            }
+            MuSigContract muSigContract = message.getContract();
+            makerCreatesProtocol(muSigContract, message.getTradeId(), message.getSender(), message.getReceiver())
+                    .ifPresent(protocol -> handleMuSigTradeMessage(message, protocol));
```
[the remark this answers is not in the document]
balag3: Confirmed. Fixed in `7aefaa264c` by serializing offer-state changes with the final non-consuming offer claim and using the claimed local offer for maker-side trade creation. Added deterministic coverage for deactivation winning the state lock and for removal.

trade/src/main/java/bisq/trade/mu_sig/MuSigTradeService.java
KimStrand: The separation sounds reasonable to me, since this PR prevents unverified requests from being persisted and the remaining case concerns multiple otherwise-valid admitted takes.

Should we create and link the follow-up issue before resolving this thread, so the pending (offerId, takerNetworkId) bound and its retry, resume, and cleanup semantics are not lost?

KimStrand:
utACK

trade/src/main/java/bisq/trade/mu_sig/MuSigTradeService.java
balag3: Filed as #4945.

KimStrand:
utACK

PR #4940, opened 2026-08-17
Add two-party private chat to the API

salottodev:
## Summary by CodeRabbit

* **New Features**
  * Added two-party private chat via REST, including channels, messages, reactions, leaving channels, and notification retrieval.
  * Added real-time updates for private-chat channels, messages, reactions, and unread counts.
  * Added private-chat permissions and feature configuration.
  * Enhanced mobile notifications with chat channel and peer details for deep linking.
* **Bug Fixes**
  * Improved notification eligibility for private messages.
  * Prevented private messages and reactions from being sent when a participant is banned.
  * Prevented stale trade-chat entries after channels are cleared.
  * Reduced repeated security warnings for devices using fallback encryption.


api/src/main/java/bisq/api/access/permissions/Permission.java
```diff
@@ -52,7 +52,8 @@ public enum Permission implements ProtoEnum {
     TRADES(7, Kind.STANDARD),
     USER_IDENTITIES(8, Kind.STANDARD),
     USER_PROFILES(9, Kind.STANDARD),
-    MOBILE_DEVICES(10, Kind.STANDARD);
+    MOBILE_DEVICES(10, Kind.STANDARD),
+    PRIVATE_CHAT_CHANNELS(11, Kind.STANDARD);
```
rodvar: heads up - #4948 adds `CLIENT_MANAGEMENT` with this same id 11 (and proto 12) on the same base branch. The id is real wire format: `PairingCodeEncoder` writes `permission.getId()` into the pairing code and the decoder resolves via `Permission.fromId`, which returns the first matching constant - two constants sharing id 11 compiles fine and just silently resolves wrong, so a QR meant to grant private chat could decode as client revocation power (or vice versa). Whichever PR merges second please renumber to 12 (proto 13) as part of its own rebase, not as conflict-resolution afterthought. While here: worth a one-liner javadoc on why STANDARD is right for a content-privacy permission - existing pairings gain DM access on node upgrade without re-pairing, which is intended (paired devices are operator-approved), but let's write that down at the declaration.

api/src/main/java/bisq/api/web_socket/subscription/Topic.java
```diff
@@ -30,4 +30,7 @@ public enum Topic {
     ALERT_NOTIFICATIONS,
     TRADE_RESTRICTING_ALERT,
     NETWORK_INFO,
+    PRIVATE_CHAT_CHANNELS,
```
rodvar: as per the discussion in bisq-network/bisq-mobile#1750: these three topics ride the subscription path that has no per-topic permission check yet, so a pairing that withheld `PRIVATE_CHAT_CHANNELS` still receives the DM payloads over WS. We're tackling subscription authorization separately (topic->permission mapping mirroring `RestPermissionMapping`), nothing in this PR makes that harder - but please drop a short comment here referencing the gap so it doesn't get lost between the two efforts.

api/src/main/java/bisq/api/web_socket/subscription/SubscriptionService.java
```diff
@@ -231,6 +263,15 @@ private Optional<BaseWebSocketService> findWebSocketService(Topic topic) {
             case NETWORK_INFO -> {
                 return Optional.of(networkInfoWebSocketService);
             }
+            case PRIVATE_CHAT_CHANNELS -> {
```
KimStrand, line 268: The REST resource is protected by PRIVATE_CHAT_CHANNELS, but these native WebSocket subscriptions are only authenticated; subscribe() never checks the authenticated client's granted permissions. A paired client with an explicit grant that omits PRIVATE_CHAT_CHANNELS can therefore subscribe to all three topics and receive its complete private-chat state while the equivalent REST requests return 403.

Should we enforce the permission before adding the subscriber, ideally through a central topic-to-permission mapping, and add a test covering a denied subscription?

api/src/main/java/bisq/api/web_socket/domain/chat/private_chat/PrivateChatReactionsWebSocketService.java
```diff
@@ -0,0 +1,257 @@
...
+                chatMessagesPinsByChannelId.compute(channelId, (key, oldPin) -> {
+                    if (oldPin != null) {
+                        oldPin.unbind();
+                    }
+
+                    return channel.getChatMessages().addObserver(new CollectionObserver<>() {
+                        @Override
+                        public void onAdded(TwoPartyPrivateChatMessage message) {
+                            String messageId = message.getId();
+                            Map<String, Pin> reactionPins = chatMessageReactionsPinsByChannelId
```
KimStrand: This registration can race unbindChannelPins(). ObservableCollection uses copy-on-write observer snapshots, so a message callback that was already captured can continue after channel removal has removed both map entries. It can then recreate chatMessageReactionsPinsByChannelId here and install a reaction pin without a corresponding chatMessagesPinsByChannelId entry.

That orphan is not cleaned by unbindAllChannelPins(), which enumerates only the message-pin keys, and it can still publish reactions for a channel that was left. Should we coordinate registration and teardown through one atomic per-channel state and add a test that forces the leave-versus-message interleaving?

api/src/main/java/bisq/api/rest_api/endpoints/chat/private_chat/PrivateChatRestApi.java
```diff
@@ -0,0 +1,355 @@
...
+        try {
+            // Jersey hands us null for an absent body, which would otherwise NPE into a 500.
+            if (request == null || StringUtils.isEmpty(request.text())) {
+                asyncResponse.resume(buildResponse(Response.Status.BAD_REQUEST, "text must not be empty."));
+                return;
+            }
+            withChannel(channelId, asyncResponse, channel -> {
+                Optional<Citation> citation = Optional.ofNullable(request.citation())
+                        .map(DtoMappings.CitationMapping::toBisq2Model);
+                channelService.sendTextMessage(request.text(), citation, channel);
```
KimStrand: PrivateChatChannelService can return an already-failed future before storing anything when either the local identity or peer is banned. Because this result is discarded, the endpoint still returns 204 even though the documented local acceptance never occurred. The reaction call below has the same behavior.

Should we distinguish immediate local rejection from asynchronous network delivery—either by validating those preconditions here or exposing a separate local-acceptance result—and return a non-2xx response when nothing was accepted locally?

KimStrand (commented):
Did we consider that adding PRIVATE_CHAT_CHANNELS raises the pairing permission count to 12, causing Connect 0.8.0, which accepts only 11, to reject new pairing codes unless the tolerant decoder is released  first?

api/src/main/java/bisq/api/web_socket/subscription/SubscriptionService.java
salottodev: Good catch. It's actually broader than these three topics. `subscribe()` has never checked the client's permissions for any topic, so the same gap exists today for `TRADES`, `OFFERS`, `TRADE_CHAT_MESSAGES`, and the others. The REST filter only sees requests proxied through JAX-RS, and a native subscription never goes through it.

`NETWORK_INFO` is the worst case: it has no REST route at all, so that data (our own address, our keyId, and the address of every peer we're connected to) is reachable with no grant whatsoever.

Because of that, I'd rather not patch it inside this PR. I already have the fix on a separate branch off the same base, and it's basically what you're describing:

*   A `SubscriptionPermissionMapping` that maps topic to permission. I kept it separate from `PermissionMapping` on purpose, since that one answers for a path plus a method and a topic is neither. Forcing both surfaces through the same path-shaped interface is what let this one slip: a WebSocket-only feature never touches the REST mapping, so nothing ever asks it for a permission.
*   The mapping is a switch expression with no default, so adding a topic without declaring its permission doesn't compile. That way it can't happen again by accident.
*   The check runs at the top of `subscribe()`, before the topic is routed and before the subscriber is added, same as a filter running before the resource. It fails closed at every step: no `clientId` on the upgrade request means denied, no permissions registered for that client means denied. The identity is read from the upgrade request and not from the frame, because that's where the handshake authenticated it.
*   The error reuses the REST wording (`permission_not_granted`) so the client can tell a missing permission apart from a connection problem and prompt for re-pairing. The reason stays in the log, like the REST filter does.
*   Tests cover: permission not granted (refused, no subscriber added), missing `clientId` (refused), permission granted (proceeds), and a node running without authorization (proceeds).

About ordering: that branch already leaves permission id 11 free for `PRIVATE_CHAT_CHANNELS` as declared here, so there's no collision either way.

Does splitting it like that work for you, or would you prefer I include the enforcement here?

salottodev:
> Did we consider that adding PRIVATE_CHAT_CHANNELS raises the pairing permission count to 12, causing Connect 0.8.0, which accepts only 11, to reject new pairing codes unless the tolerant decoder is released first?

Yes, we did, we found this issue with @rodvar and he worked on a solution.

rodvar:
> > Did we consider that adding PRIVATE_CHAT_CHANNELS raises the pairing permission count to 12, causing Connect 0.8.0, which accepts only 11, to reject new pairing codes unless the tolerant decoder is released first?
> 
> Yes, we did, we found this issue with @rodvar and he worked on a solution.

Indeed @KimStrand and thanks for reviewing with prod backward compatibility in mind 💪 

In short: yes - sequencing: `Connect 0.8.1` (w/tolerant decoder, already on main - probably going on today) releases before any node build containing this PR; the decoder skips unknown ids so `0.8.1` pairs fine against both old and new nodes 👍 


Since we are talking about this, worth mentioning that [this proto change](https://github.com/bisq-network/bisq2/pull/4948/changes#diff-21830f2b9c415dd070207a77617a477f76ec2640a02d24cc06f10a6d14223ddc) in a PR that comes after this one would need updating after this merge. Thanks again

api/src/main/java/bisq/api/rest_api/endpoints/chat/private_chat/PrivateChatRestApi.java
salottodev: Confirmed, and fixed — including the reaction call, which fails the same way in `sendMessageReaction` before `addMessageReaction`, and whose future the endpoint discarded too.

  It was quieter than it looks. Nothing is thrown: `failedFuture` is a normal return, so the endpoint's own `try/catch` never saw it, and a `CompletableFuture` whose exception nobody consumes is dropped with
  no uncaught-exception handler and no log. And because the refusal returns before `createAndGetNewPrivateChatMessage`, we never reached `addMessage`, `confidentialSend`, or the `registerResendMessageData`
  inside it — so the message left no trace in any subsystem: no store entry, no `ResendMessageData`, no `MessageDeliveryStatus`, no log line. The 204 was the only artifact it produced.

  I took your first option, as an extraction rather than a second copy of the condition. `findSendRejection(channel, receiver)` returns an `Optional` of a small enum; `sendMessage` and `sendMessageReaction`
  both guard on it, and the endpoints query it before sending. That block was already duplicated across the two send methods, so this removes a copy rather than adding one, and the previously anonymous `new
  RuntimeException()` gains its reason on the way. A refusal answers 409 rather than 403 — the caller is authorised, and 403 already means a missing permission in this API, so reusing it would be ambiguous.

  One detail worth flagging: on the reaction endpoint the check sits inside the branch that actually sends, not before it. A duplicate reaction neither sends nor stores anything, so its documented idempotent
  204 stays honest and only a real send can answer 409. There is a test pinning that, alongside three that fail without the guards.

  Reading the future could not have resolved this, for what it's worth. `isCompletedExceptionally()` happens to work today only because `confidentialSend` goes through an executor and an `orTimeout(2,
  SECONDS)` (`NetworkService:306-312`) — a timing property of another module that nothing documents. And resuming from `whenComplete` is worse: a delivery failure and a local refusal arrive identically as

  Reading the future could not have resolved this, for what it's worth. `isCompletedExceptionally()` happens to work today only because `confidentialSend` goes through an executor and an `orTimeout(2,
  SECONDS)` (`NetworkService:306-312`) — a timing property of another module that nothing documents. And resuming from `whenComplete` is worse: a delivery failure and a local refusal arrive identically as
  `throwable != null`, so we would fail a message the client is about to receive over the WebSocket, and a false failure produces a duplicate on retry where the current false success does not.

  Two things I left out deliberately. `TradeChatMessagesRestApi:95` and `:172` discard the future exactly as this did — `BisqEasyOpenTradeChannelService` now inherits `findSendRejection` as well, so wiring
  it there is a few lines, but it changes the response contract of an endpoint clients already ship against and I'd rather that be reviewed as its own change. And desktop is where this shape came from and is
  only half-handled: `ChatMessageContainerController:292` checks the ban in the caller and returns before reaching the service — the same approach — but only for my own profile, with `// If I am banned, we
  silently skip sending the message`. The peer-banned case isn't checked there at all. That drift is really the argument for having one shared answer; fixing it means deciding whether desktop should start
  showing something, which felt like a product call.

  On the other half of your distinction: asynchronous delivery still has no channel in the API. `NetworkService:468` already exposes `getMessageDeliveryStatusByMessageId()` as an observable map keyed by
  message id — desktop consumes it, we don't — so a delivery-status topic would be its natural home, and would give the mobile client the sent/failed feedback it currently cannot get. Separate from this, but
  it's the piece that makes the local-versus-asynchronous split visible end to end.

api/src/main/java/bisq/api/web_socket/subscription/SubscriptionService.java
KimStrand: A separate PR works for me.

api/src/main/java/bisq/api/rest_api/endpoints/chat/private_chat/PrivateChatRestApi.java
KimStrand: Thanks, the normal refusal path is covered now. The new helper still leaves a check then act window: if a ban lands after the endpoint pre check, the service returns an already failed future, this endpoint discards it, and it still resumes 204 without storing. Should we make the send call expose a synchronous typed local acceptance outcome so the authoritative check and the HTTP response cannot diverge?

api/src/main/java/bisq/api/web_socket/domain/chat/private_chat/PrivateChatReactionsWebSocketService.java
```diff
@@ -0,0 +1,321 @@
...
+        channelsPin = channelService.getChannels().addObserver(new CollectionObserver<>() {
+            @Override
+            public void onAdded(TwoPartyPrivateChatChannel channel) {
+                bindChannel(channel);
+            }
+
+            @Override
+            public void onRemoved(Object element) {
+                if (element instanceof TwoPartyPrivateChatChannel channel) {
+                    unbindChannelPins(channel.getId());
```
KimStrand: An old channel’s removal can still tear down a newer same-ID binding here. ObservableCollection removes from the backing set before notifying observers, so an inbound message can recreate the deterministic-ID channel while this callback is pending; the ID-only teardown then closes the new pins while the new channel remains live. PrivateChatMessagesWebSocketService has the same ABA window. Should we associate each pin bundle with the exact channel instance and only remove it when that instance still owns the current binding, with a remove/re-add-same-ID test?

salottodev: Fixed as you suggest. Each pin bundle holds the channel it was registered for and the teardown compares by reference, never by equals. It reads the entry, checks the owner, and removes conditionally, so a bind that published its own bundle in between keeps it and the departing channel leaves with nothing to close.

  The same identity question showed up in the bind path, which was asking `getChannels().contains(channel)`. That compares by equals and answers for the successor rather than for the instance being bound, so
  it is now a reference scan.

  `PrivateChatMessagesWebSocketService` had it too, as you say, and it was also missing the removed-while-binding guard the reactions service already had. It has both now, in the same shape. It also stopped
  registering its observer inside a `compute()` on the map: `addObserver` replays every message on the channel and sends each to the subscribers, so those callbacks ran with a `ConcurrentHashMap` bin lock
  held. The conditional writes are what make the interleavings safe, so nothing is lost by dropping it.

  `PrivateChatChannelsWebSocketService` keeps no per-channel pins, so it is not affected.

  Both services get the remove-and-re-add-same-id test. Mockito leaves equals as reference identity on the mocks, so the two instances coexist in the set and the leave can be issued after the replacement is
  bound, which makes the reproducer single-threaded. Each fails without the ownership check.

api/src/main/java/bisq/api/web_socket/domain/chat/private_chat/PrivateChatReactionsWebSocketService.java
salottodev: Fixed as the single `Map<String, ChannelPins>` you describe, though merging the maps is not what makes it safe. Dropping the id lookup is. `addObserver` replays synchronously, so the reaction binding runs
  inside the message binding, and resolving the channel through the map there is a recursive update on the key that call is already holding, which `ConcurrentHashMap` documents as unsupported. Two maps only
  avoided it by being separate instances. It also rules out re-checking `containsKey` in the callback: during the replay the key is not mapped yet, so it cannot tell "not registered yet" from "torn down".

  So the callbacks capture the `ChannelPins` instance instead of looking it up, and no `compute` is left in the class. One arriving after teardown writes into the instance that was closed underneath it, is
  told so, and unbinds the pin it just created. The `closed` flag also covers what `compute` gave: of two racing binds, the loser is closed by the winner's `put` and cleans up after itself.

  The interleaving is covered on one thread by having the reactions `ObservableSet` perform the leave inside `addObserver`, which lands between creating the observer and storing its pin. It fails on the
  previous code.

api/src/main/java/bisq/api/rest_api/endpoints/chat/private_chat/PrivateChatRestApi.java
salottodev: `trySendTextMessage` and `trySendTextMessageReaction` now return a `SendOutcome` with the local rejection, if any, and the delivery future. The endpoint reads the rejection off the send itself and has
  nothing left to check, so there is no second decision to diverge from the first.

  They have to be separate in the type because the future cannot carry both. The rejection is settled when the call returns and says whether the message exists at all; the future resolves later and says only
  whether it reached the peer. Both arrive as an exceptionally completed future, and answering 409 for a delivery failure would make the client retry into a duplicate, which is worse than the 204 it used to
  get.

  `findSendRejection` is private again, so what I said earlier about `BisqEasyOpenTradeChannelService` inheriting it no longer holds. `sendTextMessage` and `sendTextMessageReaction` stay as one-line
  delegates, so desktop and `ModeratorService` are untouched and still compile. On the reaction endpoint only the branch that sends can answer 409, so a duplicate still gets its documented 204. Tests cover
  both.

api/src/main/java/bisq/api/web_socket/domain/chat/private_chat/PrivateChatMessagesWebSocketService.java
```diff
@@ -0,0 +1,263 @@
...
+                // Private chat messages cannot be removed
+            }
+
+            @Override
+            public void onCleared() {
+                // Private chat messages cannot be removed
+            }
+        });
+
+        ChannelBinding binding = new ChannelBinding(channel, messagesPin);
```
KimStrand: The teardown side now protects a newer owner, but I think the inverse ordering is still open. CollectionObserver callbacks are synchronous per operation but are not serialized across threads: an old onAdded callback can resume after that channel was removed and its same ID successor was already bound. This unconditional put then replaces and unbinds the successor; isLive(old) removes the stale binding, leaving the live channel with none. PrivateChatReactionsWebSocketService has the same ordering at its put. Should we make installation atomically preserve a binding owned by the live successor, and add a test that forces remove/re-add before the old onAdded callback resumes?

salottodev: Right, the put was the mirror of the teardown problem and it was still open. Both services now check inside the same compute that publishes whether the owner is still in the channel collection, and decline otherwise. Since at most one channel per id can be live, a live owner is enough to know that whatever is filed under the id belongs to an instance that is gone. The teardown moved to computeIfPresent on the same key so the two cannot cross.

While tracing this I found a related gap at shutdown: it does not touch the channel collection, so a bind straddling the sweep still saw its owner as live and published into a map already emptied. A generation counter bumped before each bulk teardown and re-read at publish refuses those, and the messages sweep now removes key by key instead of snapshot-then-clear.

Added single-threaded reproducers for both interleavings in each service. All four fail on the previous commit.

api/src/main/java/bisq/api/web_socket/subscription/SubscriptionService.java
salottodev: https://github.com/bisq-network/bisq2/pull/4961

api/src/main/java/bisq/api/web_socket/domain/chat/private_chat/PrivateChatMessagesWebSocketService.java
```diff
@@ -0,0 +1,321 @@
...
+     * <p>
+     * Both reads run under the bin lock of this key, so they must stay lock-free and must never reach
+     * back into this map: {@link #isLive} is a plain scan of the channel set, and the domain service
+     * holds its own monitor while it notifies the removal that takes this same lock in
+     * {@link #unbindChannel}.
+     *
+     * @return false if the owner is gone or a teardown ran meanwhile, leaving the pin for the caller
+     * to unbind.
+     */
+    private boolean install(String channelId, ChannelBinding binding, long generation) {
```
KimStrand: Should we also reject an onAdded callback that starts after shutdown from an already-captured CopyOnWriteArrayList snapshot, since it reads the new generation, still sees the channel live, and can publish after the sweep, with a terminal/accepting-bind flag and a regression test for both services?

chat/src/main/java/bisq/chat/priv/SendOutcome.java
```diff
@@ -0,0 +1,55 @@
...
+ * rejection is settled by the time the send returns and says whether the message exists at all; the
+ * future resolves later and says only whether it reached the peer. A caller that has to answer
+ * synchronously — an HTTP endpoint — can read the first without waiting for the second, which for a DM
+ * over Tor may take seconds or go to a mailbox.
+ * <p>
+ * The future alone cannot carry both: a rejection and a delivery failure both surface as an exceptionally
+ * completed future, and telling them apart is the difference between "nothing was stored, say so" and
+ * "it was stored and is on its way, saying otherwise makes the client retry into a duplicate".
+ */
+public record SendOutcome(Optional<SendRejection> rejection,
```
KimStrand: Should we make SendOutcome a final class with a private constructor, or a sealed result, rather than a record? Accepted versus rejected is an invariant, while the public record constructor allows inconsistent combinations, and our guideline reserves records for simple value objects.

api/src/main/java/bisq/api/rest_api/endpoints/chat/private_chat/PrivateChatRestApi.java
```diff
@@ -0,0 +1,392 @@
...
+            if (userIdentityService.findUserIdentity(peerProfileId).isPresent()) {
+                asyncResponse.resume(buildResponse(Response.Status.BAD_REQUEST,
+                        "Cannot open a private chat with yourself."));
+                return;
+            }
+            userProfileService.findUserProfile(peerProfileId)
+                    .ifPresentOrElse(peer -> {
+                        // The ChatService wrapper, not channelService.findOrCreateChannel: selecting the
+                        // channel is what keeps the next DM bound correctly. See the class javadoc.
+                        chatService.createAndSelectTwoPartyPrivateChatChannel(ChatChannelDomain.DISCUSSION, peer)
```
KimStrand: Should we avoid selecting the channel in this REST operation and call channelService.findOrCreateChannel(...) instead? DISCUSSION shares one selection between the public and private desktop controllers, so this API call can replace the conversation currently displayed on the public Discussion page and can also consume an existing channel’s notifications. The lower-level method already binds creation to the currently selected identity, and later sends use the identity stored on the channel.

api/src/main/java/bisq/api/rest_api/endpoints/chat/private_chat/PrivateChatRestApi.java
```diff
@@ -0,0 +1,392 @@
...
+            }
+    )
+    public void leaveChannel(@PathParam("channelId") String channelId,
+                             @Suspended AsyncResponse asyncResponse) {
+        applyTimeout(asyncResponse);
+        try {
+            withChannel(channelId, asyncResponse, channel -> {
+                // Not channelService.leaveChannel: the manager additionally re-selects the next channel
+                // and consumes the departed channel's notifications, as desktop does.
+                leavePrivateChatManager.leaveChannel(channel);
```
KimStrand, line 331: Should we only reselect when the channel being left is actually the selected channel? LeavePrivateChatManager assumes the desktop flow is leaving the current channel and unconditionally selects the first remaining private channel, or null. This endpoint can leave any channel by ID, so leaving a background DM switches or clears the desktop Discussion view.

api/src/main/java/bisq/api/rest_api/endpoints/chat/private_chat/PrivateChatRestApi.java
```diff
@@ -0,0 +1,392 @@
...
+                    @ApiResponse(responseCode = "500", description = "Unexpected internal error")
+            }
+    )
+    public void sendTextMessage(@PathParam("channelId") String channelId,
+                                SendChatMessageRequest request,
+                                @Suspended AsyncResponse asyncResponse) {
+        applyTimeout(asyncResponse);
+        try {
+            // Jersey hands us null for an absent body, which would otherwise NPE into a 500.
+            if (request == null || StringUtils.isEmpty(request.text())) {
```
KimStrand, line 180: Should we reject text longer than ChatMessage.MAX_TEXT_LENGTH here and add boundary tests? The desktop enforces the 10,000-character limit, but TwoPartyPrivateChatMessage construction and serialization do not call ChatMessage.verify(), so this API currently persists and sends messages that violate the domain limit.

api/src/main/java/bisq/api/web_socket/domain/chat/private_chat/PrivateChatReactionsWebSocketService.java
```diff
@@ -0,0 +1,449 @@
...
+    @Override
+    public Optional<String> getJsonPayload() {
+        return getJsonPayload(channelService.getChannels().stream());
+    }
+
+    private Optional<String> getJsonPayload(Stream<TwoPartyPrivateChatChannel> channels) {
+        ArrayList<TwoPartyPrivateChatMessageReactionDto> payload = channels
+                .flatMap(channel ->
+                        channel.getChatMessages().stream()
+                                .flatMap(message -> message.getChatMessageReactions().stream()
```
KimStrand, line 327: Should we filter at the parent-message level here and also filter the reactions embedded in the message DTO, so both topics expose consistent state? A reaction’s sender is not necessarily the message’s sender: after the peer is banned, my reaction to their hidden message still passes this filter and becomes orphaned, while their reaction to my visible message still arrives embedded through PRIVATE_CHAT_MESSAGES.

KimStrand (commented):
utACK

PR #4944, opened 2026-08-18
bugfix/android_api_33_bootstrap_tor_fails

rodvar:
- resolves https://github.com/bisq-network/bisq-mobile/issues/1741
 - used JdkFacade to separate Android-specific Java8 compatible code o avoid tor bootstraping issue in Android node API 33 + tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved asynchronous error handling for HTTP requests across supported platforms.
  * Preserved request retry, failover, timeout, cancellation, and error propagation behavior.
  * Enhanced Android compatibility when recovering from failed operations.

* **Tests**
  * Added coverage for successful operations, failure recovery, wrapped exceptions, cancellation scenarios, and recovery-handler failures.

### DIFF
--- a/common/src/main/java/bisq/common/facades/android/AndroidJdkFacade.java
+++ b/common/src/main/java/bisq/common/facades/android/AndroidJdkFacade.java
@@ -21,6 +21,10 @@ import bisq.common.facades.JdkFacade;
 
 import java.io.File;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
 import java.util.stream.Stream;
 
 public class AndroidJdkFacade implements JdkFacade {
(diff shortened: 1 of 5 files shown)


rodvar:
@KimStrand

common/src/main/java/bisq/common/facades/android/AndroidJdkFacade.java
```diff
@@ -70,4 +73,14 @@ public <T> void addFirst(List<T> list, T element) {
     public <T> T getFirst(List<T> list) {
         return list.get(0);
     }
+
+    @Override
+    public <T> CompletableFuture<T> exceptionallyCompose(CompletableFuture<T> future,
+                                                         Function<Throwable, ? extends CompletionStage<T>> fn) {
+        // CompletableFuture.exceptionallyCompose crashes Android below API 34
+        return future.handle((result, throwable) -> throwable == null
```
KimStrand: handle(...).thenCompose(...) creates an intermediate future. Canceling the returned future only cancels the thenCompose stage, so the recovery handler can still run if the source later fails. Native CompletableFuture.exceptionallyCompose does not invoke the handler after its dependent future has been canceled.

In HttpRequestService, that handler can start another provider request. This means a canceled Android request could continue through the failover sequence.

Could the fallback use an explicit result future and check whether it has already been completed or canceled before invoking the recovery function?

rodvar: thanks for this Kim, I'm about to finish the improvement push incoming

common/src/main/java/bisq/common/facades/android/AndroidJdkFacade.java
```diff
@@ -70,4 +73,38 @@ public <T> void addFirst(List<T> list, T element) {
...
+            }
+            if (throwable == null) {
+                result.complete(value);
+            } else {
+                try {
+                    fn.apply(throwable).whenComplete((recovered, recoveryThrowable) -> {
+                        if (recoveryThrowable == null) {
+                            result.complete(recovered);
+                        } else {
+                            result.completeExceptionally(recoveryThrowable);
```
KimStrand: Could recovery failures be relayed with the same wrapping semantics as native exceptionallyCompose?

The Android implementation passes recoveryThrowable directly to completeExceptionally. Native exceptionallyCompose wraps a recovery stage’s raw failure in CompletionException. This produces observable platform differences:
      - Failed recovery: Java SE exposes CompletionException(cause); Android exposes the raw cause.
      - Canceled recovery: Java SE returns isCancelled() == false; Android returns isCancelled() == true.

This could make downstream exception or cancellation handling behave differently between Java SE and Android. The same wrapping would also be needed when the recovery handler itself throws at line 104. Could parity tests be added for failed and canceled recovery stages?

rodvar: great catch Kim, easy fix - will push in a bit

KimStrand (commented):
utACK

PR #4948, opened 2026-08-19

wodoro:

Revoking a paired client only removed its profile and sessions. WebSocket authentication happens at the handshake, so an established connection kept working, and push registrations are keyed by device and survived entirely.

ClientRevocationService now defines what revocation means and runs a list of ClientRevocationHandlers (close live sockets, drop push registrations), each guarded so one failure cannot leave a client half revoked. ApiAccessService is the single entry point and delegates to it; Desktop no longer reassembles the steps itself, so REST and Desktop cannot drift apart.

Add GET /access/clients, paginated and sorted (the profiles come from a map, so paging an unspecified order would duplicate or skip clients). It maps to ClientProfileDto before paging, so no ClientProfile with its plaintext secret can reach the response body.

/access/clients was absent from RestPermissionMapping, and unmapped paths fail closed, so the existing DELETE answered 403 for every client. Map both to a new CLIENT_MANAGEMENT permission rather than to SETTINGS, since revoking another device is a distinct capability. It is auto-grantable, so already paired clients gain it without re-pairing.

Tie push registrations to the API client that created them, which is what makes revocation able to find them. Registering a device ID owned by another client is now refused (device IDs are client chosen, so otherwise any client could claim another's device and redirect its notifications), and unregistering is limited to the owner. Registrations persisted before the ownership link have no owner and stay claimable, so the set only shrinks.

## Summary by CodeRabbit

* **New Features**
  * Added paginated, sorted viewing of paired clients without exposing sensitive credentials.
  * Added client-management permissions for client administration endpoints.
  * Revoking a client now closes active connections, invalidates sessions, and removes associated mobile-device registrations.
  * Mobile-device registrations are linked to clients, preventing unauthorized registration or removal.

* **Bug Fixes**
  * Blank client names are rejected without consuming pairing codes, and overly long names are safely limited.
  * Device ownership checks prevent clients from accessing one another’s registrations.
  * Revoked clients must pair again before registering devices.
  * Revoked connections no longer process incoming messages.


api/src/main/java/bisq/api/access/permissions/Permission.java
```diff
@@ -52,7 +52,8 @@ public enum Permission implements ProtoEnum {
     TRADES(7, Kind.STANDARD),
     USER_IDENTITIES(8, Kind.STANDARD),
     USER_PROFILES(9, Kind.STANDARD),
-    MOBILE_DEVICES(10, Kind.STANDARD);
+    MOBILE_DEVICES(10, Kind.STANDARD),
+    CLIENT_MANAGEMENT(11, Kind.STANDARD);
```
rodvar, line 74: making a cross-client destructive capability auto-grantable is a real decision, and I think the right one under the current all-or-nothing pairing model: revoking another device is availability-only (recoverable from Desktop / the node side), and auto-grant is exactly what lets an existing pairing revoke a lost device without having to re-pair first. Could we please capture that reasoning in a short javadoc / md ? 

Otherwise the next cross-client permission rides this one as silent precedent. Worth noting there too that a future granular/guest grant must not include this permission.

api/src/main/java/bisq/api/ApiService.java
```diff
@@ -157,7 +160,32 @@ public ApiService(ApiConfig apiConfig,
...
+        }
+
+        // Deliberately not exposed: callers reach revocation through ApiAccessService, so the
+        // access layer keeps a single entry point.
+        ClientRevocationService clientRevocationService = new ClientRevocationService(pairingService,
+                sessionService,
+                List.of(
+                        // WebSocket auth happens at the handshake only, so a revoked client keeps
+                        // receiving data until the socket is closed explicitly.
+                        clientId -> webSocketService.ifPresent(service -> service.disconnectClient(clientId)),
```
rodvar, line 185: please fix: I think if `webSocket.close()` throws on the first matching socket inside `WebSocketConnectionHandler.disconnectClient`, the stream aborts and any other live socket of the same client stays connected. The guard in `ClientRevocationService` saves the push-registration cleanup but not the remaining sockets, and multiple sockets per client is an anticipated case per the `onConnect` todo. A per-socket try/catch inside that forEach keeps the "cannot leave a client half revoked" promise.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/settings/bisq_connect/BisqConnectController.java
```diff
@@ -127,16 +130,10 @@ void onRevokeClient(BisqConnectView.ClientListItem item) {
...
-                        optionalWebSocketService.ifPresent(ws -> ws.disconnectClient(clientId));
-                        if (profileRemoved) {
-                            log.info("Revoked client {} ({})", item.getClientName(), clientId);
-                        } else {
-                            log.warn("Client profile not found for {}, but session/connection cleaned up", clientId);
-                        }
+                        // Session and connection cleanup is part of revokeClient, so the same
+                        // revocation semantics apply here and on the REST endpoint.
+                        apiAccessService.revokeClient(clientId);
+                        log.info("Revoked client {} ({})", item.getClientName(), clientId);
```
rodvar: nit - the old code branched on the result, so a double-click or stale list entry now logs "Revoked ..." for a no-op. `revokeClient` returns whether the profile existed - maybe use it to keep this info log honest? (the service already warns on the not-found path).

rodvar (changes requested):
overall a very sound implementation @wodoro - thank you
please see inline comments for your review, hope its helpful.

api/src/main/java/bisq/api/ApiService.java
wodoro: WebSocketConnectionHandler.disconnectClient is unlikely to throw, but still wrapped it's close in try catch to prevent this issue from happening
edit: reviewing CRAI response

notifications/src/main/java/bisq/notifications/mobile/registration/DeviceRegistrationService.java
```diff
@@ -67,25 +81,94 @@ public void register(String deviceId,
...
-                symmetricKeyBase64);
-        MobileDeviceProfile previous = persistableStore.getDeviceByDeviceId().put(deviceId, mobileDeviceProfile);
-        if (previous == null || !previous.equals(mobileDeviceProfile)) {
-            persist();
+                symmetricKeyBase64,
+                Optional.of(clientId));
+        // The owner check and the write have to be one step: otherwise two clients registering
+        // the same new device ID both pass the check and the later write silently takes the
+        // device, which is the takeover the check exists to prevent.
+        synchronized (persistableStore) {
```
KimStrand, line 100: A request can pass authentication, pause, then resume after ClientRevocationService has removed the client’s registrations. Lines 89–100 do not verify that the client remains paired, so the request recreates a persistent push registration for the revoked client. That client can continue receiving notifications despite successful revocation.

Registration and revocation need coordinated client liveness validation, such as a revocation generation/tombstone checked atomically when storing the registration.

wodoro:
closed for https://github.com/bisq-network/bisq2/pull/4960

PR #4949, opened 2026-08-21
feature/granular_based_api_sub_versioning_for_dockerized_distros

rodvar:
- ade version api return system variable versioning config if available, otherwise default to current impl
 - setup docker file to allow IMAGE_VERSION argument and share as env var BISQ_IMAGE_VERSION so that IMAGE_VERSION is used on each platform setup
 - this allows for dockerised distrib to return granular versioning first hand (e.g. 2.1.11.3), making distribution of new features detection more robust

## Summary by CodeRabbit

* **New Features**
  * Version reporting now includes a release-specific image version when available, alongside the core application version.
  * Image version values are trimmed for consistent display, with blank or missing values omitted.
  * Container builds can provide a release version for runtime reporting.

* **Documentation**
  * Updated version endpoint documentation to describe the optional image version field.

### DIFF
--- a/api/src/main/java/bisq/api/rest_api/endpoints/settings/SettingsRestApi.java
+++ b/api/src/main/java/bisq/api/rest_api/endpoints/settings/SettingsRestApi.java
@@ -70,15 +70,25 @@ public class SettingsRestApi extends RestApiBase {
 
     @GET
     @Path("/version")
-    @Operation(description = "Get the current Bisq2 version")
+    @Operation(description = "Get the current Bisq2 version. Containerised distributions additionally report " +
+            "their granular image version (e.g. 2.1.11.2) as imageVersion via BISQ_IMAGE_VERSION.")
     public Response getVersion() {
         try {
             String version = ApplicationVersion.getVersion().toString();
-            return Response.ok(new SettingsApiVersionDto(version)).build();
+            String imageVersion = resolveImageVersion(System.getenv("BISQ_IMAGE_VERSION"));
+            return Response.ok(new SettingsApiVersionDto(version, imageVersion)).build();
         } catch (Exception e) {
             log.error("Error getting version", e);
             return buildErrorResponse("An unexpected error occurred: " + e.getMessage());
         }
+    }
+
+    // Clients key caches of static node config on the reported version, so a distribution that can be
+    // updated in place under the same core version (docker image rebuilds) reports the per-image release
+    // in a separate field. version itself must stay the plain core version: deployed clients parse it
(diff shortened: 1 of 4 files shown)


apps/api-app/docker/Dockerfile
```diff
@@ -76,6 +76,13 @@ RUN chmod +x /usr/local/bin/entrypoint.sh
 # entrypoint's chown is a no-op. See entrypoint.sh.
 WORKDIR /opt/bisq2-api
 
+# Granular distribution version (e.g. 2.1.11.2) reported by the /settings/version endpoint in
+# place of the core version. Paired clients key config caches on that string, so it must change
+# with every image release even when the core version does not. Stamped by the release workflow;
+# empty (core-version fallback) for ad-hoc local builds.
+ARG IMAGE_VERSION=""
+ENV BISQ_IMAGE_VERSION=$IMAGE_VERSION
+
```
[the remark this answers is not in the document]
rodvar: this is already done on bisq-mobile automatic release building action - no need to add here

rodvar:
@KimStrand small but important one

api/src/main/java/bisq/api/rest_api/endpoints/settings/SettingsRestApi.java
```diff
@@ -70,17 +70,27 @@ public Response getSettings() {
 
     @GET
     @Path("/version")
-    @Operation(description = "Get the current Bisq2 version")
+    @Operation(description = "Get the current Bisq2 version. Containerised distributions report their " +
+            "granular image version (e.g. 2.1.11.2) via BISQ_IMAGE_VERSION; otherwise the core version.")
     public Response getVersion() {
         try {
-            String version = ApplicationVersion.getVersion().toString();
+            String version = resolveVersion(System.getenv("BISQ_IMAGE_VERSION"));
```
KimStrand: This breaks existing mobile clients because they only accept a three part version such as 2.1.11. Returning 2.1.11.2 makes the client treat the node as incompatible and disconnect.

Should the existing `version` field remain the three part API version. And should the Docker image revision instead be returned in a separate field, for example `imageVersion: "2.1.11.2"`.

rodvar: This is correct, I was thinking of adding support for the extra ".x" number in the version but I think the `imageVersion` idea is much better - I'll get this done in a tick

KimStrand:
utACK

rodvar:
couldn't make it to this release but will be included in the next dockerised distros `2.1.12.1` target for first week of September

PR #4960, opened 2026-08-24

wodoro:

Revoking a paired client only removed its profile and sessions. WebSocket authentication happens at the handshake, so an established connection kept working, and push registrations are keyed by device and survived entirely.

ClientRevocationService now defines what revocation means and runs a list of ClientRevocationHandlers (close live sockets, drop push registrations), each guarded so one failure cannot leave a client half revoked. ApiAccessService is the single entry point and delegates to it; Desktop no longer reassembles the steps itself, so REST and Desktop cannot drift apart.

Add GET /access/clients, paginated and sorted (the profiles come from a map, so paging an unspecified order would duplicate or skip clients). It maps to ClientProfileDto before paging, so no ClientProfile with its plaintext secret can reach the response body.

/access/clients was absent from RestPermissionMapping, and unmapped paths fail closed, so the existing DELETE answered 403 for every client. Map both to a new CLIENT_MANAGEMENT permission rather than to SETTINGS, since revoking another device is a distinct capability. It is auto-grantable, so already paired clients gain it without re-pairing.

Tie push registrations to the API client that created them, which is what makes revocation able to find them. Registering a device ID owned by another client is now refused (device IDs are client chosen, so otherwise any client could claim another's device and redirect its notifications), and unregistering is limited to the owner. Registrations persisted before the ownership link have no owner and stay claimable, so the set only shrinks.

## Summary by CodeRabbit

- **New Features**
  - Added paginated client listing with sorted names and IDs, excluding sensitive credentials.
  - Added client-management permissions for listing and revoking paired clients.
  - Added client ownership tracking for mobile device registrations.
  - Added automatic cleanup of sessions, WebSocket connections, and device registrations when a client is revoked.

- **Bug Fixes**
  - Prevented unauthorized device registration and removal.
  - Rejected blank client names and safely limited overly long names.
  - Improved WebSocket cleanup and blocked messages from revoked clients.


wodoro:
closed waiting for https://github.com/bisq-network/bisq2/pull/4961 to merge to rebase and redo the PR

PR #4965, opened 2026-08-25
chore: sync #4949 to main — feature/granular_based_api_sub_versioning_for_dockerized_distros

rodvar:
Automated sync of #4949 from `for-mobile-based-on-2.1.12` to `main`.

## Summary by CodeRabbit

* **New Features**
  * The settings API now reports a separate container image version when available.
  * Version values are trimmed and omitted when unavailable or blank.
  * Container builds can provide the image version for runtime display.

### DIFF
--- a/api/src/main/java/bisq/api/dto/settings/SettingsApiVersionDto.java
+++ b/api/src/main/java/bisq/api/dto/settings/SettingsApiVersionDto.java
@@ -17,5 +17,9 @@
 
 package bisq.api.dto.settings;
 
-public record SettingsApiVersionDto(String version) {
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+// imageVersion is only present on containerised distributions (see SettingsRestApi.resolveImageVersion)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public record SettingsApiVersionDto(String version, String imageVersion) {
 }
--- a/apps/api-app/docker/Dockerfile
+++ b/apps/api-app/docker/Dockerfile
@@ -76,6 +76,13 @@ RUN chmod +x /usr/local/bin/entrypoint.sh
 # entrypoint's chown is a no-op. See entrypoint.sh.
 WORKDIR /opt/bisq2-api
 
+# Granular distribution version (e.g. 2.1.11.2) reported by the /settings/version endpoint as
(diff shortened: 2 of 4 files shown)


apps/api-app/docker/Dockerfile
```diff
@@ -76,6 +76,13 @@ RUN chmod +x /usr/local/bin/entrypoint.sh
 # entrypoint's chown is a no-op. See entrypoint.sh.
 WORKDIR /opt/bisq2-api
 
+# Granular distribution version (e.g. 2.1.11.2) reported by the /settings/version endpoint as
+# imageVersion, alongside the unchanged core version. Paired clients key config caches on it, so
+# it must change with every image release even when the core version does not. Stamped by the
+# release workflow; empty (field omitted) for ad-hoc local builds.
+ARG IMAGE_VERSION=""
```
KimStrand, line 83: Should we also update the manual publishing command in apps/api-app/docker/README.md to pass --build-arg IMAGE_VERSION="$VERSION"? As written, manually published images receive the granular tag but leave BISQ_IMAGE_VERSION empty, so /settings/version omits imageVersion and clients cannot invalidate their caches for image-only releases.

api/src/main/java/bisq/api/dto/settings/SettingsApiVersionDto.java
```diff
@@ -17,5 +17,9 @@
 
 package bisq.api.dto.settings;
 
-public record SettingsApiVersionDto(String version) {
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+// imageVersion is only present on containerised distributions (see SettingsRestApi.resolveImageVersion)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public record SettingsApiVersionDto(String version, String imageVersion) {
```
KimStrand, line 24: Should we mark imageVersion as @Nullable? It is intentionally null for non-container builds, while Bisq treats unannotated values as non-null by default. This would also match the convention used by other DTO record components.

rodvar:
@KimStrand Kim thanks for the review - I agree with both comments, could we please do that as a follow-up against `for-mobile` to follow the normal flow since they are not critical and utACK this one?

KimStrand:
utACK

Issue #4883, opened 2026-07-22
Complete the MuSig create/take offer rewrite


balag3:
HenrikJannsen started a rewrite of the MuSig create/take offer flow with the goal that
JavaFX controllers stay presentation oriented and domain validation and state live in use
cases. Parts of it are already on main (#4686, #4688), the rest lives on his branch
[Rebased-Apply-create-offer-workflow-to-UI-3](https://github.com/HenrikJannsen/bisq2/tree/Rebased-Apply-create-offer-workflow-to-UI-3)
(earlier WIP PRs #4687, #4658 were closed). This issue is the collection point for status,
specs, findings and the planned workflow.

Working branch: [offer-rewrite-continuation](https://github.com/balag3/bisq2/tree/offer-rewrite-continuation)
(Henrik's branch, commits and authorship preserved, rebased onto current main, compiles
and tests run green).

Since MuSig is devMode gated, none of this is reachable in release builds.

## Current status of the code

Create offer:
- The domain side is nearly complete: DraftOfferUseCase/CreateOfferUseCase with per
  concern components (market, direction, price and limits, payment method, amount and
  limits) under `offer/.../mu_sig/use_case/create_offer/`. Most controllers are thin
  bindings; the review controller still owns substantive logic (materializes the
  amount/price specs, builds the account options, creates and publishes the offer).
- Observed working in a live A/B against main: amounts always publish what the screen
  shows, and the payment method dependent limits are real (5000 USD max with Wise, 10000
  with Advanced Cash, matching the chargeback risk rails in code). Both of these are
  broken on main (details under Bugs).
- The old create/take UI was restructured under `mu_sig/offer/draft/` (renames plus new
  amount/price components), navigation targets unchanged.

Take offer:
- `TakeOfferUseCase` is a skeleton: `initialize(MuSigOffer)` has an empty body, the
  mutators are no-ops and `toInputAmount`/`toPassiveAmount` return null.
- The per concern services are thinner than they look: `PaymentMethodSelectionService` is
  meaningfully implemented, but `TakeOfferAmountService` is just model setters,
  `TakeOfferPriceService` is a single setter, and `TakeOfferPaymentMethodService` has
  empty recomputation hooks where the spec requires limit updates on payment method or
  account changes.
- Live behavior matches: taking a fixed amount offer works end to end (with the mocked
  musigd backend and the dev skip button) because the controller bypasses the amount step
  and sets review amounts directly from the offer when a `FixedAmountSpec` is present.
[...]

PR #4966, opened 2026-08-25
Complete the MuSig create and take offer rewrite

balag3:
Completes the MuSig create/take-offer rewrite started by @HenrikJannsen (his branch `Rebased-Apply-create-offer-workflow-to-UI-3`), which I continued under #4883. Create-offer is Henrik's work; I added the take-offer use case and per-concern services, the two review slices, and the spec.

**What's here**
- Henrik's create-offer workflow rewrite (per-concern Selection components, DraftOfferUseCase, presentation-only controllers, validation in use cases; one exception left: the create-offer review controller still assembles the offer options, tracked in #5000).
- `TakeOfferUseCase` + per-concern take-offer services: init/validation, payment + eligibility, price, amount + limits, review + handoff, controller wiring.
- Spec: `docs/specifications/offer/mu-sig/take-offer.md`.
- Reconciled with current `main` via one merge commit (main's fixes since the branch — PriceQuote overflow, offer-list validity, amount-limit policy move — are preserved; the create-offer area is superseded by the rewrite). Not rebased, to keep Henrik's authorship intact — squash or rebase on merge if you prefer linear history.

**Tests:** offer, bisq-easy, trade, api, mu-sig and desktop suites green.

**Deferred (product decisions, not blockers):** the take-offer review shows `N/A` for the trade fee until the review and the protocol share one fee source — pending the fee-policy call. Tracking + open items in #4883.

Part of #4883.


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/draft/components/MuSigPriceInput.java
```diff
@@ -163,31 +174,39 @@ private void updateFromMarketPrice() {
 
         @Override
         public void onActivate() {
+            model.reset();
+            updateFromMarketPrice();
             model.isPriceValid.set(true);
             updateFromMarketPrice();
 
             marketPricePin = marketPriceService.getMarketPriceByCurrencyMap().addObserver(() -> {
```
KimStrand: Should we avoid writing the current market price back into `PriceSelection` on every market price map update?

`onSetPriceQuote()` also recalculates and persists the percentage, so this appears to reset a fixed price, or a non-zero floating markup, to the current market price / 0% without a user action. The same mutation also happens twice during activation above.

Should we preserve a fixed quote, and for floating pricing recompute the derived quote while preserving the selected percentage?

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/limits/AmountLimitsProvider.java
```diff
@@ -0,0 +1,192 @@
...
+        checkNotNull(directionService, "directionService must not be null");
+        checkNotNull(paymentMethodService, "paymentMethodService must not be null");
+        checkNotNull(priceService, "priceService must not be null");
+
+        absoluteAmountLimitsProvider = new AbsoluteAmountLimitsProvider(marketPriceService, marketService, priceService);
+        paymentMethodSpecificAmountLimitsProvider = new PaymentMethodBasedAmountLimitsProvider(marketPriceService, marketService, paymentMethodService, priceService);
+        userSpecificAmountLimitsProvider = new UserSpecificAmountLimitsProvider(marketPriceService, marketService, directionService, priceService);
+    }
+
+    @Override
```
KimStrand, line 65: Should we explicitly calculate all three limit providers from the current market, direction, price, and payment state during initialization?

`AbsoluteAmountLimitsProvider` and `UserSpecificAmountLimitsProvider` only register custom listeners, but those listeners do not replay their current values. On the normal default path, those values were established before these listeners were attached, so `AmountLimitsProvider` can remain uninitialized and the amount input handlers silently ignore edits.

Could we also add a test proving that `AmountSelection.initialize()` becomes usable without manually firing dependency listeners?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/draft/take_offer/review/MuSigTakeOfferReviewController.java
```diff
@@ -259,10 +344,18 @@ public void takeOffer() {
                 delayedSuccessScheduler.stop();
             }
             delayedSuccessScheduler = UIScheduler.run(() -> {
```
KimStrand, line 347: Should we keep this attempt in `SENT` or a pending state until we receive an actual protocol or delivery success signal?

`MuSigTradeService.takeOffer()` dispatches protocol handling asynchronously, so the absence of an error after 200 ms does not establish success. A rejection or delivery failure can arrive after this scheduler has already shown the success state and stopped the timeout.

If no completion signal is currently available, would a pending state be safer than reporting success based on elapsed time?

offer/src/main/java/bisq/offer/mu_sig/use_case/take_offer/payment_method/TakeOfferPaymentMethodService.java
```diff
@@ -0,0 +1,219 @@
...
+    public void putAllAccountsByPaymentMethod(Map<PaymentMethod<?>, List<Account<?, ?>>> selectedAccountByPaymentMethod) {
+        checkNotNull(selectedAccountByPaymentMethod, "selectedAccountByPaymentMethod must not be null");
+        model.putAllAccountsByPaymentMethod(selectedAccountByPaymentMethod);
+    }
+
+    public void clearAccountsByPaymentMethod() {
+        model.clearAccountsByPaymentMethod();
+    }
+
+    public void putSelectedAccountByPaymentMethod(PaymentMethod<?> paymentMethod, Account<?, ?> account) {
```
KimStrand, line 101: Should we enforce the payment-selection invariant at this public mutation boundary?

Could we verify that the account's payment method equals `paymentMethod`, that the account belongs to the eligible account list for that method, and that the method is offered? `getSelectedPaymentMethodSpec()` later derives the spec from the map key while `getSelectedAccount()` returns the value, so an inconsistent entry would be handed to the protocol as a mismatched spec/account pair.

The current controller passes matching values, but the take offer specification identifies the use case as the enforcement point for this guarantee.

common/src/main/java/bisq/common/monetary/PriceQuote.java
```diff
@@ -199,6 +248,34 @@ public double asDouble(int precision) {
...
+        checkNotNull(min, "min must not be null");
+        checkNotNull(max, "max must not be null");
+
+        checkArgument(this.getBaseSideMonetary().getCode().equals(max.getBaseSideMonetary().getCode()),
+                "this and max base side codes must match. this.base=%s; max.base=%s",
+                this.getBaseSideMonetary().getCode(), max.getBaseSideMonetary().getCode());
+        checkArgument(min.getBaseSideMonetary().getCode().equals(max.getBaseSideMonetary().getCode()),
+                "this and max base side codes must match. this.base=%s; max.base=%s",
+                min.getBaseSideMonetary().getCode(), max.getBaseSideMonetary().getCode());
+        checkArgument(min.getQuoteSideMonetary().getCode().equals(max.getQuoteSideMonetary().getCode()),
```
KimStrand, line 269: Should we also verify that this quote's quote side currency matches the min/max quote side currency?

The current checks compare this quote's base currency and verify that min/max match each other, but never compare `this.getQuoteSideMonetary().getCode()` with the limits. As a result, a BTC/EUR quote whose numeric value is inside BTC/USD limits can be returned unchanged.

Would comparing both sides of all three quotes here make this clamp fail closed across markets?

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/AmountSelection.java
```diff
@@ -0,0 +1,601 @@
...
+            applyMaxAmountAndSliderValue(model.getMaxTradeAmount());
+        }
+    }
+
+
+    /* --------------------------------------------------------------------- */
+    // User interaction
+    /* --------------------------------------------------------------------- */
+
+    public void onSetUseBaseCurrencyForAmountInput(boolean value) {
```
KimStrand, line 279: Should we recompute the input amount range, user limit marker, and normalized slider values when the active input side changes?

At present this only changes the boolean and persists it, leaving the derived state denominated and mapped using the previous input side until another limits update occurs. The create offer specification explicitly lists this recomputation as an update trigger, and the take flow performs the corresponding recalculation in `TakeOfferUseCase.setUseBaseCurrencyForAmountInput()`.

common/src/main/java/bisq/common/monetary/TradeAmountRange.java
```diff
@@ -38,4 +39,28 @@ public TradeAmountRange(TradeAmount min, TradeAmount max) {
                 "min and max quote side codes must match. min.quote=%s; max.quote=%s",
                 min.getQuoteSideAmount().getCode(), max.getQuoteSideAmount().getCode());
     }
+
+    public String printRelevantStringa(Market market) {
```
KimStrand: `printRelevantStringa` appears to be a misspelled duplicate of `printRelevantString` immediately below it, and I could not find any callers.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/price/PriceSelection.java
```diff
@@ -0,0 +1,262 @@
...
+    public void onSetPricePercentage(double pricePercentage) {
+        applyPricePercentage(pricePercentage, true);
+
+        PriceQuote priceQuote = percentageToPriceQuote(getPricePercentage());
+        applyPriceQuote(priceQuote, true);
+    }
+
+    private void applyPricePercentage(double pricePercentage, boolean notifyListeners) {
+        if (Double.compare(pricePercentage, model.getPricePercentage()) != 0) {
+            pricePercentage = priceLimits.clamp(pricePercentage);
```
KimStrand: Should the create-offer UI route out-of-range percentage input through this clamp?

`MuSigCreateOfferPriceController.applyPercentageString()` retains the previous behavior of validating and returning before `onSetPricePercentage()` is called. That means this newly added clamp, and the specification saying that UI input is clamped, does not currently apply to percentage input.

Is the intended behavior to change the UI to use this clamp, or should we retain rejection and adjust the specification accordingly?

common/src/main/java/bisq/common/monetary/PriceQuote.java
balag3: Confirmed. The clamp now also compares this quote's quote side against the limits, so a quote from another market fails closed instead of being returned unchanged. Fixed in: 4d1799db91

common/src/main/java/bisq/common/monetary/TradeAmountRange.java
balag3: Confirmed, removed. Dead misspelled duplicate with no callers (the same method was also in PriceQuoteRange). Fixed in: 0f0adbdc65

offer/src/main/java/bisq/offer/mu_sig/use_case/take_offer/payment_method/TakeOfferPaymentMethodService.java
balag3: Confirmed. Fixed in 9c43f9dd13. Checks the account matches the method key and is in the eligible list (which is offered-only, so that's covered too).

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/AmountSelection.java
balag3: Confirmed. Fixed in c01269fb47. The switch now recomputes range/marker/slider on the new side, and the amount + slider controllers use the take flow's origin separation so the programmatic re-write on the switch no longer feeds back against the stale range (verified live for both fixed and range offers).

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/draft/components/MuSigPriceInput.java
balag3: Confirmed. Fixed in e4865f3c54 — the market observer now leaves a fixed price alone and, for a floating price, recomputes the quote from the kept percentage; activation no longer pushes the market price. While doing this I found a deeper, separate issue: the price-step controller round-trips every published quote back through onSetPricePercentage (no origin separation), which drifts the floating percentage on a market update and rewrites a fixed price on re-entry. Filed it as #4967 to fix on its own with proper testing.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/limits/AmountLimitsProvider.java
balag3: Confirmed. Fixed in 11a20d7336. Each sub-provider now computes its value in initialize() instead of waiting for a listener that never replays. Added a test.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/draft/take_offer/review/MuSigTakeOfferReviewController.java
balag3: Agreed, elapsed time isn't a real success signal. The protocol handler doesn't surface a delivery/completion event yet (that's the todo here), so there's nothing to drive a genuine pending→success transition — the 200 ms + error observer + timeout is the interim. Leaving it until the protocol reports delivery and fixing it there; I'll track it as a rewrite item on #4883.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/price/PriceSelection.java
balag3: Agreed, clamp it, spec stays. Can't do it on its own though — it hits the same feedback loop as #4967: even routing the clamped percentage into the domain, the quote observer derives it back from the rounded quote and validateQuote rejects it again (a valid small percentage on a low-sat crypto market gets rewritten the same way). So I'll fold the clamp into #4967 — the origin separation there is what makes it stick.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/draft/components/MuSigPriceInput.java
```diff
@@ -220,20 +235,31 @@ private void onFocusedChanged(boolean isFocused) {
...
+            priceUseCase.ifPresent(priceSelection -> {
+                if (priceSelection.getUseFixPrice()) {
+                    // A fixed price is an absolute value; a market move must not change it.
+                    return;
+                }
+                // Floating price: recompute the quote from the selected percentage against the new
+                // market price, keeping the percentage. Only when this market has a price available.
+                Market market = draftOfferUseCase.getMarket();
+                if (market != null && marketPriceService.findMarketPrice(market).isPresent()) {
+                    priceSelection.onSetPricePercentage(priceSelection.getPricePercentage());
```
KimStrand: Should we refresh the price limits from the same market price update before recomputing the floating quote?

PriceLimits.initialize() only reacts to market selection, so its PriceQuoteRange remains based on the previous market price. This call computes a quote from the new market price, but PriceSelection.applyPriceQuote() then clamps it against that stale range. For example, if the market moves from 100 to 200, a 0% quote is calculated as 200 but clamped to the old +50% maximum of 150, while the stored percentage remains 0%.

Should this market-price update live in PriceSelection/PriceLimits so the limits and quote can be refreshed together? Could we also cover upward and downward market ticks, including the -10% and +50% endpoints, with domain tests?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/draft/create_offer/amount_and_price/price/MuSigCreateOfferPriceController.java
```diff
@@ -115,48 +117,80 @@ public ReadOnlyBooleanProperty getIsOverlayVisible() {
...
+            UIThread.run(() -> {
+                if (pricePercentage != null) {
+                    model.getPercentage().set(pricePercentage);
+                    model.getPercentageInput().set(PercentageFormatter.formatToPercent(pricePercentage));
+                }
+            });
+        }));
+
+
+        pins.add(priceSelection.priceQuoteObservable().addObserver(priceQuote ->
```
KimStrand: Should we keep domain published quote updates display only before this PR merges?

This observer calls onQuoteInput() → applyPercentageFromQuote() → PriceSelection.onSetPricePercentage(), writing a displayed projection back into the domain. The percentage observer has a similar path through applyPercentageString() and setQuote(). Registration and market price updates can therefore rewrite a fixed price on re-entry or drift a floating percentage through monetary rounding.

#4967 documents the reproductions, but this controller is introduced by this PR and the behavior changes trade prices without user input. Should the origin separation and its regression tests land here rather than being deferred?

balag3: I filed #4967 to document the issue and make scoping and review focused, since this refactor is estimated to be quite big chunk. The plan is to deliver the fix on this branch and close the issue pointing at that commit. IS that OK?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/draft/components/MuSigPriceInput.java
balag3: Confirmed. Fixed in: 7674aa47eb. Details and testing on #4967.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/AmountSelection.java
```diff
@@ -0,0 +1,673 @@
...
+    /* --------------------------------------------------------------------- */
+    // Amount as input
+    /* --------------------------------------------------------------------- */
+
+    public void onSetFixTradeAmountFromInputAmount(Monetary inputAmount) {
+        synchronized (draftLock) {
+            checkNotNull(inputAmount, "inputAmount must not be null");
+            Market market = marketSelection.getMarket();
+            PriceQuote priceQuote = priceSelection.getPriceQuote();
+            if (amountLimits.isInitialized() && market != null && priceQuote != null) {
```
KimStrand: Should we base these input guards on the currently available limit ranges rather than the latched `isInitialized()` flag? The invalid-dependency path now clears `effectiveTradeAmountLimits` and `potentialTradeAmountLimits` while leaving `initialized` true.

After a valid USD draft is switched to EUR while BTC/USD is temporarily unavailable, entering an amount reaches `clamp(..., null)` and throws an NPE. I reproduced this through `onSetFixTradeAmountFromInputAmount`.

Could we fail softly until the current limits are available, use both ranges where required by the slider handlers, and add a market-switch regression test?

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/AmountSelection.java
```diff
@@ -0,0 +1,673 @@
...
+
+    /* --------------------------------------------------------------------- */
+    // Apply default amount and slider value
+    /* --------------------------------------------------------------------- */
+
+    private Optional<TradeAmount> findDefaultTradeAmountForMarket(Market market) {
+        // Empty when a required market price is missing (e.g. the draft started before the
+        // first price arrived): the default seed is applied when the limits first initialize.
+        try {
+            return MarketBasedAmountConversion.findTradeAmountFromUsdAmount(marketPriceService, market, DEFAULT_TRADE_AMOUNT_IN_USD);
```
KimStrand: Should we treat every non-positive required price quote as unavailable here, matching `PriceSelection` and `TradeAmountLimitUtils.findRates`? `MarketPrice.verify()` validates only the timestamp.

With a positive BTC/EUR quote but a negative BTC/USD quote, default seeding produces a negative `TradeAmount`. The limit providers reject that rate, but `isDraftReadyForReview()` still returns true because it checks only for non-null values; materializing the negative amount spec later fails verification.

Should we validate every conversion leg as `> 0`?

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/limits/RatesCache.java
```diff
@@ -0,0 +1,48 @@
...
+    private Market market;
+
+    Optional<TradeAmountLimitUtils.Rates> resolve(MarketPriceService marketPriceService, Market market) {
+        Optional<TradeAmountLimitUtils.Rates> fresh = TradeAmountLimitUtils.findRates(marketPriceService, market);
+        if (fresh.isPresent()) {
+            this.rates = fresh.get();
+            this.market = market;
+            return fresh;
+        }
+        if (rates != null && market.equals(this.market)) {
```
KimStrand: Should we invalidate the cached rates as soon as the requested market differs, even when no fresh rates are available?

With A(valid) → B(no rates), the A rates remain cached. Returning B → A while rates are still unavailable then returns the pre-switch A rates. This means retention does span a market change despite the class contract.

Could we clear the cache on the first market change and add an A → B → A regression test?

offer/src/main/java/bisq/offer/amount/OfferAmountUtil.java
```diff
@@ -295,4 +296,45 @@ public static Monetary calculateSecurityDepositAsBTC(Monetary btcSideMonetary,
         checkArgument(btcSideMonetary.getCode().equals("BTC"));
         return Monetary.from(MathUtils.roundDoubleToLong(btcSideMonetary.getValue() * securityDeposit), btcSideMonetary.getCode());
     }
+
+    /* --------------------------------------------------------------------- */
+    // Variants converting with an explicitly resolved price quote. The create-offer review uses
+    // these with a snapshot so a concurrent market-price update cannot produce mixed values.
+    /* --------------------------------------------------------------------- */
+
+    public static Optional<Monetary> findBaseSideFixedAmount(PriceQuote resolvedPriceQuote, AmountSpec amountSpec, Market market) {
```
KimStrand, line 314: Should we verify that `resolvedPriceQuote.getMarket().equals(market)` before performing these conversions?

Unlike the service-backed overloads, these six public methods accept the quote and market independently. `PriceQuote` conversion checks only the `Fiat`/`Coin` runtime type, not the currency code, so a BTC/USD quote passed with a BTC/EUR market silently treats an EUR amount as USD. A focused test expecting an `IllegalArgumentException` confirmed that no exception is currently thrown.

Should we centralize the preconditions for all six overloads and add mismatched-market coverage?

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/AmountSelection.java
balag3: Confirmed. Fixed in: 59c43689da.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/AmountSelection.java
balag3: Confirmed. Fixed in: 59c43689da.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/limits/RatesCache.java
balag3: Confirmed. Fixed in: 59c43689da.

offer/src/main/java/bisq/offer/amount/OfferAmountUtil.java
balag3: Confirmed. Fixed in: 59c43689da.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/AmountSelection.java
```diff
@@ -0,0 +1,683 @@
...
+        } catch (ArithmeticException e) {
```
KimStrand: Should we remove this catch, or update it to describe the exception it is still intended to handle? `MarketBasedAmountConversion` now filters zero and negative quotes to `Optional.empty()` before conversion, so a zero price no longer reaches this `ArithmeticException` branch. The current comment therefore describes control flow that no longer occurs.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/trade/trade_limits/simulation/MuSigTradeLimitsSimulationController.java
```diff
@@ -95,9 +97,11 @@ void onSelectFiatPaymentRail(FiatPaymentRail selectedItem) {
         }
     }
 
-    private void applySelectFiatPaymentRail(FiatPaymentRail fiatPaymentRail) {
-        model.getSelectedFiatPaymentRail().set(fiatPaymentRail);
-        String maxTradeLimit = MuSigTradeAmountLimits.getFormattedMaxTradeLimitInUsd(fiatPaymentRail);
+    private void applySelectFiatPaymentRail(FiatPaymentRail paymentRail) {
+        model.getSelectedFiatPaymentRail().set(paymentRail);
+
```
KimStrand: Should we remove the trailing spaces from this blank line so `git diff --check` is clean?

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/AmountSelection.java
balag3: Confirmed. Fixed in: bbbc435aa3.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/trade/trade_limits/simulation/MuSigTradeLimitsSimulationController.java
balag3: Confirmed. Fixed in: bbbc435aa3.

balag3:
Open product related question. None of them requires an immediate decision.

MuSig trade fee policy/schedule                                                                                                                       
Description: MuSig will have a trade fee, but the schedule and basis are undecided. The UI and the protocol have no shared fee source: the protocol          
hard-codes 5,000 sats and a fixed receiver address in NonceSharesRequestUtil.                                                                                
Current solution: The take-offer review shows a localized N/A for the fee until a single authoritative fee policy exists that both the review display and the
protocol consume.                                                                                                                                            
                                                                                                                                                               
Max price-deviation bound                                                                                                                             
Description: Offers priced too far from the market price should be treated as invalid in the offerbook and blocked from being taken, but the maximum         
deviation value has not been chosen. (The related warning threshold is decided: user setting, default 10%.)                                                  
Current solution: The full mechanism is specified (offerbook invalidation, take-side init reject, runtime confirmation block) but stays dormant until the bound is defined; today only the non-blocking warning fires.                                                                                                 
                                                                                                                                                               
Reputation source for the buyer amount cap                                                                                                            
Description: The user-specific buyer amount limit needs a real reputation input; whether the limits concept from discussion #4164 is final still has to be confirmed.                                                                                                                                                   
Current solution: UserSpecificAmountLimitsProvider uses a hardcoded mock value (4000 USD); replacement comes as a separate PR. The seller cap is deliberately absent by design (intent note in the spec).                                                                                                                  
                                                                                                                                                               
Wallet affordability check                                                                                                                            
Description: The take flow should verify the taker can actually fund the trade, without blocking fundless new users from exploring the UI (so it must not act as an amount-limits provider).                                                                                                                               
Current solution: Designed as an end-of-flow confirmation check with a wallet-funding popup; dormant until a real wallet service exists to check against, currently no affordability check runs.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/limits/TradeAmountLimitUtils.java
```diff
@@ -0,0 +1,113 @@
...
+        checkNotNull(market, "market must not be null");
+        checkNotNull(priceQuote, "priceQuote must not be null");
+        checkNotNull(usdAmount, "usdAmount must not be null");
+
+        Market usdBitcoinMarket = MarketRepository.getUSDBitcoinMarket();
+        PriceQuote btcUsdPriceQuote = marketPriceService.getMarketPriceQuoteOrThrow(usdBitcoinMarket);
+        Monetary quoteSideAmount;
+        if (market.isBtcFiatMarket()) {
+            PriceQuote btcFiatPriceQuote = marketPriceService.getMarketPriceQuoteOrThrow(market);
+            quoteSideAmount = AmountConversion.usdToFiat(btcUsdPriceQuote, btcFiatPriceQuote, usdAmount);
```
KimStrand: This “exact” take-side limit can still wrap before the final checked conversion. `AmountConversion.usdToFiat()` uses the legacy `PriceQuote` conversions, which end in `longValue()`.

A focused test with positive rates turned the intended $10,000 cap into a positive wrapped limit of €155,325,592,629,044.8384 / 1,553,255,926.29044838 BTC instead of throwing. Because the result remains positive, the downstream guard accepts it and the absolute cap can effectively be bypassed.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/limits/TradeAmountLimitUtils.java
```diff
@@ -0,0 +1,113 @@
...
+                                                      Market market,
+                                                      PriceQuote priceQuote,
+                                                      Fiat usdAmount) {
+        checkNotNull(marketPriceService, "marketPriceService must not be null");
+        checkNotNull(market, "market must not be null");
+        checkNotNull(priceQuote, "priceQuote must not be null");
+        checkNotNull(usdAmount, "usdAmount must not be null");
+
+        Market usdBitcoinMarket = MarketRepository.getUSDBitcoinMarket();
+        PriceQuote btcUsdPriceQuote = marketPriceService.getMarketPriceQuoteOrThrow(usdBitcoinMarket);
```
KimStrand: One amount constraint recomputation can combine different market-price snapshots. This reads BTC/USD and the offer market separately; for a BTC/USD offer it reads the same map entry twice. A sequential 100,000 → 200,000 mock turns a $100 cap into $200. `computeAmountConstraints()` then calls this helper independently for the absolute minimum, absolute maximum, payment-rail limit, and user limit.

`findRates()` above already captures and reuses one context to avoid this race. Should we capture one `Rates` snapshot at the start of `computeAmountConstraints()`, pass it through a checked overload for every limit?

common/src/main/java/bisq/common/monetary/TradeAmountConversion.java
```diff
@@ -37,6 +38,24 @@ public static TradeAmount toTradeAmount(Market market, PriceQuote priceQuote, Mo
         }
     }
 
+    /**
+     * Like {@link #toTradeAmount(Market, PriceQuote, Monetary)}, but fails with an
+     * ArithmeticException when the converted side does not fit into a long.
+     */
+    public static TradeAmount toTradeAmountExact(Market market, PriceQuote priceQuote, Monetary amount) {
```
KimStrand, line 47: `toTradeAmountExact()` checks that the amount belongs to `market`, but it does not verify that `priceQuote` belongs to the same market. A focused test with a BTC/EUR market, BTC/USD quote, and EUR amount succeeds and returns a BTC/EUR pair calculated using the USD rate, because `PriceQuote` checks only the `Fiat`/`Coin` runtime type.

The same invariant is now enforced in `OfferAmountUtil`. Should we validate `market.equals(priceQuote.getMarket())` in both public conversion overloads and add mismatched-market coverage?

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BitcoinAmountDisplay.java
```diff
@@ -145,7 +148,10 @@ private void updateDisplay() {
         }
 
         valueTextFlow.setVisible(true);
-        formatBtcAmount(amount);
+        try {
+            formatBtcAmount(amount);
+        } catch (Exception ignored) {
```
KimStrand: This empty broad catch leaves the previously rendered amount visible. If the property changes from a valid value to malformed text, the control is made visible before parsing fails, so stale BTC data appears current. It also hides unexpected programming failures.

Should we catch only the expected parse failure, explicitly clear or hide an invalid value, and let unexpected exceptions remain visible?

common/src/main/java/bisq/common/monetary/TradeAmount.java
```diff
@@ -53,4 +54,54 @@ public TradeAmount clamp(TradeAmount min, TradeAmount max) {
...
+    public int compareToRange(TradeAmount min, TradeAmount max) {
+        MonetaryRange baseSideLimits = new MonetaryRange(min.getBaseSideAmount(), max.getBaseSideAmount());
+        MonetaryRange quoteSideLimits = new MonetaryRange(min.getQuoteSideAmount(), max.getQuoteSideAmount());
+        int clampedBaseSideAmount = baseSideAmount.compareToRange(baseSideLimits);
+        if (clampedBaseSideAmount != 0) {
+            return clampedBaseSideAmount;
+        }
+        return quoteSideAmount.compareToRange(quoteSideLimits);
+    }
+    //
```
KimStrand: is this needed?

balag3: Confirmed. Fixed in: 95485a18a8.

common/src/main/java/bisq/common/monetary/TradeAmountConversion.java
balag3: Confirmed. Fixed in: 95485a18a8.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/limits/TradeAmountLimitUtils.java
balag3: Confirmed. Fixed in: 95485a18a8.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/limits/TradeAmountLimitUtils.java
balag3: Confirmed. Fixed in: 95485a18a8.

apps/desktop/desktop/src/main/java/bisq/desktop/components/controls/BitcoinAmountDisplay.java
balag3: Confirmed. Fixed in: dd3cdda7df.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/limits/TradeAmountLimitUtils.java
```diff
@@ -0,0 +1,114 @@
...
+     * closed refuses the take when the limit cannot be represented. The caller captures one
+     * {@link Rates} snapshot for a whole limit set so its limits never mix market prices.
+     */
+    public static TradeAmount toTradeAmountLimitExact(Rates rates,
+                                                      Market market,
+                                                      PriceQuote priceQuote,
+                                                      Fiat usdAmount) {
+        checkNotNull(rates, "rates must not be null");
+        checkNotNull(market, "market must not be null");
+        checkNotNull(priceQuote, "priceQuote must not be null");
```
KimStrand, line 102: The public limit conversion currently checks only for nulls. `PriceQuote.toBaseSideMonetaryExact` validates the monetary class rather than the currency code, so, for example, a BTC/GBP quote can be accepted for a BTC/EUR market and produce a numerically valid but incorrect limit. This relies on every caller keeping `rates`, `market`, and `priceQuote` coherent, which is fragile for code used to enforce create/take amount caps.

Should we validate that `priceQuote.getMarket()` matches `market` and that the fiat rate, when present, belongs to the same market in both conversion methods?

settings/src/main/java/bisq/settings/Cookie.java
```diff
@@ -118,6 +118,28 @@ public Optional<Double> asDouble(CookieKey key, String subKey) {
                 }).stream().findAny();
     }
 
+    public Optional<Long> asLong(CookieKey key) {
+        return asString(key)
+                .flatMap(stringValue -> {
+                    try {
+                        return Optional.of(Long.parseLong(stringValue));
+                    } catch (Throwable t) {
```
KimStrand: The newly added long-cookie API currently has no call sites. It also extends the existing broad `Throwable` catch pattern, which can suppress VM or linkage failures even though malformed `Long.parseLong` input only requires handling `NumberFormatException`.

Should we remove these unused overloads for now, or, if they are intentionally retained, narrow both new catches to `NumberFormatException`?

KimStrand (commented):
Six newly added test classes start directly with the package declaration.
Can we please add the standard licence header to these files as well?

settings/src/main/java/bisq/settings/Cookie.java
balag3: Confirmed. Fixed in: 679c27caf6.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/limits/TradeAmountLimitUtils.java
balag3: Confirmed. Fixed in: ac9c1490a1.

balag3:
Confirmed. Fixed in: 4683bc531e.

offer/src/main/java/bisq/offer/amount/spec/AmountSpecFactory.java
```diff
@@ -48,6 +49,15 @@ public static BaseSideAmountSpec createBaseSideAmountSpec(boolean useRangeAmount
                                                               TradeAmount maxTradeAmount,
                                                               TradeAmount fixTradeAmount) {
         if (useRangeAmount) {
+            // The base side (sats) of two quote side amounts that are equal once rounded to the
+            // fiat display precision can still differ by a few units due to price conversion
+            // rounding, which would create a spurious range offer. Decide range vs fixed on the
+            // rounded quote side value so a range only remains when the fiat amounts actually differ.
+            Monetary minQuoteSideAmount = minTradeAmount.getQuoteSideAmount();
+            Monetary maxQuoteSideAmount = maxTradeAmount.getQuoteSideAmount();
+            if (minQuoteSideAmount.isEqual(maxQuoteSideAmount, minQuoteSideAmount.getLowPrecision())) {
```
HenrikJannsen, line 58: We would need to take altcoins into account here as well. Using `TradeAmount.getNonBitcoinSideAmount()` would do that switch internally. 

But I am not sure if that override would cause problems elsewhere. 
Did that case that rounding causes same display amounts but underlying quote and base amounts are not the same actually happened? I assume that is super rare. Introducing a potential inconsistency at the domain structure might not be a good trade-off to fix a super rare confusing UI state.
Also I assume the btc amount has still differences.

I would only consider it a fixed price offer if both are exactly the same.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/AmountSelection.java
```diff
@@ -165,6 +165,20 @@ private void handlePotentialTradeAmountLimitsChange(TradeAmountRange potentialTr
     private void applyPotentialTradeAmountLimits(TradeAmountRange potentialTradeAmountLimits) {
         MonetaryRange inputAmountRange = toInputSideMonetaryRange(potentialTradeAmountLimits);
         model.setInputAmountRange(inputAmountRange);
+        // All slider values are fractions of the input amount range. When the range changes but the
```
HenrikJannsen: How can it be that the slider value change and the amount do not change?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/draft/take_offer/amount/container/fix/slider/MuSigFixAmountSliderView.java
```diff
@@ -49,29 +51,45 @@ public MuSigFixAmountSliderView(MuSigFixAmountSliderModel model,
...
         subscriptions.clear();
-        slider.valueProperty().unbindBidirectional(model.getGetSliderValue());
+        slider.valueProperty().removeListener(userGestureListener);
     }
 
-    private void updateSliderValueAndStyle() {
-        double maxAllowedValue = model.getMaxAllowedValue().get();
-        if (slider.getValue() > maxAllowedValue) {
-            slider.setValue(maxAllowedValue);
+    private final ChangeListener<Number> userGestureListener = (observable, oldValue, newValue) -> {
```
HenrikJannsen: We put fields at the beginning of the class.

mu-sig/src/main/java/bisq/mu_sig/MuSigService.java
```diff
@@ -260,15 +256,24 @@ public void republishMyOffers() {
...
+    // specification, "Handoff to the trade protocol").
     public MuSigProtocol takerCreatesProtocol(UserIdentity takerIdentity,
                                               MuSigOffer muSigOffer,
                                               Monetary takersBaseSideAmount,
                                               Monetary takersQuoteSideAmount,
                                               PaymentMethodSpec<?> takersPaymentMethodSpec,
-                                              Account<?, ?> takersAccount)
+                                              Account<?, ?> takersAccount,
+                                              long marketPrice,
+                                              boolean proceedWithoutMediator)
```
HenrikJannsen: Do we want to allow the case to trade without mediator? I guess arbitration requires mediation to have completed.

HenrikJannsen (approved):
utACK

Only lightly reviewed

offer/src/main/java/bisq/offer/amount/spec/AmountSpecFactory.java
balag3: This is just a port from #4865 that fixes your #4410, the rewrite just dropped it.
Also this only affects the fiat market, altcoins take the quote side overload wit the exact comparison same as in #4865.
If you prefer exact equality, that reverts #4865.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/AmountSelection.java
balag3: So that the slider value represents the fraction of the range. And when the range changes the slider value is recomputed based on the amount. Example:
1st state: range: 10k, slider value: 0.4, amount: 4k
payment method changes -> range changes (from Advanced Cash (10k max) to Wise (5k max) )
2nd state: range: 5k, slider value: 0.8, amount : 4k
There was even an associated stale slider bug in #4883 (Fix the stale slider clamp bug, with a domain test.)

Altough I agree that the comment is a bit misleading and fails to explain this correctly, I will reword it

mu-sig/src/main/java/bisq/mu_sig/MuSigService.java
balag3: yes, I took the consent popup over from Bisq Easy where it is OK, but here arbitration depends on mediation, so a missing mediator should be a hard stop here. Fixing it.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/mu_sig/offer/draft/take_offer/amount/container/fix/slider/MuSigFixAmountSliderView.java
balag3: Confirmed. Fixed in: c8856bd394.

mu-sig/src/main/java/bisq/mu_sig/MuSigService.java
balag3: Fixed in: daf5a669dc.

offer/src/main/java/bisq/offer/mu_sig/use_case/create_offer/amount/AmountSelection.java
balag3: Comment reworded in d2fe46e11a.

offer/src/main/java/bisq/offer/amount/spec/AmountSpecFactory.java
HenrikJannsen: Ah was me running into that super rare case ;-)... 
I guess my initial worry is not justified as we do not use base amount, quote amount and price but only 2 of them, thus a slight difference should not cause domain inconsistency.

HenrikJannsen (approved):
utACK

HenrikJannsen (approved):
utACK

PR #4968, opened 2026-08-26
Fix trading-restriction alerts: structured errors for API clients + UNSPECIFIED wildcard matching

rodvar:
### Problem

When the security manager publishes an EMERGENCY alert with `haltTrading` or `requireVersionForTrading`, the trusted node (api-app registers its trade service as `AppType.DESKTOP`) rejects take-offer with a raw `checkArgument` message written for desktop users ("For trading you need to have version X installed…"). That text travels verbatim over the API, so Connect apps have no reliable way to classify the failure — mobile users got a cryptic error and a blocking dialog, which forced the Bisq Connect 0.8.2 hotfix (client-side string matching on the English backend text).

While fixing this we found the opposite gap: an EMERGENCY alert published with app type `UNSPECIFIED` shows a notification banner (`AlertNotificationsService` treats `UNSPECIFIED` as a wildcard) but does **not** restrict trading - `BisqEasyTradeService`/`MuSigTradeService` matched the app type with strict equality. A network-wide halt would silently fail open in the trade services.

### Dev logs

 - long term fix to avoid https://github.com/bisq-network/bisq-mobile/issues/1775 ocurrence
 - Implemented TradeRestrictedException and usages to handle that case appropriately
 - closed gap with Unespecified restriction to afloat handling
 - handled structured error in Trade take offer
 - harden alerts bookeeping (current toggle impl make alerts overlap one another potentially causing missings)i
 - improved semantics on security alerts checks
 - added AI generated tests

## Summary by CodeRabbit

- **New Features**
  - Added clearer handling for trading restrictions caused by trading halts or required minimum versions.
  - Trade APIs now return structured error details, including restriction type and required version when applicable.
  - Alerts applying to all application types are now recognized consistently.
  - Trading version requirements are updated automatically when authorized alerts change.

- **Bug Fixes**
  - Improved detection and enforcement of applicable trading restrictions across desktop and trading services.
  - Preserved existing error messages while providing more precise restriction information.

### DIFF
--- a/api/src/main/java/bisq/api/rest_api/endpoints/trades/TradeRestApi.java
+++ b/api/src/main/java/bisq/api/rest_api/endpoints/trades/TradeRestApi.java
@@ -50,6 +50,7 @@ import bisq.offer.price.spec.PriceSpec;
 import bisq.presentation.formatters.AmountFormatter;
 import bisq.support.SupportService;
 import bisq.support.mediation.bisq_easy.BisqEasyMediationRequestService;
+import bisq.trade.TradeRestrictedException;
 import bisq.trade.TradeService;
 import bisq.trade.bisq_easy.BisqEasyTrade;
 import bisq.trade.bisq_easy.BisqEasyTradeService;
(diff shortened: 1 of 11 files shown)


KimStrand (commented):
utACK

PR #4972, opened 2026-08-27
chore: sync #4940 to main — Add two-party private chat to the API

salottodev:
Manual sync of #4940 from `for-mobile-based-on-2.1.12` to `main` (with conflict resolution).

## Summary by CodeRabbit

* **New Features**
  * Added REST API support for two-party private chats, including channel creation, messaging with citations, reactions, leaving channels, and notification consumption.
  * Added real-time updates for private chat channels, messages, reactions, and unread counts.
  * Added private-chat feature access and permission controls.
  * Mobile notifications now support direct-chat routing with channel and peer information.

* **Bug Fixes**
  * Leaving a background chat no longer changes the selected conversation.
  * Banned senders and recipients receive clear send-refusal responses.


api/src/main/java/bisq/api/web_socket/domain/chat/trade/TradeChatMessagesWebSocketService.java
```diff
@@ -109,8 +109,12 @@ public void onRemoved(Object element) {
 
             @Override
             public void onCleared() {
+                // Cleared as well as unbound, matching shutdown() below. A bulk clear is reachable:
+                // the channel store's applyPersisted replaces the set through setAll, and
+                // CollectionObserver#onAllSet is onCleared() followed by onAllAdded(values). Channels
+                // that are not re-added would otherwise leave a dead entry behind under their id.
                 new ArrayList<>(messagesByChannelIdPins.values()).forEach(Pin::unbind);
-
+                messagesByChannelIdPins.clear();
```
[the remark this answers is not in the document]
salottodev: Not reachable. `onCleared()` only fires from `setAll`, and the only `setAll` on this store is `applyPersisted`. That runs from `Executable` via `readAllPersisted().join()` before `launchApplication`, so before any service calls `initialize()`. This service registers its channel observer in `initialize()`, which means there is no observer yet when that `setAll` happens, and `onCleared()` never runs concurrently with `onAdded()`. There is no other `clear()` or `setAll` on that collection anywhere in the repo. The `clear()` added here is defensive cleanup rather than a fix for a real window.

chat/src/main/java/bisq/chat/priv/PrivateChatChannelService.java
```diff
@@ -98,13 +98,28 @@ protected CompletableFuture<SendMessageResult> sendMessage(String messageId,
...
-        if (bannedUserService.isUserProfileBanned(myUserIdentity.getUserProfile())) {
-            return CompletableFuture.failedFuture(new RuntimeException());
-        }
-        if (isPeerBanned(receiver)) {
-            addPeerBannedSystemMessage(channel, myUserIdentity, receiver, date);
-            return CompletableFuture.failedFuture(new RuntimeException("Peer is banned"));
+        Optional<SendRejection> rejection = findSendRejection(channel, receiver);
+        if (rejection.isPresent()) {
+            if (rejection.get() == SendRejection.PEER_BANNED) {
+                addPeerBannedSystemMessage(channel, myUserIdentity, receiver, date);
```
KimStrand, line 126: `PEER_BANNED` now stores a protocol log message before returning `SendOutcome.rejected`. The REST endpoint maps that outcome to 409 but documents that nothing was stored and nothing will appear on the message stream. On the first banned-peer send, subscribers can therefore receive a new system message after the API reported a no-storage rejection.

This combination is specific to the sync: #4940 returned directly on rejection, while main supplied the banned peer notice. Should we either keep rejection side effect free here, or update the 409 contract and add a test covering the system message side effect?

salottodev: Confirmed. I updated the contract rather than dropping the notice, since it is a real desktop feature (#4769). The 409 on `sendTextMessage` now says the message itself is neither stored nor sent, but that a `PEER_BANNED` refusal can still add a system message to the channel which does reach the stream. The reaction endpoint says the opposite explicitly, so the difference does not read as an oversight.

Added four tests in `PrivateChatChannelServiceTest.RejectedSend`. Removing the `addPeerBannedSystemMessage` call fails two of them, so they bite.

KimStrand (commented):
utACK

PR #4976, opened 2026-08-31
feature/my_contacts_api_support

rodvar:
- implemented APIs to support connect apps "my contacts" feature + gating + permissions

## Summary by CodeRabbit

* **New Features**
  * Added API support for adding, removing, and annotating contacts with tags, notes, and trust scores.
  * Added contact reasons and real-time contact-list updates through subscriptions.
  * Added contact-related API permissions and feature configuration.
* **Validation**
  * Added validation for unknown profiles, duplicate or invalid contacts, annotation lengths, and trust-score limits.
* **Tests**
  * Added coverage for contact endpoints, permissions, configuration, and real-time updates.


rodvar:
@KimStrand

api/src/main/java/bisq/api/rest_api/endpoints/contacts/ContactsRestApi.java
```diff
@@ -0,0 +1,197 @@
...
+                && (request.trustScore() < ContactListService.CONTACT_LIST_ENTRY_MIN_TRUST_SCORE
+                || request.trustScore() > ContactListService.CONTACT_LIST_ENTRY_MAX_TRUST_SCORE)) {
+            return buildResponse(Response.Status.BAD_REQUEST, "Trust score must be between 0 and 1");
+        }
+        try {
+            Optional<ContactListEntry> entry = contactListService.findContactListEntry(userProfileId);
+            if (entry.isEmpty()) {
+                return buildNotFoundResponse("The profile is not a contact");
+            }
+            if (request.tag() != null) {
```
KimStrand: ContactListEntry hashes every field, including mutable annotations, while entries are stored in a ConcurrentHashMap-backed set. Mutating an annotation changes the hash without moving the entry to its new bucket, so DELETE can return changed=false while retaining the contact. I reproduced the exact add → setTag → remove failure. Should we give entries stable identity-based equality/hash, or replace set elements instead of mutating them?

api/src/main/java/bisq/api/rest_api/endpoints/contacts/ContactsRestApi.java
```diff
@@ -0,0 +1,197 @@
...
+            }
+            Optional<UserProfile> userProfile = userProfileService.findUserProfile(userProfileId);
+            if (userProfile.isEmpty()) {
+                return buildNotFoundResponse("No user profile found for the given profile ID");
+            }
+            UserProfile myUserProfile = userIdentityService.getSelectedUserIdentity().getUserProfile();
+            boolean changed = contactListService.addContactListEntry(userProfile.get(),
+                    myUserProfile,
+                    ContactListEntryDtoMapping.toBisq2Model(request.contactReason()));
+            return buildOkResponse(new AddContactResponse(changed,
```
KimStrand: An idempotent add returns the complete stored entry, and the WebSocket REST proxy logs the complete response, including tag, notes, and trust score, at INFO. PUT bodies are also logged at DEBUG and at WARN on forwarding failures. This defeats the endpoint’s effort to keep user authored identifying information out of logs. Should we make proxy logging metadata only or redact bodies before exposing annotations through it?

api/src/main/java/bisq/api/rest_api/endpoints/contacts/ContactsRestApi.java
KimStrand: The endpoint invokes three independent setters. Each requests persistence and emits an edit notification, but persistence drops calls made within one second or while the first write is active. The first asynchronous, shallow snapshot can serialize between mutations; an unclean shutdown before another persistence operation can therefore restore only some fields from a Save that returned 204. Subscribers can likewise receive intermediate snapshots. Should we add one domain-level batch update that applies all supplied fields, persists once, and emits one notification?

api/src/main/java/bisq/api/rest_api/endpoints/contacts/ContactsRestApi.java
```diff
@@ -0,0 +1,197 @@
...
+        }
+        if (request.tag() != null && request.tag().length() > ContactListService.CONTACT_LIST_ENTRY_MAX_TAG_LENGTH) {
+            return buildResponse(Response.Status.BAD_REQUEST,
+                    "Tag must not exceed " + ContactListService.CONTACT_LIST_ENTRY_MAX_TAG_LENGTH + " characters");
+        }
+        if (request.notes() != null && request.notes().length() > ContactListService.CONTACT_LIST_ENTRY_MAX_NOTES_LENGTH) {
+            return buildResponse(Response.Status.BAD_REQUEST,
+                    "Notes must not exceed " + ContactListService.CONTACT_LIST_ENTRY_MAX_NOTES_LENGTH + " characters");
+        }
+        if (request.trustScore() != null
```
KimStrand, line 188: Both range checks rely on < and >, which are false for Double.NaN. The configured Jackson mapper accepts "NaN" as a Double, after which the value is persisted and serialized back as a JSON string rather than a number. This violates the wire contract and can break consumers of the contacts snapshot. Should we reject non-finite values with Double.isFinite in both API and domain validation?

api/src/main/java/bisq/api/rest_api/endpoints/contacts/ContactsRestApi.java
```diff
@@ -0,0 +1,197 @@
...
+    public Response addContact(@PathParam("userProfileId") String userProfileId, AddContactRequest request) {
+        // Jersey hands back null for an absent entity rather than rejecting the request.
+        if (request == null || request.contactReason() == null) {
+            return buildResponse(Response.Status.BAD_REQUEST, "A contact reason is required");
+        }
+        try {
+            if (userIdentityService.findUserIdentity(userProfileId).isPresent()) {
+                return buildResponse(Response.Status.BAD_REQUEST, "Cannot add an own identity as contact");
+            }
+            Optional<UserProfile> userProfile = userProfileService.findUserProfile(userProfileId);
```
KimStrand, line 112: The profile service removes expired network profiles, while the contact list retains its embedded profile. POST checks the transient profile service before checking whether the contact already exists, so an already present contact receives 404 instead of the documented changed=false response. Should we return the stored contact first when it already exists, consulting the network profile only for a new addition?

KimStrand (commented):
General API semantics question: could we let the HTTP status communicate the add outcome instead of duplicating it with changed?

- Return 201 Created, with a Location header and the current entry, when a contact is newly added.
- Return 200 OK with the current entry when it already exists.

For DELETE, 200 {"changed": ...} is valid. If clients do not need to distinguish “removed” from “already absent,” could we simplify both successful outcomes to 204 No Content?

I would keep POST for adding because this request behaves like a creation command and does not replace the complete existing contact representation. This would preserve the idempotent behavior while expressing outcomes through standard HTTP semantics.

api/src/main/java/bisq/api/rest_api/endpoints/contacts/ContactsRestApi.java
rodvar: I verified this finding correct, but the fix is unrelated to this PR. Could we please create an issue for it instead of fixing here?

rodvar:
> General API semantics question: could we let the HTTP status communicate the add outcome instead of duplicating it with changed?
> 
> * Return 201 Created, with a Location header and the current entry, when a contact is newly added.
> * Return 200 OK with the current entry when it already exists.
> 
> For DELETE, 200 {"changed": ...} is valid. If clients do not need to distinguish “removed” from “already absent,” could we simplify both successful outcomes to 204 No Content?
> 
> I would keep POST for adding because this request behaves like a creation command and does not replace the complete existing contact representation. This would preserve the idempotent behavior while expressing outcomes through standard HTTP semantics.

Thanks for the review and great findings as usual @KimStrand  - I'm addressing most of it in my next push with the exception of the ones commented on + this general comment:

Please consider:

 1. the WebSocket proxy that Connect clients actually use forwards only requestId, statusCode, and body - a Location header is dropped completely, so half of his proposed 201 idiom/semantics is invisible to the primary consumer
 2. DELETE→204-always deliberately erases the removed-vs-already-absent distinction, which the mobile client uses (analytics tracks Contact.Removed only when changed), and an empty 204 body would also need special-casing in the client's typed deserialization. The changed-boolean style also mirrors ContactListService's own idempotent booleans end to end. 

Regarding return 201 vs 200 to express created-vs-existing at the status level while keeping the same body — that one we can do. However we should consider that the initial intention of the `RestAPI` implementation was to be REST; which means technically we shouldn't be using POST for creation, only for update and use PUT for that purpose.
But since its generally accepted anyways, I think the 201 vs 200 is better although doesn't change anything for mobile clients at themoment - will include in my push :)

rodvar:
@KimStrand all done. Also re-tested [client PR](https://github.com/bisq-network/bisq-mobile/pull/1791) and no changes are needed 👍

KimStrand (commented):
utACK

PR #4977, opened 2026-08-31
Make chat mention tagging follow the caret

### LINKED ISSUE #4942: Tagging @username in chat is flaky
I confirmed that if you try to tag a user, for example typing `@` and then  `sudd...` it will only show the list of matching profile names if the `@` is at the very end of the current string, so for example if you want to insert a tag before other characters already on the screen, it will not show the names list.

Also it would be really cool if, once the list appears, you can ArrowUp and ArrowDown and autocomplete with Tab and/or Enter.

balag3:
Fixes #4942

The mention popup and the completion were both anchored to the end of the input text: `StringUtils.deriveWordStartingWith` searched from the last `@` and required the word to end the string, and completing a suggestion replaced an end-anchored regex match with the caret forced to the end. So a mention typed in front of existing text never showed suggestions, and completing one would have replaced the wrong range.

### What changed

- The mention token is now parsed at the caret (new `ChatMentionParser`): the word between the nearest `@` before the caret and the caret itself. The popup filter observes both the text and the caret position, so clicking into the middle of a token re-opens the suggestions for it.
- Completion replaces exactly the parsed range — never text past the caret (IntelliJ-style), and the trailing space is only inserted in front of letters or digits, so punctuation stays attached to the mention.
- Keyboard navigation on the popup: ArrowUp/ArrowDown move the selection, Enter or Tab complete it, Shift+Enter still inserts a newline, and Enter with no match sends the message as usual. The keys are handled on the popup window itself because JavaFX redirects the owner window's key events there while the popup is showing — an input-field handler never sees them.
- Escape dismisses the popup for the current mention token; editing the same token keeps it dismissed, a fresh `@` re-arms it. Completing behaves like Escape for that token.
- The selected row is highlighted and kept in view without re-anchoring the list to the top on every key press.
- `StringUtils.deriveWordStartingWith` had a single caller and is replaced by the caret-aware parser, so it is removed.

### Tests

- `ChatMentionParserTest` covers the token parsing (mid-text, caret bounds, adjacent words, punctuation).
- Controller tests cover the caret-aware insertion and dismissal state.
- View tests drive the real popup event path (arrow selection, Enter/Tab completion, Escape, empty-list passthrough) — they go through the popup window's key handling, which is the load-bearing piece.

### Manual testing

Two clients (Alice + Bob) on a local CLEAR network, both in the same public discussion channel. All steps on Alice unless noted. Client logs were checked afterwards — no exceptions.

**The reported bug — mention in front of existing text**
1. Type `hello world`, press Home, then type `@` and the first letters of Bob's nickname → the suggestion popup opens and filters while typing (it never opened mid-text before).
2. Click Bob's name → the input becomes `@Bob hello world`: single space after the mention, the existing text untouched, caret right after the inserted space.

image

**Caret-bounded completion**
1. Type `hi @jjj there`, click between the `j`s → the popup opens, filtering on the part before the caret. Completing replaces only the part before the caret; the remaining `j`s and ` there` stay.
2. Type `Hi @xx, more` with the caret right after `xx`, complete → `Hi @Name, more`, no extra space in front of the comma.

**Keyboard navigation**
1. Type `@` alone → popup with the full user list.
2. ArrowDown twice, ArrowUp once → the selection highlight moves accordingly; typing focus stays in the input.
3. Enter → the highlighted name completes into the input, the message is not sent.
4. `@` + letters, Tab → same completion.
5. `@zzzz` (no match) → arrows move the caret in the text as usual, Enter sends the message with the literal `@zzzz`.
6. `@` + letters, Shift+Enter → a newline is inserted, the popup is unaffected.

image

**Escape — dismissal is per token**
1. `@` + letters, Escape → popup closes; typing more letters of the same token or backspacing keeps it closed.
2. A space and a fresh `@` re-open it.
3. Compose `@aa bbb @cc`, Escape on the last token, click into the middle of `@aa` → the popup opens for `@aa` (the dismissal does not leak between tokens).

**Regressions**
1. Plain text + Enter → sends.
2. Empty input + ArrowUp → edit-last-message behavior unchanged.
3. Reply flow that inserts `@Bob` from a message's context menu → still works.
4. `x@name` (no space before the `@`) → no popup, as before.
5. Popup open, switch channel → popup closes.
6. On Bob: a received message containing `@Bob` highlights as before.

Screenshot from 2026-08-31 12-09-58

## Summary by CodeRabbit

- **New Features**
  - Improved chat mention completion with accurate caret-based matching.
  - Added keyboard navigation and completion using Arrow keys, Tab, and Enter.
  - Added Escape dismissal and reliable popup reopening for new mention searches.
  - Mention completion now preserves surrounding text and places the caret correctly.

- **Bug Fixes**
  - Prevented accidental deletion of text following a mention.
  - Improved handling near punctuation, newlines, and partially edited mentions.
  - Added clearer visual highlighting for the selected mention result.

### DIFF
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/ChatMessageContainerController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/ChatMessageContainerController.java
@@ -46,8 +46,10 @@ import bisq.settings.SettingsService;
 import bisq.user.banned.BannedUserService;
 import bisq.user.identity.UserIdentity;
 import bisq.user.identity.UserIdentityService;
+import bisq.desktop.main.content.chat.message_container.components.ChatMentionParser;
 import bisq.user.profile.UserProfile;
 import bisq.user.profile.UserProfileService;
+import com.google.common.annotations.VisibleForTesting;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
@@ -211,10 +213,37 @@ public class ChatMessageContainerController implements bisq.desktop.common.view.
         return chatMessagesListController.scrollPageDown();
     }
 
-    void onUserProfileSelected(UserProfile user) {
-        String content = model.getTextInput().get().replaceAll("@[a-zA-Z\\d]*$", "@" + user.getUserName() + " ");
-        model.getTextInput().set(content);
-        model.getCaretPosition().set(content.length());
+    @VisibleForTesting
+    ChatMessageContainerModel getModel() {
+        return model;
+    }
+
+    void onUserProfileSelected(UserProfile user, ChatMentionParser.MentionMatch match) {
+        String text = model.getTextInput().get();
+        if (text == null
+                || match.indicatorIndex() < 0
+                || match.caretPosition() > text.length()
+                || match.indicatorIndex() >= match.caretPosition()) {
+            // The text changed between parsing and selection; dropping the completion is safer
+            // than replacing an arbitrary range.
+            return;
+        }
+        String prefix = text.substring(0, match.indicatorIndex());
+        String suffix = text.substring(match.caretPosition());
+        String mention = "@" + user.getUserName();
+        int caretPosition = prefix.length() + mention.length();
+        String separator = "";
+        if (suffix.startsWith(" ")) {
+            // Reuse the existing space and continue typing after it.
+            caretPosition++;
+        } else if (suffix.isEmpty() || Character.isLetterOrDigit(suffix.charAt(0))) {
+            // A separating space is only needed before a word; punctuation and line breaks
+            // must stay attached to the mention.
+            separator = " ";
+            caretPosition++;
+        }
+        model.getTextInput().set(prefix + mention + separator + suffix);
+        model.getCaretPosition().set(caretPosition);
     }
 
     void onOpenProfileCard(UserProfile userProfile) {
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/ChatMessageContainerController.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/ChatMessageContainerController.java
@@ -243,7 +243,7 @@ public class ChatMessageContainerController implements bisq.desktop.common.view.
             caretPosition++;
         }
         model.getTextInput().set(prefix + mention + separator + suffix);
-        model.getCaretPosition().set(caretPosition);
+        model.getCaretPositionRequest().set(ChatMessageContainerModel.CaretPositionRequest.of(caretPosition));
     }
 
     void onOpenProfileCard(UserProfile userProfile) {
--- a/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/components/ChatMentionPopupMenu.java
+++ b/apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/components/ChatMentionPopupMenu.java
@@ -22,10 +22,11 @@ import bisq.desktop.common.utils.ImageUtil;
 import bisq.desktop.components.controls.BisqPopup;
 import bisq.desktop.components.controls.BisqTextArea;
 import bisq.i18n.Res;
+import bisq.desktop.main.content.chat.message_container.components.ChatMentionParser.MentionMatch;
 import bisq.user.profile.UserProfile;
 import javafx.beans.binding.Bindings;
-import javafx.beans.property.SimpleStringProperty;
-import javafx.beans.property.StringProperty;
+import javafx.beans.property.ObjectProperty;
+import javafx.beans.property.SimpleObjectProperty;
 import javafx.beans.value.ChangeListener;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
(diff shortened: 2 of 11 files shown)


apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/ChatMessageContainerController.java
```diff
@@ -211,10 +213,37 @@ boolean onPageDownKeyPressed() {
...
+            // Reuse the existing space and continue typing after it.
+            caretPosition++;
+        } else if (suffix.isEmpty() || Character.isLetterOrDigit(suffix.charAt(0))) {
+            // A separating space is only needed before a word; punctuation and line breaks
+            // must stay attached to the mention.
+            separator = " ";
+            caretPosition++;
+        }
+        model.getTextInput().set(prefix + mention + separator + suffix);
+        model.getCaretPosition().set(caretPosition);
```
KimStrand: Repeated completions can leave the caret at the start of the input. `caretPosition` is a persistent `IntegerProperty`, and the view observes it through a change listener. Updating the bound text resets JavaFX's caret to 0; if a later completion computes the same position as the previous one, for example, completing `@alice` at index 0 in two consecutive messages, this `set(...)` emits no change and `positionCaret(...)` is never called. This is a merge blocker because subsequent typing then occurs at the beginning of the message.

Should we model caret positioning as an event-like request, such as resetting the property before setting it or including a sequence number?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/components/ChatMentionPopupMenu.java
```diff
@@ -77,30 +86,149 @@ public ChatMentionPopupMenu(BisqTextArea inputField, Consumer<UserProfile> userP
...
+        // continue to the input field through the same redirection.
+        addEventFilter(KeyEvent.KEY_PRESSED, this::handleKeyPressed);
+
+        mentionMatchChangeListener = (observableValue, oldValue, newValue) -> {
             if (newValue != null) {
-                filteredList.setPredicate(item -> item.matchUserName(newValue));
-                sortedList.setComparator(sortByPrefixMatchingQuery(newValue));
+                // The caret can jump from one mention token straight into another; the token
+                // identity, not a null transition, decides when the dismissal is re-armed.
+                boolean isNewToken = oldValue == null || oldValue.indicatorIndex() != newValue.indicatorIndex();
```
KimStrand: Dismissal is keyed only by `indicatorIndex`, so it can suppress an unrelated fresh token created at the same offset. For example, after Escape dismisses `@al`, selecting and replacing it with `@bo` leaves both matches at index 0. Because `lastStableMatch` prevents a null transition during the replacement, `dismissed` never resets and the new popup remains closed.

Should we track token replacement/generation rather than only the absolute `@` index, and cover same-offset replacement in a test?

KimStrand (commented):
The PR description says the view tests cover ArrowUp/ArrowDown, Tab, Shift+Enter, and empty-result passthrough, but the added tests only exercise Enter and Escape related behavior. This leaves most of the keyboard acceptance criteria unverified.

Should we add coverage for selection movement, Tab completion, modifier passthrough, and no match event propagation?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/components/ChatMentionPopupMenu.java
```diff
@@ -77,30 +90,162 @@ public ChatMentionPopupMenu(BisqTextArea inputField, Consumer<UserProfile> userP
...
+        hide();
+    }
+
+    private static boolean isWithinTokenEdit(String oldQuery, String newQuery) {
+        if (newQuery.isEmpty()) {
+            // The token content was cleared; whatever comes next is a fresh mention attempt.
+            return oldQuery.isEmpty();
+        }
+        // Typing extends the query and Backspace truncates it; anything else replaced it.
+        return oldQuery.startsWith(newQuery) || newQuery.startsWith(oldQuery);
```
KimStrand: This still infers token identity from query prefixes, which produces conflicting dismissal behavior. After dismissing `@a`, Backspace to the bare `@` reopens the popup, although backspacing within the current token is supposed to keep it dismissed. Conversely, replacing the dismissed `@al` with a fresh `@alice` at the same offset keeps it dismissed because the queries are prefix-related, although a fresh `@` should re arm it.

Should we track the actual token/edit generation instead of inferring identity from the query strings, or explicitly define and document a different dismissal contract?

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/ChatMessageContainerController.java
balag3: Confirmed. Fixed in: 6d19d27548.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/components/ChatMentionPopupMenu.java
balag3: Confirmed. Fixed in: 6d19d27548.

balag3:
Confirmed. Fixed in: 6d19d27548.

apps/desktop/desktop/src/main/java/bisq/desktop/main/content/chat/message_container/components/ChatMentionPopupMenu.java
balag3: Confirmed. Fixed in: 935cb1565e.

KimStrand (commented):
utACK

PR #4979, opened 2026-09-01
Expose public chat over REST and websocket

salottodev:
## Summary by CodeRabbit

* **New Features**
  * Added public chat REST APIs for channels, message history, sending, editing, deletion, reactions, and notifications.
  * Added live WebSocket updates for public chat channels, messages, and reactions, including channel-specific subscriptions.
  * Added contacts and public chat API availability with access controls.
  * Added citation support and selectable sender identities when sending messages.
* **Bug Fixes**
  * Improved chat request validation and standardized asynchronous request timeouts.
  * Filters expired, banned, or unavailable authors and reactions from public chat results.


api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatReactionsWebSocketService.java
```diff
@@ -0,0 +1,279 @@
...
+     * {@code findDto}, which needs the sender's profile, so a sender pruned from the profile store
+     * after the reaction was pushed produces no {@code REMOVED} at all and the client keeps an orphan
+     * until it reconnects. Carrying the removal without the profile would mean a dto that admits an
+     * absent author, which is a wider contract change than the window is worth.
+     */
+    private void handleRemovedReaction(CommonPublicChatMessageReaction reaction) {
+        if (shutdownStarted) {
+            return;
+        }
+        push(reaction, ModificationType.REMOVED);
```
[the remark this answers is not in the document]
salottodev: The removal path skips `isVisible` on purpose, and the DTO does carry the sender profile. The subscriber, though, is the node operator's own paired client, and `GET /user-profiles?ids=` (`UserProfileRestApi.java:136`) already returns any profile in the store with no ban filter. So the profile in that payload tells that client nothing it could not fetch directly. What the REMOVED adds is the tuple (banned sender, reactionId, messageId), and the client drops it: it deletes by id, and it never held that one.

Both suggested fixes are worse than what they fix:

- Filtering the removal by visibility brings back the orphan this shape avoids: a reaction pushed before the ban would never leave the client. `PrivateChatReactionsWebSocketService:350` gates the removal path and has exactly that behaviour today.
- Tracking prior recipients means per-subscriber state with no bound, which is the kind of leak `install`, `unbindMessage` and `isLive` already exist to prevent.

So I am keeping the behaviour, but the javadoc was thin here: it covered the pruned-profile window and not this one. Extended it with:

> The same goes for a subscriber that connected after the ban and never saw the reaction in its snapshot: the sender's profile rides along in the payload, but `GET /user-profiles` already serves it to that client without filtering banned users.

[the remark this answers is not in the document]
salottodev: The mappings are right, but that client cannot be produced: `ApiConfig.java:64` hardcodes `grantedPermissions` to `Permission.autoGrantable()`, and `ApiService.java:357` is the only `createPairingCode` call site. Every permission is `Kind.STANDARD`, so the grant is always the whole enum. (Once the `TODO` at `ApiConfig.java:62` moves that to config, the restricted client becomes real.)

Pairing aside, this topic ships profiles by design: `CommonPublicChatMessageDto:35` carries the author's, `:38` the citation author's, and every embedded reaction its sender's. What REMOVED adds is banned users specifically.

You are still pointing at something real, and it is wider than this thread: `PublicChatMessagesWebSocketService:145` skips the same check, and its payload carries the author's profile, the citation author's and every reactor's. Neither of your fixes works, though. Tracking recipients breaks on reconnect, where a fresh subscriber id has no history. Gating removals on visibility guarantees the orphan, which is what `PrivateChatReactionsWebSocketService:350` does today.

The fix is that a removal should not carry the entity at all, since the client deletes by id. That drops the disclosure instead of filtering it, and closes the limitations at `PublicChatReactionsWebSocketService:252` and `PublicChatMessagesWebSocketService:140` as well. Not in this PR, though: every topic in the module sends the full DTO on REMOVED, even the ones with no visibility filter to bypass, so this is a convention change. Public chat having no released consumer makes now the cheapest time to decide it.

rodvar:
@KimStrand

api/src/main/java/bisq/api/rest_api/endpoints/chat/ChatRequestValidation.java
```diff
@@ -0,0 +1,89 @@
...
+        if (StringUtils.isEmpty(text)) {
+            return Optional.of("text must not be empty.");
+        }
+        if (text.length() > ChatMessage.MAX_TEXT_LENGTH) {
+            return Optional.of("text must not be longer than " + ChatMessage.MAX_TEXT_LENGTH + " characters.");
+        }
+        return Optional.empty();
+    }
+
+    public static Optional<String> citationError(@Nullable CitationDto citation) {
```
rodvar, line 64: please fix: citationError bounds authorUserProfileId and text but not citation.chatMessageId - that one would go out to the whole network unbounded via the public message. A length cap here would do 👍

rodvar (changes requested):
overall a very solid PR; just one validation missing I think we can fix easily and saves us from network-type headaches.

I know from client side (mobile) this is guarded, but since we are not the only clients is important to fix 🙏

api/src/main/java/bisq/api/rest_api/endpoints/chat/ChatRequestValidation.java
salottodev: I put the cap in `Citation.verify()` with the existing `NetworkDataValidation.validateId` (50 chars), rather than only in the endpoint. `Citation` is also built by the desktop and by `fromProto`, so a check in the API layer alone would leave those two open.

The API keeps a check of its own for the same reason the author id has one: `validateId` appends the rejected id to its message, so leaving it to the constructor answers a 400 that hands the client back what it just sent.

`fromProto` calls `verify()`, so a node now also drops incoming citations whose id is over 50. That is intended, but it is a change on the network side and not only in the API.

rodvar: great 👍

rodvar (approved):
ACK

 - Tested on both Android and iOS from [mobile PR](https://github.com/bisq-network/bisq-mobile/pull/1804) that depends on this change.
 - one comment worth noting: the tightening-on-persisted-data landmine: `verify()` now rejects data that was previously accepted and `CommonPublicChatChannelStore.fromProto` has no per-entry try/catch, so a store containing a pre-upgrade overlong citation fails wholesale. I think impact would be a local history reset in this edge case (has to be a malicious peer doing it), no risk of crash loops so its acceptable 👍

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatMessagesWebSocketService.java
```diff
@@ -0,0 +1,165 @@
...
+        super(subscriberRepository, PUBLIC_CHAT_MESSAGES, channels, userProfileService, bannedUserService);
+    }
+
+    /**
+     * The channel set is fixed by {@code CommonPublicChatChannelService.initialize}, which completes
+     * before this service starts, and no public channel is added or removed afterwards, so the messages
+     * of each channel are observed once and for all.
+     */
+    @Override
+    public CompletableFuture<Boolean> initialize() {
```
KimStrand, line 73: The service observes only message additions/removals, while visibility depends on the independently changing profile and banned-user stores (api/src/main/java/bisq/api/chat/common/PublicChatDtoFactory.java:100). If an author is banned after their message was pushed, or a missing profile arrives after an addition was skipped, no websocket event reconciles the client. Reactions have the same problem. The live client therefore disagrees with REST and a fresh subscription until it reconnects. Should we observe those dependencies and emit the appropriate upserts/removals or force a topic resync when visibility changes?

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatReactionsWebSocketService.java
```diff
@@ -0,0 +1,282 @@
...
+ * <p>
+ * The unbind rides on the removal actually being processed, so it inherits the limitation
+ * {@code PublicChatMessagesWebSocketService#handleRemovedMessage} describes: a message the domain
+ * refuses to remove — its author banned or rate limited at that moment — keeps both its place in the
+ * channel and its binding here, for the life of the process. Bounded by how many authors are in that
+ * state, not by the message count, so it is a stale entry rather than the leak the unbind exists for.
+ */
+public class PublicChatReactionsWebSocketService extends ChannelScopedWebSocketService {
+    private final List<Pin> messagesPins = new ArrayList<>();
+    private final Map<String, MessageBinding> bindingsByMessageId = new ConcurrentHashMap<>();
```
KimStrand: Reaction observers are indexed solely by message ID, although message IDs are not enforced as globally unique and reactions identify both channel and message. A valid Support message reusing a Discussion message’s ID displaces and unbinds the Discussion observer. Removing the Support message then removes the remaining binding without restoring the Discussion observer. A peer can therefore suppress live reaction events for a targeted message. Should we key bindings by channel ID plus message ID?

api/src/main/java/bisq/api/dto/mappings/chat/common/CommonPublicChatChannelDtoMapping.java
```diff
@@ -0,0 +1,59 @@
...
+     * consolidation.
+     * <p>
+     * So this is desktop's behaviour rather than a divergence introduced here, and resolving the strings
+     * in a dto mapper would fix one surface of a shared problem while putting i18n key construction where
+     * it does not belong. Whoever confirms a store like that still exists should fix
+     * {@code getDisplayString()} for both.
+     *
+     * @param unreadCount from {@code ChatNotificationService}, which the channel does not know about
+     */
+    public static CommonPublicChatChannelDto fromBisq2Model(CommonPublicChatChannel value, long unreadCount) {
```
KimStrand, line 50: On upgraded stores, a deprecated survivor such as DISCUSSION_BITCOIN reports the migrated ID/domain for discussion.bisq, but getDisplayString() and getDescription() still use its legacy fields. The API can consequently return discussion.bisq labelled “Bitcoin,” giving mobile clients contradictory channel metadata. Should we normalize migrated channels before mapping their labels?

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatReactionsWebSocketService.java
salottodev: Confirmed. Bindings are now keyed by channel ID plus message ID, as `PrivateChatReactionsWebSocketService` already does by nesting its reaction pins per channel. Tests cover the displacement and the removal, and a second message in the same channel.

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatMessagesWebSocketService.java
salottodev: Agreed. The ban half is shared: `PrivateChatMessagesWebSocketService:347` and `PrivateChatReactionsWebSocketService:384` also read the banned-user store at push time without observing it.

The profile half is only in the public topics. Private chat messages carry `getSenderUserProfile()`, while the public ones resolve the author from the profile store. An addition skipped for a missing profile is not replayed when that profile arrives, so the message never reaches the client at all.

Should we cover all four services in a follow-up rather than observing the stores in this PR?

api/src/main/java/bisq/api/dto/mappings/chat/common/CommonPublicChatChannelDtoMapping.java
salottodev: Confirmed. `getId()` and `getChatChannelDomain()` migrate, but `getDisplayString()` and `getDescription()` build their i18n key from the raw `chatChannelDomain` field. A surviving DISCUSSION_BITCOIN therefore reads discussion.bisq under the Bitcoin title.

Nothing repairs that state. `Migration.migrate` has no caller, and `maybeAddDefaultChannels` returns early on a non-empty store. The desktop reads the same getters, so the profile card shows it too (`ChannelMessagesDisplayList:119`).

Should we fix the two getters on `CommonPublicChatChannel` in a follow-up, rather than normalize in the mapper?

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatReactionsWebSocketService.java
```diff
@@ -0,0 +1,300 @@
...
+     * is a message of the same channel with the same id, so it is an instance that is gone and its
+     * observer is unbound — see {@link BindingKey} for why the channel has to be in the key for that to
+     * hold.
+     *
+     * @return false if the message is gone or the shutdown has started, leaving the pin for the caller
+     * to unbind
+     */
+    private boolean install(MessageBinding binding) {
+        AtomicReference<MessageBinding> displaced = new AtomicReference<>();
+        MessageBinding current = bindingsByMessage.compute(keyOf(binding.owner()), (key, present) -> {
```
KimStrand, line 152: The composite key prevents collisions across channels, but it still assumes IDs are unique within a channel. That is not enforced: ChatMessage.verify() only validates the ID, while ChatChannel stores messages using full-object equality. A peer can therefore add a different message in the same channel with the target ID. Both remain live, but this replaces and unbinds the first binding; removing the second then leaves the original unobserved. The new same-channel test uses "m2", so it does not cover this collision. Should we enforce one live message per (channelId, messageId), or support multiple live bindings for that key?

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatMessagesWebSocketService.java
KimStrand: The broader ban state reconciliation across all four services is appropriate for a follow-up. But the missing profile behavior is specific to these new public WebSocket topics and silently loses events for the lifetime of the subscription. I recommend fixing that part in this PR.

api/src/main/java/bisq/api/dto/mappings/chat/common/CommonPublicChatChannelDtoMapping.java
KimStrand: Agreed that the domain getters are the right layer. Since this issue predates the API and also affects desktop, I do not consider it blocking for this PR. Should we open and link a follow-up issue for getDisplayString() and getDescription() before resolving this thread?

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatReactionsWebSocketService.java
rodvar: 👍  Enforcing uniqueness at the domain would make acceptance order-dependent (two nodes receiving the colliding messages in opposite order would diverge), so that's a protocol change, no? I'd rather not smuggle into an API PR. I'll go with multiple live bindings per key: the map value becomes a small list, install appends, and unbind only removes entries whose owner is no longer live - isLive already does the identity check needed for that. Will add the same-id/different-content test to cover it.

rodvar:
@KimStrand thanks for the review, I'm taking over this as discussed on matrix. Will create a new PR from my fork cherry-picking current commits to finalise the work together. Thank you 🙏

rodvar:
@HenrikJannsen please close in favour of https://github.com/bisq-network/bisq2/pull/4983

PR #4980, opened 2026-09-01

wodoro:
same as #4948 and #4960 (was waiting for #4961 to merge first)

Revoking a client only removed its profile and sessions. Subscriptions are authorized when they are taken out and never re-checked, so an established WebSocket kept delivering them, and push registrations are keyed by device and survived entirely. ClientRevocationService now defines what revocation means and runs a list of ClientRevocationHandlers, each guarded so one failure cannot leave a client half revoked, with the grant removed before the socket is closed so the re-read in SubscriptionService.subscribe sees it gone. ApiAccessService is the single entry point and delegates to it; Desktop no longer reassembles the steps itself.

Restore GET and DELETE on /access/clients, removed upstream as unreachable because no rule mapped them and unmapped paths fail closed. They are reachable here through a new CLIENT_MANAGEMENT permission, whose javadoc records why a cross-client capability is auto-grantable. The listing is paginated and sorted, and maps to a secret-free DTO before paging, so no ClientProfile with its plaintext secret can reach the response body.

Tie push registrations to the API client that created them, which is what lets revocation find them. Registering a device ID owned by another client is refused, unregistering is limited to the owner, and a registration stored while its client was being revoked is swept by re-reading the pairing store afterwards.

## Summary by CodeRabbit

* **New Features**
  * Added client-management endpoints for listing paired clients and revoking access with secure management IDs.
  * Revocation now cleans up sessions, WebSocket connections, and device registrations, with retry support for incomplete cleanup.
  * Added public chat API availability.
  * Linked mobile device registrations to API clients and prevented unauthorized changes.
* **Bug Fixes**
  * Improved handling of revoked clients across WebSocket and device APIs.
  * Pairing now rejects blank names and safely limits overly long names.
* **Permissions**
  * Added dedicated client-management access controls.


api/src/main/java/bisq/api/rest_api/endpoints/devices/DevicesRestApi.java
```diff
@@ -133,14 +153,31 @@ public Response registerDevice(RegisterDeviceRequest request) {
...
+            DeviceRegistrationResult result = deviceRegistrationService.register(
                     deviceId,
                     deviceToken,
                     publicKeyBase64,
                     deviceDescriptor,
                     platform,
-                    symmetricKeyBase64
+                    symmetricKeyBase64,
+                    clientId
             );
```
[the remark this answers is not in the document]
wodoro: This is intended behavior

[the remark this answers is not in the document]
wodoro: pre-existing registrations without an owner fail open, and the reasoning behind that still holds: register now requires a clientId, so no new ownerless rows can be created, the set is closed and shrinks as each real owner registers again and claims its row

api/src/main/java/bisq/api/access/pairing/PairingService.java
```diff
@@ -112,7 +126,7 @@ public ClientProfile requestPairing(byte version,
         String clientSecret = Base64.getUrlEncoder().withoutPadding().encodeToString(secret);
         ClientProfile clientProfile = new ClientProfile(clientId,
                 clientSecret,
-                clientName);
+                cappedClientName);
         apiAccessStoreService.putClientProfile(clientId, clientProfile);
```
KimStrand: The profile and permission grant are written separately. The new listing endpoint can expose the profile between those writes; a concurrent DELETE then removes the profile and nonexistent grant and returns REVOKED, after which pairing writes the grant. With shipped session handling disabled, that orphan grant is sufficient for authorization. Should we persist profile and permissions atomically under the same ?lock as removal

notifications/src/main/java/bisq/notifications/mobile/registration/DeviceRegistrationService.java
```diff
@@ -67,25 +80,94 @@ public void register(String deviceId,
...
+     * matched here; they have to be removed by device ID.
+     *
+     * @param clientId The API client whose registrations should be removed
+     * @return The device IDs that were removed
+     */
+    public Set<String> unregisterByClientId(String clientId) {
+        checkArgument(StringUtils.isNotEmpty(clientId), "clientId must not be null or empty");
+
+        synchronized (persistableStore) {
+            Set<String> deviceIds = persistableStore.getDeviceByDeviceId().values().stream()
```
KimStrand, line 160: Existing persisted registrations deserialize without clientId, while this filter removes only registrations with a matching present owner. Revoking a lost device immediately after upgrade therefore returns 204, yet that device continues receiving encrypted chat bodies and trade identifiers. Ownerless records are also claimable/removable by any client knowing the device ID, so that fallback is not a safe migration. Should we migrate or clear ownerless registrations, or report revocation as incomplete until they are safely handled?

api/src/main/java/bisq/api/web_socket/WebSocketConnectionHandler.java
```diff
@@ -104,44 +113,62 @@ public void onMessage(WebSocket webSocket, String message) {
...
+     * exception escaping here would leave the client's remaining sockets connected, and a revoked
+     * client must not keep a live connection. A close that fails cannot be forced at this level,
+     * so the socket is unsubscribed and deregistered anyway, which stops the data it had already
+     * subscribed to. New subscriptions are refused by the authorization in
+     * {@code SubscriptionService}, which reads the grant this revocation has removed.
      *
-     * @param clientId The client ID whose connection should be closed
+     * @param clientId The client ID whose connections should be closed
      */
     public void disconnectClient(String clientId) {
```
KimStrand, line 147: disconnectClient snapshots registered sockets, but a handshake can reach onConnect after that snapshot and be registered without any revocation check. The comment says onConnect catches this case, but api/src/main/java/bisq/api/web_socket/WebSocketConnectionHandler.java:73 currently performs no such check. DELETE can consequently return 204 while a socket for the revoked ID remains connected. Should we revalidate the authoritative client state during connection registration?

notifications/src/main/java/bisq/notifications/mobile/registration/DeviceRegistrationService.java
wodoro: Migration added, but this can result in missed notifications until the next registration by client

api/src/main/java/bisq/api/access/pairing/PairingService.java
wodoro: done

api/src/main/java/bisq/api/web_socket/WebSocketConnectionHandler.java
wodoro: fixed

notifications/src/main/java/bisq/notifications/mobile/registration/DeviceRegistrationService.java
```diff
@@ -43,13 +44,51 @@ public DeviceRegistrationService(PersistenceService persistenceService) {
...
+    public void onPersistedApplied(DeviceRegistrationStore persisted) {
+        synchronized (persistableStore) {
+            Set<String> unowned = persistableStore.getDeviceByDeviceId().values().stream()
+                    .filter(profile -> profile.getClientId().isEmpty())
+                    .map(MobileDeviceProfile::getDeviceId)
+                    .collect(Collectors.toSet());
+            if (unowned.isEmpty()) {
+                return;
+            }
+            unowned.forEach(persistableStore.getDeviceByDeviceId()::remove);
```
KimStrand: Loading this version deletes every legacy push registration, not only registrations belonging to a revoked client.

All records written by previous versions have an empty `clientId`, so the first node restart after upgrading clears the entire existing registration store. Recovery requires every mobile app to launch and successfully register again. That is not guaranteed: Connect 0.8.2 can attempt registration before the selected profile loads and never retry; this was fixed only after that release: https://github.com/bisq-network/bisq-mobile/commit/83c7f2df3b8f751bce745a988dad438cf319de67. Notifications can therefore stop silently for existing users.

wodoro: Yes, thats the trade-off I mentioned:

> Migration added, but this can result in missed notifications until the next registration by client

But I have a few ideas, which I will implement soon about this

wodoro: It didn't need to be either this or that (could do better)

wodoro: check [aa6572c](https://github.com/bisq-network/bisq2/pull/4980/commits/aa6572cf8e3288c86ea136569fbb047035fb1367)

We simply drop them when the user takes a security action, which gives us favorable behavior of pushes continuing normally till such security action is needed

api/src/main/java/bisq/api/rest_api/endpoints/access/AccessApi.java
```diff
@@ -191,4 +205,90 @@ public Response requestSession(
...
+    @ApiResponse(responseCode = "500", description = "Unexpected internal server error")
+    public Response getClients(@QueryParam("page") Integer page,
+                               @QueryParam("pageSize") Integer pageSize) {
+        try {
+            // Mapped to the secret-free DTO before paging, not via the mapping overload of
+            // buildPaginatedResponse: nothing that reaches the response body is then a
+            // ClientProfile, so no later edit can serialize the client secret by accident.
+            // Sorted because the profiles come from a map, and paging an unspecified order
+            // would let one client appear on two pages and another on none.
+            List<ClientProfileDto> clients = apiAccessService.getClientProfiles().stream()
```
KimStrand: Both shipped app configurations set supportSessionHandling=false, so Bisq Client Id alone is sufficient to authorize requests. This endpoint now exposes every paired client’s ID.

A client can save another client’s ID before being revoked, then continue accessing the API using that ID without its secret or session. Revoking the original client therefore does not reliably remove its access while another paired client remains.

Should we enforce session authentication throughout API access before exposing these IDs, or return separate management identifiers that cannot authorize requests?

wodoro: I will ask rodvar for details of why supportSessionHandling is false by default, I assume this has been slipped the previous work reviews, doesnt make any sense to me that a client id be enough to authorize requests, and make that the case by default. I will ping you when my investigation is over.

rodvar: Dug into this one, here's what I found.

Why is it false by default: history rather than a fresh decision. The flag comes from the Jan 2026 rename "Replace signatureBasedAuthenticationRequired with supportSessionHandling" (3deb9ce3ee) and inherited the off-default from that equally-optional mechanism. It stayed off because flipping it **fail-closes every already-paired client that doesn't do the session handshake** - with it true the node registers RestApiSessionAuthenticationFilter + WebSocketSessionAuthenticationFilter and rejects session-less requests. 

So it looks like a staged rollout ("ship the capability, flip the default once clients catch up") whose second half never happened.

The bad news? we are not flipping it true anywhere, no distro. The good news? We have all the code in place both sides (bisq2 api and the clients) to do it.

Until now this was no big deal but this PR exposes the clientId as Kim correctly mentioned, so its ok that now is the time to address this I guess.

So, what I'd suggest:

 1. **For this PR**: go with KimStrand's second option -> don't return raw clientIds; return non-authorizing management identifiers (hash/display id) and let the revoke endpoint accept those. That fixes the leak independently of any session rollout and makes the CLIENT_MANAGEMENT auto-grant argument valid again.
 2. **As a follow-up** (pre-existing issue, not this PR's to fix): stage the supportSessionHandling default flip for api_app behind the capability probe `like everything else`, **with a deprecation window for old paired apps**. Note sessions only fully close the stolen-ID attack once the session-less fallback is retired, so point 1 is needed regardless.
 3. please check - CLIENT_MANAGEMENT claims Permission ordinal 14 and the public chat API work is allocating in the same range. Ordinals feed the proto enum and the pairing-code decoder, so whichever PR merges second needs to renumber (and the mobile Permission pins follow). Let's coordinate so we don't land 14 twice. just a heads up ;)

wodoro: The pre-existing issue is serious: a client id is basically a bearer token and we never intended this. But the highest impact is limited to nodes exposed over LAN. I will make this PR work without exposing client id, but we need another PR enabling the session handling as soon as possible.

rodvar: agreed - thanks wod 💪

api/src/main/java/bisq/api/access/ClientRevocationService.java
```diff
@@ -0,0 +1,84 @@
...
+     * @param clientId The client ID to revoke
+     * @return {@code true} if the client was found and revoked; {@code false} if not found
+     */
+    public boolean revokeClient(String clientId) {
+        boolean removed = pairingService.revokeClientProfile(clientId);
+        sessionService.removeSessionByClientId(clientId);
+        revocationHandlers.forEach(handler -> {
+            try {
+                handler.onClientRevoked(clientId);
+            } catch (Exception e) {
```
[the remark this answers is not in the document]
KimStrand: The new management ID lookup prevents retrying an incomplete revocation. ClientRevocationService removes the profile before running cleanup handlers. If a handler fails, DELETE returns 500 with “retry”, but the next DELETE returns 404 because the management ID can no longer resolve to a profile. The failed cleanup never runs again.

Should we keep incomplete revocations addressable by management ID until cleanup succeeds, without preserving authentication access?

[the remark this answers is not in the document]
wodoro: @coderabbitai check 920b01f

api/src/main/java/bisq/api/rest_api/endpoints/devices/DevicesRestApi.java
```diff
@@ -133,14 +153,31 @@ public Response registerDevice(RegisterDeviceRequest request) {
...
+            if (result == DeviceRegistrationResult.DEVICE_OWNED_BY_ANOTHER_CLIENT) {
+                return buildResponse(Response.Status.FORBIDDEN, "Device is registered to another client");
+            }
+            if (!clientPairedCheck.test(clientId)) {
```
KimStrand: Because revocation now removes the profile after cleanup, this check can still pass after permissions have been withdrawn. An already authorized request can register a device after unregisterByClientId() has run but before profile removal. Revocation then returns 204 while leaving that push registration active.

Should we make clientPairedCheck also require the client’s permissions to still exist?

api/src/main/java/bisq/api/ApiService.java
```diff
@@ -159,7 +165,44 @@ public ApiService(ApiConfig apiConfig,
 
         SessionAuthenticationService sessionAuthenticationService = new SessionAuthenticationService(pairingService, sessionService);
 
-        ApiAccessService apiAccessService = new ApiAccessService(pairingService, sessionService);
+        // Nothing holds a grant when authorization is off, so asking for one would refuse every
+        // client. SubscriptionService short-circuits on the same flag.
+        Predicate<String> clientAuthorizedCheck = apiConfig.isAuthorizationRequired()
```
KimStrand: This unconditional true also applies when supportSessionHandling=true and authorizationRequired=false. Clients still have to authenticate in that configuration, but the post registration revocation check is bypassed.

An already authenticated request can register its device after the cleanup pass, leaving that push registration active even though revocation returns 204.

Should we bypass this check only when both session handling and authorization are disabled?

wodoro: was still running reviews locally, flagged this as well, will push fixes shortly

KimStrand:
utACK

PR #4983, opened 2026-09-03
Expose public chat over REST and websocket (v2)

rodvar:
continuation of work on https://github.com/bisq-network/bisq2/pull/4979

## Summary by CodeRabbit

* **New Features**
  * Added REST API support for browsing and managing public chat channels, messages, citations, edits, deletions, reactions, and notifications.
  * Added WebSocket updates for public chat channels, messages, and reactions.
  * Added the `public-chat` API feature and permission support.
* **Bug Fixes**
  * Improved chat input validation and asynchronous request timeout handling.
  * Added validation for cited message IDs.
  * Improved filtering and visibility handling for expired, banned, or unavailable content.
* **Tests**
  * Added comprehensive coverage for public chat REST, WebSocket, permissions, validation, visibility, and moderation behavior.


rodvar:
@KimStrand please review

KimStrand (commented):
I found several paths that can omit state until reconnect, including a normal inventory ordering used by the linked mobile client. The inline comments also cover convergence under re-delivery and collisions, duplicate observer registration, and an HTTP/WebSocket ordering guarantee that is not currently enforced.

Verification: the API test suite passes (497 tests, 1 skipped), as do the citation tests and `git diff --check`; these failures are in interleavings and cross-client behavior not covered by those tests.

api/src/main/java/bisq/api/chat/common/PublicChatDtoFactory.java
```diff
@@ -0,0 +1,204 @@
...
+    public boolean awaitsAuthorProfile(CommonPublicChatMessage message) {
+        return !message.isExpired()
+                && !bannedUserService.isUserProfileBanned(message.getAuthorUserProfileId())
+                && userProfileService.findUserProfile(message.getAuthorUserProfileId()).isEmpty();
+    }
+
+    /**
+     * The reaction counterpart of {@link #awaitsAuthorProfile}: the message itself is visible and only
+     * the reaction sender's profile is missing. A reaction skipped because the <em>message author</em>
+     * is missing needs no parking here — when that author arrives, the messages topic replays the
```
KimStrand: This assumption does not hold for the [linked mobile client](https://github.com/salottodev/bisq-mobile/blob/feature/connect-public-chat/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/chat/public_chat/ClientPublicChatServiceFacade.kt#L464-L470): its `toDomain` path explicitly replaces the DTO's embedded reactions with `reactionDtosByMessageId`. If the message author is unresolved when a reaction arrives, this topic neither sends nor parks the reaction; the later author replay only updates the messages topic, leaving the reaction missing until reconnect. Should we park this case for author-profile replay, or change the client to consume the embedded reactions and add integration coverage?

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatMessagesWebSocketService.java
```diff
@@ -0,0 +1,323 @@
...
+        return toJson(dtoFactory.visibleMessageDtosNewestFirst(messages));
+    }
+
+    private void handleAddedMessage(CommonPublicChatMessage message) {
+        if (shutdownStarted) {
+            return;
+        }
+        if (dtoFactory.isVisible(message)) {
+            push(message, ModificationType.ADDED);
+        } else if (dtoFactory.awaitsAuthorProfile(message)) {
```
KimStrand: There is still a lost-wakeup window before `park()`: `isVisible` can observe the profile as absent, then the profile can be stored and invoke `replayFor` while the pending map is empty, after which `awaitsAuthorProfile` observes it as present and returns false. The message is then neither pushed nor parked. The reactions path has the same two-lookup pattern. Should we classify the state from one lookup, or recheck visibility before falling through, with a deterministic interleaving test?

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatMessagesWebSocketService.java
```diff
@@ -0,0 +1,323 @@
...
+                .stream()
+                .flatMap(channel -> channel.getChatMessages().stream())
+                .filter(live -> live.getId().equals(removed.getId()))
+                .max(Comparator.comparingLong(CommonPublicChatMessage::getDate));
+    }
+
+    /** A message that leaves its channel while parked is not coming back; equality drops the parked copy. */
+    private void dropParked(CommonPublicChatMessage message) {
+        awaitingAuthor.computeIfPresent(message.getAuthorUserProfileId(), (id, present) -> {
+            present.remove(message);
```
KimStrand, line 315: Removing by equality can erase pending work for a still-live equal successor. In the supported P2P re-delivery ordering, an earlier removal observer can add and park a fresh equal successor; that add deduplicates against the predecessor already in this set, and this delayed removal then deletes their shared entry. When the profile arrives, nothing is replayed until reconnect. Should we retain the entry while an equal message remains live, or track pending instances or generations separately?

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatMessagesWebSocketService.java
```diff
@@ -0,0 +1,323 @@
...
+     * {@code PublicChatReactionsWebSocketService#handleRemovedReaction}, which carries the same note.
+     */
+    private void handleRemovedMessage(CommonPublicChatMessage message) {
+        if (shutdownStarted) {
+            return;
+        }
+        dropParked(message);
+        // An invisible survivor falls through to REMOVED on purpose: the snapshot would not show it
+        // either, and if it is merely waiting for its author it gets its ADDED from the replay.
+        Optional<CommonPublicChatMessage> survivor = survivorSharingId(message).filter(dtoFactory::isVisible);
```
KimStrand: Visibility is applied after selecting the newest same-ID survivor. If the newest collider is unresolved, banned, or expired while an older visible collider remains, this filters the result to empty and emits `REMOVED`, even though a fresh snapshot still contains the older visible message. Should we filter candidates for visibility before selecting the newest survivor?

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatReactionsWebSocketService.java
```diff
@@ -0,0 +1,475 @@
...
+     * after the reaction was pushed produces no {@code REMOVED} at all and the client keeps an orphan
+     * until it reconnects. Carrying the removal without the profile would mean a dto that admits an
+     * absent author, which is a wider contract change than the window is worth.
+     */
+    private void handleRemovedReaction(CommonPublicChatMessageReaction reaction) {
+        if (shutdownStarted) {
+            return;
+        }
+        dropParked(reaction);
+        push(reaction, ModificationType.REMOVED);
```
KimStrand, line 471: Reaction removal still has the ID-collision problem handled above for messages. Reaction IDs are peer-controlled and not validated as unique, so distinct live reactions can share an ID. Removing one sends delete-by-ID, and the [linked mobile client](https://github.com/salottodev/bisq-mobile/blob/feature/connect-public-chat/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/chat/public_chat/ClientPublicChatServiceFacade.kt#L369-L375) removes the shared map entry, hiding the surviving reaction until reconnect. Should we use a collision-safe wire identity or re-push a same-ID survivor here?

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatReactionsWebSocketService.java
```diff
@@ -0,0 +1,475 @@
...
+    private boolean install(MessageBinding binding) {
+        List<MessageBinding> dead = new ArrayList<>();
+        AtomicBoolean installed = new AtomicBoolean();
+        bindingsByMessage.compute(keyOf(binding.owner()), (key, present) -> {
+            List<MessageBinding> live = new ArrayList<>();
+            if (present != null) {
+                present.forEach(each -> (isLive(each.owner()) ? live : dead).add(each));
+            }
+            if (!shutdownStarted && isLive(binding.owner())) {
+                live.add(binding);
```
KimStrand, line 209: The same message instance can be bound twice during observer registration. `ObservableCollection.addObserver` registers the observer before iterating the live collection, so a concurrent addition can arrive through both `onAdded` and the initial replay. This method retains both bindings, causing every later reaction to be emitted twice for that message. Should we reject a binding whose owner is already present by identity while continuing to preserve genuinely distinct same-ID messages?

api/src/main/java/bisq/api/rest_api/endpoints/chat/public_chat/PublicChatRestApi.java
```diff
@@ -0,0 +1,598 @@
...
+ * deleting and un-reacting use the author's identity; and messages are removed for real, which the
+ * messages topic reports as REMOVED events. The history is delivered whole, see
+ * {@link PublicChatDtoFactory#visibleMessagesNewestFirst}.
+ * <p>
+ * Every path addresses a channel by id, so the two channels differ only in which
+ * {@code CommonPublicChatChannelService} publishes for them; {@link PublicChatChannels} resolves that.
+ * <p>
+ * A 204 means the node applied the change locally and started broadcasting it, never that peers have
+ * it. Where the local change happens asynchronously (edit, delete, un-react all go through the P2P
+ * store's remove), the response waits for it, so that the 204 never arrives before the matching
```
KimStrand: The stated client-visible ordering is not enforced. `BaseWebSocketService.send` discards the future from `Subscriber.send`, which only queues the actual WebSocket write on another executor; the REST callback can therefore resume with 204 first when that executor or socket is delayed. Should we await the matching send before completing the response, or remove this guarantee and require clients to tolerate either ordering?

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatMessagesWebSocketService.java
rodvar: I don't think this is a problem, it relies on that shared-lock serialization, since it's invisible from this file alone.

api/src/main/java/bisq/api/web_socket/domain/chat/public_chat/PublicChatMessagesWebSocketService.java
rodvar: this is the most serious one, thank you I'm already working on it.

api/src/main/java/bisq/api/rest_api/endpoints/chat/public_chat/PublicChatRestApi.java
rodvar: blocking the rest response for this would be a no-go IMHO, I will fix the Java-doc

KimStrand (commented):
utACK

PR #4984, opened 2026-09-04
Stop deriving proto registry keys from lambda class names

wodoro:
## Why this is needed

Two runtime registries derived their lookup keys from the class name of a lambda. Lambda class names are not specified by the JVM, and optimizers merge lambdas into shared synthetic classes, so under an optimizing build both registries silently produce garbage keys.

The consequences are the worst kind: silent, and expensive. A wrong persistence key means every `PersistableStore` fails to resolve on the next launch, falls back to defaults, and the store file plus all its backups are moved into `corruptedFilesAtRead`. That rotates the user's key bundle and discards local trade data. A missing whitelist entry means `StorageService` refuses that store key, so the node stops accepting prices and offers from the P2P network.

Nothing is broken today, because current javac happens to produce stable names and nothing optimizes these jars. But this is unspecified behaviour with a data-loss failure mode, and it is what forces `-dontoptimize` in the Android Node build (bisq-network/bisq-mobile#1725). The fix is worth making on its own merits, independent of whether anyone ever enables R8.

Keys are byte-identical after this change, so no migration and no compatibility shim is needed.

---

## What was wrong

**`ProtoResolver#getProtoType(ProtoResolver)`** took the resolver object, read its class name, and sliced out the package and enclosing class:

```java
// resolver.getClass() is bisq.user.identity.UserIdentityStore$$Lambda$31/0x0000000800c8b720
String className = resolver.getClass().getSimpleName().split("\\$")[0];
return resolver.getClass().getName().split("\\.")[1] + "." + className;
```

It worked only because every store declares `getResolver()` as an inline lambda in its own body, so the synthetic class happened to carry the store's name.

**`NetworkStorageWhiteList#add(String, ProtoResolver)`** did the same with a different split, cross-checked the recovered name against the second token of the proto type name, added the entry on a match, logged a warning and added *nothing* on a mismatch, and carried hardcoded escape hatches for `support.MediationRequest` and `support.MediatorsResponse`, all wrapped in a catch-all `try/catch`. It ran as a side effect of `NetworkProtoResolverMap#addProtoResolver`.

These are the only two places in the repository that derive anything from a lambda class name.

---

### Pass the store, not a (class, resolver) pair

```java
public static void addResolver(PersistableStore<?> store) {
    protoResolverMap.addProtoResolver(ProtoResolver.getProtoType(store.getClass()), store.getResolver());
}
```

The obvious signature is `addResolver(Class<?>, ProtoResolver<PersistableStore<?>>)`, but `persistableStore.getClass()` erases to `Class<? extends PersistableStore>`, which does not fit `Class<? extends PersistableStore<?>>`, so it has to be widened to `Class<?>` and the type safety is thrown away. Taking the store instead removes the problem: there is one argument, so the key and the resolver cannot disagree, and no cast is needed. It also matches what `PersistenceService` already does two lines away, where the store's own `getClass().getSimpleName()` is the file name.

### Keep the proto type name as an explicit argument on the network side

The two network registrations now look like this:

```java
DistributedDataResolver.addResolver("user.UserProfile", UserProfile.class, UserProfile.getResolver());
```

That is the same fact spelled twice for 38 of the 40 registrations, which is a real cost, and deriving the name from the class would delete 38 duplicated literals. It was deliberately not done.

The proto type name is the wire contract. `support.MediationRequest` is registered with `BisqEasyMediationRequest` precisely because the class was renamed after the name was already on the wire. Under an explicit literal, renaming a class leaves the wire name alone, which is the safe default and is what actually happened. Under derivation, that same rename silently changes the wire name and the author has to notice and opt out. For a P2P protocol the default has to be the safe one.

Instead of deriving, the pairs are pinned by a test (see below), which catches both a wrong class and a mistyped name without inverting that default.

### Delete the match check and the legacy escape hatches rather than port them

The old check compared a name recovered from a lambda against the proto type name. It was a check on the *derivation*, not on the data: it asked "did I recover the right class?" With the class supplied directly there is nothing left to check it against, and the two `support.*` special cases existed only to excuse a mismatch that no longer exists. `BisqEasyMediationRequest.class.getSimpleName()` is correct by construction.

The `try/catch` guarded the parsing, and there is no parsing left. `@Slf4j` went with it, since nothing logs any more.

The one thing lost is that the check also incidentally caught a typo in the proto type name literal. That is now covered by the test, at build time, instead of a warning at runtime that nobody reads.

### `Class<? extends NetworkProto>`, not `Class<? extends DistributedData>`

`ChatMessage` and `ChatMessageReaction` are registered as abstract bases that implement `NetworkProto` but not `DistributedData`, so `NetworkProto` is the tightest bound that accepts every registration. `NetworkStorageWhiteList#add` was narrowed from `Class<?>` to the same bound while it was being touched.

### Move the whitelist call into the two network resolvers

It used to happen inside `NetworkProtoResolverMap#addProtoResolver`, in `common`, as an unnamed side effect of registering a resolver. It now happens in `DistributedDataResolver` and `NetworkMessageResolver`, which are the classes that have the payload class to register, and which live in the `network` module where a network storage concern belongs. Both methods are three lines where both effects are visible.

This leaves `NetworkProtoResolverMap` as an empty subclass. Deleting it was considered and rejected: the `T extends NetworkProto` bound is real compile-time separation that stops a persistence type being used in a network resolver map, and the same holds for `PersistableProtoResolverMap`.

### Fail with the class name when the package is missing

`clazz.getName().split("\\.")[1]` threw `ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1` from inside `getOrCreatePersistence`, naming nothing. That is exactly the failure an Android integrator hits with R8 repackaging and a missing keep rule, and the message gave them no way to find it. It now throws `IllegalArgumentException` naming the class and stating the constraint.

### Make the remaining five stores final

The key now follows the runtime class of the store, whereas the old derivation followed the class declaring the resolver lambda. For every store today these are the same class, which is why the keys did not change, but the old code was subclass-safe by construction and the new code is not: a subclass inherits `getResolver()`, which still unpacks the parent's proto, so the type URL stays the parent's while the registered key becomes the subclass's.

Five stores were `public class` rather than `public final class`. Two of them, `CommonPublicChatChannelStore` and `TwoPartyPrivateChatChannelStore`, are registered under an explicit file name rather than the default derived from the class, so a subclass would have read the parent's file under a mismatched key. The other 33 stores were already final, so this makes the property uniform.

### Pin the invariants with tests, and make them actually run

Three things had no coverage: the derivation rule itself, the contents of the whitelist, and whether a registration names one type consistently.

- `ProtoResolverTest` pins the derivation rule and the new guard.
- `ResolverConfigTest#whiteListMatchesResolverRegistrations` pins the 69 whitelisted names. Adding a payload type is meant to fail it, which is the moment the whitelist should be looked at.
- `ResolverConfigTest#everyRegistrationNamesOneTypeConsistently` reads whole registration calls and checks the wire name, the class and the decoder owner against each other. The two frozen names sit in `LEGACY_PROTO_TYPE_NAMES`, where a reader looks for them, rather than buried in a utility in `common`.
- `ResolverConfigTest#everyRegistryHoldsExactlyTheExpectedTypes` pins the proto type names per registry. Per registry rather than as one set, because a type dropped from one of the two still leaves that registry unable to decode it while the shared whitelist looks untouched.
- `PersistableStoreContractTest` asserts that the name a store registers under equals the proto type of the message its own `getResolver` unpacks, read from the `.proto` sources. A store which inherits its resolver, or is nested, or is not final fails too, since each is a way to make those two disagree.

The source-reading tests parse with the java compiler rather than matching text. An earlier revision used regular expressions and a review demonstrated that they silently skipped legal code: `PersistableStore <T>`, an annotation on the declaration line, a nested declaration, and a call written `addResolver (`. A pattern that misses a case is worse than no guard, because it reports success, and patching the known cases only moves the blind spot. Sources are read rather than the classpath because the stores live in 16 modules and no module depends on all of them, and 18 of the 37 are package-private so a test elsewhere could not construct them anyway. Every scan asserts up front that it reached known entries, so one that finds nothing fails instead of passing vacuously. The scanned sources are declared as task inputs, because without that Gradle keeps the task up to date and the test silently does not run when a store in another module changes; that was verified by mutation, not assumed.

### Run the checks in CI

Every workflow runs `./gradlew build -x test`, so no test in this repository runs on a pull request. These all guard silent, data-losing failures, so they get their own job that runs them by name, along with the reader/writer round trip. It is a separate job rather than a step in `build_bisq_module.yml`, which is called four times with different build roots across four operating systems: only the root build has these tasks, and one Linux run is enough for a source-level check. Enabling tests in CI generally is a separate decision and is not made here.

---

## Compatibility

**No migration.** The key is derived at runtime and never written anywhere. The persisted file carries the protobuf type URL, which is unchanged.

**Keys verified, not assumed.** Every `.proto` in the repository was parsed to build the real proto message name for each generated Java class, and for all 37 production `PersistableStore` implementations `ProtoResolver#getProtoType(storeClass)` equals the type URL suffix of the message that store's resolver unpacks. That is stronger than "unchanged": it matches the actual read path.

**Whitelist verified.** Dumped from `ResolverConfig.config()` before and after: 69 entries, byte-identical. The arithmetic agrees, 40 registrations covering 38 distinct classes (`ChatMessage` and `ChatMessageReaction` are each registered twice) plus 31 explicit `add(Class)` calls, no overlap.

**API breaks, deliberately loudly.** `PersistableStoreResolver#addResolver`, `DistributedDataResolver#addResolver`, `NetworkMessageResolver#addResolver`, `ProtoResolver#getProtoType` and the removed `NetworkStorageWhiteList` overload are all source and binary incompatible for an outside caller. That is intended: anyone affected finds out at compile time. bisq-mobile, the only known consumer, uses none of them and defines no `PersistableStore`.

**Obfuscation constraint.** Both the old and the new code depend on `getSimpleName()` being stable, so bisq2 core cannot be obfuscated. This constraint already existed, since `MetaData.className` is wire-visible and the explicit `NetworkStorageWhiteList.add(Class)` calls already relied on it, but it was written down nowhere. It is now stated in javadoc on both `getProtoType` and `NetworkStorageWhiteList`, and enforced by the new guard.

---

## Testing

`./gradlew compileJava compileTestJava` passes for the whole project. `:common:test`, `:application:test`, `:persistence:test`, `:chat:test` and `:network:network:test` pass.

Every test here was checked by mutation rather than only by passing, and each mutation reverts clean. For the store contract: a non-final store, a nested one, one declaring no resolver of its own, one whose resolver unpacks another module's proto, plus the three declaration forms the earlier regex missed. For the registrations: a registry losing a type while the other keeps it, a class paired with another type's decoder, and a mistyped wire name written with a space before the parenthesis. Each fails the assertion meant for it.

Not covered here: the Android acceptance criteria from bisq-network/bisq-mobile#1725, that a release build with R8 optimization enabled starts, keeps its stores across a restart and syncs P2P. That can only be verified in bisq-mobile against republished jars.

---

## Out of scope, worth separate issues

Private keys reaching `bisq.log` on a failed store read was listed here in an earlier revision. It was fixed separately in #4985, which is merged and is the base of this branch, so it is no longer open.

- **CI runs no tests at all.** `build -x test` in every workflow. This PR adds one job for its own checks rather than changing that policy.
- **Mailbox store keys.** For mailbox data the class name travels on the wire and `onAddMailboxRequest` routes on it. The whitelist answers only whether a name is known, never whether it is the right name for this payload. The deleted method was the only place asserting that a resolver and a proto type name belonged together, so it is worth stating explicitly whether anything else is meant to enforce the pairing or whether per-store validation is what binds payload to store. Pre-existing and unchanged.

## Bug Fixes
- Improved consistency when recognizing network messages and persisted data across nodes.
- Added validation to detect mismatches between registered message types and their corresponding classes.

## Refactor
- Standardized resolver registration across application, networking, and persistence components.
- Persistable storage implementations are now protected from unintended extension.

## Documentation
- Updated guidance for protobuf registrations and safe Java class renames.

## Tests
- Added automated checks for resolver registrations, protobuf type generation, storage compatibility, and store requirements.
- Added CI coverage for targeted persistence and resolver tests.


persistence/build.gradle.kts
```diff
@@ -2,3 +2,15 @@ plugins {
...
 }
+
+// PersistableStoreFinalityTest scans the java sources of every module, including the separate `apps` and `network`
+// builds, so those sources have to be declared as inputs. Without this the task stays up-to-date and the test silently
+// does not run when a store in another module changes.
+tasks.test {
+    inputs.files(
+        fileTree(rootDir) {
+            include("**/src/main/java/**/*.java")
+            exclude("**/build/**", "**/.git/**", "**/.gradle/**")
```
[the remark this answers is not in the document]
wodoro: Fixed in a872f4dd0a, though in the test rather than here, so the diff on this file is empty.

tasks.test already restricted its inputs to **/src/main/java/**/*.java. The mismatch was on the other side: PersistableStoreFinalityTest walked every .java under the repo root. The scan is now restricted to paths containing src/main/java, so the files it reads match the declared inputs.

I did not add the .git and .gradle exclusions. Requiring src/main/java already excludes them, and neither directory contains a path matching it, so those conditions could never fire. The build exclusion is kept, since generated protobuf sources live under build/generated/source/proto/main/java.

One consequence worth flagging: this drops the TimestampStore fixture in persistence/src/test/java from the scan, so it now finds 37 stores rather than 38. That is intentional. A non-final test fixture cannot cost a user their data, and covering test sources would rerun :persistence:test whenever any test file in the repository changes.

rodvar:
@coderabbitai full review

rodvar:
this solves https://github.com/bisq-network/bisq-mobile/issues/1725 and contribs to https://github.com/bisq-network/bisq-mobile/issues/1820

application/src/test/java/bisq/application/ResolverConfigTest.java
```diff
@@ -0,0 +1,195 @@
...
+        for (String line : source) {
+            Matcher matcher = IMPORT.matcher(line);
+            if (matcher.find()) {
+                packageBySimpleName.put(matcher.group(2), matcher.group(1));
+            }
+        }
+
+        List<String> registrations = new ArrayList<>();
+        List<String> mismatches = new ArrayList<>();
+        for (String line : source) {
```
KimStrand: The parser checks each source line separately, silently skipping registrations whose arguments span multiple lines. I reproduced this: misspelling the market-price protobuf name fails the test, but wrapping that same registration before the class argument makes both tests pass. Such a typo prevents incoming price data from resolving while CI stays green.

Should we inspect complete registration calls and add coverage for wrapped arguments?

wodoro: Thanks
Fixed
Also it was silently skipping the test, now a `Resolver.addResolver` call that the pattern cannot read fails as well

wodoro:
rebased on main

KimStrand:
utACK

persistence/src/main/java/bisq/persistence/PersistenceService.java
```diff
@@ -100,7 +100,7 @@ public <T extends PersistableStore<T>> Persistence<T> getOrCreatePersistence(Per
                                                                                  String fileName,
                                                                                  PersistableStore<T> persistableStore,
                                                                                  MaxBackupSize maxBackupSize) {
-        PersistableStoreResolver.addResolver(persistableStore.getResolver());
+        PersistableStoreResolver.addResolver(persistableStore);
```
HenrikJannsen, line 103: I would find `PersistableStoreResolver.register(persistableStore);` more clear. Before we passed the resolver directly, not we pass the store and the method derives the params.

HenrikJannsen (commented):
Provided author a codex review in matrix.

wodoro:
@HenrikJannsen thank you
addressed review concerns I thought is important to address in this PR
please check again

HenrikJannsen (approved):
utACK

HenrikJannsen:
@wodoro can you resolve conflicts?

wodoro:
MuSigMediationResultAcceptanceMessage has been removed in f832784519,  rebased and adjusted tests

HenrikJannsen (approved):
utACK

PR #4986, opened 2026-09-05
Make published jars reproducible across build machines

wodoro:
Gradle copies file modes from disk into jar entries, so the umask of the machine that compiled the classes and checked out the resources ends up in every published jar. Entry order and timestamps were already pinned, but building the same commit under umask 022 and then 002 still produced 27 differing jars out of 27.

Pin the entry modes to 0644/0755 in `bisq.java-conventions`. The 204 tracked resources carrying an exec bit are images, fonts and text, so normalising them loses nothing.

Separately, give the bundled `tor.zip` the fixed entry order and timestamps the jar tasks already have. It ships as a resource inside `tor.jar` on macOS and Windows, and its permissions stay as extracted so the Tor binaries remain executable, which
is why it is handled apart from the jar convention.

Verified with four clean builds, `--no-build-cache`, fresh daemon per umask: 27/27 jars differ before the change, 0/27 after.

Needs Gradle 8.3 or newer for `dirPermissions`/`filePermissions`; the pinned 8.9
satisfies it.

This is done as part of preparations for f-droid release of android apps

## Summary by CodeRabbit

* **Build Improvements**
  * Built JAR and Tor binary archives now use consistent file permissions.
  * Archive contents are packaged in a stable order without preserving build-machine timestamps, improving reproducibility across builds.

### DIFF
--- a/build-logic/commons/src/main/kotlin/bisq.java-conventions.gradle.kts
+++ b/build-logic/commons/src/main/kotlin/bisq.java-conventions.gradle.kts
@@ -28,6 +28,10 @@ tasks {
     withType<Jar> {
         isPreserveFileTimestamps = false
         isReproducibleFileOrder = true
+        // Pinned so the entry modes don't follow the umask of the machine that compiled the classes
+        // and checked out the resources.
+        dirPermissions { unix("0755") }
+        filePermissions { unix("0644") }
     }
 }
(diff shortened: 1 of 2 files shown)


HenrikJannsen (approved):
utACK

PR #4990, opened 2026-09-09
Sync reproducible jar changes from main to for-mobile-based-on-2.1.12

wodoro:
Brings the reproducible jar work from main (#4986) to the mobile branch to prepare for f-droid release

The mobile branch never received the `withType<Jar>` block from "fix reproducible build issues" (5f0ad3c393), which #4986 builds on, so that commit is cherry-picked first. All three commits are cherry-picked with `-x`:

- 5f0ad3c393 fix reproducible build issues
  Only conflict was `gradle.properties`; resolved by keeping the mobile version, which already has `org.gradle.java.installations.auto-detect=false` and its own `tor.version`.
- 4c53a4a894 Pin jar entry permissions to make jars reproducible
- e23c12da04 Make the bundled Tor archive reproducible

## Summary by CodeRabbit

* **Build Improvements**
  * Archives and application packages are now generated reproducibly, with consistent file ordering and timestamps.
  * Packaged files use standardized permissions, reducing differences between builds on different machines.

### DIFF
--- a/build-logic/bi2p/src/main/kotlin/bisq/gradle/bi2p/Bi2pAppPlugin.kt
+++ b/build-logic/bi2p/src/main/kotlin/bisq/gradle/bi2p/Bi2pAppPlugin.kt
@@ -29,6 +29,8 @@ class Bi2pAppPlugin : Plugin<Project> {
             destinationDirectory.set(project.layout.buildDirectory.dir("generated"))
             include("bi2p-$version-all.jar")
             from(project.layout.buildDirectory.dir("libs"))
+            isPreserveFileTimestamps = false
+            isReproducibleFileOrder = true
         }
 
         val copyBi2pAppVersionToResources = project.tasks.register<Copy>("copyBi2pAppVersionToResources") {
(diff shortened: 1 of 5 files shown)


build-logic/bi2p/src/main/kotlin/bisq/gradle/bi2p/Bi2pAppPlugin.kt
```diff
@@ -29,6 +29,8 @@ class Bi2pAppPlugin : Plugin<Project> {
             destinationDirectory.set(project.layout.buildDirectory.dir("generated"))
             include("bi2p-$version-all.jar")
             from(project.layout.buildDirectory.dir("libs"))
+            isPreserveFileTimestamps = false
```
KimStrand, line 32: The ZIP still inherits the generated JAR’s filesystem permissions. In a Gradle 8.9 fixture, umasks 022 and 002 produced identical shadow JARs but different ZIPs because the entry permissions were 0644 versus 0664. This difference also changed the enclosing JAR’s hash.

Should we add `filePermissions { unix("0644") }` here and in `zipWebcamAppShadowJar`? Adding that override made the fixture’s archives identical across both umasks.

wodoro: The finiding is correct, but for bisq-mobile only the published module jars matter, and those are covered by this PR
This is a sync, so if we want to fix this, it makes sense to create a PR to main first, then update this PR with the fix, so that sync is correct
Since we care about the reproducibility on main as well, I'll send a PR with the suggested fix to main and change this to draft, once that's merged, I'll update and reopen this PR

KimStrand (commented):
utACK

PR #4993, opened 2026-09-09
Pin zip entry permissions of the bundled shadow jars

wodoro:
zipBi2pAppShadowJar and zipWebcamAppShadowJar wrap the shadow jar into a Zip that ends up as a resource inside the desktop jar. The shadow jar entries are already pinned by the jar convention, but the Zip copies the on-disk mode of the shadow jar file itself, which follows the umask of the building machine. Building under umask 022 and then 002 produced identical shadow jars but different zips (0644 versus 0664 entry), and so a different desktop jar.

Pinning the entry mode makes both zips identical across umasks. Note the Gradle daemon keeps the umask it was started with, so comparing umasks needs a daemon restart between builds.

suggested by @KimStrand at https://github.com/bisq-network/bisq2/pull/4990#discussion_r3971538637

## Summary by CodeRabbit

- **Build Improvements**
  - Standardized file permissions to `0644` for entries in generated application archives, ensuring consistent archive permissions across build environments.


KimStrand (commented):
utACK

PR #4994, opened 2026-09-10
Judge the MuSig minimum on the fiat obligation

balag3:
Taking a MuSig offer at its 10 USD minimum fails at the maker with "The maker has rejected the take offer request because the offer is not available anymore." The maker log shows the real reason: `The trade amount lies below the absolute minimum. usd=99998.3253`.

The taker's 10 USD converts to 8,977 sats at the current price, and the maker's validation converts those sats back to 9.9998 USD and compares that with the 10 USD minimum. The check judged the Bitcoin-side value with no tolerance, while the rail cap check next to it already takes the fiat obligation into account. Introduced with #4938.

### Fix

For a USD quote the absolute minimum now bounds the quote amount itself, which is the obligation the policy limits express. The rail cap keeps bounding the larger of the quote amount and the Bitcoin-side value, as before. For non-USD fiat quotes the Bitcoin-side value still stands in for the obligation, since that needs a fiat/USD conversion (the follow-up already noted in the code). Applying the price tolerance to the minimum instead would also have worked; I went with the obligation because it mirrors the cap and needs no threshold.

One consequence worth stating: within the maker's price tolerance the Bitcoin side of a 10 USD take can now sit a few percent under 10 USD, where before it was the quote side that could. The two sides are bound to each other by the tolerance check either way, so this is a side swap rather than a loosening. If you'd rather have the minimum hold on both sides, the strict form is the smaller of the two values with a one-satoshi allowance; say so and I'll switch.

### Testing

Two validator tests: a 10 USD take at 111,393.92375 USD/BTC (8,977 sats, worth 9.9998 USD on the way back) is accepted, and a 9.99 USD take whose Bitcoin side is worth 10.03 USD is rejected. Both fail on the unfixed validator. `:trade:test` is green.

Found while manually testing #4992 with two local clients: the first take at the offer minimum was rejected, a take at 50 USD went through.

## Bug Fixes

- Corrected minimum trade amount validation for USD-quoted Bitcoin markets to use the quote-side obligation, preventing rounding from incorrectly accepting or rejecting offers.
- Improved rail-limit validation by considering the greater of the trade obligation and Bitcoin-side value.
- Rejects offers when the BTC/USD price is zero or otherwise non-positive, instead of bypassing Bitcoin-side limit checks.
- Added coverage for edge cases involving minimum USD obligations and Bitcoin value rounding.

### DIFF
--- a/trade/src/main/java/bisq/trade/mu_sig/messages/network/handler/maker/MuSigTakeOfferRequestValidator.java
+++ b/trade/src/main/java/bisq/trade/mu_sig/messages/network/handler/maker/MuSigTakeOfferRequestValidator.java
@@ -344,28 +344,30 @@ public final class MuSigTakeOfferRequestValidator {
             // the absolute and rail limits, so require a fresh one before enforcing them.
             throw reject("No fresh BTC/USD price is available to validate the absolute trade limits");
         }
-        // Exact BigDecimal (PriceQuote's longValue would wrap at an extreme price); the result is
-        // a USD Fiat atomic value comparable to the policy limits.
-        BigDecimal usdAtomic = BigDecimal.valueOf(btcAmount.getValue())
+        // Exact BigDecimal; the result is a USD Fiat atomic value comparable to the policy limits.
+        BigDecimal btcValueUsd = BigDecimal.valueOf(btcAmount.getValue())
                 .multiply(BigDecimal.valueOf(btcUsdMarketPrice.get().getPriceQuote().getValue()))
                 .movePointLeft(btcAmount.getPrecision());
-        if (usdAtomic.compareTo(BigDecimal.valueOf(MuSigTradeAmountLimitsPolicy.MIN_USD_TRADE_AMOUNT.getValue())) < 0) {
-            throw reject("The trade amount lies below the absolute minimum. usd=" + usdAtomic.toPlainString());
+        // The limits bound the fiat obligation. For a USD quote that is the quote amount itself;
+        // the Bitcoin-side value only stands in for it otherwise (non-USD fiat needs a fiat/USD
+        // conversion - a disclosed follow-up). Judging the minimum on the Bitcoin side rejects a
+        // take at exactly the minimum, whose satoshi rounding lands a fraction below it; that
+        // remains the case for non-USD fiat until the conversion exists.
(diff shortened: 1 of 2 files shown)


trade/src/main/java/bisq/trade/mu_sig/messages/network/handler/maker/MuSigTakeOfferRequestValidator.java
```diff
@@ -344,28 +344,30 @@ private static void validateUsdLimits(MarketPriceService marketPriceService,
...
+        // take at exactly the minimum, whose satoshi rounding lands a fraction below it; that
+        // remains the case for non-USD fiat until the conversion exists.
+        BigDecimal obligationUsd = market.isBaseCurrencyBitcoin() && "USD".equals(market.getQuoteCurrencyCode())
+                ? BigDecimal.valueOf(quoteSideAmount)
+                : btcValueUsd;
+        if (obligationUsd.compareTo(BigDecimal.valueOf(MuSigTradeAmountLimitsPolicy.MIN_USD_TRADE_AMOUNT.getValue())) < 0) {
+            throw reject("The trade amount lies below the absolute minimum. usd=" + obligationUsd.toPlainString());
+        }
+        // The tolerance lets the Bitcoin value sit above the obligation, so cap on the larger of the two.
+        BigDecimal cappedUsd = obligationUsd.max(btcValueUsd);
```
KimStrand, line 369: A fresh zero or negative BTC/USD quote now fails open for USD markets. MarketPrice validates only timestamp freshness, and when btcValueUsd <= 0, this max discards the invalid Bitcoin-side value and applies only the fiat obligation. I reproduced this with a fixed price 10 BTC contract and a $5,000 ACH obligation: using a fresh zero BTC/USD snapshot, the validator accepted the request instead of rejecting it, so the Bitcoin side rail cap was not meaningfully enforced. Before this change, the minimum comparison implicitly rejected that snapshot. Should we explicitly reject BTC/USD quotes whose value is non positive before calculating btcValueUsd?

balag3: Confirmed. Fixed in: 6d682b95f5.

KimStrand (commented):
utACK

Issue #4796, opened 2026-06-02
Bisq2 v2.1.11 Linux DEB differs from a clean build of the release tag


xrviv:
Hi Bisq maintainers,

I'm Danny and I work with [WalletScrutiny.com](https://walletscrutiny.com) and we verify the reproducibility of Bitcoin Apps and Devices. 
First of all, kudos to open sourcing one of the most iconic Bitcoin projects in the space.

While checking the reproducibility of the Bisq2 v2.1.11 Linux DEB, we found that the published `Bisq-2.1.11.deb` differs from a clean build of the public `v2.1.11` tag.

We are sharing the findings in case they help improve the release build process.

## Build details

- Repository: https://github.com/bisq-network/bisq2
- Tag: `v2.1.11`
- Public tag commit: `505e4f48a46afd3bf87918aa4869acbceb965f03`
- Build task: `:apps:desktop:desktop-app-launcher:generateInstallers`
- Build environment: Ubuntu 22.04 container with Zulu JDK 21.0.11
- Official artifact: `Bisq-2.1.11.deb`
- Built artifact: `bisq2_2.1.11-1_amd64.deb`

| | Official | Clean rebuild |
|---|---:|---:|
| DEB size | 423,135,576 bytes | 345,271,924 bytes |
| Extracted size | 646,716,818 bytes | 566,064,043 bytes |
| SHA256 | `4ecd2ed516a981fcffcf5ed4dd548dd7bfeb7400abfca3869a26635a718999cc` | `93c10eb26fbe971352a0375b7bcf57df03aa9e9743fdce7a9b4e3417d7f96890` |

## Main difference

Most of the extracted size difference is inside `/opt/bisq2/lib/app/desktop.jar`.

| | Official | Clean rebuild |
|---|---:|---:|
| `desktop.jar` size | 329,108,468 bytes | 248,455,691 bytes |

The official `desktop.jar` contains both of these files:

```text
webcam-app/webcam-app-1.0.0.zip
webcam-app/webcam-app-2.0.0.zip
```

The clean rebuild contains only:

```text
webcam-app/webcam-app-2.0.0.zip
```

The additional `webcam-app-1.0.0.zip` file in the published DEB is `80,608,344` bytes.

This looks consistent with a generated file from an earlier build remaining in the release workspace, but we have not confirmed the release environment.

## Embedded revision difference

The `ApplicationVersion.class` string inside `common-2.1.11.jar` also differs:

```text
[...]

PR #5005, opened 2026-09-13
Ship only the current webcam archive to desktop

majesticsky:
Fixes the stale webcam archive part of #4796.

"processWebcamForDesktop" copies the whole generated directory, so nothing removes an archive from a previous webcam version. If the build directory survives a version bump, both archives are copied into the desktop resources and shipped in "desktop.jar" — the pair @xrviv found in the official 2.1.11 DEB, where the stale 1.0.0 archive is ~80 MB but is never read: "WebcamJarProvider" resolves the exact resource path from "version.txt".

The task is now a "Sync" restricted to the current version's archive. A plain "Sync" would also remove "version.txt" from the target, though, which "copyWebcamAppVersionToDesktop" previously wrote there. The new task therefore copies "version.txt" as well, and the separate task is dropped.

Checked on this branch:

- with a stale "webcam-app-1.0.0.zip" placed in "webcam-app/build/generated", the desktop resources contain only "version.txt" and "webcam-app-2.0.0.zip";
- before the change, the same setup produced both archives;
- ":apps:desktop:desktop:processResources" is green;
- a repeated run is "UP-TO-DATE".

One limitation worth stating: if a stale archive is already present in the build directory, a clean build is still required. Gradle can keep the task "UP-TO-DATE" when its inputs have not changed. This change prevents stale archives from being copied; it does not retroactively remove one from an existing build directory.

## Summary by CodeRabbit

- **Chores**
  - Streamlined desktop webcam app packaging.
  - Packaging now synchronizes only the required versioned archive and source assets, reducing unnecessary build output.
  - Removed a redundant intermediate copy step.


majesticsky:
Anything needed from my side to get the checks running here?

Meanwhile the same commit is green in my fork:
https://github.com/majesticsky/bisq2/actions/runs/35207221085

"Build Bisq 2" passed all 16 jobs — ubuntu-24.04, macos-15, macos-15-intel and windows-2025. The commit built there is "be9a8379", the one this PR points at.

KimStrand:
@majesticsky 
If you are on Matrix feel free to reach out. I am @midnightdeer:matrix.org there.

build-logic/webcam-app/src/main/kotlin/bisq/gradle/webcam_app/WebcamAppPlugin.kt
```diff
@@ -41,15 +33,18 @@ class WebcamAppPlugin : Plugin<Project> {
...
-            dependsOn(copyWebcamAppVersionToDesktop)
 
             val desktopProject = project.parent?.childProjects?.filter { e -> e.key == "desktop" }?.map { e -> e.value.project }?.first()
             desktopProject?.tasks?.let {
-                from(project.layout.buildDirectory.dir("generated"))
-                exclude("sources")
+                from(project.layout.buildDirectory.dir("generated")) {
+                    exclude("sources")
+                    include("webcam-app-" + VersionUtil.getVersionFromFile(project) + ".zip")
+                    include("src/**")
```
KimStrand: is this needed?

PR #5025, opened 2026-09-19
fix: whitelist the two missing MuSig mediation messages

### LINKED ISSUE #5024: Two MuSig mediation messages cannot be stored as mailbox messages
## Summary

`MuSigCustomPayoutPsbtMessage` and `MuSigMediationResultRejectionMessage` are missing from `NetworkStorageWhiteList`. Both are mailbox messages, so when the counterparty is offline the store lookup is rejected and the message is not stored. The MuSig mediation custom payout flow loses those two messages instead of delivering them later.

## Affected versions

Present since `f832784519` ("Add MuSig mediated custom payout flow"), which added both message types without adding them to the registration block in `ResolverConfig`.

## Why the message is dropped

`MuSigTradeMessage` extends `TradeMessage`, which implements `MailboxMessage`, so both types take the mailbox path when the peer is not reachable:

`TradeMessageSender.send` → `NetworkService.confidentialSend` → `ConfidentialMessageService.storeMailBoxMessage` → `DataService.addMailboxData(...)`.

That resolves a store by the payload's class name, and `StorageService` gates every store lookup on the whitelist:

```java
public CompletableFuture<MailboxDataStorageService> getOrCreateMailboxDataStore(String storeKey) {
    if (isStoreKeyNotAllowed(storeKey)) {
        return CompletableFuture.failedFuture(new IllegalArgumentException("Unsupported storeKey"));
    }
    ...
```

```java
private static boolean isStoreKeyNotAllowed(String storeKey) {
    Set<String> allowed = NetworkStorageWhiteList.getClassNames();
    return !allowed.isEmpty() && !allowed.contains(storeKey);
}
```

Neither class name is in `allowed`, so the future fails and the mailbox message is never stored.

## Reproduction

1. Start a MuSig trade and take it to mediation.
2. Have the mediator accept a custom payout, or have a peer reject the mediation result, while the counterparty is offline.
3. The sender's send path fails on the mailbox store instead of storing the message. When the counterparty comes back online there is nothing to collect.

## Scope

`MuSigTradeMessage` has 12 concrete subclasses; `ResolverConfig` registers 10. `BisqEasyTradeMessage`'s 10 subclasses are all registered. So these two are the only gap.

## Fix

Add both to the `addStorageType` block in `ResolverConfig`:

```java
addStorageType(MuSigCustomPayoutPsbtMessage.class);
addStorageType(MuSigMediationResultRejectionMessage.class);
```

and to `ResolverConfigTest.EXPECTED_CLASS_NAMES`.

Worth checking whether any other whitelist entry is missing at the same time, and whether existing installations need anything beyond the code fix (they do not: nothing was ever stored under these keys, so there is no data to migrate).

## Preventing a recurrence

Nothing currently compares the whitelist against the types that actually resolve a store key. `ResolverConfigTest` pins the set of registered names, which confirms what is registered but cannot report what is missing, and no compile time check can see an absent entry in a hand maintained list.

PR #5023 adds the comparison, in `StoragePolicyCoverageTest`:

```java
@Test
void everyStoredTypeIsAKnownStoreKey() throws Exception {
    ResolverConfig.config();
    Set<String> allowed = NetworkStorageWhiteList.getClassNames();
    List<String> unknown = new ArrayList<>();
    for (Class<?> clazz : BisqClasses.bisqClasses()) {
        // ... concrete types which resolve their own policy
        if (!allowed.contains(clazz.getSimpleName())) {
            unknown.add(clazz.getName());
        }
    }
    assertEquals(List.of(), unknown, "Stored types missing from NetworkStorageWhiteList ...");
}
```

That test is red on #5023 today and names exactly these two classes. It turns green once this issue is fixed and #5023 is rebased on top.

wodoro:
resolves https://github.com/bisq-network/bisq2/issues/5024

MuSigCustomPayoutPsbtMessage and MuSigMediationResultRejectionMessage were added in f832784519 but never added to NetworkStorageWhiteList.

Both extend MuSigTradeMessage, so they extend TradeMessage and are MailboxMessage. When the peer is offline the send path stores them as mailbox data, and StorageService gates every store lookup on the whitelist:

    if (isStoreKeyNotAllowed(storeKey)) {
        return CompletableFuture.failedFuture(new IllegalArgumentException("Unsupported storeKey"));
    }

Neither class name was allowed, so the mailbox store could not be opened and the message was not stored. The mediation custom payout and mediation rejection messages were lost whenever the counterparty was not online, instead of being delivered when it came back.

## Summary by CodeRabbit

* **New Features**
  * Added network support for additional MuSig trade messages related to custom payout processing and mediation result rejections.
  * These messages can now be resolved and stored as part of supported MuSig trade workflows.

* **Tests**
  * Added coverage to verify recognition and distribution of the newly supported message types.
  * Added validation to ensure all eligible stored message types are included in the supported storage registry.


HenrikJannsen (approved):
utACK

PR #5029, opened 2026-09-21
enhancement/for-mobile-based-on-2.1.13_do_not_store_plain_secrets_for_paired_devices

rodvar:
- resolves https://github.com/bisq-network/bisq-mobile/issues/1885
 - Stores a SHA-256 of the client secret in ClientProfile instead of he plaintext; new proto field clientSecretHash, legacy clientSecret is read-only and never written again
 - Authenticates session requests by hashing the presented secret and comparing in constant time
 - Hashes legacy plaintext secrets on load and persist on the same boot so the plaintext leaves the disk on first start of this version
 - Drops entries left without any credential by a downgrade, together with their grant, since nothing can authenticate as or revoke them
 - Returns the one-time plaintext to the pairing response via a transient NewPairing record with a redacted toString; ClientManagementId keys its HMAC with the stored hash, so existing management IDs are unchanged
 - Adds AI generated tests for hashing, migration write-back, field precedence, dead-entry drop and management ID stability

## Summary by CodeRabbit

* **Security**
  * Client secrets are now stored as hashes rather than plaintext.
  * Pairing responses provide the secret once, with sensitive values redacted from logs.
  * Authentication continues to work with migrated credentials.

* **Compatibility**
  * Existing stored credentials are migrated automatically while preserving access.
  * Invalid or malformed credentials are safely discarded with their permissions.

* **Bug Fixes**
  * Cleanup now removes sessions and push registrations for discarded client profiles.
  * Obsolete backups containing plaintext credentials are removed after migration.
  * Persistence failures are now reported instead of being silently ignored.


api/src/main/java/bisq/api/access/ApiAccessService.java
```diff
@@ -114,9 +112,7 @@ public SessionResponse requestSession(String clientId, String clientSecret) thro
         ClientProfile clientProfile = pairingService.findClientProfile(clientId)
                 .orElseThrow(() -> new InvalidSessionRequestException("No client profile found for Client ID"));
```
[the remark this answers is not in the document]
rodvar: he endpoint already returns the same 401 body for every rejection, so the message does not leak. This is not needed.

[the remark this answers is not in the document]
rodvar: fair, my point was about the body and yours is timing. I think is still not needed though: the gap is one sha256 over ~43 bytes plus a 32-byte compare vs a map miss, sub-microsecond behind an https/tor round trip with ms of jitter; and client ids are random 122-bit uuids, so even a perfect "this id exists" oracle leaves the 256-bit secret to guess. `client-id` enumeration is an accepted risk here. leaving as is.

rodvar:
@KimStrand

api/src/main/java/bisq/api/access/persistence/ApiAccessStoreService.java
```diff
@@ -48,16 +48,21 @@ public Map<String, PermissionSet> getPermissionsByClientId() {
...
+     * otherwise stay on disk until the next pairing or revocation.
+     * <p>
+     * Relies on the transport starting only after all stores are read: the load replaces the live
+     * maps without the monitor the write paths take, so a pairing landing during it could be lost.
      */
     @Override
     public void onPersistedApplied(ApiAccessStore persisted) {
-        if (persisted.hadPromotedEntriesDuringLoad()) {
-            log.info("Persisting grantAll promotions computed while loading the store");
+        if (persisted.needsWriteBackAfterLoad()) {
```
KimStrand: persist() backs up the original store before replacing it. The active file contains hashes, but the backup retains secrets that still authenticate. Older backups also remain untouched. Reading the data directory therefore still exposes existing clients’ credentials. Should we migrate or remove plaintext bearing backups as part of the successful migration?

api/src/main/java/bisq/api/access/persistence/ApiAccessStore.java
```diff
@@ -123,22 +125,57 @@ public bisq.api.protobuf.ApiAccessStore toProto(boolean serializeForHash) {
...
+        for (Map.Entry<String, bisq.api.protobuf.ClientProfile> e : proto.getClientProfileByIdMapMap().entrySet()) {
+            bisq.api.protobuf.ClientProfile profileProto = e.getValue();
+            if (!ClientProfile.hasUsableCredential(profileProto)) {
+                // Either written by a version that predates hashed secrets after it loaded a hashed
+                // store (it rebuilt the entry from its own model and dropped the hash), or a hash of
+                // the wrong length. No secret can ever match, so the client has to pair again; kept,
+                // the entry would only be a listing that cannot be revoked, because the management
+                // ID is keyed by the missing hash. Dropped rather than failing the load, so one bad
+                // entry does not take every other client's access with it.
+                log.warn("Dropping paired client {} whose stored credential is unusable", e.getKey());
```
KimStrand, line 141: In the downgrade/upgrade case handled here, removing the profile and permissions bypasses client revocation cleanup. Startup cleanup only scans remaining profiles, so the old push registration survives. It remains eligible for notifications and blocks the same device from registering after pairing again: its owner is still the deleted client, resulting in HTTP 403. Should we complete registration cleanup before permanently discarding these client IDs?

api/src/main/java/bisq/api/access/persistence/ApiAccessStoreService.java
rodvar: good catch: added Persistence.deleteBackups(): after the hashed store is written we drop every backup of this store and write once more so a hashed backup exists. `awaited` so nothing paired in between can be swept. 

Any backup that survives deletion is logged at error with its path since the next boot has no plaintext left to trigger it again.

api/src/main/java/bisq/api/access/persistence/ApiAccessStore.java
rodvar: 👍 , the drop bypassed the handlers. the store now records the ids it dropped and `completeInterruptedRevocations` runs revokeClient for them, so the push registration goes with the profile. residual: no retry on a later boot if a handler fails on that one, the id is not persisted. Acceptable IMHO, it needs a downgrade cycle plus a handler failure on the same boot and the cost is a stale registration, not access.

api/src/main/java/bisq/api/access/persistence/ApiAccessStoreService.java
```diff
@@ -47,19 +52,62 @@ public Map<String, PermissionSet> getPermissionsByClientId() {
...
+                persisted.hadPlaintextSecretsDuringLoad(),
+                persisted.getClientIdsDroppedDuringLoad().size());
+        if (!persisted.hadPlaintextSecretsDuringLoad()) {
             persist();
+            return;
         }
+        persist()
+                .thenCompose(written -> {
+                    log.info("Deleting backups of the store, they still hold plaintext client secrets");
+                    return persistence.deleteBackups();
```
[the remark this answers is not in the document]
KimStrand: I found the same issue

api/src/main/java/bisq/api/access/ClientRevocationService.java
```diff
@@ -112,5 +112,12 @@ public void completeInterruptedRevocations() {
                     log.warn("Completing the interrupted revocation of client {}", clientId);
                     revokeClient(clientId);
                 });
+        // Dropped by the store for an unusable credential, so profile and grant are already gone
+        // and only the handlers have work left: a push registration still owned by such a client
+        // would keep receiving notifications and refuse the same device after it pairs again.
+        pairingService.getClientIdsDroppedDuringLoad().forEach(clientId -> {
```
KimStrand, line 120: If unusable profiles are migrated while REST and WebSocket are disabled, ApiService.initialize() skips cleanup. The profiles are already removed from disk, and the dropped IDs exist only in memory. After restarting with the API enabled, their push registrations remain orphaned and block registration after pairing again. Should we run cleanup while the API is disabled or persist pending IDs until cleanup completes?

api/src/main/java/bisq/api/access/persistence/ApiAccessStoreService.java
```diff
@@ -47,19 +53,69 @@ public Map<String, PermissionSet> getPermissionsByClientId() {
...
+                    log.info("Deleting backups of the store, they still hold plaintext client secrets");
+                    return persistence.deleteBackups().thenCompose(nil -> {
+                        // Loud rather than silent: the store no longer holds a plaintext after this
+                        // boot, so nothing would try the deletion again.
+                        List<Path> remaining = persistence.getBackups().stream().map(BackupFileInfo::getPath).toList();
+                        if (!remaining.isEmpty()) {
+                            log.error("Backups still holding plaintext client secrets could not be deleted, " +
+                                    "remove them manually: {}", remaining);
+                        }
+                        return persist();
```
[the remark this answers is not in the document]
rodvar: valid but accepted: deleting our own backup file only fails on a broken data dir or a transient lock. a broken dir fails the hashed write first, which now keeps the backups and retries next boot. the residual case is logged at error with every remaining path. persisting a cleanup flag isn't worth a schema field for that; happy to add a boot-time backup scan in a follow-up if we see it in the field. In short, does not worth it.

KimStrand (commented):
utACK

